Issue #201, opened 2020-04-06
Rand::Random(const double p_Lower, const double p_Upper) returns values greater than expected when p_Lower > 0


jeffriley:
Rand::Random(const double p_Lower, const double p_Upper) returns values greater than expected when p_Lower > 0. 
 The return statement in Rand::Random(const double p_Lower, const double p_Upper) needs to change from

    return (gsl_rng_uniform(m_Rng) + lower) * (upper - lower);

to

    return (gsl_rng_uniform(m_Rng) * (upper - lower)) + lower;

SimonStevenson:
I confirm this is indeed an issue, and the correct solution

SimonStevenson:
@jeffriley Similarly, in 
Rand::RandomInt(const int p_Lower, const int p_Upper)
should
`gsl_rng_uniform_int(m_Rng, upper) + lower;`
which would generate an integer in the range [lower, upper+lower)
be
`gsl_rng_uniform_int(m_Rng, (upper-lower)) + lower;`
which would generate an integer in the range [lower, upper]

Do we ever actually use this function?

jeffriley:
Fix for Rand::RandomInt(const int p_Lower, const int p_Upper) is to return:

    return gsl_rng_uniform_int(m_Rng, (upper - lower)) + lower;

No, not used currently (so no harm done with that one) - there in case anyone wants it.

PR #203, opened 2020-04-07
hotfix-Issue_201_Rand

jeffriley:
Sync dev with master after hotfix applied (Issue #201)


ilyamandel:
The description above makes it sound like this is a merge from the hot fix branch into dev -- but it should be a merge from master to ensure that master and dev are identical, right?  Or is this just Github's way of breaking down the info?

ilyamandel:
Ahh, I see, it says master->dev at top.  Never mind.

PR #206, opened 2020-04-07
Update issue templates

SimonStevenson:


SimonStevenson:
One of the recommended items here https://github.com/TeamCOMPAS/COMPAS/community

ilyamandel:
Good idea, but since this isn't a hot fix, I suggest merging into dev first.  This will also help us keep master and dev synchronised.

reinhold-willcox:
This looks good, thanks Simon! I'll give the Open Source Guides a read when I have the chance.

ilyamandel:
I still suggest that this be changed to pull into dev, not master -- @SimonStevenson ?

SimonStevenson:
Yeah I agree, this should be into dev

Issue #207, opened 2020-04-07
Include link to git dev documentation to make it easy for people to find


SimonStevenson:
We should include a link to the git dev documentation
https://github.com/TeamCOMPAS/COMPAS/blob/master/docs/dev-git-workflow.md

Either on the homepage README.md
https://github.com/TeamCOMPAS/COMPAS/blob/master/README.md

Or from the Getting Started 
https://github.com/TeamCOMPAS/COMPAS/blob/master/docs/getting_started.md

This will make it easier for people to find, and thus contribute to COMPAS.

PR #219, opened 2020-04-11
Add links to git workflow document in both README.md and gettin_started.md

themikelau:
This fixes issue #207


ilyamandel:
@themikelau , do you really want this to go into master, not dev?  If so, you'll need to pull it into dev to avoid them becoming synchronised.

Issue #218, opened 2020-04-11
Lookup table for logfile definitions file parser changed erroneously


jeffriley:
COMPASUnorderedMap<STAR_PROPERTY, std::string> STAR_PROPERTY_LABEL in constants.h was changed erroneously in commit:

https://github.com/TeamCOMPAS/COMPAS/commit/aeabad6d0d78b7d7d3ab0d23b857f96951071e98

This map does not contain header strings - it is the lookup table for the logfile definitions parser, so should only be changed if properties are being added or deleted (or their names changed).  

STAR_PROPERTY::MEAN_ANOMALY had its lookup string changed - needs to be restored
STAR_PROPERTY::METALLICITY had its lookup string changed - needs to be restored
STAR_PROPERTY::MZAMS had its lookup string changed - needs to be restored
STAR_PROPERTY::OMEGA_ZAMS had its lookup string changed - needs to be restored

Header strings are in the following maps, and should be changed there:

std::map<ANY_STAR_PROPERTY, PROPERTY_DETAILS> ANY_STAR_PROPERTY_DETAIL
std::map<ANY_STAR_PROPERTY, PROPERTY_DETAILS> BINARY_PROPERTY_DETAIL

ilyamandel:
@jeffriley -- this has been fixed, right?  Can it be closed?

jeffriley:
Fixed in v02.09.09

PR #220, opened 2020-04-11
Issue #218 - Logfile definitions parser lookup table changed erroneously

jeffriley:
Defect repair:
(1) restored property names in COMPASUnorderedMap<STAR_PROPERTY, std::string> STAR_PROPERTY_LABEL in constants.h (issue #218) (was causing logfile definitions files to be parsed incorrectly)

See Issue #218


jeffriley:
Opps - not what I meant

PR #221, opened 2020-04-11
Issue 218 - logfile definitions parser lookup table changed erroneously

jeffriley:
(1) restored property names in COMPASUnorderedMap<STAR_PROPERTY, std::string> STAR_PROPERTY_LABEL in constants.h (issue #218) (was causing logfile definitions files to be parsed incorrectly) 

See issue #218


ilyamandel:
Don't forget to pull into dev to avoid master and dev diverging.

PR #225, opened 2020-04-11
Implementing Muller and Mandel SN fallback prescription

ilyamandel:
At this stage, the fallback has been tested and produces sensible results.  The kicks have not yet been tested, but will require more invasive changes to test properly, so I would like to pull this first in order to avoid a more significant divergence from dev.

Outputs are only affected when using the MULLERMANDEL prescription.

Example of SSE evolution showing the remnant mass distribution as a function of ZAMS mass is attached.  

MullerMandelRemnantMass

P.S. This fix also addresses issue # 215.


ilyamandel:
The while loops should be just fine if we use reasonable values — in fact, a single draw will be enough in >99% of the cases.  I suppose if someone picks crazy inputs, we’d get in trouble.  For now, though, the inputs are hard-coded, so I am not too concerned.

Issue #105, opened 2020-03-05
Error in output of He core mass in supernovae.csv


reinhold-willcox:
Looking at the Supernovae.csv file outputed from a default run of 100 systems, it appears that the He core mass is sometimes incorrect, usually when the supernova-ing star is HeHG. 

In these situations, the value of Mass_Total@CO_SN (which I believe is the total mass of the star pre-supernova) is very slightly less than Mass_He_Core@CO_SN (which should be the He core mass at the same moment). The difference is about a few tenths of a solar mass, and it does not happen for all HeHG stars, but it seems to me that there is a bug either in the calculation or in the printing of this parameter. 

Weirdly, of the systems in this output, this happens on all but one of the HeHG (#8) stars, as well as 2 (but not all) TPAGB (#6) stars. It does not happen for EAGB (#5) stars, and there are no other types in the output. I have not yet found a correlation with any other data to explain this. 

Running on default pythonSubmit settings with any of the following seeds (or just the standard run up to 100 stars) should reproduce this bug.
 1  52 2  4  6  10 13 28 24 18 29 30 44 45 70 73 77 85 91 94

SimonStevenson:
Does anyone (@jeffriley @ilyamandel ) have an idea of where to make this change to fix this issue? I have also seen it in recent test runs. I'm happy to make the changes but thought I'd check if someone had figured it out first. Do we just need to implement `CalculateHeCoreMassOnPhase` for HeGB and HeHG stars, and make it return the total mass of the star?

jeffriley:
HeHG.h implements CalculateHeCoreMassOnPhase():

    double CalculateHeCoreMassOnPhase() { return m_HeCoreMass; }

> make it return the total mass of the star

Just change the line shown above to

    double CalculateHeCoreMassOnPhase() { return m_Mass; }

SimonStevenson:
Ah yes, I was looking in the wrong place for the implementation (I looked in the .cpp file instead of the .h file). I'll implement this change on a branch and test it before/after the change.

jeffriley:
Yeah - if the implementation is small enough the convention is to just have it in the .h file...  Perhaps we could note that in the developer part of the documentation.

ilyamandel:
[...]

PR #231, opened 2020-04-15
Fix core and envelope calculations: HeHG and TPAGB

SimonStevenson:
This PR addresses issue #105 

See that issue for a description of the problem (including plots).

Up to the EAGB, the envelope and cores look ok. Then, for type 6 (TPAGB) things go wrong.

For this 7.5 Msol Star, McBAGB ~2 Msol

See Section 5.4 in the SSE paper (Hurley et al. 2000, https://ui.adsabs.harvard.edu/abs/2000MNRAS.315..543H/abstract)

Using current dev, I tried simulating the following mass stars: 7.0, 7.1, 7.2, 7.3, 7.4, 7.5, 7.6 Msol , all at solar metallicity.

Here is the output:

0: RandomSeed = 1, Initial Mass = 7,   Metallicity = 0.0142, Oxygen-Neon_White_Dwarf
1: RandomSeed = 2, Initial Mass = 7.1, Metallicity = 0.0142, Oxygen-Neon_White_Dwarf
2: RandomSeed = 3, Initial Mass = 7.2, Metallicity = 0.0142, Oxygen-Neon_White_Dwarf
3: RandomSeed = 4, Initial Mass = 7.3, Metallicity = 0.0142, Oxygen-Neon_White_Dwarf
4: RandomSeed = 5, Initial Mass = 7.4, Metallicity = 0.0142, Oxygen-Neon_White_Dwarf
5: RandomSeed = 6, Initial Mass = 7.5, Metallicity = 0.0142, Neutron_Star
6: RandomSeed = 7, Initial Mass = 7.6, Metallicity = 0.0142, Neutron_Star

Looks like currently in COMPAS, the stars up to (and including) 7.4 Msol form ONeWDs, whilst the stars 7.5 Msol and above form NSs. 

It seems that this issue is only really an issue with book-keeping (properly calculating and recording the envelope mass), and does not affect the evolution of the stars.

On the EAGB:
- Mc = McHe = McBAGB
- CO core mass increases from 0 for the first time. Should increase to McDU (roughly 1.328 Msol). When it reaches this point, second dredge up happens, reducing Mhe to McDU and we switch to the TPAGB. This all looks correct, as far as I can tell.

On the TPAGB:
- CO core mass, He core mass and core mass are not the same, should be I think (check Hurley 2000). Previously, CalculateHeCoreMassOnPhase was simply doing return m_HeCoreMass; , which meant that m_HeCoreMass was not being updated.
Solution :
CalculateHeCoreMassOnPhase should return CoreMass

This is the bug responsible for the titular issue in #105 (orange curve above blue curve in the plots).

- envelope mass at this stage should be total mass - core mass
Solution :
CalculateEnvelopeMassOnPhase return Mass - CoreMass

This resolves the green curve being above the blue curve in the plots.

- star should be stripped/become WD when Env mass = 0 (or alternatively, where currently core mass > total mass)
Solution :
ResolveEnvelopeMassOnPhase should check when core mass > mass (or equivalently, when envelope mass <= 0)

Previously checked for stuff to do with supernovae (McSN). I wasn't sure if this should remain, so I have removed it.

Peforming the same run of 7.0-7.6 Msol stars as above including these fixes yields:

0: RandomSeed = 1, Initial Mass = 7, Metallicity = 0.0142, Oxygen-Neon_White_Dwarf
1: RandomSeed = 2, Initial Mass = 7.1, Metallicity = 0.0142, Oxygen-Neon_White_Dwarf
2: RandomSeed = 3, Initial Mass = 7.2, Metallicity = 0.0142, Oxygen-Neon_White_Dwarf
3: RandomSeed = 4, Initial Mass = 7.3, Metallicity = 0.0142, Oxygen-Neon_White_Dwarf
4: RandomSeed = 5, Initial Mass = 7.4, Metallicity = 0.0142, Oxygen-Neon_White_Dwarf
5: RandomSeed = 6, Initial Mass = 7.5, Metallicity = 0.0142, Neutron_Star
6: RandomSeed = 7, Initial Mass = 7.6, Metallicity = 0.0142, Neutron_Star


reinhold-willcox:
@SimonStevenson 

Thanks for doing this! I'm not sure I fully follow your comments though.

Your before and after runs appear to be the same, unless I'm missing something. I'm also not clear on what the correct answer should be (i.e what is the minimum single star zams mass at solar metal. that we expect theoretically should become a NS?)

SimonStevenson:
Indeed, them being the same before/after is evidence that this issue was not affecting the evolution of these stars, just the book keeping of their core/envelope masses. Regarding the separation between WD/NS (which is not exactly the point of this issue), I'm also not sure what the "right" answer is, but this seems reasonable (we expect the transition around 8 Msol), and my changes don't affect it. To see the impact of the changes, see the plots in the issue.

SimonStevenson:
@jeffriley version bumped, please check :)

SimonStevenson:
I don't follow your concern @ilyamandel . These are the functions which calculate these quantities.

ilyamandel:
@SimonStevenson -- you replaced 

double          CalculateEnvelopeMassOnPhase(const double p_Tau)                                 { return m_EnvMass; } 

with

double          CalculateEnvelopeMassOnPhase(const double p_Tau)                                 { return m_Mass - std::max(m_COCoreMass, m_HeCoreMass);}       

in TPAGB.h.  Great, I agree this function returns the right quantity -- but it doesn't seem to update m_EnvMass.  So after running this function, can it be that it's output is different from m_EnvMass?  Can you be sure that the if statement check you use in TPAGB.cpp, "(utils::Compare(m_EnvMass, 0.0) <= 0)", is using the correct envelope mass?

ilyamandel:
What if we do away with m_EnvMass altogether?  It seems superfluous since we can always get the correct value by calling CalculateEnvelopeMassOnPhase, so it only makes life needlessly complicated by requiring us to sync another variable which doesn't hold any info that can be readily obtained...

SimonStevenson:
ResolveEnvelopeMassOnPhase in BaseStar is the function that updates EnvMass by calling CalculateEnvelopeMassOnPhase. In particular:
ResolveEnvelopeMassOnPhase(const double p_Tau){ m_EnvMass = CalculateEnvelopeMassOnPhase(p_Tau); }
I pointed this out in the issue.
All I did was fix the implementation of  CalculateEnvelopeMassOnPhase. Previously, it wasn't doing anything, now it does the right thing.

SimonStevenson:
> What if we do away with m_EnvMass altogether? It seems superfluous since we can always get the correct value by calling CalculateEnvelopeMassOnPhase, so it only makes life needlessly complicated by requiring us to sync another variable which doesn't hold any info that can be readily obtained...

This is beside the point. Either way, it still wasn't being calculated properly; it wouldn't matter if this was precomputed, or calculated when needed, its still wrong

ilyamandel:
I agree that you are fixing things, I am just thinking that we should fix it for good once we are at it. :) 

Your TPAGB.h ResolveEnvelopeMassOnPhase will override the definition of ResolveEnvelopeMassOnPhase from the parent class BaseStar.h -- so, unless I am mistaken, you are no longer updating the m_EnvMass variable when calling ResolveEnvelopeMassOnPhase for TPAGB stars.

SimonStevenson:
There is no implementation of ResolveEnvelopeMassOnPhase in TPAGB. I didn't edit one in this PR, so not sure what you are referring to.

I thought TPAGB will use the BaseStar one. I must admit I did get a bit lost in inheritance and had to ask @jeffriley for help on this.

ilyamandel:
You are fine, I got confused between the Resolve... and Calculate... functions.  Spotted my mistake already before your response. :D Mea culpa.

SimonStevenson:
Thanks @ilyamandel . If you still think that we should reconsider how we calculate envelope masses (or whether we should calculate them at all) please feel free to raise a new issue, and we can figure out if there is an alternate proposal that makes more sense than what we are doing now.

SimonStevenson:
@jeffriley no I added it in response to your comment, you're not going crazy :)

Issue #237, opened 2020-04-18
Documentation lags code changes for v02.09.07 - changes to <2ndSN properties


jeffriley:
**Describe the bug**

A change was made to "<2ndSN" properties in v02.09.07 - "Change eccentricity, semi major axis and orbital velocity pre-2nd supernove to just pre-supernova everywhere in the code".  The documentation (specifically the list of properties available for printing and their associated heading strings in Appendix C) still refers to the pre-change properties and heading strings and needs to be updated.

**To Reproduce**
n/a

**Expected behavior**
Documentation to reflect code and output file headers

**Screenshots**

**Versioning (please complete the following information):**
 - OS: Ubuntu 18.04
 - COMPAS v02.09.10, but present since (and including) v02.09.07

SimonStevenson:
This is my fault (I updated the code but not the docs). As penance, I can volunteer to fix this issue.

SimonStevenson:
I noticed a possible procedural issue with this, not sure if I should start an issue: the documentation is stored as .docx and .pdf, which github treats as binary files. I think this means that if someone else had updated the documentation in a commit after the one I branched from, that my commit would wipe out their changes, rather than incorporating them/creating a conflict. I'm not sure what the right way to deal with version control for these file types is; the only solution I can think of is switching to something like a latex file that can be properly version controlled. As usual, thoughts welcome!

jeffriley:
Yeah it kind of grew.  It probably should be LaTex. Is there a MSWord to
LaTex converter?

SimonStevenson:
Has this been resolved now? I think the PR has been merged. We should start a seperate issue for the issue of maintaining the COMPAS_doc file

jeffriley:
> We should start a seperate issue for the issue of maintaining the COMPAS_doc file

Yes, let's do that

SimonStevenson:
I have opened issue #252 to discuss this. Closing this issue.

PR #242, opened 2020-04-20
Update docs to reflect pre2ndSN -> preSN change

SimonStevenson:
hopefully fixes issue #237


ilyamandel:
Silly question, but is there a way to view the changed file via the web page (rather than checking out a copy of Simon's branch)?

SimonStevenson:
There usually is (clicking on a commit will show you the diff side by side). However in this case github treats docx files as binary files, so there is no diff available (see discussion of this problem in the linked issue)

Issue #238, opened 2020-04-18
SSE code samples kicks (since v02.10.01) - add kicks to SSE grid file


jeffriley:
**Is your feature request related to a problem? Please describe.**

No.  @ilyamandel moved the sampling of kicks into the SSE code path so that SSE would sample kicks - previously kicks were only sampled in the BSE code path.  The SSE grid file does not allow kick values to be supplied (c.f. the BSE grid file).

**Describe the solution you'd like**

Add ability for users to supply kick values via the SSE grid file

**Describe alternatives you've considered**

None (other than no action).

**Additional context**

jeffriley:
Added kick functionality to SSE Grid file.  Updated docs.

SimonStevenson:
The associated pull request has now been merged, so closing this issue

PR #250, opened 2020-04-27
Enhancement: Issue #238 add supernova kick functionality to SSE grid file

jeffriley:
(1) Issue  #238 - add supernova kick functionality to SSE grid file (+ updated docs)


jeffriley:
Good call re updating the example file - I'll fix that.

Theta, phi and mean anomaly.  Of course.  Doh.  I just adapted one half of the BSE code and didn't think it through - I will take all three out.

Yes, I think we should have an SSE supernovae file - and yes I think that should be a separate issue.

I'll make the changes to this PR over the weekend.  Thanks...

jeffriley:
Requested changes completed. 

Also fixed defect relating to printing of actual random seed - see change log in constants.h

jeffriley:
Sigh... I think I forgot to update the table.  I'll push an update version of the doc.

jeffriley:
Ok, should be done now - unless I forgot something else...  Thank goodness for reviews :-)

PR #257, opened 2020-05-07
removed COMPASparser.py

reinhold-willcox:
it has been superseded by postProcessing.py


reinhold-willcox:
Thanks @avivajpeyi for bringing this to our attention

reinhold-willcox:
I hadn't realized the parser was still usable, I thought it was a relic from legacy. If you'd prefer to keep it, we can move it into defaults.

jeffriley:
I don't really mind.  I wrote it as an example of how to parse the new
COMPAS output files.  It needs to be updated - things have changed a little
- I could do that but it would be low priority (though it wouldn't take
long).  I have my own copy of course, so if it's not useful to other people
now that the post-processing scripts work with the new output files, I'm
happy for it to go.  If you keep it in defaults I'll update it some time in
the next few weeks.

reinhold-willcox:
I'll go ahead and delete it so there's no confusion. When yours is updated, we can add it back in.

PR #258, opened 2020-05-07
Update supernova outputs

reinhold-willcox:
Fixed 2nd SN printing bug
Updated SN_STATE parameter
Added SN_STATE to COMPAS_Doc.docx
Removed ID from BSE output files


reinhold-willcox:
@jeffriley Sorry, that comment slipped through as one of my notes to self - I'm planning to test that one part later to see if it ever gets raised, and if not if we could get rid of it. 

I have run tests that show that the 2nd supernova is always printed, and the physics was not intended to change in this PR, but I'm happy for others to review as well.

jeffriley:
> I'm planning to test that one part later to see if it ever gets raised, and if not if we could get rid of it.

Well the point of defensive programming in general and this check in particular is to avoid problems if anybody does something they shouldn't (in this case call that function when there is no SN event).  Just because nobody does now doesn't mean nobody will make a mistake in the future.

I'd leave it in - it's not a lot of overhead.  It actually came from the legacy code, so maybe there were times it was being called erroneously, but it has another use: it also allows us to use the function in a "if there is a supernova then resolve it" mode (as the BaseStar equivalent is used) as well as "I know there was a supernova, resolve it" mode, so it is a useful check.

jeffriley:
> I have run tests that show that the 2nd supernova is always printed, and the physics was not intended to change in this PR, but I'm happy for others to review as well.

But since you've changed the flow of the code (you removed a call to EvaluateSupernova() because it wasn't called in the tests you did), do we know that large runs over a wider set of parameters than your tests with these changes produce the expected results - the same results as previous large runs (except for the printing of the 2nd supernova)?

Presumably the call to EvaluateSupernova() was there (in the legacy code) for a reason (though I know that doesn't mean it should have been there) - my concern is that even though you didn't see it called in your tests, larger tests might actually reveal that it was called and that we should re-evaluate the code flow in that function.

reinhold-willcox:
I agree with your points. I will not remove the check, but I might expand it to output an error if this happens (so that we can find the bug sooner).

Regarding times when the other EvaluateSupernova might get called, this may definitely happen in some corner of the parameter space that I didn't test (we can never be 100% sure). But the new implementation is physically justifiable, since any SN _should_ get picked up here. Leaving the original in is more of a band-aid fix, and leaves the code more disorganized and harder to follow. I definitely think we should run expanded tests, but if those fail we can use them to identify new bugs.

jeffriley:
Yep, agree an error message is a good idea.

I have some thoughts about the removal of the EvaluateSupernovae().  I think it's fine to remove it as you have done, but I think the fact that it is never called helps us identify the bug in the legacy code that led to the inclusion of the second call (the one you're leaving in).  

I think putting some thought into why, after adding the second call (in the legacy code) :

(a) the problem (that initiated the addition of the second call - it was put there for a reason) apparently went away, and
(b) the first call was no longer called

might help to identify what the actual code path should be.  Maybe that is how you have it now, but it seems to me that adding the second call was the bandaid and masked the underlying problem.  Removing the first call leaves the bandaid in place - maybe we can indentify the underlying problem and fix that.

It seems to me that by (pseudo?) advancing the timestep then calling EvaluateSupernovae() the second time (in the legacy code) picked up SN events that were not being picked up on the first call (because of mass changes perhaps - don't know), but it also picked up SN events that would have been picked up by the first call to EvaluateSupernovae() in the next call to EvaluateBinary() (or whatever the legacy equivalent is) - which is why the first call is not picking anything up.  I think.  Maybe.  I think it's worth thinking about (by people other than just me because I don't have the depth of physics knowledge here) to see if we can identify what the original problem was, and so make sure it isn't present in the new code.

jeffriley:
I just ran a test with the legacy code.  For 1000 binaries using pythonSubmitdefault.py:

The legacy equivalent to the first call to EvaluateSupernovae() was called 356 times (328 times for which_star = 1, 28 times for which_star = 2, and 0 times for which_star = 3).

The legacy equivalent to the second call to EvaluateSupernovae() was called 35 times (21 times for which_star = 1, 14 times for which_star = 2, and 0 times for which_star = 3).

That plus your tests that show the first call to EvaluateSupernovae() is not being exercised in the new code suggests to me that:

(a) there may be a problem with the porting of the legacy code to the new code - we should try to ascertain why that first call is not being exercised (maybe the new code fixes the original problem with the legacy code, but I don't think that should be our first assumption).

(b) there really was an underlying problem in the legacy code - or at least an underlying reason for the second call to EvaluateSupernovae() to be there.  We should try to figure out what that is - I believe the second call was added to the legacy code because some SN events were being missed (per Simon).

I don't think simply removing the first call to EvaluateSupernovae() is the best solution here without doing some more investigation.  I think we should try to understand what our code is doing, and why, before making this change - otherwise how can we be sure that it is the right change?

reinhold-willcox:
@jeffriley 

I finally sorted out where the problem lies in the 2 different calls to EvaluateSupernovae(). 

In BaseBinaryStar::EvaluateBinary(), the IsSupernova flag is raised in the UpdateAttributes function (more precisely, it's in a chain of function calls that, when appropriate, includes GiantBranch::ResolveSupernova(), which calls IsSupernova()). The supernova is flagged correctly for both SN events, and ResolveSupernova() immediately updates the stellar type of the star to the remnant type, so the EvaluateBinary loop will end with either one or both compact objects before returning to BaseBinaryStar::Evolve().

When EvaluateSupernovae() is placed after the UpdateAttributes() call, then the Supernova is evaluated before the end of the loop, but when it is placed between ResolveCommonEnvelopeEvent() and ResolveMassChanges(), it has to wait for the next loop. For the case of the second supernova, the previous loop closes out with 2 compact objects, which BaseBinaryStar::Evolve() flags as a DCO, and stops the simulation.

The easiest solution I can think of is to keep the EvaluateSupernovae() function after UpdateAttributes(). Otherwise, we'd have to add into DCO some sort of check for whether both Supernovae have been resolved.

jeffriley:
Good job @reinhold-willcox - chasing that down can't have been straight-forward.  I figured the second call was soaking up the SNs and so depriving the first call of anything to do - that all makes sense.  What worries me is that wasn't (and isn't) happening in the legacy code - that first call was (is) being exercised.  However, you've done enough work chasing down what's happening to understand if it's right, so I'm happy to go with it.  Do we know if making this change causes any of the formation rates for DCOs to change?

reinhold-willcox:
I'm not sure, I haven't done any checks on population changes.

jeffriley:
Ok.  I'm happy with the coding changes, but I'd like one of the other reviewers to look over the changes in the code flow.  The new code should be (until we start changing it significantly) a copy of the legacy code - though there will be changes simply because of the new architecture.  The legacy code uses both of those calls to EvaluateSupernovae() and I'd just like to be sure that I didn't break something when I ported the code that causes the first not to be exercised in the new code.

reinhold-willcox:
Thanks Jeff!

This pull request is superceeded by #263. I'll add a comment to the other one that describes the changes first implemented here, and close this one.

PR #263, opened 2020-05-20
Ns kicks

reinhold-willcox:
- Added in vector3d.h/.cpp and functionality for vector addition of velocities from both SNe (for more accurate system and component velocities).
- Fixed 2nd SN printing bug
- Updated SN_STATE parameter
- Added SN_STATE to COMPAS_Doc.docx
- Removed ID from BSE output files


reinhold-willcox:
@jeffriley all good points, I've addressed them in the latest commits. 

I originally changed to ResolveSupernovaInBinary() because I was getting confused about which ResolveSupernova()  was supposed to be called where. They're in different classes, but if you don't know the class hierarchy it can become unclear which function is getting called by the children classes. But the return type makes them clearly distinguishable, so I changed it back.

reinhold-willcox:
This pull request supersedes #258, so that one will be closed.

For clarity, the other pull request addressed the following:
- Fixed 2nd SN printing bug
- Updated SN_STATE parameter
- Added SN_STATE to COMPAS_Doc.docx
- Removed ID from BSE output files

The final comments from jeff were that the code changes were fine but physics needs to be tested (same as this current PR).

src/BaseBinaryStar.cpp
```diff
@@ -1390,366 +1390,388 @@ void BaseBinaryStar::ResolveCoalescence() {
...
- * Simplifies to Brandt & Podsiadlowski 1995 (http://arxiv.org/abs/astro-ph/9412023), eq 2.8 when e = 0, beta = pi/2
- * Also given in Hurley et al 2002 (http://arxiv.org/pdf/astro-ph/0201220v1.pdf), eq A.12
+ * Note: the systemic speed is only valid for intact binaries, and component speeds are only valid for disrupted binaries.
+ * 
+ *  /////////////////////////////////
+ *  // Logic
+ *  // 
+ *  // If (Unbound before SN):
+ *  //
+ *  //         Must be 2nd SN, only need to update starSN component velocity (rotated into previous reference frame).
```
avigna, line 1409: Why must it be 2nd SN? Couldn't it be disrupted in the first one?

src/BaseBinaryStar.cpp
```diff
@@ -1390,366 +1390,388 @@ void BaseBinaryStar::ResolveCoalescence() {
...
+ *  //        Update binary systemic velocity (even if disrupted, just for consistency) - rotate into previous reference frame if needed.
+ *  // 
+ *  //        If now unbound:
+ *  //
+ *  //                Set m_Unbound to True - should be the only place in the code this is done.
+ *  //
+ *  //                Continue vector algebra to find v1inf and v2inf.
+ *  //                Add these values to previous component velocities (rotated if need be) which will be the systemic velocity if this is the 2nd SN. 
+ *  //
+ *  //                For unbound binary, new Euler Angles should be randomized.
```
avigna: Maybe point out to Euler Angles in the `vector3d.cpp` function.

src/BaseBinaryStar.cpp
```diff
@@ -1390,366 +1390,388 @@ void BaseBinaryStar::ResolveCoalescence() {
...
- *
- * bool ResolveSupernova()
- *
- * @return                                      True if a supernova event occurred, otherwise false
- */
-bool BaseBinaryStar::ResolveSupernova() {
+        m_Supernova->UpdateComponentVelocity( natalKickVector.RotateVector(m_ThetaE, m_PhiE, m_PsiE));    // yes - only need to update the velocity of the star undergoing SN
 
-    if (!m_Supernova->IsSNevent()) return false;                                                                                    // not a supernova event - bail out (or bale out depending whence you hail...) passively
+        // The quantities below are mostly meaningless
```
avigna, line 1466: For an unbound binary my "Legacy approach" would be to set the values for a particular number, such as the NEVER_SET -7.333 from  Legacy. It doesn't seem to appear in Legacy, which makes me believe there is a reason for that.

src/BaseBinaryStar.cpp
```diff
@@ -1390,366 +1390,388 @@ void BaseBinaryStar::ResolveCoalescence() {
...
+        double m1 = m_Supernova->MassPrev();                             // Mo  - SN star pre-SN mass
+        double m2 = m_Companion->MassPrev();                             // Mo  - CP star pre-SN mass
+        double mb = m1 + m2;                                             // Mo  - Total binary pre-SN mass
+        
+        // Functions of Eccentric Anomaly
+        m_Supernova->CalculateSNAnomalies(e);
+        double cosEA = cos(m_Supernova->SN_EccentricAnomaly());        
+        double sinEA = sin(m_Supernova->SN_EccentricAnomaly());
+
+        double omega = sqrt(G*mb /(a*a*a));                              // orbits/s  - Keplerian orbital frequency
```
avigna, line 1506: orbits/s -> rad/s?

src/BaseBinaryStar.cpp
```diff
@@ -1390,366 +1390,388 @@ void BaseBinaryStar::ResolveCoalescence() {
...
-                                                             totalMass,
-                                                             totalMassPrime,
-                                                             m_Supernova->SN_Theta(),
-                                                             m_Supernova->SN_Phi());
-        m_SystemicVelocity /= KM;                                                                                                   // convert to km s^-1
-    }
-    else {                                                                                                                          // no longer bound
-        m_Unbound = true;
-    }
+            // If N = H x H_ is not well-defined, need to account for degeneracy between eccentricities
```
avigna, line 1621: Why wouldn't it be well defined? Maybe explain or put a link to some literature at some point in the code.

src/BaseBinaryStar.cpp
```diff
@@ -1390,366 +1390,388 @@ void BaseBinaryStar::ResolveCoalescence() {
...
+    // Do for all systems 
 
-    m_EPrime            = m_Supernova->PostSNeOrbitalEnergy();
+    // Set Prime values
+    m_EccentricityPrime = m_EccentricityPostSN;
+    m_SemiMajorAxisPrime = m_SemiMajorAxisPostSN;
 
-    m_Eccentricity      = ePrime;
-    m_EccentricityPrime = ePrime;
+    double totalMassPrime = m_Supernova->MassPrev() + m_Companion->MassPrev();                                            // Total Mass preSN
```
avigna, line 1692: Why are you calculating prev hhere? Should't it just be: `m_Supernova->Mass()`?

src/BaseBinaryStar.cpp
```diff
@@ -1390,366 +1390,388 @@ void BaseBinaryStar::ResolveCoalescence() {
...
-    if (m_Star1->IsSNevent() && (p_Resolve2ndSN || (utils::Compare(aPrime, 0.0) > 0))) {                                // star1 supernova
-        m_SupernovaState = SN_STATE::STAR1;                                                                             // star1
+    // Resolve the SN(e)
+    if (m_Star1->IsSNevent()) { 														                                // star1 supernova
 
         // resolve star1 supernova
         m_Supernova = m_Star1;                                                                                          // supernova
         m_Companion = m_Star2;                                                                                          // companion
-        (void)ResolveSupernova();                                                                                       // resolve supernova
+        (void)ResolveSupernova();                                                                               // resolve supernova
```
avigna: Why is void in parenthesis and what does that means?

src/constants.h
```diff
@@ -452,8 +460,10 @@ extern OBJECT_ID globalObjectId;
...
 
-constexpr double FLOAT_TOLERANCE_ABSOLUTE               = 0.0000005;                                                // Absolute tolerance for floating-point comparisons if COMPARE_WITH_TOLERANCE is defined
-constexpr double FLOAT_TOLERANCE_RELATIVE               = 0.0000005;                                                // Relative tolerance for floating-point comparisons if COMPARE_WITH_TOLERANCE is defined
+//constexpr double FLOAT_TOLERANCE_ABSOLUTE               = 0.0000005;                                                // Absolute tolerance for floating-point comparisons if COMPARE_WITH_TOLERANCE is defined
+//constexpr double FLOAT_TOLERANCE_RELATIVE               = 0.0000005;                                                // Relative tolerance for floating-point comparisons if COMPARE_WITH_TOLERANCE is defined
+constexpr double FLOAT_TOLERANCE_ABSOLUTE               = 0.0000000;                                                // Absolute tolerance for floating-point comparisons if COMPARE_WITH_TOLERANCE is defined
```
avigna: Is this right? No tolerance there...

src/typedefs.h
```diff
@@ -92,9 +92,9 @@ typedef struct SupernovaDetails {                           // Holds attributes,
     double           kickVelocity;                          // Kick velocity the system received during the supernova (km s^-1)
     double           kickVelocityRandom;                    // Random number U(0,1) for choosing the supernova kick velocity magnitude - drawn once at star creation
     double           meanAnomaly;                           // Mean anomaly at instantaneous time of the SN - uniform in [0, 2pi]
-    double           phi;                                   // Angle between 'x' and 'y', both in the orbital plane of supernovae vector (rad)
-    SN_STATE         supernovaState;                        // indicates which star (or stars) are undergoing / hove undergone a supernova event
-    double           theta;                                 // Angle between the orbital plane and the 'z' axis of supernovae vector (rad)
+    double           phi;                                   // Kick angle in the orbital plane, defined CCW from the radial vector pointed away from the Companion (rad) [0, 2pi)
```
avigna, line 95: I'd maybe call them `kick_angle_phi` and `kick_angle_psi` or so, but whatever.

src/vector3d.cpp
```diff
@@ -0,0 +1,288 @@
...
+void Vector3d::operator =( Vector3d vNew ) {
+    updateVector( vNew[0], vNew[1], vNew[2] );
+}
+
+// Overload vector addition operator
+Vector3d Vector3d::operator +(Vector3d vec) {
+    Vector3d newVect;
+
+    for (int i=0; i<3; i++) {
+        newVect[i] = (*this)[i] + vec[i];
```
avigna, line 75: What/where does (*this) points to/at?

src/vector3d.cpp
```diff
@@ -0,0 +1,288 @@
...
+     * @param   [IN]   b                            second vector
+     * @return                                      angle between them, in radians
+     */
+    double angleBetween(const Vector3d& a, const Vector3d& b) {
+        // Angle between 2 vectors, between [0, 2pi] 
+        return acos(linalg::dot(a,b));
+    }
+
+}
+
```
avigna, line 288: Did you do sanity checks for these linear algebra functions? If so, good. If not, please do.

jeffriley:
Don't forget the code changes for pr#258 (superseded by this and now closed).  I think someone (@ilyamandel, @avigna  or @SimonStevenson) should at least take a quick look...

avigna:
Hi Jeff... you mean a quick look to pr#258? I don't understand why would it be closed if no one checked it out or why would there be changes that I need to check there that are not here. Summary: I am confused.

src/BaseBinaryStar.cpp
reinhold-willcox: This first if statement tests if the system is unbound before the SN, which can only happen if it's the 2nd SN.

jeffriley:
Hi Alejandro,

The changes for PR#258 got rolled up into this PR, so I don't think anybody but me looked at the changes.  The code changes are fine, I just have a feeling there's it's not quite right - there was what looks like a kludge in place in the legacy code and I'm just not sure the underlying problem has been fixed.  Maybe just have a read of the comments for PR#258

src/vector3d.cpp
reinhold-willcox: It's a c++ thing that refers to the specific instance of the object, similar to `self` in python. Jeff probably can explain it better if you're curious. Here, it just means that if I have a Vector3d + Vector3d, it adds them in the expected way.

src/BaseBinaryStar.cpp
jeffriley: @avigna , '(void)' here tells the compiler to ignore and discard any return value from the ResolveSupernova() function.  It's not strictly necessary (the compiler will ignore it anyway if you don't do something like 'x = ResolveSupernova()', but explicitly adding the '(void)' tells readers of the code that you didn't accidentally forget - you really do just want to ignore (and discard) the return value of the function (just so someone looking at the code doesn't think there might be a defect there).

src/vector3d.cpp
jeffriley: 'this' is a pointer to the object - I assume here the Vector3d object (I don't remember the code)

jeffriley: Oh - now I understand the question :-)  We're in an overload operator - 'this' points to the subect of the operator.  So here the '+' operator for the vector3d class is declared as:

Vector3d Vector3d::operator +(Vector3d vec)

so it would be used like a normal '+' in the code e.g. a = b + c, and
'this' refers to 'b', and 'vec' refers to 'c'  (and 'newVec' is what 'a' will get set to...

src/BaseBinaryStar.cpp
reinhold-willcox: Could you clarify: are you asking me if this is defined correctly, or suggesting that I convert to different units?

src/BaseBinaryStar.cpp
reinhold-willcox: I'm not sure what legacy would choose for these, but this section sets binary parameters for already disrupted binaries, so they don't mean much physically. Orbital velocity (pre and post) and dimensionless kick don't have a meaning, so I set them to NaN, but I keep the e>1 and a<0 values since it's clear by eye that these apply to disrupted binaries. 

I doubt this will impact anything else, but if it does, I'd rather it throw an error so we can properly debug it.

src/BaseBinaryStar.cpp
reinhold-willcox: If H is parallel or anti-parallel to H_ (H prime), then the cross product is not well-defined (e.g if the binary orbital plane is the same, or completely inverted, during the SN).

The degeneracy between eccentricities refers to the special case where the cross product isn't well-defined AND both ecc<SN and ecc>SN are >0. Then the individual Euler angles phi and psi are not defined (since they are defined relative to the N vector), but either the sum or difference is (for parallel and anti-parallel respectively). 

I don't know any literature that refers to this, it's not really an open problem. It might be a good idea to put this into the COMPAS methods paper, so that it's clear (to us all of us and everyone else) what's happening.

src/constants.h
reinhold-willcox: Ah good catch, this should probably be removed. We actually don't have tolerance implemented right now, so I had these turned on for testing while I built up the ResolveSupernova() function. 

This is something @ilyamandel , @jeffriley and I have discussed, particularly what is the correct way to apply a tolerance. If I remember correctly, Ilya had a solution which handled relative and absolute tolerance simultaneously, but we haven't coded this up yet. We also wanted to start a discussion with the group on what values to choose for the tolerances. 

I'll remove these lines and push back up.

reinhold-willcox:
@avigna I closed the PR when I pushed up this one - the same changes are incorporated in this PR (they were related to SNe), and it was cleaner to push all at once.

src/BaseBinaryStar.cpp
reinhold-willcox: The void means ignore whatever is returned from the function. In this case, ResolveSupernova returns a boolean, but running the function by itself would leave a line which is just True/False and I think would throw an error. 

I'll leave it to @jeffriley to comment on why we need ResolveSupernova to be a boolean though. In principle, the function could return a void and I think it would all still work. 

Edit: sorry jeff, I hadn't seen your response when I posted mine

jeffriley: It returns true if a SN event actually occurred, false if not (see the first IF statement in the function) - so the caller can check if anything actually changed during the call.  I think that's a hangover from the legacy code.

src/typedefs.h
reinhold-willcox: Looking through where these are called in the code, they always appear as either KickParameters.theta or SupernovaDetails.theta (or phi, ...), so I think these are safe to leave as is.

src/vector3d.cpp
reinhold-willcox: Yes, all checks passed.

ilyamandel:
@jeffriley , @reinhold-willcox , @avigna -- I am happy to check particular bits of code for "physics", as Jeff suggests, but I doubt that I would be able to spot an issue in the code flow that you have not spotted (I am now referring to the appropriate placement of EvaluateSupernovae(), for example).  Are there specific things I should check?  Otherwise, I agree with @jeffriley that the most sensible approach here is a population check and comparison, perhaps even a 3-way comparison between Reinhold's code, a previous version of new COMPAS, and Legacy.

ilyamandel:
@jeffriley , @reinhold-willcox , @avigna -- see my latest comment / query.

avigna:
A population run sounds reasonable. Not sure how to practically proceed with it (i.e. which particular population to target the study to, so it doesn't become a very massive/difficult text). One test I can suggest is checking for DNSs and plot the P-e diagram; that could be compared to results from Vigna-Gomez+2018 (where kicks were also varied).

It seem also that there are some conflicts "that must be resolved"? @reinhold-willcox

reinhold-willcox:
I think a population of any size would work - the only physical change should be the final system velocities, but all other attributes should be identical. This doesn't touch the kick prescriptions, so the P-e plot might be really useful for this.

As far as the conflicts to be resolved - I think I responded to all of them. If you are happy with the responses, you can mark them as "resolved" at the bottom of the chats. If not, let me know if I need to clarify/fix anything.

The reason this PR is still outstanding is that we haven't run the tests (individual or population). It's a bit a bit of a low priority, but maybe we can get to this at next week's hackathon.

avigna:
I have approved the changes I asked for. Sorry about the delay.

reinhold-willcox:
@ilyamandel I've updated to the latest changes from dev, and run a 10 systems comparison between ns_kicks and dev.

Unfortunately, there are some glaring issues that I think need to be addressed first.

Here are my notes:
*Minor things*
- There are some minor differences due to differing variable names or meanings (e.g SN type), or new variables (ComponentSpeed)
- There are some minor differences in values of some parameters, most likely due to random number generation differences (if the generator gets an extra call somewhere)
- The systemic_speed (or _velocity) parameter is 0 for all of `dev`, but this is likely because the parameter is only set for intact binaries, and none of the binaries in my test run were intact for `dev`.

*Major things*
- None of the binaries from `dev` produced a 2nd supernova, even though the evolve_unbound flag is True. This is not correct. `ns_kicks` does this, but its hard to say if it does it correctly (there is nothing to compare to).
- There are some larger discrepancies between the variables `Separation>SN` and `Eccentricity>SN`, which 
 may or may not be due to the previous random number generation differences. These parameters are functions of the previous ones, but the differences are substantial enough that there could be errors in the functions. This may be a bit of a headache to sort out.
- In one particular case (out of 10), a Supernova resulted in an unbound binary in `dev` and a bound binary in `ns_kicks`. This is very bad. Almost definitely related to the issue with `Separation>SN` and `Eccentricity>SN`, and my guess is due to an error in calculation somewhere, but I'll have to confirm this.

Issue #264, opened 2020-06-02
Mass transfer: irregular stripping and evolution


avigna:
**Describe the bug**
During a mass transfer episode, the total mass of the system drops during one timestep and then increases on the next timestep. This seems to do with the donor giving away the envelope mass in the first timestep, without affecting the accretor or the orbit. Then, in the next timestep, it seems the accretor and orbit are solved as if part of the mass transfer episode. I only tested this in a non-conservative mass transfer episode, so the total mass dropped and then increased a bit, but not back to what the pre-mass-transfer value was. The behaviour in the HR diagram is also odd, as stripping episodes lead to a vertical drop in luminosity and only posterior evolution towards the Helium tracks.

**Label the issue**
`urgent`
`severity_moderate` - This is a moderately severe bug

**To Reproduce**
1. Run the pythonScript attached in the next message and run it with the COMPAS version specified.
2. Using `detailed_output = True` verify **any** case B or case C mass transfer episode.

**Expected behavior**
Total mass should never increase in COMPAS and both donor, accretor and orbit should be solved in the same timestep. The HR diagram shouldn't have that odd drop and go straight from post-main-sequence to Helium star (during case B and C).

**Screenshots**
Total evolution and zoomed in evolution for the same system as in the pythonSubmit attached.
![REFACTORED_1766_ZOOM](https://user-images.githubusercontent.com/22151352/83497748-31d36a80-a4bb-11ea-932f-f627d7f23c9c.png)
![REFACTORED_1766](https://user-images.githubusercontent.com/22151352/83497754-3435c480-a4bb-11ea-83af-b57c6443aa0f.png)

**Versioning (please complete the following information):**
 - OS: macOS Catalina
 - COMPAS [v02.11.01]

**Additional context**
- I think this applies to any mass transfer episode, but didn't have time to test many cases. 
- I did a small run with COMPAS_Legacy to double check this issue was not present there, and it wasn't. It seems like a COMPAS_refactored issue. 
- Note the IMF in the pythonSubmit has masses above 20 Msol, so would be good to check with lower mass systems too.

avigna:
Apparently I can't straightforwardly add python files here. I will therefore paste the code. Apologies.

```
import numpy as np
import subprocess
import sys
import os
import pickle
import itertools 
from subprocess import call

# Check if we are using python 3
python_version = sys.version_info[0]
print("python_version =", python_version)

class pythonProgramOptions:
    """
[...]

PR #268, opened 2020-06-04
Mass transfer bug

reinhold-willcox:
Fixes the mass transfer bug in detailed output, issue #264.

Note that detailed printing is done by default in every timestep, so we could delete the new m_PrintExtraDetailedOutput flag and it's functionality. The benefit to keeping it is that timesteps which invoke big changes (e.g mass transfer) will have 2 lines in detailed output, and plots of these events will show vertical changes in these parameters (as opposed to having slopes which may appear to be physical but are only artefacts of the timestepping)


reinhold-willcox:
@ilyamandel my bad, all fixed up now.

Issue #275, opened 2020-06-25
Previous stellar type sometimes not set when stellar type is changed


jeffriley:
**Describe the bug**

The stellar type of a star is usually only modified in the Star::EvolveOneTimestep() function - as it should be.  If the stellar type is changed in Star::EvolveOneTimestep() the stellar variable m_StellarType is changed to the new stellar type - and importantly the the stellar variable m_StellarType_Prev has already been set to the current stellar type (at the start of the timestep).

There is one other place in the code the stellar type is changed - it probably shouldn't be: I don't think changing the stellar type of a star mid-timestep is valid, but it is as it is for the moment (I think longer-term we should figure out how to remove the change of stellar type mid-timestep). The problem (for now) is that when the stellar type is changed mid-timestep, the previous stellar type is not also changed. This was notived becuase in a test a HeMS star switched mid-timestep to a HeHG star, then the HeHG star went SN - but since the previous stellar type was still HeMS the output files recorded that a star that went SN has a previous stellar type of HeMS, which was wrong.

The fix is to update the stellar variable m_StellarType_Prev when the stellar type is changed mid-timestep.  That will fix the immediate problem (of having what looks like an HeMS star go SN), but it changed the meaning of (one of) the '_Prev' variables - they are supposed to be the values at the previous timestep.

**Label the issue**
`urgency_low`      - This issue is not urgent
`severity_minor`    - This is a minor bug with minimal impact

**To Reproduce**
Steps to reproduce the behavior):
1. Run COMPAS using the attached python submit file - it will produce 1 binary
2. Look in the supernovae output file
3. Should see Stellar_Type_Prev_SN = 7 - that's not correct

**Expected behavior**
1. Run COMPAS using the attached python submit file - it will produce 1 binary
2. Look in the supernovae output file
3. Should see Stellar_Type_Prev_SN = 8 - that's correct

**Versioning (please complete the following information):**
 - OS: Ubuntu 18.04
 - COMPAS v02.11.03

jeffriley:
[pythonSubmit_STPrevIssue.py.txt](https://github.com/TeamCOMPAS/COMPAS/files/4829307/pythonSubmit_STPrevIssue.py.txt)

(might have to remove .txt suffix first... git wouldn't let me upload a .py file)

jeffriley:
Fixed in v02.11.04

Issue #266, opened 2020-06-02
KROUPA IMF returns incorrect value for mass


jeffriley:
**Describe the bug**
KROUPA IMF returns incorrect value for mass when initial_mass_min < 0.08 (first break) and initial_mass_max > 0.5 (second break).  Test of 10,000,000 stars created yields total mass too low by a factor of 10.

This does not affect the default in pythonSubmitDefault.py (initial_mass_min = 5.0 and initial_mass_max = 150.0).

**Label the issue**
urgency_low - This issue is not urgent
severity_minor - This is a minor bug with minimal impact

**To Reproduce**
1. Set the following values in the pythonSubmit file:

    initial_mass_function = 'KROUPA'
    initial_mass_min = 0.0001
    initial_mass_max = 150.0

2. Run COMPAS
3. Check total mass created

**Expected behavior**
KROUPA IMF returns correct value for mass for all input values

**Versioning (please complete the following information):**
 - Ubuntu 18.04
 - COMPAS v02.11.01

**Suggested fix**

Change line 1031 of BaseBinaryStar.cpp from:

                    double term3 = ONE_OVER_KROUPA_POWER_3_PLUS1 * KROUPA_BREAK_1_POWER_1_2 * KROUPA_BREAK_2_POWER_2_3 * (pow(OPTIONS->InitialMassFunctionMax(), KROUPA_BREAK_2_POWER_2_3) - KROUPA_BREAK_2_PLUS1_3);

to

                    double term3 = ONE_OVER_KROUPA_POWER_3_PLUS1 * KROUPA_BREAK_1_POWER_1_2 * KROUPA_BREAK_2_POWER_2_3 * (pow(OPTIONS->InitialMassFunctionMax(), KROUPA_POWER_PLUS1_3) - KROUPA_BREAK_2_PLUS1_3);

jeffriley:
Fixed in v02.11.04

Issue #260, opened 2020-05-08
When using a grid file, ZAMS values not set when equilibrating masses and circularising a binary at birth


jeffriley:
**Describe the bug**
When either star of a binary undergoes RLOF at birth and the allow-rlof-at-birth program option is true we equilibrate the masses of the stars and and circularise the binary, then we recalculate other attributes of the stars (radius, luminosity etc.)based on the new masses.  When the initial parameters of the binaries are supplied by the user via a grid file we do not reset the ZAMS values for the attributes of the stars to the newly calculated values - we should.

**Label the issue**
urgency_low - This issue is not urgent
severity_minor - This is a minor bug with minimal impact

**To Reproduce**

Use a grid file with the following values:
Mass_1,Mass_2,Metallicity_1,Metallicity_2,Separation,Eccentricity
5.01898,0.131986,7.94E-04,7.94E-04,0.01198,0.0

and run COMPAS.  The detailed output file will show that the masses were equilibrated and the binary circularised, but that the ZAMS values were not reset.

**Expected behavior**
When using a grid file, ZAMS values reset when equilibrating masses and circularising a binary at birth

**Screenshots**
n/a

**Versioning (please complete the following information):**
 - OS: Ubuntu 18.04
 - COMPAS v02.11.00

**Additional context**
n/a

jeffriley:
Fixed in v02.11.04

PR #276, opened 2020-06-25
Defect repairs: Issues #260, #266 & #275

jeffriley:
(a) Issue #260 - Corrected recalculation of ZAMS values after eqilibration and cicularisation at birth when using grid files
(b) Issue #266 - Corrected calculation in BaseBinaryStar::SampleInitialMassDistribution() for KROUPA IMF distribution
(c) Issue #275 - Previous stellar type not set when stellar type is switched mid-timestep - now fixed


jeffriley:
Tested with a run of 10,000 binaries - no surprises.  Fix for #260 was also used in 12 million binary run for CHE tests.

Fixed comment in constants.h (talk about eagle eyes...)

PR #282, opened 2020-06-29
Fixing issue 277

ilyamandel:
moved UpdateAttributesAndAgeOneTimestepPreamble() to after ResolveSupernova() to avoid inconsistency

### DIFF
--- a/src/BaseStar.cpp
+++ b/src/BaseStar.cpp
@@ -3392,14 +3392,16 @@ STELLAR_TYPE BaseStar::UpdateAttributesAndAgeOneTimestep(const double p_DeltaMas
     STELLAR_TYPE stellarType = m_StellarType;                                               // default is no change
 
 
-    UpdateAttributesAndAgeOneTimestepPreamble(p_DeltaMass, p_DeltaMass0, p_DeltaTime);      // apply mass changes and save current values if required
-
     if (ShouldBeMasslessRemnant()) {                                                        // ALEJANDRO - 02/12/2016 - Attempt to fix updating the star if it lost all of its mass
-        stellarType = STELLAR_TYPE::MASSLESS_REMNANT;                                       // JR: should also pik up already massless remnant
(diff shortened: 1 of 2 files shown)


reinhold-willcox:
@ilyamandel I don't think this fully solves all the problems. 

Using the pythonsubmit (with a particular trouble seed) and demoplotter posted below (for detailed output), this system still undergoes SN the same timestep that it becomes a HeHG star. I think this is because the CO core mass is set wrong when it changes type, and the incorrect core mass is so high that it triggers the SN early. 

ilyaPR

This is a zoom in of the SN - you can see in the bottom right that the star spends no time as a HeHG star, and in the top left that the CO Core mass jumps from 0 to the total mass of the star (18Mo). In the detailed output file, there are 3 lines that correspond to this SN timestep (Time = 4.009...). I think this is an issue with how we print multiple detailed outputs in one timestep, since some of the Age parameters seem to increase even though the Time parameter does not. Here you can check that the CO core mass jumps up to 14Mo, followed by 18Mo. I'm not sure why it jumps up twice, so this needs investigation. 

Ultimately, I think the long term fix is to make sure we're only changing mass and stellar type once, at the end of each tiemstep.

(Note that In GiantBranch.cpp, I've commented the lines at the very bottom where Simon asks "Why reset things to 0", since this was not working properly either. The values were reset immediately after, so this just added a confusing vertical spike to the plots).

[pythonSubmitMMsolar.py.txt](https://github.com/TeamCOMPAS/COMPAS/files/4843512/pythonSubmitMMsolar.py.txt)
[demo_single_sys_plotter.py.txt](https://github.com/TeamCOMPAS/COMPAS/files/4843513/demo_single_sys_plotter.py.txt)

ilyamandel:
Very interesting.  I agree this is worth investigating further, but I think this is a separate issue, and is probably an issue with the choice of way/how we output information rather than a lack of internal self-consistency.  At first glance, I do not expect this to lead to incorrect remnant mass and remnant kick calculations.

PR #292, opened 2020-07-18
Starting to clean up mass transfer functionality

ilyamandel:
Slowly starting to clean up mass transfer treatment.  For now, no actual changes in behaviour (tested with regression testing), small changes in variable names that require a tweak in the pythonSubmit function (or command-line calls).  The goal is to continue cleaning to make mass transfer transparent and consistent with the OOP paradigm, to allow for easier implementation of future changes.


jeffriley (approved):
All good.  Did you mean to add a second "both" to the comment on line 2444 in BinaryBaseStar.cpp?

ilyamandel:
Thanks for the catch, @jeffriley -- fixed!

PR #294, opened 2020-07-23
Change to thermal timescale MT for both donor and accretor to determine MT efficiency 

ilyamandel:
This PR changes the estimate of the MT efficiency (ultimately used to measure MT stability) to always use the thermal timescale of the donor and accretor within the Hurley prescription.  A moderate size run shows identical outputs to the previous prescription, so the only impact (except for possible rare cases) is likely to be a slightly cleaner code that avoids an unnecessary special case for Case C MT.


jeffriley (approved):
Approved

PR #296, opened 2020-07-23
Preparation for alternative envelope state prescriptions

ilyamandel:
This PR introduces a new ENVELOPE_STATE_PRESCRIPTION to deal with different prescriptions for convective vs. radiative envelopes.  It also removes the unused COMMON_ENVELOPE_PRESCRIPTION.  

There is plenty of code cleanup, but no actual changes in behaviour as long as ENVELOPE_STATE_PRESCRIPTION::LEGACY is chosen, which is set as the new default.  A diff between pre and post change runs indicates no changes in behaviour, as expected, with the exception of circularisation and synchronisation timescales -- the latter change in some cases because this PR assumes that the convective or radiative nature of the envelope is computed consistently for each star.  The synchronisation and circularisation timescale calculation will need to be checked/updated separately.

### DIFF
--- a/defaults/pythonSubmitDefault.py
+++ b/defaults/pythonSubmitDefault.py
@@ -89,6 +89,7 @@ class pythonProgramOptions:
     common_envelope_mass_accretion_prescription = 'ZERO'
     common_envelope_mass_accretion_min = 0.04           # For 'MACLEOD+2014' [Msol]
     common_envelope_mass_accretion_max = 0.10           # For 'MACLEOD+2014' [Msol]
+    envelope_state_prescription = 'LEGACY'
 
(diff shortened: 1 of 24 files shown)


jeffriley (approved):
Approved, but...

We really need to document changes we are making - not just in the change log, but in the documentation.  What is the "LEGACY" prescription?  "HURLEY"?  "FIXED_TEMPERATURE"?  How would a new COMPAS user know?  Were all the options ever documented anywhere?  We need to bring our documentation up-to-date wrt to options (amongst other things).

(This is a general comment - not specifically aimed at these changes)

ilyamandel:
Excellent point, will work on documentation updates shortly.

Issue #278, opened 2020-06-25
Change default units for output effective temperature to Kelvin (and not solar temperature )


demink:
My suggestion is to change the default unit for outputting  effective temperatures  in Kelvin: Teff (K) or possibly log10(Teff/K) following standard conventions in stellar astrophysics.   It currently is given in units of `Tsol`.  This is highly unconventional and will likely introduce confusion. 
 
**Example**: `Teff_1` and `Teff_2` that are written to `COMPAS_Output/Detailed_Output/BSE_Detailed_Output_X.csv`

**Context**: Using solar units is indeed common practice for Luminosity, Radius and Mass, probably mostly because standard units require very large numbers of 1e30 kg etc. . This is not the case for the (effective) temperature. Most stellar astronomers know reasonable temperature ranges by heart in units of T(k) or log_10 T (K).

ilyamandel:
Fixed while I was in the middle of doing something else, will include in next PR.

ilyamandel:
Fixed in PR #297 .  Closing issue.  Thank you for raising it, @demink !

PR #297, opened 2020-07-24
Fix for temperature output units (issue 278) and documentation update

ilyamandel:
Minor update to fix the temperature output units from Tsol to K (see issue #278 ).  Also, took the opportunity to update the documentation to reflect recent changes in program options.


jeffriley (commented):
Not worth providing both "TSOL" and "K" output options (with "K" in the default record specification) so the user can choose at runtime?

ilyamandel:
I can't imagine why anybody would want to measure temperature in Tsol.  We only use it internally in the code for the sake of convenience, because of the simple scaling L/Lsol = (R/Rsol)^2 (T/Tsol)^4, without the need for more constants.

jeffriley (approved):
Approved

PR #298, opened 2020-07-25
Remove the numerical Roche lobe calculation and introduce a determination of envelope state based on temperature

ilyamandel:
This PR removes the numerical Roche lobe calculation, which was consistent with the analytical one but superfluous and implemented without necessary care in taking numerical derivatives.  It also introduces a determination of the envelope type based on the temperature (FIXED_TEMPERATURE prescription).   Results with the LEGACY envelope state prescription  continue to reproduce results from earlier runs.


jeffriley (approved):
Approved

PR #299, opened 2020-07-26
Extends use of zetaRadiativeEnvelopeGiant (formerly zetaHertzsprungGap) for all radiative envelope giant-like stars

ilyamandel:
As with my other recent PRs, this doesn’t change the behaviour if using the LEGACY option for the envelope state (confirmed).  

There is one slightly unsightly exemption in HG::CalculateZeta() when using the LEGACY prescription, stemming from the decision to label HG envelopes as radiative in LEGACY, yet to apply a zeta=6.5 rather than the Soberman zeta prescription to them.  The exemption can be made less unsightly, but I intentionally kept it as a reminder to myself to create a more permanent fix, which requires fixing BaseBinaryStar::CalculateMassTransferOrbit() [next on my to-do list, but not for today].


jeffriley (approved):
Approved, but...

A reminder that this PR removes one option and adds another - that needs to be added to the documentation

ilyamandel:
Added documentation, now merging.

Issue #295, opened 2020-07-23
Binaries experiencing post-unbound interactions


avigna:
**Describe the bug**
When activating the `evolve_unbound_systems = False` flag, I noticed that some systems entering the CEE function had negative separations. My guess is that a check for RLOF in the form `is R_star>R_RL' keeps happening and will be true if semi-major axis < 0. Regardless, there were unbound systems engaging getting all over the CEE funtion.

**Label the issue**
`urgency_moderate` - This is a moderately urgent issue
`severity_minor`    - This is a minor bug with minimal impact

**To Reproduce**
1. Set `evolve_unbound_systems = True` in the pythonSubmit.
2. Suggestion: Add a print statement at the beginning of the CEE function (this is how I found the bug, but there are many ways to do it). Alternatively, play with a very wide binary and see if it happens after the first supernova.
3. If you followed the suggestion, run _N_ number of different binaries until you see a negative number.

**Expected behavior**
Unbound systems shouldn't interact with their companions. The only thing they should experience is a supernova (I think).

**Screenshots**
NA

**Versioning (please complete the following information):**
 - OS: MacOS Catalina
 - COMPAS v02.12.00

**Additional context**
NA

ilyamandel:
Fixed in PR #301 , closing.

PR #301, opened 2020-07-26
Defect repair for issue 295: do not engage in mass transfer if the binary is unbound

ilyamandel:
Check if binary is unbound; if it is, do not engage in mass transfer.  No change in behaviour for bound binaries.

### DIFF
--- a/src/BaseBinaryStar.cpp
+++ b/src/BaseBinaryStar.cpp
@@ -2640,10 +2640,11 @@ void BaseBinaryStar::InitialiseMassTransfer() {
  */
 void BaseBinaryStar::CheckMassTransfer(const double p_Dt) {
 
-    if(Unbound())                          
-        return;                                                                                                                 // do nothing for unbound binaries
     InitialiseMassTransfer();                                                                                                   // initialise - even if not using mass transfer (sets some flags we might need)
 
+    if(Unbound())
+        return;                                                                                                                 // do nothing for unbound binaries
+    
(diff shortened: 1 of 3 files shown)


jeffriley (commented):
Was this code actually being called for unbound binaries, or is this just defensive programming?  Happy for the check to go in, but I didn't think we could get to this code for an unbound binary.

jeffriley (commented):
Ah - unless we're evolving unbound binaries...

If that's the case - if we can actually call this code for unbound binaries, then we can probably call ResolveMassChanges() in BaseBinaryStar::EvaluateBinary(), and if we do that we might need to initialise m_aMassTransferDiff and m_OmegaMassTransferDiff to zero.  That's currently done in BaseBinaryStar::InitialiseMassTransfer(), so maybe that should be called and a check done there?

Just a thought - I haven't done a rigorous code review

ilyamandel:
@jeffriley -- Yes, this code was being called, and it was creating a problem when unbound binaries continued to be evolved -- see the reported issue #295 .  I confirm that this fix solves the issue, and doesn't impact bound binaries.  
But you make a good point about being safe -- I'll make sure that BinaryConstituentStar::SetRocheLobeFlags sets the flags correctly to avoid RLOF in this case.

ilyamandel:
@jeffriley -- this should address your concern, please check again when you have a chance (no rush); I confirm it still has the desired behaviour in practice.

jeffriley (approved):
Approved

Issue #269, opened 2020-06-10
Common-envelope event: Legacy bug in eccentric RLOF leading to a CEE


avigna:
**Describe the bug**
This is a Legacy code that need to be checked and verified in COMPAS. Eccentric binaries which enter a common-envelope event might not be circularised (e.g. if `circularise_binary_during_mass_transfer = False`). If they are not, the energy formalis is solved using the semi-major axis instead of the periastron. 

**Label the issue**
`urgency_moderate` - This is a moderately urgent issue
`severity_moderate` - This is a moderately severe bug

**To Reproduce**

**Expected behavior**
The common-envelope function should use the periastron as initial separation instead of the semi-major axis. This avoids the issue for eccentric binaries.

**Screenshots**
N/A

**Versioning (please complete the following information):**
N/A

**Additional context**
Found in Legacy. Need to check in COMPAS and fix if needed.

avigna:
This was present in COMPAS and has now been fixed in as part of  https://github.com/TeamCOMPAS/COMPAS/pull/303 .

PR #302, opened 2020-07-26
PR fixing bug #269: Common-envelope event: Legacy bug in eccentric RLOF leading to a CEE

avigna:


avigna:
Incorrectly pulling from and to the right place. Will close this PR without merger and make a new one.

PR #303, opened 2020-07-26
PR fixing bug #269: Common-envelope event: Legacy bug in eccentric RLOF leading to a CEE

avigna:


jeffriley (changes requested):
The changes are good, except for the following:

(a) I don't see updates to the change log and version number in constants.h.  You can put them in, and if there are conflicts when merging we can make them right then

(b) what is the test file?

avigna:
Thanks Jeff!
> (a) I don't see updates to the change log and version number in constants.h. You can put them in, and if there are conflicts when merging we can make them right then
I added the strings. I unfortunately lost track of Ilya's changes and not sure what number it should be (I probably didn't fetched to the right dev branc when I started working on this bug).

> (b) what is the test file?
It was a random txt file I made just to be sure I was pushing to the right place. I should have deleted it (and assumed I did as I do not see it in the changed files).

jeffriley (approved):
Merge conflicts resolved.

Issue #267, opened 2020-06-02
Common-envelope event: consistent calculation of final separation


avigna:
**Describe the bug**
In the ```BaseBinaryStar::ResolveCommonEnvelopeEvent``` function, the lambda structure parameter is calculated using the radius of the stars, but the final separation is calculated using the Roche lobe of the stars. This is wrong and particularly harmful when the radius is not simillar to the Roche-lobe radius. 

**Label the issue**
`urgency_high`     - This is a very urgent issue and should be resolved as soon as possible
`severity_moderate` - This is a moderately severe bug

**To Reproduce**
Run any system with a CEE, easier to notice with a double-core CEE (will add an example later).

**Expected behavior**
1. All the ```BaseBinaryStar::ResolveCommonEnvelopeEvent``` function should only read with Rsol units, instead of a mixture between AU and Rsol.
~2. User should be allowed to choose if they want to use the stellar radius or Roche lobe radius for both the lambda and binding energy calculations, as well as for the calculation of the post-CEE separation.~
2. Use the stellar radius as an input (instead of Roche-lobe radius).

**Versioning (please complete the following information):**
 - OS: MacOS Catalina
 - COMPAS [e.g. v02.11.01]

**Additional context**
N/A

avigna:
> All the BaseBinaryStar::ResolveCommonEnvelopeEvent function should only read with Rsol units, instead of a mixture between AU and Rsol.

This was sorted out as part of https://github.com/TeamCOMPAS/COMPAS/pull/303.

avigna:
> Use the stellar radius as an input (instead of Roche-lobe radius).

This is already done as has been fixed during the rewriting of the code.

```
	m_Star1->CalculateLambdas(m_MassEnv1);                                                                              // calculate lambdas for star1
	m_Star2->CalculateLambdas(m_MassEnv2);                                                                              // calculate lambdas for star2

    m_Star1->CalculateBindingEnergies(m_Mass1Final, m_MassEnv1, m_Star1->Radius());                                     // calculate binding energies for star1 (uses lambdas)
    m_Star2->CalculateBindingEnergies(m_Mass2Final, m_MassEnv2, m_Star2->Radius());                                     // calculate binding energies for star2 (uses lambdas)

    m_Star1->CalculateCommonEnvelopeValues();                                                                           // calculate common envelope values for star1
    m_Star2->CalculateCommonEnvelopeValues();                                                                           // calculate common envelope values for star2
```

[...]

PR #307, opened 2020-08-06
Use radius of the star instead of Roche-lobe radius throughout Resolv…

avigna:
…eCommonEnvelopeEvent()

This addresses https://github.com/TeamCOMPAS/COMPAS/issues/267


avigna:
I made a quick test to check that pre-post bug BSE_Common_Envelopes.csv files are different. They calculate the binding energy with different values for radius and work as expected.

ilyamandel (approved):
We need to think more about very eccentric CEs (if they circularise at periastron, they dissipate energy -- does that not also go into unbinding the envelope, or is it radiated away?), but good enough for now. :)

PR #310, opened 2020-08-07
SN_muller2016_229

### LINKED ISSUE #229: Incorrect fitting parameters in Muller 16 SN kick function
In BaseStar.cpp +2791-2801

The constants in the function don't line up with those reported in Vigna-Gomez 2018 (eq. B5). 

The piecewise ranges on m_CO are mostly* correct, but the constants that determine the kick in the code are double what they are in the paper (except for line 2794, which has a 250, where 200 would be the appropriate double). I'm not sure why this is, but it seems like a bug.

*Somewhat related, the lowest limit in both (B4) and (B5) in the paper is m_CO = 1.372 (below which there is no SN), but in the code, we have m_CO < 1.44 statements instead (GiantBranch.cpp +1170 has the analogous lower limit for remnant mass). Alejandro comments that this lower limit is a first approach, so this may be a feature.

avigna:
Fixing of "Incorrect fitting parameters in Muller 16 SN kick function" #229. 

I tested an old and new run to verify that kicks are now half of what they were, consistent with the numbers from Vigna-Gomez+2018.

It also introduces a new feature. Before, with the Muller+2016 prescriptions, stars with CO core mass less than 1.44 Msol had zero kick. Now, using what it is on Vigna-Gomez+2018, stars with 1.37 < m_CO/Msol < 1.44 get a non-zero. I think this is the correct thing to do.

@reinhold-willcox I would be grateful if you can double check the actual numbers with the ones from my paper and then we can move on having you, Ilya or Jeff approving the request.


ilyamandel (changes requested):
I suggest following the equations of Vigna-Gomez et al., with explicit reference to those equations (B4 or B5, as relevant), verbatim -- no changes or extra rounding (e.g., 1.37 vs. 1.372).

Update comments to just point to the equations and state that BHs get no natal kick, e.g., in lieu of 

"* For BH assume complete fallback so no ejecta hence no rocket effect
 * If BH doesnt assumes complete fallback, e.g. neutrino mass loss, a Blauuw Kick should be calculated		 * If BH doesnt assumes complete fallback, e.g. neutrino mass loss, a Blauuw Kick should be calculated
 *		 *
 * ALEJANDRO - 17/03/2017 - It seems from Bernhard's notes he gives us V_kick = V_kick_3D.		 * ALEJANDRO - 17/03/2017 - It seems from Bernhard's notes he gives us V_kick = V_kick_3D.
 * Checking with some of the highest kicks he gives, we get V_kick>1000 km s^-1		 * Checking with some of the highest kicks he gives, we get V_kick>1000 km s^-1"

should just say 
" * Kicks implemented following Equation (B5) of Vigna-Gomez et al., 2018, arXiv:***
 * BHs receive no natal kicks in this model"

ilyamandel:
Also, it appears that you have not updated      GiantBranch::CalculateRemnantTypeByMuller2016() .  You seem to want to treat <1.44 CO core mass objects as NSs, but they will be labeled as WDs under this prescription.  

if (utils::Compare(p_COCoreMass, MCH) < 0) {
        stellarType = STELLAR_TYPE::OXYGEN_NEON_WHITE_DWARF;
}

I suggest re-working the code so that the type is predicted by the mass returned by CalculateRemnantMassByMuller2016(), as is done for other prescriptions.

avigna:
@ilyamandel I made the changes; mostly updating comments and deleting the condition from your last comment.

ilyamandel:
Looks nice, but repeating from previous comment:

"I suggest following the equations of Vigna-Gomez et al., with explicit reference to those equations (B4 or B5, as relevant), verbatim -- no changes or extra rounding (e.g., 1.37 vs. 1.372)."

avigna:
Ahhh, I see. The arXiv version has 1.372 and the MNRAS version has 1.37 (https://ui.adsabs.harvard.edu/abs/2018MNRAS.481.4009V/abstract).

Which one do you prefer? We currently pointing to arXiv version in the comments, so probably that one?

Edit: Updated the numbers to the arXiv version.

ilyamandel:
Ahh, I hadn't realised there was a difference between the MNRAS and arXiv versions, sorry. :)

ilyamandel (approved):
Changes look good!

PR #312, opened 2020-08-08
Fixed mass transfer with fixed accretion rate, cleaned up various bits of code

ilyamandel:
- Simplified random draw from Maxwellian distribution to use gsl libraries
 - Fixed mass transfer with fixed accretion rate
 - Cleaned up code and removed unused code
 - Updated documentation


ilyamandel:
P.S. Again, enough small changes that direct diffs fail, but differences are not statistically significant.

jeffriley (approved):
Approved, with comments (not directly related, but purely to capitalise on you having eyes on this code...)

BaseStar line 1219 - my comment  "JR: todo: is this really necessary?"
Is it?  If it is, we should remove my comment.  If it isn't, we should remove the code.

BaseStar line 1223 // mass before fake mass change - due to order of updating radius bug     JR: todo: check this
My goal when refactoring was to change a little as possible so that I could check that the new code produced consistent results with legacy.  Wrt to the comment "due to order of updating radius bug", I didn't know then, and don't know now, what the "order of updating radius" bug is, but I suspect with the changes between legacy and the current version of the new code, odds are it no longer exists - or exists in some other form.  Should we fix this code?

Should we just submit issues to look at and clean the places in the code I've left comments like that (whenever we come across them)?

And, if you happen to be in the code, could you fix my typo in BaseBinaryStar at line 2150 - "arbitray" -> "arbitrary"

ilyamandel:
Thank you, @jeffriley !  

- I have seen the queries about UpdateAttributesAndAgeOneTimestep() in both ZetaThermal and ZetaNuclear calculations.  I'll try to understand these, too, but not on this commit -- one step at a time.  No need for extra issues, I think -- the code comment is enough to flag them in this case. 
- Typo in "arbitrary" fixed.

Issue #300, opened 2020-07-26
Adiabatic Zeta calculation should be based on the true core mass, not the He core mass


ilyamandel:
The functions CalculateZadiabaticSPH(const double p_CoreMass) and CalculateZadiabaticHurley2002(const double p_CoreMass) expect a core mass on input.  When calling them to determine the star’s zeta (and hence stability) via GiantBranch::CalculateZeta() pre-PR #299 (or via BaseStar::CalculateZadiabatic() after PR #299 ), we always pass in the He core mass, not the core mass — meaning that, say, for HeHG stars, we pass in the total mass, rather than the mass of the CO core.  

I suggest using m_CoreMass rather than m_HeCoreMass when calling these; any objections, @SimonStevenson , @avigna , @jeffriley ?

ilyamandel:
Since I've not heard any objections, I will implement this change in the next PR.  It should not make much practical difference, since we generally run with stable case BB MT, anyway, i.e., the value of Zeta for HeHG stars is irrelevant.

ilyamandel:
Fixed in PR #313 ; no impact for our default model with stable case BB, as expected.

Issue #277, opened 2020-06-25
Inconsistent He core mass values


ilyamandel:
The He core mass value appears to be greater than the overall core mass or even the total mass in some cases.

I was checking the BSE_Supernovae.csv file.  First, the things which do make sense:
- the total pre SN mass, Mass_Total@CO_SN,  is always greater than or equal to the core pre SN mass, Mass_Core@CO_SN
- the core pre SN mass, Mass_Core@CO_SN, is always greater than or equal to the CO core pre SN mass, Mass_CO_Core@CO_SN

Now, the things which do not make sense:
- the pre SN He core mass, Mass_He_Core@CO_SN, sometimes exceeds the total pre SN core mass, Mass_Core@CO_SN.  In fact, the pre SN He core mass sometimes exceeds the total pre SN mass, Mass_Total@CO_SN!  This is particularly true for ultra-stripped SN leading to NS formation: for these, the total pre-SN mass always equals the pre-SN CO core mass, indicating ultra-stripping, yet the He core mass is 1 to 3 solar masses larger than these.
- the remnant mass, Mass_SN, is sometimes larger than the total pre SN mass, Mass_Total@CO_SN.  

The last bullet point may be at least partly tied to the penultimate one, because I use the He core mass in GiantBranch::CalculateRemnantMassByMullerMandel() to calculate the remnant mass.  We need to make sure all internal mass variables are self-consistent. 

This bug has a potentially significant impact on remnant mass and kick calculations, at least for the Mandel & Mueller prescription.

COMPAS v02.11.03

reinhold-willcox:
To add, Jeff and I looked into this and (P)PISN are also dependent on the HeCore mass.

ilyamandel:
This issue is fixed through PR #313 , closing.

Issue #189, opened 2020-04-02
USSN terminology


ilyamandel:
Suggest changing code terminology regarding USSN to avoid future confusion about when the star is ultra-stripped (so that when it explodes in the future, it will be a USSN) vs. the supernova itself.

avigna:
Agreed! My suggestion is for each star to have a `stripped` and `ultra-stripped` flag. And then in the case of an USSN, we can check for the `ultra-stripped` flag.

ilyamandel:
Fixed in PR #313 , closing.

PR #313, opened 2020-08-09
Bug fixes and code streamlining

ilyamandel:
- Use total core mass rather than He core mass in calls to CalculateZAdiabtic (see Issue #300 )
- Fixed inconsistent values of He core mass in stripped stars (see issue #277)
- Fixed USSN nomenclature -- now refers to supernova rather than to ultra-stripping (see issue #189 ).


jeffriley (approved):
Approved

Issue #317, opened 2020-08-13
Boost bracket_and_solve_root issue: unable to bracket root


reinhold-willcox:
**Describe the bug**
Very rarely, the following error occurs.

```
COMPAS v02.13.03
Compact Object Mergers: Population Astrophysics and Statistics
by Team COMPAS (http://compas.science/index.html)
A binary star simulator

Start generating binaries at Thu Aug 13 15:04:13 2020

terminate called after throwing an instance of 'boost::exception_detail::clone_impl<boost::exception_detail::error_info_injector<boost::math::evaluation_error> >'
  what():  Error in function boost::math::tools::bracket_and_solve_root<double>: Unable to bracket root, last nearest value was 3.4657688153706713e-24
Aborted (core dumped)
```

**Label the issue**

`urgency_moderate` - This is a moderately urgent issue

`severity_moderate` - This is a moderately severe bug

**To Reproduce**
On the dev branch, version v02.13.03, run the generic pythonSubmitDefault on the attached grid file.

[bad_grid.txt](https://github.com/TeamCOMPAS/COMPAS/files/5067221/bad_grid.txt)

**Expected behavior**
The binary would complete without the boost error.

**Screenshots**
N/A

**Versioning (please complete the following information):**
 - OS: Ubuntu 18.04 & OzStar
 - COMPAS v2.13.03

**Additional context**
Happens extremely rarely (once in a 1e6 binary run).

reinhold-willcox:
What I notice is that the dM at each iteration is going down. It starts at 0.00205... (our guess), and although factor = 2 and is_rising = 1, the dM drops by half each time (instead of doubling), so boost throws an error when we drop below the system minimum float (or something like that).

ilyamandel:
The solution is simple (catching an exception thrown by the boost root finder).  This, indeed, is very rare.

However, the ultimate reason for the issue is more complex, and has to do with our failing to properly set the core mass of newly minted HG and HeHG stars.  Fixing this has a potentially more significant impact.

Both issues will be addressed by PR #319 .

ilyamandel:
Solved by PR #319 , closing.

PR #319, opened 2020-08-13
Initialise HG and HeHG core masses correctly, catch exceptions thrown by boost root finder

ilyamandel:
- Catch exception in boost root finder for mass transfer (resolve issue #317)
- Update core masses during Initialisation of HG and HeHG stars to be consistent with Hurley models
- Avoid division by zero in mass transfer rates of WDs
- Removed POSTITNOTE prescription for BH masses


jeffriley (commented):
I understand why you have done this:

double CalculateThermalMassLossRate() { return std::numeric_limits<double>::max(); }

but I'm not sure I like it much.  It will solve the problem of divide by zero, and you say in practice it will be Eddington limited, but is it possible that in the future someone may make a change not knowing that this exists and end up being bitten by this?

I suppose if the rate really should be (effectively) infinite, then this is the way to do it - but I want to be sure this is really the best fix and not just expedient.

ilyamandel:
White dwarfs lose mass on either the nuclear timescale or the dynamical timescale, and gain it on the dynamical timescale (see, e.g., last paragraph of section 2.6.4 of Hurley+ 2002).  I've therefore replaced the thermal timescale with the dynamical timescale for WDs.

jeffriley (approved):
Ok, I like that better :-)

+remove POSTITNOTE

Approved

Issue #315, opened 2020-08-09
ResolveSupernova() can leave higher remnant masses than initial pre-SN masses for SSE


ilyamandel:
At the end of GiantBranch::ResolveSupernova(), m_Mass can be larger than m_SupernovaDetails.totalMassAtCOFormation, despite the latest code fixes.  This leads to nonsensical negative ejecta masses being passed along.  This only seems to be an issue for SSE, not BSE.

ilyamandel:
(Note to self: confirmed this is not an issue for BSE, so need to understand why/where SSE and BSE paths diverge.)

jeffriley:
Do you have a pythonSubmit file and seed to duplicate?

ilyamandel:
Add the following at line 1733 of GiantBranch.cpp and run with the attached pythonSubmit file.

        if(m_Mass>m_SupernovaDetails.totalMassAtCOFormation){
            std::cout<<"SNERROR: "<< m_Mass<<" "<<m_SupernovaDetails.totalMassAtCOFormation<<std::endl;
        }

[pythonSubmitMandelMuller.py.txt](https://github.com/TeamCOMPAS/COMPAS/files/5047141/pythonSubmitMandelMuller.py.txt)

ilyamandel:
[I suspect there's just one extra step of evolution happening -- the remnant mass grows between the start and end of ResolveSupernova; if so, it may be easy enough to fix, but I want to actually use this as an excuse to better align SSE and BSE rather than putting a patch on SSE.]

jeffriley:
@ilyamandel I have not done anything with this yet, but @reinhold-willcox is making changes to SN handling - his nskicks branch makes some fairly significant changes to the BSE SN code flow (I don't remember is it touches the SSE code flow) - so you and he might want to discuss before you embark on your cleanup.

reinhold-willcox:
My PR doesn't touch the Giantbranch::ResolveSupernova function, just the BaseBinaryStar::ResolveSupernova.

That said, if Ilya is correct that some mass quantity is changing mid-timestep, then the changes I'm proposing to the structure of a full timestep would fix that. I haven't coded these up though, I'm still in the early stages of that.

ilyamandel:
Fixed in PR #322 , closing.

PR #322, opened 2020-08-16
General cleanup and bug fixes, including issue 315, removal of ZetaThermal and ZetaNuclear calculations

ilyamandel:
- General code cleaning
- Removed some redundant variables (e.g., m_EnvMass, which can be computed from m_Mass and m_CoreMass)
- Removed calculations of ZetaThermal and ZetaNuclear (these were previously incorrect because they relied on the evolution of a stellar copy which reverted to BaseStar and therefore didn't have the correct behaviour)
- Fixed CalculateZadiabatic to use ZetaAdiabaticArbitrary rather than ZetaThermalArbitrary; removed the latter
- Capped He core mass gain during shell H burning for CHeB and TPAGB stars, whose on-phase evolution now ends promptly when this limit is reached; this change also resolves issue #315  (higher mass SN remnants than total stellar mass)
- Documentation update


jeffriley (approved):
Approved.

I particularly like the renaming of the "Is...SN" functions to "Resolve..SN."

Just one suggestion:

On line 80 in GiantBranch.h and on line 317 of BaseStar.cpp we use the quantity Mass() - CoreMass() (and possibly elsewhere).  Should we have a convenience/getter for EnvMass() that returns Mass() - CoreMass().  That way the envelope mass is computed in one place only (not that I can think of a good reason to change how it is computed, but if we have the getter then we know that it is always computed the same way).

Just a suggestion - not a requirement for approval.

ilyamandel:
Yes, I thought about that, too -- but then decided it was actually more clear this way what exactly we are using.  So that was a conscious choice.  I am happy to be overruled, but for now, I'll merge. :)

PR #323, opened 2020-08-19
Overhauled Linux instructions to better distinguish different distro families

warrickball:
A user was confused by literally following the "Instructions for Ubuntu", which had at the end a few instructions for RHEL-based distros. I've tried to re-arrange that section of "Getting Started" into a less Ubuntu-specific form without removing the actual Ubuntu command line instructions.

### DIFF
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -3,7 +3,7 @@
 # Getting Started
 
   * [1. Installing COMPAS and Dependencies](#1-installing-compas-and-dependencies)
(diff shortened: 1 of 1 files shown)


ilyamandel:
Hi Warrick,
These changes are great, thank you!  However, pulling these directly into production will leave dev out of sync.  Can you change the PR to pull into dev -- and then we can pull dev into production?
Thank you,
Ilya

ilyamandel (approved):
Actually, never mind, I just took the liberty of changing this to pull into dev.

warrickball:
Rookie error on my part, so thanks for handling it.

PR #324, opened 2020-08-19
Pull dev into production

ilyamandel:
Mostly for the sake of incorporating Warrick's Linux installation changes.


jeffriley (approved):
Approved

Issue #65, opened 2020-02-19
Feature request: option ECSN threshold mass


TeamCOMPAS:
Add the ECSN threshold as an input parameter (currently set to Hurley model, should have Fryer model as option).

ilyamandel:
Implemented in PR #326 , closing.

PR #326, opened 2020-08-20
Supernova enhancements and cleaning

ilyamandel:
- ONeWDs can now undergo ECSN if their mass rises above MECS=1.38 solar masses (previously, they could only undergo CCSN on rising above 1.38 solar masses).  ONeWD::CalculateInitialSupernovaMass now returns MCBUR1 rather than 5.0 to ensure this happens
- BaseStar::CalculateMaximumCoreMassSN() has been removed — it’s superfluous since  GiantBranch::CalculateCoreMassAtSupernova_Static does the same thing
- Some misleading comments in TPAGB dealing with SNe have been clarified
- Option to set MCBUR1 [minimum core mass at base of the AGB to avoid fully degenerate CO core formation] to a value different from the Hurley default of 1.6 solar masses added, Issue #65 resolved
- Removed unused Options::SetToFiducialValues()
- Documentation updated


jeffriley (approved):
Approved

Issue #320, opened 2020-08-13
Exception handling


ilyamandel:
Currently, errors and warnings are reported in text outputs, but evolution continues, potentially with unpredictable results.  There is even a risk that evolution could go into an infinite loop.  Throwing and catching COMPAS exceptions, ultimately leading to a termination of evolution, could be a safer mechanism for error handling.

ilyamandel:
@jeffriley -- are we realistically going to try exception handling, or does this belong in the "won't fix" bin?

jeffriley:
It's a big job...

A few things:

> Currently, errors and warnings are reported in text outputs, but evolution continues, potentially with unpredictable results. 

That's true.  Do we have a good idea of what we should actually do with those errors and warnings?  The classification of error vs warning was mine when I refactored the code, and I knew a lot less then about what was going on than I do now, but still I think someone with more experience/knowledge could classify them better - then we could (e.g.) fail (stop) on error and continue on warnings (as a first cut).

> There is even a risk that evolution could go into an infinite loop

Possible, but I think unlikely - I think I fixed every place in the code that could do that.

> Throwing and catching COMPAS exceptions, ultimately leading to a termination of evolution, could be a safer mechanism for error handling

That's true.  The trouble is we would have to check every calculation for (e.g.) divide by zero, and throw an exception for that.  That's not only a performance issue, we either have to fail on every exception, or sit down and work out which ones are fatal and which ones should just be warnings.  

Would be terminate the evolution of a singe star/binary, and continue the rest (from --number-of-systems, or a grid file), or would we terminate the entire run since the run might be dependent upon having all stars/binaries evolved (or should that be configurable)?

As I said, a big job - even just figuring out how we want to handle different errors/warnings - but ultimately it would make the product better.

ilyamandel:
OK, leaving this alive for the future at @jeffriley 's suggestion.

jeffriley:
Closing for now - will revisit later.

ilyamandel:
I did look at this one as part of my PR and issue review, but intentionally didn't even ask. ;-)

PR #327, opened 2020-08-20
Code cleanup

jeffriley:
Code cleanup:
(1) moved BaseStar::SolveKeplersEquation() to utils
(2) changed call to (now) utils::SolveKeplersEquation() in BaseStar::CalculateSNAnomalies() to accept tuple with error and show error/warning as necessary
(3) removed call to std::cerr from utils::SolveQuadratic() - now returns error if equation has no real roots
(4) changed call to utils::SolveQuadratic() in GiantBranch::CalculateGravitationalRemnantMass() to accept tuple with error and show warning as necessary
(5) changed RadiusEqualsRocheLobeFunctor() in BinaryBaseStar.h to not use the SHOW_WARN macro (can't uset ObjectId() function inside a templated function - no object)
(6) changed COMMANDLINE_STATUS to PROGRAM_STATUS (better description)
(7) moved ERROR:NONE to top of enum in constants.h (so ERROR = 0 = NONE - makes more sense...)
(8) added new program option '--enable-warnings' to enable warning messages (via SHOW_WARN macros).  Default is false.  SHOW_WARN macros were previously #undefined


ilyamandel (approved):
These look great, and I am approving but not yet pulling because I have one optional suggestion.

I like 

root = bracket_and_solve_root(RadiusEqualsRocheLobeFunctor<double>(p_Binary, p_Donor, p_Accretor, &error), guess, factor, is_rising, tol, it);
             if (error != ERROR::NONE) SHOW_WARN(error);

more than 

std::tie(error, m_SupernovaDetails.eccentricAnomaly, m_SupernovaDetails.trueAnomaly) = utils::SolveKeplersEquation(m_SupernovaDetails.meanAnomaly, p_Eccentricity);

           if (error == ERROR::NO_CONVERGENCE) { SHOW_ERROR(error, "Solving Kepler's equation"); }        // show error
     else if (error == ERROR::OUT_OF_BOUNDS ) { SHOW_WARN(error, "Eccentric anomaly"); }                 // eccentric anomaly out of bounds - show warning

The first catches and displays all errors thrown (and if someone adds another error inside the function, they don't have to remember to add catching them in the calling functions), the second does not.  I realise the second distinguishes errors and warnings, but I feel like that should be done based on the ERROR type thrown...

jeffriley:
@ilyamandel Initially I agreed with you, but then in thinking about it more I have changed my mind :-)

In both situations shown, the functions returning the error are utility functions that are being asked to do something for the calling function, and if something that the called function couldn't handle occurs in that operation the called function returns an indication of what the problem was.  I think it is up to the caller to decide if the problem reported by the called function should be treated as an error (and maybe propagated up and the evolution aborted (eventually - when we get to that), or a warning (and a message is emitted and evolution continues).

It looks (and probably is) a little inelegant here because in BaseStar::CalculateSNAnomalies() we don't do anything different for an error vs a warning (other than label them differently when we emit the message).  Eventually, as noted above, we should actually treat them differently in the caller function.  Whereas in MassLossToFitInsideRocheLobe() we just treat the error returned as a warning - maybe that should change?

Does this make sense?

ilyamandel:
OK, can return to this when we consider COMPAS exceptions (see issue #320 ) and maybe find a more elegant solution then.

PR #328, opened 2020-08-21
Cosmetic cleanup of SN related functions and variables

### LINKED ISSUE #254: Remove ID from output files
The ID column in the output files seems to be redundant and/or misleading. It uniquely identifies binaries only on a given core, so IDs are repeated when using multi-core compas. The Seeds already provide the functionality of uniquely identifying systems (as well as providing random seeds).

My understanding is that IDs are useful for internal COMPAS calculations, so we should not get rid of them completely, but just not include them in the output files. 

I don't actually know where IDs are used internally - if they are not, we could scrap them.

### LINKED ISSUE #261: misnamed m_OrbitalVelocity
**Describe the bug**
Some of the variables and functions with names which include *OrbitalVelocity* actually refer to OrbitalAngularVelocity. These include: m_OrbitalVelocity,m_OrbitalVelocityPrev, m_OrbitalVelocityPrime, OrbitalVelocity(). However, m_OrbitalVelocityPreSN and OrbitalVelocityPreSN() are set and used independently, and are implemented correctly.

**Label the issue**

**To Reproduce**
N/A - it's just names of variables, does not affect results.

**Expected behavior**
These should be renamed to OrbitalAngularVelocity. 

**Screenshots**
N/A

 - COMPAS v02.11.00

### LINKED ISSUE #335: m_OrbitalVelocity* inconsistencies

m_OrbitalVelocity and all of its extensions (Prev, PreSN, PostSN) are somewhat inconsistent. The SN-related values are assigned to a BinaryConstituentStar and the base and Prev values to a BaseBinaryStar. Some may also be superfluous.

As the orbital velocity is a binary quantity (and not a quantity of a single star, even one in a binary), all reincarnations should be in BaseBinaryStar. Also, the postSN quantities should be replaced with just the base ones (keeping with the new changes which remove postSN quantities). This means changing what gets called in the output file and double checking that the correct quantities are being output.

I can do this, but I'm posting here to remind myself (and notify others).

reinhold-willcox:
These are cosmetic changes that are somewhat overdue (they were addressed in the ns_kicks PR, but that one has issues that still need to be addressed). The only difference between outputs here and in `dev` should be the removal of the "ID" parameter.

- Created new "changelog.h" file and moved over relevant parts of constants.h
- Changed OrbitalVelocity to OrbitalAngularVelocity where that parameter was misnamed
- Changed Pre/PostSNeOrbitalVelocity to OrbitalVelocityPre/PostSN for consistency
- Added and updated physical conversion constants for clarity (e.g MSOL to MSOL_TO_KG)
- Removed ID from output files, it is confusing and superceeded by SEED
- Removed 'Total' from TotalOrbital(Energy/AngularMomentum) since these values contribute to the totals, but are not really themselves totals.
- Typos


ilyamandel (approved):
Looks good!  I'd just remove BINARY_PROPERTY::ID rather than commenting it out (makes for shorter and cleaner code), so I am not pulling yet -- feel free to make this tweak and then pull, or ignore if you feel strongly about keeping that in.

reinhold-willcox:
I'll refer to @jeffriley on the best approach. I know he spent a lot of time putting together a clever mechanism to choose what gets printed and what doesn't, but I've never used it before.

ilyamandel:
I just literally meant removing unused code lines instead of commenting them out, that's all. :)

reinhold-willcox:
No I understand, but Jeff's scheme does this much more cleverly than removing the commented line (though I agree that mine was not the correct approach). So I defer to Jeff to decide if this is a parameter we will *ever* need. If so, we should leave it in and I should learn how to use his scheme. If not, we should delete it completely.

ilyamandel:
@jeffriley 's scheme is that users can overwrite the default outputs with text files specifying the desired output formats.  That's perfect, but doesn't impact on how we implement the default outputs in the code.

ilyamandel:
Actually, I've just done this manually, so pulling and closing. :)

PR #330, opened 2020-08-21
Update dockerhub-ci.yml

ghost:
Updating to grab the version string from changelog.h instead of constants.h


TeamCOMPAS:
Didn't realize this could be done as part of the previous PR, my bad
-Reinhold

ilyamandel (approved):
Thank you for the fix, @reinhold-willcox!

PR #331, opened 2020-08-22
Enhancements and cleanup

ilyamandel:
- Removed several stored options (e.g., m_OrbitalAngularVelocity, m_RocheLobeTracker, etc.) to recompute them on an as-needed basis
- Removed some inf values in detailed outputs
- Slight speed-ups where feasible
- Declare functions constant where feasible
- Shift various calculations to only be performed when needed, at printing, and give consistent values there (e.g., OmegaBreak, which was never updated previously)
- Remove a number of internal variables
- Remove options to calculate Zetas and Lambdas at every timestep; variables that only appear in detailed outputs should not be computed at every timestep in a standard run
- Update documentation


jeffriley (commented):
Approved, but awaiting review from Alejandro

jeffriley (approved):
Approved

Issue #288, opened 2020-07-04
We still do not output a self-consistent state in the detailed output file


ilyamandel:
We do not always output a self-consistent state on a given line of the detailed output file.  For example, look at lines 960--965 of the detailed output file produced with the attached submit file.   Radius_1/RL has a value above one for a couple of steps when a stripped naked He star of barely a solar radius supposedly overflows its Roche lobe in a binary with a separation of a couple thousand solar radii.  Now, it did overflow the Roche lobe earlier — that’s how it got stripped — but at that point Radius_1/RL < 1.  So, again, we have an inconsistency in the way variables are recorded.  This is making debugging (and understanding what’s going on) very difficult — we really have to fix this.

I expect all variables recorded on a given line to represent an internally consistent state.  E.g., I should be able to calculate Radius_1/RL from the masses and separation recorded on the same line, and get a consistent result.  This is just one example -- we should make sure this always happens. 

Using version 2.12.0.

[pythonSubmitBelcz190814.py.txt](https://github.com/TeamCOMPAS/COMPAS/files/4872719/pythonSubmitBelcz190814.py.txt)

ilyamandel:
Being more familiar with the code now, I suspect this happens because we use stored values of variables rather than re-calculating them.  Most values except the basics like mass should just be re-computed on the fly rather than stored.  I'll try that out next.

reinhold-willcox:
@ilyamandel 
Can you clarify what you mean by stored vs recalculated? Do you mean when getter functions are written as `getParam() { m_Param; }` vs `getParam() { CalculateParamValue(); }` ?

My understanding (@jeffriley should confirm) is that repeatedly calling Calculate() functions to values which have already been calculated can add up, computationally, and that it was better to return `m_Param` where possible, and when we can be sure it was calculated correctly.

ilyamandel:
Yes, I mean getParam() { CalculateParamValue(); }

Most of the recalculations are very cheap.   In the few cases when they aren't, we can try to be more careful in checking that the value is still up-to-date at the time of printing.  But we can get bigger cost savings elsewhere, I am more worried about correctness for now.

reinhold-willcox:
I think this is fine for the short term, but might get in the way with consistency later on.

[...]

PR #332, opened 2020-08-23
Correct detailed output, remove some Prime quantities

ilyamandel:
- More cleaning, removed some of the unnecessary prime quantities like m_SemiMajorAxisPrime, m_EccentricityPrime, etc.
- Thermal timescales are now correctly computed after the CE phase
- Detailed output passes a set of self-consistency checks (issue #288)


jeffriley (approved):
Approved

PR #333, opened 2020-08-23
Defect repairs: Fixed debugging and logging macros in LogMacros.h

jeffriley:


ilyamandel (approved):
Why is "'\n" preferred to std::endl?   I thought (perhaps mistakenly) that the latter resolves to the most appropriate end-line character for a given system.
Just curious, happy to approve.

jeffriley:
I haven't looked into the details, but I believe '\n' works on all implementations.  The reasons I use '\,' are:

(a) laziness..., and
(b) I saw this:

https://en.cppreference.com/w/cpp/io/manip/endl

Note the last sentence of the first paragraph in 'Notes':

"Use of std::endl in place of '\n', encouraged by some sources, may significantly degrade output performance."

PR #336, opened 2020-08-24
SN kicks and vector addition of velocities

### LINKED ISSUE #236: 2nd supernova not printed in disrupted systems
In systems which are disrupted in the first supernova, the second supernova is not printed. The evolve_unbound_systems flag is True (and these systems all exist in the DCOs output) so these disrupted systems are being evolved, but the second supernova specifically is not included in the output.

**To Reproduce**
pythonSubmit with evolve_unbound_systems = True, and number_of_binaries = 1000 (to compare with intact DCO systems). All of the seeds in the Double_Compact_Objects should exist twice in Supernovae, but there are many that do not. To see all of the ones repeated twice, run the following bash command in your COMPAS_Output:
`tail -n+4 BSE_Supernovae.csv | awk '{print $2}' | uniq -d`
(You can verify that these are the systems which either disrupt at the 2nd SN or survive both). But the list is only 14 systems long, while DCOs is 131. 

**Expected behavior**
There should be the same number of repeated systems in Supernovae as in Double_Compact_Objects.

 - OS: Ubuntu 18.04
 - COMPAS v02.10.03

### LINKED ISSUE #255: Update meaning of SN_STATE
Currently, the SN_STATE parameter (in the code and output) is initialized to 0 (for no SNs), set to 1 after star1 supernovas, set to 2 after star2 supernovas, or set to 3 if both are simultaneous. In the output, there is no way to tell which system went first by looking at a single line of output (one would need to look at the times of both systems, or e.g the stellar types). 

This is the optimal place to describe how many SNe happened and the order they occurred in. These features have been requested as other columns, but I believe that fix is not necessary. This parameter is already an integer, so I propose we expand it's possible values and their meanings. Below I've created a dictionary of useful values -> meanings. Please let me know if you think there should be changes.

| Integer Value | Meaning |
|-----|-----|
| 10 | Star 1 went SN, second has not occured |
| 20 | Star 2 went SN, second has not occured |
| 12 | Star 1 went SN first, Star 2 second |
| 21 | Star 2 went SN first, Star 2 second |
| 3  | Both stars went SN simultaneously |

The value of this is that, given 2 SNe lines in Supernovae.csv for a given system, it is immediately clear in each individual line which Star went SN in the current line and whether there was a previous SN. 

A new column could be created with information on whether there was a previous SN, but that requires more overhead and makes searching for this information slightly more cumbersome.

**Additional context**
Related to Issue #183

reinhold-willcox:
Improvements and bug fixes on the previous ns_kicks pull request. Cleans up the ResolveSupernova function, to follow the algebra in Pfahl, Rappaport, Podsiadlowki 2002, App. B.

Adds a class vector3d which is exactly what it sounds like, and defines velocities of the binary system and stellar components in terms of these vector3d's. In particular, velocities from 2 SNe are added vectorally, and so are more accurate than average-value estimates. These quantities are all internal - values which are produced in the output files are now the system and component speeds, which are of course the magnitudes of these velocities. 

Other smaller fixes include:
- SN phi value is redefined to be relative to the periapsis (or Laplace-Runge-Lens) vector, which is still sampled uniformly in [0, 2pi], but means that system-to-system comparisons will break down.
- PostSN parameters have been removed, in favor of simply calling the present values of those parameters in the Detailed or Supernovae output files. 
- Supernova_State has been redefined to the following:
  - 0 means no SN has occured
  - 10 means only star 1 has SN'd
  - 20 means only star 2 has SN'd
  - 12 means star 1, then star 2
  - 21 means star 2, then star 1
  - 3 means SNe were simultaneous (this should ideally never happen, though we have not implemented the correct workaround yet)
- The coordinate bases of the 2nd SN is written in terms of the original CoM basis using [Euler rotation angles](https://en.wikipedia.org/wiki/Euler_angles). Currently, the direction of the periapsis vector of the 2nd SN is taken to be parallel to the post-1stSN periapsis (assuming non-zero eccentricities), though the new formulism could allow for precession of the periapsis in MT binaries if we choose to implement that at a later date (see, e.g Dosopoulou & Kalogera 2016, 2018). 

As mentioned above, phi differences mean system-to-systems comparisons will not work, but the overall populations rates are still consistent. Below are the intact DCO rates from dev and ns_kicks for 1million binaries each. 

```
dev
total =  1000000
nBNS =  719 
nBBH =  8510 
nBHNS =  3006 

nsk
total =  1000000
nBNS =  698 
nBBH =  8407 
nBHNS =  2954
```


jeffriley (commented):
I'm mostly ok with this (I'm not wild about setting variables to nan, but I'll live with it :-) It's an historical thing...).  Test show good match to pre-change code, so that's all good.

I've added @ilyamandel  as a reviewer so he can review your changes from a physics standpoint.

ilyamandel (changes requested):
This may well be doing the right thing.  I trust you that it is, @reinhold-willcox .  But I found the code very difficult to read, so I stopped after the first couple of files.  A few comments, rough in order of the BaseBinaryStar files that I read:

1. Speed and velocity still seem to be intermittently used, e.g., 
```
Vector3d natalKickVector = m_Supernova->SN_KickVelocity() * Vector3d(cos(theta)*cos(phi), 
                                                                          cos(theta)*sin(phi),
                                                                          sin(theta));
```
[the SN_KickVelocity() function returns a speed, right?]


2. 

```
// The quantities below are mostly meaningless 
  * Resolves supernova event - one of the stars has gone supernova!		         m_OrbitalVelocityPreSN = -nan("");
  *		         m_uK = m_Supernova->SN_KickVelocity() / m_OrbitalVelocityPreSN; // -- - Dimensionless kick velocity
```

What happens when you divide by nan?  Why would you intentionally do that?

3.  

`double m1_ = m_Supernova->Mass();                                // Mo   - SN star postSN mass
`

We seem to be introducing different notation from the rest of the code -- this will make it harder for future maintainers (not to mention that it's really not obvious that a following underscore means post-SN, while the absence of an underscore means pre-SN).

4. 

```
Vector3d V = Vector3d( (-omega*a*a/r)*sinEA,   
                                (omega*a*a/r)*sqrt(1-e*e)*cosEA,  
                                 0.0                            );        // km/s      - Relative velocity vector
```

I find single-letter variable names are quite confusing (and then there's dV1, dV2, etc.), and this isn't helped by the comments ("Relative velocity vector" relative to what?)

5.  

```
Vector3d Vcm_ = (-m2*dm1/(mb*mb_) + m1*dm2/(mb*mb_)) *V 
 	//                       SHOULD BE BACK TO NICE UNITS NOW                        //		                                                  + (m1_/mb_) *dV1 
 	///////////////////////////////////////////////////////////////////////////////////		                                                  + (m2_/mb_) *dV2;       // km/s       - PostSN center of mass velocity vector
```

If I understand correctly, this is meant to be in the ZAMS CM frame.  If so, this ignores all information about earlier kicks (e.g., if this is the second SN).  [Also, "SHOULD BE BACK TO NICE UNITS NOW" in all caps seems superfluous -- it's enough to state what the units are.]

6.     

`double              m_SemiMajorAxisPreSN;                                               // Semi-major axis prior to 2nd supernova`

Really, just before the second one?  Why does that need a special class level variable (and shouldn't 2nd be reflected in the variable name)?


7.    

` double              m_SemiMajorAxisPrev;                                                // Semi-major axis at previous timestep		     double              m_SemiMajorAxisPrev;                                                // Semi-major axis at previous timestep double              m_SemiMajorAxisPrime;                                               // Semi-major axis `

m_SemiMajorAxisPrime shouldn't even be in the comments

reinhold-willcox:
Thanks Ilya, I'll respond point by point.

> This may well be doing the right thing. I trust you that it is, @reinhold-willcox . But I found the code very difficult to read, so I stopped after the first couple of files. A few comments, rough in order of the BaseBinaryStar files that I read:
> 
> 1. Speed and velocity still seem to be intermittently used, e.g.,
> 
> ```
> Vector3d natalKickVector = m_Supernova->SN_KickVelocity() * Vector3d(cos(theta)*cos(phi), 
>                                                                           cos(theta)*sin(phi),
>                                                                           sin(theta));
> ```
> 
> [the SN_KickVelocity() function returns a speed, right?]
> 
Agreed, that is inconsistent. It requires several larger changes though (the Options input, and samplers as well), but maybe it is a good time to do that.
 
> 2. 
> ```
> // The quantities below are mostly meaningless 
>   * Resolves supernova event - one of the stars has gone supernova!		         m_OrbitalVelocityPreSN = -nan("");
>   *		         m_uK = m_Supernova->SN_KickVelocity() / m_OrbitalVelocityPreSN; // -- - Dimensionless kick velocity
> ```
> 
> What happens when you divide by nan? Why would you intentionally do that?
>
Agreed, I think m_uK is just nan as well, but I should probably just set it this way instead. 

> 3.  
> `double m1_ = m_Supernova->Mass(); // Mo - SN star postSN mass `
> 
> We seem to be introducing different notation from the rest of the code -- this will make it harder for future maintainers (not to mention that it's really not obvious that a following underscore means post-SN, while the absence of an underscore means pre-SN).
> 
> 4. 
> ```
> Vector3d V = Vector3d( (-omega*a*a/r)*sinEA,   
>                                 (omega*a*a/r)*sqrt(1-e*e)*cosEA,  
>                                  0.0                            );        // km/s      - Relative velocity vector
> ```
> 
> I find single-letter variable names are quite confusing (and then there's dV1, dV2, etc.), and this isn't helped by the comments ("Relative velocity vector" relative to what?)
> 
See longer comments at the bottom.

I will fix the comment about the relative velocity vector to clarify that it's M2 relative to M1. 

> 5. 
> ```
> Vector3d Vcm_ = (-m2*dm1/(mb*mb_) + m1*dm2/(mb*mb_)) *V 
>  	//                       SHOULD BE BACK TO NICE UNITS NOW                        //		                                                  + (m1_/mb_) *dV1 
>  	///////////////////////////////////////////////////////////////////////////////////		                                                  + (m2_/mb_) *dV2;       // km/s       - PostSN center of mass velocity vector
> ```
> 
> If I understand correctly, this is meant to be in the ZAMS CM frame. If so, this ignores all information about earlier kicks (e.g., if this is the second SN). [Also, "SHOULD BE BACK TO NICE UNITS NOW" in all caps seems superfluous -- it's enough to state what the units are.]
> 
This is meant to be in the current CoM frame. If it's the 2nd SN, the final velocity (either systemic or component) is added to the previous systemic velocity to give a total velocity in the ZAMS CoM. 

I'm not sure where the "SHOULD BE BACK..." part comes from, I don't see that in my branch. Did you copy from the Github files changed page? It might have copied over the full line, including the old code. 

> 6. 
> `double m_SemiMajorAxisPreSN; // Semi-major axis prior to 2nd supernova`
> 
> Really, just before the second one? Why does that need a special class level variable (and shouldn't 2nd be reflected in the variable name)?
> 
> 7. 
> `double m_SemiMajorAxisPrev; // Semi-major axis at previous timestep double m_SemiMajorAxisPrev; // Semi-major axis at previous timestep double m_SemiMajorAxisPrime; // Semi-major axis`
> 
> m_SemiMajorAxisPrime shouldn't even be in the comments

Agreed, I didn't clean up some of the old comments properly. I'll get on that.

---

The only major issue I gather from this is notation. The variable names were chosen to match that of Pfahl, Rappaport, Podsiadlowski 2002 as closely as possible, with underscores replacing primes. I had thought that made it easier to follow, but maybe that's because I spent so long looking at it. If you think it is unreadable and inconsistent with the rest of the code, I can rewrite with longer names.

ilyamandel (approved):
@reinhold-willcox , you are absolutely right about most of your responses -- the NICE UNITS was in the old version of dev and not in yours; you do add the old and new system velocities (I missed the +=)...  I apologise, reading big code changes from the Files changed page is clearly not optimal.  I am changing this to approve.  I can go through the rest more easily (for my own knowledge) after doing a dev pull.  Feel free to tweak the minor things at your discretion before pulling (e.g., the comment mentioning m_SemiMajorAxisPrime really does seem to be a new addition -- it's not in the current dev).

jeffriley:
Ok, I'll add a comment on the use of nan().

> What happens when you divide by nan? Why would you intentionally do that?

Part (i): you get NaN

Part (ii): Reinhold's in a bit of a bind here (well, not Reinhold personally, but he's writing this code).

The problem is that now the the binary is unbound, m_OrbitalVelocityPreSN and m_uK are meaningless and shouldn't be used.  Reinhold is hoping to use NaN as a sentinel - so if someone does use those variables, and uses them to calculate other variables, then eventually what will be printed in the output files is NaN, and that will signal that somebody did the wrong thing.  At least that's what I think Reinhold's motivation is.  He's in a bind because, although that's not really best practice (I don't think), what else can he do?  he can set both values to 0.0 and hope nobody uses them when they shouldn't.  Or he could just leave them as they are and hope nobody uses them when they shouldn't.  Either way we might end up with NaN or INF anyway, or (probably worse), we might just end up with numbers that are wrong but we don't recognise as being wrong.

Setting NaN and getting NaN in the output means we recognise that there's a problem.  The problem with that strategy though is we don't know whether the NaNs we see are a result of (this) design (a feature), or some other defect.  So, hard to track down.  Not setting NaN means, as mentioned above, we might end up in the same position, or worse, with an insidious defect that we might not recognise.

So, that's why I chose not to get upset by it :-)

I think the real answer (in this case, and in the future), is to have another class of binary (the UnboundBinary class), that we evolve to when the binary becomes unbound - and in that class the variable that we shouldn't be using don't exist...  But that's probably a future enhancement.

reinhold-willcox:
@jeffriley @ilyamandel fixes were made (except to the single letter variable names which will come later). If you like them, feel free to squash & merge!

jeffriley (changes requested):
Some questions:

(i) you have changed some options (velocity -> magnitude).  Do you need to change the corresponding options in pythonSubmitdefault?  I didn't see changes to pythonSubmitDefault, but maybe I just missed them.  Similarly, the documentation will need to be changed to reflect the changed options (sorry if you changed it and I missed it).

(ii) you have changed some of the columns in the grid file (again, velocity->magnitude).  Did you change the sample grid file to reflect the changes?  Did you change the documentation to reflect the changes? (and again sorry if you did and I just missed it).

(iii) You need to undo the changes you made to the historical changelog entries (where you changed velocity->magnitude etc.).  You can't change history!  Those comments relate to the source at the time, which still has "velocity", not "magnitude" (and you can't change that...) - so for the comments to be consistent with the source they need to remain as they were written to the changelog.  I have actually used the changelog to guide me to changes that were made in the source at some earlier time to find the source of a change (that may have resulted in a defect).  If the comments don't match the source anymore that would be problematic.

(iv) have you run a test with a grid file with the new headers/variable names?  Since you changed that code you should exercise it to see if it does what you expect.

(v) have you run a test with the new/changed options and pythonSubmit?  Again, since you changed that code you should exercise it to see if it does what you expect.

Otherwise, all good as far as I can see.

reinhold-willcox:
> Some questions:
> 
> (i) you have changed some options (velocity -> magnitude). Do you need to change the corresponding options in pythonSubmitdefault? I didn't see changes to pythonSubmitDefault, but maybe I just missed them. Similarly, the documentation will need to be changed to reflect the changed options (sorry if you changed it and I missed it).
> 
I thought I did, but it didn't make it up the ladder. It's there now.

> (ii) you have changed some of the columns in the grid file (again, velocity->magnitude). Did you change the sample grid file to reflect the changes? Did you change the documentation to reflect the changes? (and again sorry if you did and I just missed it).
> 
Sample grids are fixed, though other places (e.g stroopwafel) probably need to be changed separately. 

> (iii) You need to undo the changes you made to the historical changelog entries (where you changed velocity->magnitude etc.). You can't change history! Those comments relate to the source at the time, which still has "velocity", not "magnitude" (and you can't change that...) - so for the comments to be consistent with the source they need to remain as they were written to the changelog. I have actually used the changelog to guide me to changes that were made in the source at some earlier time to find the source of a change (that may have resulted in a defect). If the comments don't match the source anymore that would be problematic.
> 
Sorry about that - had a little too much fun with sed regex...

> (iv) have you run a test with a grid file with the new headers/variable names? Since you changed that code you should exercise it to see if it does what you expect.
> 
Confirmed, the systems run as expected.

> (v) have you run a test with the new/changed options and pythonSubmit? Again, since you changed that code you should exercise it to see if it does what you expect.
> 
Yes, all good on that front!

> Otherwise, all good as far as I can see.

jeffriley (approved):

jeffriley:
We're good to go :-)

ilyamandel:
And merged. 👍

ilyamandel:
Hmm.  @reinhold-willcox , a run with pythonSubmitDefault.py fails -- did you check it?

$ python3 pythonSubmitDefault.py 
python_version = 3
compas_executable_override None
/Users/ilyam/Work/COMPAS/COMPAS/src/COMPAS --use-mass-loss --massTransfer --pair-instability-supernovae --pulsational-pair-instability --common-envelope-allow-main-sequence-survive --number-of-binaries 10 --metallicity 0.0142 --common-envelope-alpha 1.0 --common-envelope-lambda 0.1 --common-envelope-slope-Kruckow -0.8333333333333334 --common-envelope-alpha-thermal 1.0 --common-envelope-lambda-multiplier 1.0 --luminous-blue-variable-multiplier 1.5 --wolf-rayet-multiplier 1.0 --mass-transfer-fa 0.5 --mass-transfer-jloss 1.0 --maximum-evolution-time 13700.0 --maximum-number-timestep-iterations 99999 --initial-mass-min 5.0 --initial-mass-max 150.0 --initial-mass-power 0.0 --semi-major-axis-min 0.01 --semi-major-axis-max 1000.0 --mass-ratio-min 0.0 --mass-ratio-max 1.0 --minimum-secondary-mass 0.1 --eccentricity-min 0.0 --eccentricity-max 1.0 --pulsar-birth-magnetic-field-distribution-min 11.0 --pulsar-birth-magnetic-field-distribution-max 13.0 --pulsar-birth-spin-period-distribution-min 10.0 --pulsar-birth-spin-period-distribution-max 100.0 --pulsar-magnetic-field-decay-timescale 1000.0 --pulsar-magnetic-field-decay-massscale 0.025 --pulsar-minimum-magnetic-field 8.0 --orbital-period-min 1.1 --orbital-period-max 1000 --kick-magnitude-sigma-CCSN-NS 265.0 --kick-magnitude-sigma-CCSN-BH 265.0 --fix-dimensionless-kick-magnitude -1 --kick-direction-power 0.0 --random-seed 0 --mass-transfer-thermal-limit-C 10.0 --eddington-accretion-factor 1 --PISN-lower-limit 60.0 --PISN-upper-limit 135.0 --PPI-lower-limit 35.0 --PPI-upper-limit 60.0 --maximum-neutron-star-mass 2.5 --kick-magnitude-sigma-ECSN 30.0 --kick-magnitude-sigma-USSN 30.0 --kick-scaling-factor 1.0 --maximum-mass-donor-Nandez-Ivanova 2.0 --common-envelope-recombination-energy-density 15000000000000.0 --common-envelope-mass-accretion-max 0.1 --common-envelope-mass-accretion-min 0.04 --zeta-main-sequence 2.0 --zeta-radiative-envelope-giant 6.5 --kick-magnitude-max -1.0 --log-level 0 --debug-level 0 --single-star-mass-steps 10 --single-star-mass-min 1.0 --single-star-mass-max 75.0 --case-bb-stability-prescription ALWAYS_STABLE --chemically-homogeneous-evolution NONE --mass-loss-prescription VINK --mass-transfer-angular-momentum-loss-prescription ISOTROPIC --mass-transfer-accretion-efficiency-prescription THERMAL --mass-transfer-rejuvenation-prescription STARTRACK --initial-mass-function KROUPA --semi-major-axis-distribution FLATINLOG --mass-ratio-distribution FLAT --eccentricity-distribution ZERO --rotational-velocity-distribution ZERO --remnant-mass-prescription FRYER2012 --fryer-supernova-engine DELAYED --black-hole-kicks FALLBACK --kick-magnitude-distribution MAXWELLIAN --kick-direction ISOTROPIC --outputPath /Users/ilyam/Work/COMPAS/COMPAS/src --common-envelope-lambda-prescription LAMBDA_NANJING --stellar-zeta-prescription SOBERMAN --mass-transfer-thermal-limit-accretor CFACTOR --pulsational-pair-instability-prescription MARCHANT --neutron-star-equation-of-state SSE --pulsar-birth-magnetic-field-distribution ZERO --pulsar-birth-spin-period-distribution ZERO --common-envelope-mass-accretion-prescription ZERO --envelope-state-prescription LEGACY --logfile-delimiter COMMA 
Program Options error: unrecognised option '--kick-magnitude-sigma-CCSN-NS'

ilyamandel:
Ignore -- sorry!  I was in such a rush to explore that I forgot to make clean before compiling. :( :)

All is well so far!  Nicely done, @reinhold-willcox !

reinhold-willcox:
Phew, making me nervous! Glad it worked :)

PR #341, opened 2020-08-28
Revert "SN kicks and vector addition of velocities"

reinhold-willcox:
Reverts TeamCOMPAS/COMPAS#336

Some new bugs cropped up with these changes. We'll submit another PR once those have been addressed.


ilyamandel (approved):
Temporarily approving until we've sorted this out.

Issue #306, opened 2020-07-30
Unphysical hard binaries are created


ilyamandel:
COMPAS finally succeeded in breaking through the speed of light barrier!  It can create binaries with orbital velocities of >10^{10} km/s, and separations of <1 m.  Of course, such binaries are unphysical.  An example is seed 606249 from the Mandel+ stochastic BPS solar metallicity run.

Preliminary testing indicates that such binaries arise when we enforce the stability of case BB mass transfer. In this case, we can have extreme hardening if the donor is much more massive than the accretor (by a factor of ~40 in this example, which has mass transfer from a HeHG star onto a 0.1 solar mass MS accretor). The extreme mass ratio in the context of essentially fully non-conservative mass transfer leads to extreme hardening.  The proposed solution is to include a test in the mass transfer function that enforced stable MT resulting in an RLOF configuration should be declared a merger.

ilyamandel:
I've implemented a test for stars touching (a proxy for L2 overflow and hence merger) after stable mass transfer, which resolves this issue, in PR #343 .  Closing.

Issue #70, opened 2020-02-20
Check for merger during mass transfer episode


SimonStevenson:
From @serenavinciguerra89 

To check whether both stars fit within the binary at all 'timesteps' during a mass transfer episode.

Since these stars are not in equilibrium, we don't know their 'real' radii. But we can at least make some guess as to what their radii are. 

Options include:

1) check that both stars fit in the orbit at all times based on their initial radii
2) check that both stars fit in the orbit at all times based on their final radii
3) check that both stars fit in the orbit at all times assuming their largest radii (pessimistic)
4) check that both stars fit in the orbit at all times assuming their smallest radii (optimistic)

I favour 1) for simplicity.

avigna:
I guess we could add a `contact_during_mass_transfer` flag and set it to true if it occurs.

ilyamandel:
The trouble with option 1 is that a star may be stripped of its envelope during a MT episode.  We don't necessarily expect the entire initial star to fit in, just the final star, which would suggest using option 4 from @SimonStevenson 's list.  However, that radius is typically not available during MT itself, as the stars haven't yet been adjusted in the code, i.e., the check must happen later.

Another challenge is that we do allow for contact binaries at ZAMS now, as part of CHE modelling.  So it's not as simple as disallowing all binaries that overflow RLOF at the end of the MT episode, even if we do use final radii.

I am still looking for an elegant solution (this is also connected to Issue #306 ).

ilyamandel:
As mentioned above, we do not keep track of the detailed stellar evolution during mass transfer.  Therefore, we can only require that stars do not overfill their Roche lobes immediately after a mass transfer episode.  However, even that is allowed: e.g., we do allow for stars to overfill Roche lobes after a common-envelope event, and we do allow stars to be long-term over-contact in the context of CHE evolution.  Therefore, the only requirement we can reasonably make is that stars should not be "touching" right after a mass transfer episode -- which is really a rough proxy for L2 overflow, which we assume leads to merger.  This is the check I've implemented in PR #343 .  I am therefore closing this issue.

[@serenavinciguerra89 and @SimonStevenson -- if you have specific proposals that can be realistically implemented within the context of COMPAS, please feel free to open a new issue with such a proposal.]

PR #343, opened 2020-08-29
Minor updates and bug fixes for merged binaries

ilyamandel:
- Address issue #306  by removing detailed printing of merged binaries
- Address issue #70  by stopping evolution if the binary is touching
- Check for merged binaries rather than just touching binaries in Evaluate
- Minor cleaning (e.g., removed unnecessary CheckMassTransfer, which just repeated the work of CalculateMassTransfer but with a confusing name)

Checks from comparison runs against previous code version: 
- No differences in BSE_DoubleCompactObjects and BSE_Supernovae output files
- Expected minor differences in BSE_CommonEnvelopes output files due to moving stellar updates in response to CE evolution inside this function (before printing)
- Expected minor differences to DetailedOutputs (records no longer printed once binaries merge, since such records do not contain meaningful values, anyway)


jeffriley (approved):
Approved

Issue #347, opened 2020-08-30
CalculateMassTransferOrbit may use the incorrect accretion efficiency fraction


ilyamandel:
CalculateMassTransfer() correctly computes the accretion efficiency fraction, based in part on the prescription for MassTransferThermallyLimitedVariation().  However, CalculateMassTransferOrbit() recomputes the accretion efficiency fraction without taking this into account, leading to incorrect results for MT_THERMALLY_LIMITED_VARIATION::RADIUS_TO_ROCHELOBE. 

Solution: pass the correctly computed mass transfer efficiency fraction from CalculateMassTransfer() to CalculateMassTransferOrbit() as an additional parameter.

ilyamandel:
Addressed in PR #348, closing.

PR #348, opened 2020-08-30
Fixing CalculateMassTransferOrbit bug #347

ilyamandel:
- Fixed issue #347: CalculateMassTransferOrbit was not correctly accounting for the MT_THERMALLY_LIMITED_VARIATION::RADIUS_TO_ROCHELOBE option
- Assorted very minor cleaning, including comments

Checks: no difference to previous runs under default settings.


jeffriley (approved):
Approved

PR #357, opened 2020-09-03
Add optimistic CEE and immediateRLOFpostCEE masks to ClassCOMPAS.py and update the python notebook

themikelau:
At @jeffriley 's request. This ClassCOMPAS.py has been used for his work.


ilyamandel (approved):
Looks good to me (though I am not a python expert).  A few optional comments:

"raise ValueError("types=%s not in BBH, BNS, BHNS" % (types))" -- isn't "all" an allowed input?

"self.optimisticmask = maskPessimistic" -- yikes.  That a very confusing variable name.  Should we change it?

"I assume each metallicity has the same limits, and does correction factor" -- edit for grammar

"so dont redo if we reset the data" - don't missing apostrophe :)

"sorry not the prettiest line is a boolean slice of seeds
this only works because seeds in systems file and DCO file are printed in same order"

I do not understand what line this refers to; in any case, (A) we probably should not assume this; (B) this is not a good style of comments for a public code

"Stuff of data I dont need for integral
 but I might be to laze to read in myself
and often use. Might turn it of for memory efficiency". -- again, edit for grammar and style appropriate to public code

"Possibility to test assumptions of True solar mass evolved" -- I did not understand what this comment referred to

themikelau:
Hi Ilya, thanks for going into so much detail. I will fix the ValueError and possibly rename self.optimisticmask to self.pessimistic mask.

As for the commenting style and grammar, I think I can work on making these comments on all the post-processing code and notebooks styled more for a public code in the future.

ilyamandel:
Sounds good, @themikelau .  Note that this is already approved, so you can pull as soon as you feel ready.

PR #358, opened 2020-09-05
reduce HYDROGEN_RICH and HYDROGEN_POOR flags to a single IS_HYDROGEN_POOR

### LINKED ISSUE #354: Combine HYDROGEN_RICH and HYDROGEN_POOR output variables into a single variable

We currently have separate variables, and so separate columns in the output file, for the flags HYDROGEN_RICH and HYDROGEN_POOR.  One is the boolean complement of the other - we only need one flag, and so one column.

The legacy code included both rich and poor flags.  I left it like that when I refactored for the sake of backwards compatibility, but I don't see any need for that now.  I think it's time to consolidate them.

**Additional context**

themikelau:
Would close Issue #354 . @jeffriley please let me know if you think this is a sensible implementation.


jeffriley (changes requested):
Changes look ok.  Need to update the changelog and version number.

themikelau:
Thanks Jeff, I have updated the changelog and version number to 2.14.01

jeffriley (approved):

PR #359, opened 2020-09-08
CompasHPC improvements

### LINKED ISSUE #12: Move enhancement STROOPWAFEL from Legacy COMPAS to new COMPAS
this is for me

### LINKED ISSUE #18: double check STROOPWAFEL works in newCOMPAS

### LINKED ISSUE #154: Test compatibility of CompasHPC and BSE_Grid.txt
Users can specify a list of binaries to evolve using a `BSE_grid.txt` file. 

I suspect that this may not work as intended in conjunction with CompasHPC.

My concern is that if the user runs a grid file of N_lines run on HPC using N_batches, presumably the intended result is for each batch to run N_lines/N_batches (in order to speed up the run). Currently, I think you would get N_batches versions of the same run. I haven't tested this though, but unless someone else has implemented the splitting of the file, I think this will be an issue.

reinhold-willcox:
Changes related to CompasHPC and stroopwafel sampling. Also, to pythonSubmitDefault.py.

Removed CompasHPC directory - all functionality now in `defaults/stroopwafel_interface_default.py`.

`defaults/stroopwafel_interface_default.py` runs an interface to stroopwafel, so will sample over a list of input parameters to produce grid files, then run these grids across a specified number of cores. Need to run `pip install stroopwafel` first. 

The file works on its own, but if there is a `pythonSubmit.py` in the same dir, and the appropriate flag has been set, `interface.py` will run with the options in the pythonSubmit (i.e combining stroopwafel sampling with pythonSubmit list of options).

In order to combine them, the structure of pythonSubmit has been changed so that compas arguments are now collected into a dictionary. This doesn't change how most users interact with it, but it means the "hypergrid" and "hyperlist" functionality are no longer supported. I know some people wanted to keep this for the time being, until stroopwafel sampling is more mainstream, so the old file has been renamed to "pythonSubmitDefault_old.py". 

Works also on ozstar, but you'll need to run `pip install stroopwafel --user` there first. 

Note, the "Adaptive Importance" part of stroopwafel is still not working, that will come later.

Documentation added in `docs/sampling.md` and `docs/compasHPC.md`.


jeffriley (commented):
The changes look good to me, but I know next to nothing about OzStar or the Stroopwafel implementation, especially Loki's recent work - I'll leave the decision to approve to Loki and Simon.

reinhold-willcox:
Ok, cool. It currently conflicts because of the changelog, but I'll wait for approval from the maestro's  before updating.

ilyamandel:
Pinging @lokiysh , @SimonStevenson , @FloorBroekgaarden .

defaults/stroopwafel_interface_default.py
```diff
@@ -0,0 +1,286 @@
...
+            It also has a subprocess which will run under the key process. Rest, it depends on the user. User is free to store any information they might need later 
+            for each batch run in this dictionary. For example, here I have stored the 'output_container' and 'grid_filename' so that I can read them during discovery of interesting systems below
+    OUT:
+        compas_args (list(String)) : This defines what will run. It should point to the executable file along with the arguments.
+        Additionally one must also store the grid_filename in the batch so that the grid file is created
+    """
+    batch_num = batch['number']
+    grid_filename = os.path.join(output_folder, 'grid_' + str(batch_num) + '.csv')
+    output_container = 'batch_' + str(batch_num)
+    compas_args = [compas_executable, '--grid', grid_filename, '--output-container', output_container, '--random-seed' , np.random.randint(2, 2**63 - 1)]
```
lokiysh: Do we really require random seed here? Wont it overwrite the one in pythonSubmit ?

defaults/stroopwafel_interface_default.py
```diff
@@ -0,0 +1,286 @@
...
+    # STEP 3: Initialize the stroopwafel object with the user defined functions and create dimensions and initial distribution
+    dimensions = create_dimensions()
+    #sw_object.initialize(dimensions, interesting_systems, configure_code_run, rejected_systems, update_properties_method = update_properties)
+    sw_object.initialize(dimensions, None, configure_code_run, None, update_properties_method = update_properties)
+
+
+    intial_pdf = distributions.InitialDistribution(dimensions)
+    # STEP 4: Run the 4 phases of stroopwafel
+    sw_object.explore(intial_pdf) #Pass in the initial distribution for exploration phase
+    sw_object.adapt(n_dimensional_distribution_type = distributions.Gaussian) #Adaptaion phase, tell stroopwafel what kind of distribution you would like to create instrumental distributions
```
lokiysh: There is no need to run the adaptation phase if we are not running the refinement phase. So we can comment out this line as well, it will unnecessarily use up resources

docs/sampling.md
```diff
@@ -0,0 +1,77 @@
...
+`mc_only` specifies if you would like to do naive MC sampling only. Currently, this option must be set to True
+
+`run_on_hpc` specifies if you are running on a High-Performance Computer (HPC). 
+If so, see [docs/compasHPC.md](compasHPC.md) for assistance.
+
+`output_folder` a string specifying the output folder. Relative paths will be appended onto the current directory path.
+
+`output_filename` a string specifying the name of the output samples file.
+
+`debug` for development purposes.
```
lokiysh: Good to add here is that debug is specifically used to print the output / error that COMPAS will sendover, which is otherwise hidden. It has nothing to do with debugging stroopwafel.

defaults/pythonSubmitDefault.py
```diff
@@ -498,234 +498,92 @@ def listCommands(self):
...
     programOptions = pythonProgramOptions()
+    commandOptions = programOptions.generateCommandLineOptionsDict()
+
+    #-- Convert options into a shell string
+    shellCommand = combineCommandLineOptionsDictIntoShellCommand(commandOptions)
 
-    runCompas(programOptions)
+    #-- Run exectute COMPAS shell string
+    print(shellCommand)
+    call(shellCommand,shell=True)
```
lokiysh, line 604: I assume with this we are still keeping the ability to run pythonSubmit independently ?

defaults/pythonSubmitDefault.py
```diff
@@ -498,234 +498,92 @@ def listCommands(self):
...
-    for i in range(nString):
-
-        if not stringChoices[i] == None:
-
-            command += stringCommands[i] + ' ' + stringChoices[i] + ' '
-
-    for i in range(nList):
-
-        if listChoices[i]:
+    def generateCommandLineOptionsDict(self):
```
lokiysh, line 517: nothing particularly bad, but just wondering what changed the indentation? If it works, it works!

defaults/stroopwafel_interface_default.py
```diff
@@ -0,0 +1,286 @@
...
+    debug = namespace.debug #If True, will print the logs given by the external program (like COMPAS)
+    run_on_hpc = namespace.run_on_hpc #If True, it will run on a clustered system helios, rather than your pc
+    mc_only = namespace.mc_only # If you dont want to do the refinement phase and just do random mc exploration
+    output_filename = namespace.output_filename #The name of the output file
+    output_folder = os.path.join(os.getcwd(), namespace.output_folder)
+
+    # Set commandOptions defaults - these are Compas option arguments
+    commandOptions = dict()
+    commandOptions.update({'--outputPath' : output_folder}) 
+    commandOptions.update({'--logfile-delimiter' : 'COMMA'})
```
lokiysh: This should not be hardcoded here! What if the user tries to change it from pythonSubmit, its not being respected.

defaults/stroopwafel_interface_default.py
```diff
@@ -0,0 +1,286 @@
...
+    run_on_hpc = namespace.run_on_hpc #If True, it will run on a clustered system helios, rather than your pc
+    mc_only = namespace.mc_only # If you dont want to do the refinement phase and just do random mc exploration
+    output_filename = namespace.output_filename #The name of the output file
+    output_folder = os.path.join(os.getcwd(), namespace.output_folder)
+
+    # Set commandOptions defaults - these are Compas option arguments
+    commandOptions = dict()
+    commandOptions.update({'--outputPath' : output_folder}) 
+    commandOptions.update({'--logfile-delimiter' : 'COMMA'}) 
+    compas_executable = os.path.join(os.environ.get('COMPAS_ROOT_DIR'), 'src/COMPAS') # Location of the executable
```
lokiysh: better to put this at the top of the file, since we are saying somewhere in the file "Do not change below or something".

defaults/stroopwafel_interface_default.py
reinhold-willcox: It doesn't quite work to leave the random seed in from the pythonSubmit, because the same seed is passed to each new batch (i.e for each call of COMPAS). It's not a huge problem, since the grids should all be different (which covers most of what was previously handled by the random number), but any numbers that are pulled purely from a random draw will be repeated. 

A possible fix to this would be use the random-seed from pythonSubmit, but incremented by the number of binaries in a given batch times the index of the batch. I'll give that a go and see if I run into any problems.

defaults/pythonSubmitDefault.py
reinhold-willcox: By redefining the function as a method of the pythonProgramOptions class, it made it easier to import only the class into the stroopwafel interface.

defaults/pythonSubmitDefault.py
reinhold-willcox: Yes, the internals are different, but to most users the pythonSubmit can still be configured and run like previously. The only difference is the removal of the hyper-parameter grids, which hopefully will be implementable using the new sw interface.

defaults/stroopwafel_interface_default.py
reinhold-willcox: Ya I agree there's a bit of a degeneracy here.

Ultimately, it's always going to be a little ambiguous which file wins out in the scenario where sw and pySub are both used. But in this case, since people would be actually running the sw_interface file, I figured it makes more sense to accept the sw value in stead of the pySub value. 

I'm open to discussion / disagreement on this, but in the meantime I'll try to make it clearer in the comments.

defaults/stroopwafel_interface_default.py
reinhold-willcox: Fair point

docs/sampling.md
reinhold-willcox: Oh really? I didn't know that, thanks!

defaults/stroopwafel_interface_default.py
reinhold-willcox: agreed, that's been fixed

defaults/stroopwafel_interface_default.py
reinhold-willcox: this has now been fixed

lokiysh (approved):

Issue #253, opened 2020-05-04
Implement supernova output for SSE


SimonStevenson:
**Is your feature request related to a problem? Please describe.**
For binaries, we have several summary outputs that summarise supernovae, common envelope episodes and double compact object binaries. Most of these don't make sense for single stars, so we don't have them, but a file tracking all supernovae that occur in a population of single stars would make sense.

**Describe the solution you'd like**
Implement supernova output for SSE

**Describe alternatives you've considered**
In principle, all of the data is available in the detailed outputs, but this would be a useful data product.

**Additional context**
NA

jeffriley:
@SimonStevenson @reinhold-willcox Have either of you done anything with this?  I'm doing some work on the logging code and I could implement the SSE supernova output as part of that.  Let me know if you want me to do it, and if you want to provide a list of attributes to include in the file.

SimonStevenson:
No sorry Jeff I haven't looked at this. If it would fit naturally with the restructuring you are doing, then that would be great.

I guess we want all of the parameters from the normal supernova file that make sense for a single star. I think that is:
ID     	    SEED    	Drawn_Kick_Velocity_SN	Applied_Kick_Velocity_SN	Fallback_Fraction_SN	SN_Kick_Theta_SN	SN_Kick_Phi_SN	SN_Type_SN	Mass_Total@CO_SN     Mass_CO_Core@CO_SN	   Mass_SN    		Stellar_Type_SN	Supernova_State	Stellar_Type_Prev_SN	Mass_Core@CO_SN	Mass_He_Core@CO_SN	      Time      	Systemic_Velocity	Hydrogen_Rich_SN	Hydrogen_Poor_SN

reinhold-willcox:
Haven't looked at this, thanks for taking it over Jeff!

jeffriley:
[...]

PR #360, opened 2020-09-09
Enhancements and related code cleanup:

jeffriley:
(1) implemented DETAILED_OUTPUT folder inside COMPAS_Output container for SSE output
(2) SSE Parameters files moved to DETAILED_OUTPUT folder (they are analogous to BSE_Detailed_Output files)
(3) implemented SSE Switch Log and BSE Switch Log files (record written at the time of stellar type switch - see documentation)
(4) implemented SSE Supernova log file - see documentation (issue #253)
(5) added TIMESCALE_MS as a valid property in BaseStar::StellarPropertyValue().  The TIMESCALE_MS value in the SSE_Parameters file was being printed as ERROR! and nobody noticed :-)  It now prints correctly.

Rudimentary testing done - files created correctly.  No changes made to anything other than logging code (except for trivial change for (5) above), so contents of logfiles should be unchanged. More rigorous testing is always a good thing, so reviewers please test rigorously :-)

@ilyamandel Switch Log files here - feel free to change the default record specifications.  Like other log files they can be changed at run time, so the defaults are not so important.  Use new program options --BSEswitchLog and --SSEswitchLog to enable.

@reinhold-willcox @SimonStevenson SSE Supernova file is here - feel free to change the default record specifications.  Like other log files they can be changed at run time, so the defaults are not so important.


src/GiantBranch.cpp
```diff
@@ -1721,6 +1721,15 @@ STELLAR_TYPE GiantBranch::ResolveSupernova() {
         }
             
     	CalculateSNKickVelocity(m_Mass, m_SupernovaDetails.totalMassAtCOFormation - m_Mass, stellarType);
+
+        // stash SN details for later printing to the SSE Supernova log
+        // can't print it now because we may revert state (in Star::EvolveOneTimestep())
+        // will be printed in tar::EvolveOneTimestep() after timestep is accepted (i.e. we don't revert state)
```
SimonStevenson: Typo : Star
```suggestion
        // will be printed in Star::EvolveOneTimestep() after timestep is accepted (i.e. we don't revert state)
```

src/Log.cpp
```diff
@@ -986,6 +986,181 @@ std::tuple<bool, LOGFILE> Log::GetStandardLogfileKey(const int p_FileId) {
...
+ * single star in SSE and revert to the previous state if we find we've evolved too far and
+ * possibly missed something interesting, we can't write a record to the SSE supernova file
+ * at the time it occurs because if we revert to the previous state after writing the record
+ * we can't (easily) unwind the write - and the logging code is written to only create a log
+ * file on the first write to the file (so that we don't have files created that have no
+ * records other than the header if we never write to them), so even if we could unwind the
+ * write, we might have a file created that may never have data records written to it (we
+ * could specifically check for that and delete the file, but that's a little inelegant -
+ * better to not create the file in the first place).  So, this function will enable me to 
+ * format a SSE Supernova record at the righ time, but delay writing it to after we decide
```
SimonStevenson: Typo right
```suggestion
 * format a SSE Supernova record at the right time, but delay writing it to after we decide 
```

docs/COMPAS_LaTex/sections/UserGuide/COMPASOutput.tex
```diff
@@ -0,0 +1,239 @@
...
+For BSE, if BSE Detailed Output log files (see the \textit{\texttt{-{}-}detailedOutput} program option), or BSE Switch log files (see the \textit{\texttt{-{}-}BSEswitchLog} program option), are created, they will be created inside a containing directory named `Detailed\_Output' within the COMPAS output container directory.
+
+Also created in the COMPAS container directory is a file named `Run\_Details' in which COMPAS records some details of the run (COMPAS version, start time, program option values etc.).
+
+COMPAS defines several standard log files that may be produced depending upon the simulation type (Single Star Evolution (SSE), or Binary Star Evolution (BSE), and the value of various program options. The standard log files are:
+
+\begin{itemize}
+\itemsep0pt
+\item{SSE Parameters log file \\ Records detailed information for a star during SSE.}
+\item{SSE Switch log file \\ Records detailed information for a star at the time of each stellar type switch during SSE. Enable with program option \textit{\texttt{-{}-{BSEswitchLog}}}.}
```
SimonStevenson: I think this should be SSE
```suggestion
\item{SSE Switch log file \\ Records detailed information for a star at the time of each stellar type switch during SSE. Enable with program option \textit{\texttt{-{}-{SSEswitchLog}}}.}
```

SimonStevenson (changes requested):
Looks good Jeff, I compiled and ran a bunch of single stars and checked that the SSE_Supernova file was created, and seems to look good. 

Do we need to add 

```
logfile_SSE_supernova = None
logfile_SSE_detailed_output = None
```

to `pythonSubmitDefault.py`? 

I also highlighted a few typos in some of the files below.

I found it a bit difficult to read through this changed, since there are many changed files (52!)

jeffriley:
Quite right Simon - thanks!  I've addressed the typos and updated pythonSubmitdefault.py.

Yes, 52 files changed, but most of those are the new LaTex documentation - you probably don't need to review the source of those - just check the PDF.

SimonStevenson (approved):

jeffriley:
Just noticed this:

> logfile_SSE_detailed_output = None

SSE detailed output is SSE_Parameters files - there is no mechanism to turn them off.  Up until now, they were the only artefact of SSE - but now the SSE Supernova file will be created, and the user can select the SEE Switch Log.

Given that we now have the SSE Supernova file, should we add an option to disable toe SSE_Parameters file - i.e. the SSE detailed output?

PR #364, opened 2020-09-10
Custom caller/callee profiling option

einshoe:
These changes are for the "custom caller/callee reporting" of duplicate arguments to the "pow" function.
I incorrectly applied these to my forked "production" branch but this one here in the "dev" branch is what I would like to see merged into the TeamCOMPAS "dev" branch - That's the safest way.
BTW:  The "custom caller/callee report" is enabled with the --profiling option.


ilyamandel:
Thank you, @einshoe -- @jeffriley  and I will take a look this afternoon!

ilyamandel:
Need to edit a couple of things, including the changelog -- will do that in a moment, after pulling. :)

ilyamandel (approved):

PR #368, opened 2020-09-11
Enhancement to profiling code

jeffriley:
(1) refactored profiling code
(1a) profiling code can now be #defined away for production build
(1b) added options (via #defines) to profiling code: counts only (no CPU spinning), and print calling function name
(4) removed profiling program option

@ilyamandel I haven't tested with VTunes - could you check that please (it should work ok)

Also, there are some calls to utils::intPow() that aren't profiled - feel free to add them if you want.

See comments at the top of profiling.h for instructions on turning profiling on/off and configuring options.

The code is easily extensible - we could easily add profiling for (e.g.) sqrt(), log10() etc. (though pow() is the priority)


ilyamandel (approved):
Happy to approve, but given recent changes to dev (including to changelog), please make sure you update those before pulling.  [I'd do this myself, but I don't think I can change a branch that lives in your git repo rather than the COMPAS one -- a downside of this style of branching.]

Issue #367, opened 2020-09-10
Understand changes in DNS distribution for fiducial model


SimonStevenson:
**Describe the bug**
I have run a a population of a million binaries with COMPAS using default settings for things like kicks (so Fryer + high kicks for CCSN + low kicks for ECSN/USSN), and found only a handful of fast-merging DNSs, and they are of the very high eccentricity variety. I found that the distribution of eccentricities of DNSs from this new COMPAS run is skewed to high eccentricities; we seem to have almost lost the low eccentricity population altogether.

![newCOMPAS_Porb_e_day](https://user-images.githubusercontent.com/6952112/92819231-eb6f9f00-f40b-11ea-99dc-e50f6bb46ee3.png)

I also found that the SN_Type_SN flag in BSE_Supernovae.csv has only 2 values for DNSs, 1 and 2, which I believe correspond to CCSN (type1) and ECSN (type2). I expected to see USSNe in the SN_Type_SN flag output.

![newCOMPAS_SN_type](https://user-images.githubusercontent.com/6952112/92819201-e579be00-f40b-11ea-91cc-88e303f3be98.png)

**To Reproduce**
Run COMPAS with pythonSubmitDefault, analyse DNSs

**Expected behavior**
DNS distribution would be consistent with previous COMPAS results (e.g. Vigna-Gomez et al. 2018,2020; Chattopadhyay et al. 2020; Mandel et al. 2020). I expected to see USSNe in the SN_Type_SN flag output.

**Screenshots**
See above

**Versioning (please complete the following information):**
 - COMPAS [v02.14.00]

**Additional context**
NA

ilyamandel:
Sorry, ran out of time yesterday, but self-assigning.

ilyamandel:
A short run of 10,000 binaries indicates that the total mass equals the CO core mass at the time of the SN 209 times.  These are clearly ultra-stripped, but the SN type is recorded as 1 (CCSN) in all of these cases.  I'll sort out why it is incorrectly recorded once I am done with homeschooling / other meetings.

ilyamandel:
Update: 
Ultra-stripped stars that are massive enough to become compact objects become ONeWDs after ultra-stripping.  However, ONeWDs always go through ECSNe, not USSNe.
Solution:
These stars should go through SNe right after ultra-stripping -- they are not WDs and shouldn't be called that.

ilyamandel:
[...]

PR #369, opened 2020-09-11
Correct resolution of supernovae of ultra-stripped stars

ilyamandel:
Changed ultra-stripped HeHG and HeGB stars to immediately check for supernovae before collapsing into WDs; this resolves issue #367


jeffriley (approved):
Approved

Issue #271, opened 2020-06-19
Resolve supernovae sequentially


ilyamandel:
When two supernovae happen simultaneously, the instantaneous mass loss is unrealistically too large.  In practice, we do not expect the evolution of two stellar companions to be so similar that supernovae ever happen within an orbital period of each other.  

We should instead always resolve two supernovae sequentially, adjusting the orbital parameters in between and picking a random location along the eccentric orbit for the second SN if appropriate.

reinhold-willcox:
@ilyamandel I'm wondering about the best way to approach this. A potentially simple way to do this is to evaluate the SNe in the same timestep but update the orbital parameters in between the two and randomly sample in the anomaly. But more generally, the problem arises when two stars have exactly equal mass, which is also unrealistic. 

Is it worth it to add in a check for precisely equal mass stars, and if true, to perturb one slightly? The benefit to varying the mass is that the SNe will naturally happen in different timesteps, which is more realistic and would make the detailed plots clearer.

If equal masses only come from initial conditions, we could demand that q is strictly < 1. But maybe @jeffriley can remind me if there's a condition anywhere that sets the two masses equal (e.g for contact, non-merging binaries).

jeffriley:
> @jeffriley can remind me if there's a condition anywhere that sets the two masses equal

We equilibrate the masses of the constituent stars (and circularise the orbit of the binary) in the following circumstances:

(a) At birth, if OPTIONS->AllowRLOFAtBirth() is TRUE and either of the constituent stars is overflowing its Roche lobe.  

If after equilibrating and circularising either of the stars is overflowing its Roche lobe:

If OPTIONS->AllowTouchingAtBirth() is FALSE we declare a merger and discard the binary.  
If OPTIONS->AllowTouchingAtBirth() is TRUE we allow the binary to survive, but it will merge at the first timestep (so will be in the results).

(b) During evolution (in BaseBinaryStar::InitialiseMassTransfer()), if CHE is enabled and both stars are CHE, and either of the stars is overflowing its Roche lobe (and masses are not already equal).


> Is it worth it to add in a check for precisely equal mass stars, and if true, to perturb one slightly?

[...]

PR #373, opened 2020-09-12
Fixing defect in supernova resolution

ilyamandel:
In BaseBinaryStar::ResolveSupernova, we were using PrevMass values for both the star and the companion to adjust the orbit on supernova. Those values were set at the start of the tilmestep. If something else happened to change the mass significantly (e.g., mass transfer), we could end up in a situation where the binary mistakenly thought that much more mass was lost during a supernova than was in fact lost, leading to excessive eccentricity or even unbinding. This PR implements a fix to only account for the difference in the pre- and post- SN mass of the exploding object, as reported by SN_TotalMassAtCOFormation(). It addresses issues #367 and #271 .


jeffriley (approved):
Approved

PR #377, opened 2020-09-17
New vector objects, which are used for system and component velocities

### LINKED ISSUE #314: Units of orbital velocity in output file
The column Orb_Velocity<SN in the BSE_Supernovae.csv file is labeled as having units of km/s, but in fact appears to contain the velocity in m/s.  If so, it should be recomputed (divided by 1000) before printing to contain the correct quantity.

reinhold-willcox:
- Added new vector objects for SN-related velocities.
- Rewrote the ResolveSupernova function (and related) for clarity and to align with Pfahl, Rappaport, Podsiadlowski 2002 (which handles the SN vector calculation with ease).
- Fixed the issue #314 where the orbital velocity output had the wrong units.

_Note:_ this should be the last cosmetic change from what was the NS_Kicks PR. Non-cosmetic changes to come may impact population rates, and will take more time to investigate. 

Also, @ilyamandel, I changed the variable names from the previous PR to more closely align with the COMPAS standard format. 

Edit: also changed some variables with "prime" names to remove the prime - this was decided in a previous Compas code jam.


jeffriley (commented):
The code changes look ok, but I'd prefer @ilyamandel  review them since he's actually made recent changes in this area of the code.

@reinhold-willcox have you run a sizeable test with these changes and compared the results to the current COMPAS version?

And

> this should be the last cosmetic change 

"Rewrote the ResolveSupernova function (and related) for clarity and to align with Pfahl, Rappaport, Podsiadlowski 2002" isn't just a cosmetic change, is it?  If it is, maybe you don't need the run to test whether this produces results similar to the current version.

reinhold-willcox:
By cosmetic, I mean the output should be identical. In a test of 100 binaries, it was completely identical, but a larger test is probably a good idea.

reinhold-willcox:
@ilyamandel fixed the "magnitude magnitude" that slipped through

ilyamandel (commented):
Line 1637 of BaseBinaryStar.cpp:

        double omega = sqrt(G_SN*totalMassPrev / (semiMajorAxisPrev_km*semiMajorAxisPrev_km*semiMajorAxisPrev_km));          // orbits/s  - Keplerian orbital frequency

Should be rad/s, not orbits/s

ilyamandel (commented):
Do we need to store m_TotalMass, m_TotalMassPrev, etc.?  Are these not just extra variables to maintain (and risk going out of synch)?

ilyamandel (commented):
Changelog needs to be updated with version number.

ilyamandel:
"Set the component velocites to the system velocity (to simplify the UpdateComponentVelocity function)"

Really, it's not a component velocity, it's the velocity of the star (if the star is unbound) or the binary (if the binary is bound), correct?  A more clear comment (or a more clear variable name, though I can't think of one) might help.

ilyamandel:
Note to self: ResolveCoreCollapseSN() should not take Fryer engine as argument.

ilyamandel (approved):
Approved for merging; changes mentioned above can be made post-merge.

reinhold-willcox:
@ilyamandel went to correct your comments and saw you already had! Thanks for taking care of those.

PR #388, opened 2020-10-03
Minor code cleaning and comment fixing

ilyamandel:
- Removed some unnecessary internal variables and functions (m_TotalMass, m_TotalMassPrev, m_ReducedMass, m_ReducedMassPrev, m_TotalAngularMomentumPrev, CalculateAngularMomentumPrev(), EvaluateBinaryPreamble(),...
- Cleaned up some comments
- ResolveCoreCollapseSN() no longer takes the Fryer engine as an argument (Fryer is just one of many possible prescriptions)

### DIFF
--- a/src/BaseBinaryStar.cpp
+++ b/src/BaseBinaryStar.cpp
@@ -325,15 +325,12 @@ void BaseBinaryStar::SetRemainingCommonValues() {
                                                                             gyrationRadius1,
                                                                             gyrationRadius2);
 
-    m_TotalAngularMomentumPrev                   = m_TotalAngularMomentum;
-	m_TotalMass 					             = m_Star1->Mass() + m_Star2->Mass();
-	m_TotalMassPrev						         = m_TotalMass;
-	m_ReducedMass					             = (m_Star1->Mass() * m_Star2->Mass()) / m_TotalMass;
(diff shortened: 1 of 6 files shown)


jeffriley (commented):
@ilyamandel As you have noted, you have removed the parameter to GiantBranch::ResolveCoreCollapseSN(), and now call CalculateRemnantMassByFryer2012() with the retrieved option OPTIONS->FryerSupernovaEngine().

I haven't checked the legacy code, but I suspect GiantBranch::ResolveCoreCollapseSN() was written to take the parameter because in thelegacy code different SNengines were passed to CalculateRemnantMassByFryer2012() based on option values.  It may be that's not necessary (I defer to you - I'd only be parroting what I saw in the legacy code), but I think that's the only place we call CalculateRemnantMassByFryer2012() - so why not ditch the parameter entirely and just have CalculateRemnantMassByFryer2012() retrieve the option value directly?

ilyamandel:
@jeffriley : Good suggestion.  I am in the middle of making fixes to address Tom's issues, so to avoid having too many things rolled in to one PR, can you approve this one [it really is cosmetic], and I'll address your point in the next one?

jeffriley (approved):
Approved per Ilya's reply.

PR #390, opened 2020-10-03
Updating production from dev

ilyamandel:
We have made a number of bug fixes on dev in the past month+, from the treatment of supernovae to the evolution of main sequence stars.  I think it's time to update production to incorporate these.


reinhold-willcox:
We may decide that this is the version to go in the methods paper, so is it worth having everyone run their own tests to confirm that this is a good LSR?

ilyamandel:
@reinhold-willcox -- I can bet you that there are still bugs present.  I am also sure that there are fewer bugs than in the previous iteration, because we fixed a bunch, so this will definitely be an improvement. :)

PR #396, opened 2020-10-08
Added check to avoid NaNs is fractional radial change

ilyamandel:
CalculateRadialChange() now checks to see if the previous radius is negative or zero before dividing by it.


jeffriley (commented):
Do we really want to use utils::Compare here?  It probably won't hurt (especially for now because we don't actually implement a tolerance), but it means the check will pick up radii <= +tolerance, not just <= 0.0 (absolute)

jeffriley (approved):

ilyamandel:
Yup, that was intentional -- I figured that if we are dividing by something within tolerance of zero, our results are likely to be untrustworthy to numerical error, so it's better to declare this as zero fractional change.

jeffriley:
Yeah - thought so, but mentioned it just in case...

PR #402, opened 2020-10-13
separation to semimajor axis

### LINKED ISSUE #393: Separation to SemiMajorAxis
Proposed enhancement: there are a number of places where we refer to separation when we mean semi-major axis. Most notably in the output headers, but also in some places in the code. 

I think this is needlessly misleading, but I'd like to hear what others think before making the changes. It may take some careful trawling through the code to make sure this is done right.

Correct/consistent usage of separation and semi-major axis.

Leave it as is.

Edit: I'm happy to do this, but wanted to get other's thoughts before diving in.

reinhold-willcox:
changed separation to semimajoraxis where applicable.

note: RLOF and BeXRBs still use separation for semimajoraxis, but these pertain to others' projects, so I left those as is.


jeffriley (changes requested):
Three things:

(i) You mention you didn't change "separation" -> "semi-major axis" in BeBinaries and RLOF because they are someone else's project.  I'm not sure that's the right choice.  If we implement this PR we have "semi-major axis" in some header files, as code variables in some functions, and printed in the running commentary as the evolution happens - but  in other places we have left "separation".  I think that's liable to be more confusing than what the PR was intended to address.  I think if you're going to change it, then change it everywhere - unless you truly believe that "separation" as used in the RLOF printing code and the BeBinaries code is actually not referring to semi-major axis (in which case we'd better document that because everybody seems to use them interchangeably).

(ii) you didn't update the changelog or the version string.

(iii) you didn't update the documentation.

reinhold-willcox:
@jeffriley My last few commits addressed your changes, let me know if you see other issues.

jeffriley (approved):
All good

PR #406, opened 2020-10-21
Default files cleanup

reinhold-willcox:
Removed '_default' extension from default files - it is obvious from the directory. Also fixed some of the default values in stroopwafelInterface to be more realistic.

I did not touch default values in pythonSubmit (noting because this was discussed recently).


ilyamandel (changes requested):
Are the online and latex documentation properly updated?  E.g., I see references to  pythonSubmitDefault.py in getting_started.md .

reinhold-willcox:
Ah true, I need to go fix those up.

reinhold-willcox:
Fixed up the docs. The changelog will conflict with the other PR (or vice versa depending on which gets approved first) but that's an easy fix.

ilyamandel (approved):

PR #415, opened 2020-10-29
updated demo plotting routine

reinhold-willcox:


themikelau:
Text on the new demo plot seems to be overlapping. Also, I don't think you should delete BSE_Grid_demo.txt, it is referenced in the tutorial to teach users how to use grid files.

themikelau (changes requested):
As above

themikelau:
Okay I see that you hadn't deleted BSE_Grid_demo.txt, but renamed it. In that case, all references to BSE_Grid_demo.txt in the demo markdown must also be updated

reinhold-willcox:
Text on the demo plot overlaps when first produced, but if you expand the window it doesn't overlap anymore. I'm not sure how to fix that, we might need to call on someone who's better with matplotlib. 

Agreed about the reference to the gridfile name, I'll fix that.

ilyamandel (commented):
Is the Grid_demo already aligned with Jeff's recent grid-file changes?  I am about to approve that PR, so would be good to get this one to match if necessary.

ilyamandel (commented):
@themikelau is more qualified than I am to review.  Just one question: does this already include the optimistic<->pessimistic change in the ipython notebook that I hit during yesterday's code hack?  We should fix that, too.

themikelau:
> @themikelau is more qualified than I am to review. Just one question: does this already include the optimistic<->pessimistic change in the ipython notebook that I hit during yesterday's code hack? We should fix that, too.

This pull request does not involve modifications to Coen's post-processing notebooks, so the flag is still inconsistently labelled as "optimistic" in the notebooks. But I am still waiting for changes that would allow plot_demo.png to be rendered properly.

reinhold-willcox:
@themikelau This PR is now up to date with all the other recent merges into dev. If you're happy with it, could you merge it in?

themikelau (approved):

PR #421, opened 2020-11-02
Remove tmp hack

reinhold-willcox:
Removed unnecessary supernova phi rotation - it was added to agree with Simon's original definition, and to allow for seeds to reproduce the same SN final orbit. 

Removing it means seeds won't reproduce the same systems before and after, but populations are unaffected.


jeffriley (approved):
Approved

PR #424, opened 2020-11-03
Schneider remnant mass

reinhold-willcox:
Added in the Schneider remnant mass prescription. This can be accessed in the python submit as 
`remnant_mass_prescription = 'SCHNEIDER2020'` or 
`remnant_mass_prescription = 'SCHNEIDER2020ALT'` for the alternative model for effectively single stars.

This also required a parameter for which type of MT case the SN star went through, but this can be ambiguous (different groups have different definitions of MT Cases). To avoid any complications from this, I added a new parameter `m_MassTransferDonorHistory`, which is a vector of ints (stellar types), and tracks all of the different stellar types that a star was when it was a MT donor. This can generally be converted into MT Cases, according to whichever definition you use. 

In the output, this parameter shows up as a string with format #-#-#..., so that something like 2-8 means there was HG MT followed by HeHG MT. It does not consider any events where the star was an accretor.


reinhold-willcox:
I'll add: this works, but it is not done especially well. Jeff will probably quickly see a more elegant way to implement the MassTransferDonorHistory parameter, and others will recognize that this new parameter makes several others redundant. But those might be fairly substantial changes, so I didn't want to get into all of that here.

jeffriley (changes requested):
I think the implementation is generally fine - just a few comments.

**Makefile:** 

What's the rationale for the -fcompare-debug-second compiler option?

Also, why did you remove -Wall?



**BaseBinaryStar.cpp:**

2232    // Add event to MT history of the donor
2233    m_Donor->UpdateMassTransferDonorHistory(m_Donor->StellarType());

Will the parameter passed to m_Donor->UpdateMassTransferDonorHistory() always be the stellar type of the donor? If so, the donor knows what its stellar type is - no need to pass it as a parameter.



**BaseStar.h:**

351	// Star mass transfer history 
352    INT_VECTOR              m_MassTransferDonorHistory;             // List of MT donor stellar types - mostly relevent for binary stars

m_MassTransferDonorHistory should be a vector of STELLAR_TYPE, not a vector of INT.  If it is a vector of INT, I can add a value of -2735, or 2,567,234 to it - neither of which is a valid stellar type, and both of which would cause you to print incorrect history strings.  Typing the vector as STELLAR_TYPE restricts the values that can be added to it to the STELLAR_TYPE values.

Also, I think m_MassTransferDonorHistory is more properly an attribute of a BinaryConstituentStar than a BaseStar (as evidenced by your comment) - is there a reason it is in BaseStar and not BinaryConstituentStar?  And, because m_MassTransferDonorHistory is defined in BaseStar, as a user I could request that the MT donor history be printed for a single star in SSE mode (via the --logfile-definitions option) - the output would not be meaningful.



**BaseStar.cpp**

3090 void BaseStar::UpdateMassTransferDonorHistory(const STELLAR_TYPE p_DonorType) {

I think the description of this function is incorrect (the comments just prior to the function declaration).



**GiantBranch.cpp**

1046 double GiantBranch::CalculateRemnantMassBySchneider2020(const double p_COCoreMass, const bool useSchneiderAlt)

This was probably just an oversight, but just in case: our convention is that function parameter names start with 'p_', class member variables start with 'm_', and local variables have no prefix - that way when reading the code (particularly long functions) it's easier to tell what got passed in to the function, what is a class variable that may have been set somewhere else, and what is a local function variable.


1071 else if ((mtHist.size() == 1) && (mtHist[0] < 7)) {                                 // 1 MT event, trivially from a non-stripped star

I assume 'mtHist[0] < 7' is meant to check for any stellar type up to (but not including) Hurley type 7?  What happens if we change the Hurley type constants in constants.h?  It might be unlikely, but if we do this code will break.

We should not rely on:
(a) the Hurley types continuing to have the values they have now, or
(b) the order of the Hurley types

The stellar type of Chemically Homogeneous stars is 16 - they are akin to Main Sequence stars, but their stellar type is not < 7.

Look at the 'initializer_list' constructs just after the STELLAR_TYPE enum class in constants.h.  If there is not a list that matches your requirements, create a new one.  You can then test against the list using utils::IsOneOf() (there are examples throughout the code).



**Appendices/defaultLogFileRecordSpecs.tex**

It looks like you've changed the documentation for one or more of the default logfile records, but I don't see corresponding changes to the code - was the documentation incorrect?

jeffriley:
Just a followup on this function:

double GiantBranch::CalculateRemnantMassBySchneider2020(const double p_COCoreMass, const bool useSchneiderAlt)

which you call with the 'useSchneiderAlt' parameter set true or false depending upon whether you want to use the alternative version.  That's fine, but here's a suggestion for an alternative implementation:

Declare the function in GiantBranch.h as

double CalculateRemnantMassBySchneider2020(const double p_COCoreMass, const bool p_UseSchneiderAlt = false);

Also declare another function:

double CalculateRemnantMassBySchneider2020Alt(const double p_COCoreMass) { return CalculateRemnantMassBySchneider2020(const double p_COCoreMass, true); }

Then in the code where you want to use Schneider2020, you just call:

CalculateRemnantMassBySchneider2020(p_COCoreMass);

and where you want to use Schneider2020Alt, you call:

CalculateRemnantMassBySchneider2020Alt(p_COCoreMass);


The declarations are a little more work (but not much), and then when people are reading the code it is easier to tell what the author of the code wants to happen with the calls - that is, the meaning of CalculateRemnantMassBySchneider2020(p_COCoreMass) or CalculateRemnantMassBySchneider2020Alt(p_COCoreMass) is clearer than CalculateRemnantMassBySchneider2020(p_COCoreMass, false) or CalculateRemnantMassBySchneider2020(p_COCoreMass, true)  (as a reader I don't need to go and figure out what that second parameter does).

Just a thought.

reinhold-willcox:
> I think the implementation is generally fine - just a few comments.
> 
> **Makefile:**
> 
> What's the rationale for the -fcompare-debug-second compiler option?
> 
> Also, why did you remove -Wall?
> 

That's my mistake. I have those in my Makefile to hide all the annoying notes and warnings, but it was not meant to be committed. 

> **BaseBinaryStar.cpp:**
> 
> 2232 // Add event to MT history of the donor
> 2233 m_Donor->UpdateMassTransferDonorHistory(m_Donor->StellarType());
> 
> Will the parameter passed to m_Donor->UpdateMassTransferDonorHistory() always be the stellar type of the donor? If so, the donor knows what its stellar type is - no need to pass it as a parameter.
> 

That's true, that was a relic from when I couldn't decide if this should be in BaseStar or BinaryConstituentStar. I'll remove it. 

> **BaseStar.h:**
> 
> 351 // Star mass transfer history
> 352 INT_VECTOR m_MassTransferDonorHistory; // List of MT donor stellar types - mostly relevent for binary stars
> 
> m_MassTransferDonorHistory should be a vector of STELLAR_TYPE, not a vector of INT. If it is a vector of INT, I can add a value of -2735, or 2,567,234 to it - neither of which is a valid stellar type, and both of which would cause you to print incorrect history strings. Typing the vector as STELLAR_TYPE restricts the values that can be added to it to the STELLAR_TYPE values.
> 
> Also, I think m_MassTransferDonorHistory is more properly an attribute of a BinaryConstituentStar than a BaseStar (as evidenced by your comment) - is there a reason it is in BaseStar and not BinaryConstituentStar? And, because m_MassTransferDonorHistory is defined in BaseStar, as a user I could request that the MT donor history be printed for a single star in SSE mode (via the --logfile-definitions option) - the output would not be meaningful.
> 

A vector of INT was easier to implement, but I see your point why it's not the bes choice. I'll try to fix that up. 

The parameter will only ever have the default value for single stars, but its still relevant for the remnant mass prescription, so I think it makes sense to keep it in BaseStar.

> **BaseStar.cpp**
> 
> 3090 void BaseStar::UpdateMassTransferDonorHistory(const STELLAR_TYPE p_DonorType) {
> 
> I think the description of this function is incorrect (the comments just prior to the function declaration).
> 

Yes, true, I'll fix that.

> **GiantBranch.cpp**
> 
> 1046 double GiantBranch::CalculateRemnantMassBySchneider2020(const double p_COCoreMass, const bool useSchneiderAlt)
> 
> This was probably just an oversight, but just in case: our convention is that function parameter names start with 'p_', class member variables start with 'm_', and local variables have no prefix - that way when reading the code (particularly long functions) it's easier to tell what got passed in to the function, what is a class variable that may have been set somewhere else, and what is a local function variable.
> 

Ah yes, thanks, good catch.

> 1071 else if ((mtHist.size() == 1) && (mtHist[0] < 7)) { // 1 MT event, trivially from a non-stripped star
> 
> I assume 'mtHist[0] < 7' is meant to check for any stellar type up to (but not including) Hurley type 7? What happens if we change the Hurley type constants in constants.h? It might be unlikely, but if we do this code will break.
> 
> We should not rely on:
> (a) the Hurley types continuing to have the values they have now, or
> (b) the order of the Hurley types
> 
> The stellar type of Chemically Homogeneous stars is 16 - they are akin to Main Sequence stars, but their stellar type is not < 7.
> 
> Look at the 'initializer_list' constructs just after the STELLAR_TYPE enum class in constants.h. If there is not a list that matches your requirements, create a new one. You can then test against the list using utils::IsOneOf() (there are examples throughout the code).

True, I'll add that in. 

> 
> **Appendices/defaultLogFileRecordSpecs.tex**
> 
> It looks like you've changed the documentation for one or more of the default logfile records, but I don't see corresponding changes to the code - was the documentation incorrect?

Yes, I could've explained that more explicitly, but the Supernova output was out of date. I think I forgot to fix it when I added the vector velocities. 

---

Regarding your second comment: I think that implementation would involve some repeated code, or else confusing redirects to avoid repeated code. The only difference between the standard and alt prescriptions is in the effectively single star case, so all of Cases A, B, & C are the same in both. If you can think of a way to set this up which avoids repeated or redirected code, I'm happy to consider it.

jeffriley:
> The parameter will only ever have the default value for single stars, but its still relevant for the remnant mass prescription, so I think it makes sense to keep it in BaseStar.

But a single star never engages in mass transfer, right?  Only constituent stars of binaries do that - why would we want mass transfer information in a class that doesn't engage in mass transfer?  We seem to have mass transfer code in BaseStar, so putting t there is ok for now.  We should probably take all mass transfer code out of BaseStar and put it in BinaryConstituentStar if we can.  Maybe that's for later - I'll look at the code and open an issue.  

> Regarding your second comment: I think that implementation would involve some repeated code, or else confusing redirects to avoid repeated code. The only difference between the standard and alt prescriptions is in the effectively single star case, so all of Cases A, B, & C are the same in both. If you can think of a way to set this up which avoids repeated or redirected code, I'm happy to consider it.

Repeated code?  The extra function declaration?  That isn't an issue if it makes things clearer - it's actually preferred.  That's not a confusing redirect - it's a fairly standard construct to make code easier to read, used by programmers everywhere...  Adding boolean parameters that change the code path taken in a function tends to obfuscate the intention of the function call.  As a reader of the code I have to chase down the function and figure out what the function does if the parameter is true, and what it does is false.  Having a function name that reflects what the function does is much easier to read and understand.  The aim is to write code that's easy to follow, and if adding a few lines of code achieves that, then we should add the extra lines of code rather than obscure the functionality.

ilyamandel (changes requested):
Incorrectly copied comments:

 * Convert Mass Transfer Donor History vector into string
  *
  * This is so that a string is passed to the output, not a vector of stellar types.
  *
  * std::string BinaryConstituentStar::MassTransferDonorHistoryString() 
  *
  * @return                              string of dash-separated stellar type numbers
  */
 void BaseStar::UpdateMassTransferDonorHistory

ilyamandel (changes requested):
I think we should make the decision based on the *last* MT episode, not the *first* MT episode.  We should iterate with Fabian, but I am not convinced this is the right approach on physical grounds.

reinhold-willcox:
> But a single star never engages in mass transfer, right? Only constituent stars of binaries do that - why would we want mass transfer information in a class that doesn't engage in mass transfer? We seem to have mass transfer code in BaseStar, so putting t there is ok for now. We should probably take all mass transfer code out of BaseStar and put it in BinaryConstituentStar if we can. Maybe that's for later - I'll look at the code and open an issue.
> 

I agree that it's a bit funny to have a MT parameter in a single star, but we need to have some sort of default for the Schneider prescription. I don't think that is straightforward if the parameter is in BinaryConstituentStar. 

> Repeated code? The extra function declaration? That isn't an issue if it makes things clearer - it's actually preferred. That's not a confusing redirect - it's a fairly standard construct to make code easier to read, used by programmers everywhere... Adding boolean parameters that change the code path taken in a function tends to obfuscate the intention of the function call. As a reader of the code I have to chase down the function and figure out what the function does if the parameter is true, and what it does is false. Having a function name that reflects what the function does is much easier to read and understand. The aim is to write code that's easy to follow, and if adding a few lines of code achieves that, then we should add the extra lines of code rather than obscure the functionality.

Hmm. I'll take your word that it's clearer for experienced coders. I'll give this a go.

reinhold-willcox:
@ilyamandel It's not clear to me that any of the models in Schneider should apply if we have multiple rounds of MT. If the argument is that stripping at some point changes how the core grows, then stripping again should further affect the growth. We should definitely chat with him about how best to move forward.

@jeffriley I haven't finished with the fixes just yet, I still need to figure out the clearest way to setup the different function calls for std and alt.

jeffriley:
> @jeffriley I haven't finished with the fixes just yet, I still need to figure out the clearest way to setup the different function calls for std and alt.

No worries.  You don't like my suggestion?  I'm not sure how much clearer it could be...

ilyamandel:
@reinhold-willcox -- As you saw, I pinged Fabian for his input.

reinhold-willcox:
> > @jeffriley I haven't finished with the fixes just yet, I still need to figure out the clearest way to setup the different function calls for std and alt.
> 
> No worries. You don't like my suggestion? I'm not sure how much clearer it could be...

No I mean, figure out how to implement it. It probably is easy, but I'll take another look later today.

reinhold-willcox:
@jeffriley All of your changes have been addressed, so I think this is in an approvable state. When we decide what to do about multiple MT events, I'll add in new PRs for those updates.

jeffriley (changes requested):
Looks good.  Just a few things (none too onerous I don't think):

(i) remove the .tsv files and Run_Details

(ii) fix the Makefile

(iii) I think GiantBranch::CalculateRemnantMassBySchneider2020() needs an extra 'else' at line 1103.  What happens if mtHist.size() == 1 but firstCase is not in any of the sets of stellar types?  It falls through and returns exp(logRemnantMass) - and I think at that stage logRemnantMas is not initialised.  I think we should be defensive and not rely on firstCase being one of the expected stellar types - so set a default (and show a warning if you want).  If you don't want to show a warning you could just initialise logRemnantMass to the default at the start of the function, then if it is not set after that it will return the default.

(iv) I think the warning in the SHOW_WARN doesn't match the code - it says "Mass_Remnant = 1.0", but I think you changed that to 1.25.

and we're good.

I haven't tested the code - I assume you did :-)  I just reviewed the code changes.

reinhold-willcox:
@jeffriley Pushed up some new edits, let me know if I've missed any of your requested changes. 

@ilyamandel I've added in the 2 MT episode special cases, could you double check that the logic is correct?

jeffriley:
Scratch that... I'm tired :-)  Of course you have a default on the switch...

Good to go from my side.

jeffriley (approved):
All good :-)

jeffriley (changes requested):
Actually, no, I've changed my mind back again (I really am tired...).  The short story is that you can still get to the switch statement with schneiderMassTransferCase  uninitialised - and even though you have a default on the switch, that's still not right.

schneiderMassTransferCase  can be uninitialised if firstcase or secondcase are not one of the stellar types you expect (the code may be written so they always are at the moment, but there is no guarantee it will not be changed buy someone who either doesn't understand or just gets it wrong).  It can also be uninitialised in mtHist.size() >2 - could be one day for the same reasons.

So schneiderMassTransferCase  should be initialised in all cases, not just the cases on the happy path.  You could just initialise it to a default value when it is declared, then if it is not initialised on the way through to the switch it will at least have a default value when it gets there.

(Sorry - this is what I said the first time, then in my sleep-deprived state rescinded my objection, but having thought about it again I think you should fix it)

ilyamandel (commented):
pythonSubmit:
#dtypes.append(np.dtype(('S10'), 1))
do we need this new line, since it's commented out anyway?

ilyamandel (changes requested):
I agree with @jeffriley .  Also, personally, I am not a fan of switch statements that default on expected behaviour -- I think it's much clearer to spell out the expected behaviour (and then still have a default for the cases when there's unexpected behaviour, but maybe raise a warning in that case).  So I'd suggest explicitly treating MT from HeHG and HeGB donors.

Also, the prescription is coded up in a rather convoluted manner.  All you need to consider within the current prescription is 
mtHist[mtHist.size()-1] 
-- that's your last mass transfer episode.  The only if() you need then is a check on whether mtHist.size() == 0.

reinhold-willcox:
@jeffriley @ilyamandel I think I've implemented your changes. Let me know what you think.

jeffriley (approved):
Approved

ilyamandel (approved):

Issue #387, opened 2020-10-03
Change to having a single SwitchLog command line argument and single switch log file for the run


ilyamandel:
Avoid failures if BSEswitchLog is set to true while evolving single stars, or vice versa, by having a single SwitchLog command line argument that applies to both single and binary evolution.

Also, the SwitchLog should be a single file for the run. :)

ilyamandel:
Closing following the resolution in PR #428.

PR #426, opened 2020-11-05
Switchlog fix - issue #387

jeffriley:
Enhancement -
changed switchlog implementation so that a single switchlog file is created per run
(see Issue TeamCOMPAS#387 - note: single '--switch-log' option (shared SSE/BSE) implemented in v02.16.00)


jeffriley:
Oops - wrong branch

PR #428, opened 2020-11-05
Switchlog fix - issue #387

### LINKED ISSUE #387: Change to having a single SwitchLog command line argument and single switch log file for the run

jeffriley:
Enhancement -
changed switchlog implementation so that a single switchlog file is created per run
(see Issue TeamCOMPAS#387 - note: single '--switch-log' option (shared SSE/BSE) implemented in v02.16.00)

Close issue #387 when merged


ilyamandel (approved):
These look great -- thank you, Jeff!

PR #432, opened 2020-11-06
New fast cosmic integration routine

ilyamandel:
This adds my new cosmic integration routine that is meant to be faster and is able to integrate over metallicities sampled uniformly in log Z without binning.  In fact, it's not particularly fast at the moment because the detection probability calculator is too slow (and because I am still learning python).  At least the former will be easy to fix -- to be done in the next couple of days.


jeffriley (approved):
Approved - though I have to say the lack of whitespace makes the code hard for me to read - I'm used to more whitespace, especially around arithmetic operators and such...  I know the Python parser (or the C++ parser for that matter) doesn't need the whitespace, but my brain finds it much easier to parse with whitespace separating the operators from the operands (and, for that matter, Python reserved words (a space between the 'if' and the open paren is easier for me to parse...)) - but maybe that's just me.

ilyamandel:
OK, I'll keep that in mind for a future update, thank you!

PR #448, opened 2020-11-13
Enhancements, code cleanup

### LINKED ISSUE #411: Remove AIS from the COMPAS code
Since adaptive importance sampling is now happening externally to the COMPAS C++ code, we should remove all AIS-specific code from the C++ code.  This should not impact performance, but will keep the code streamlined and easy to maintain.

### LINKED ISSUE #417: (documentation) Semi-major axes in CE output should be documented with units of Rsol and not AU
Semi-major axes in CE output should be documented with units of Rsol and not AU in the documentation

### LINKED ISSUE #443: Unimplemented eccentricity distributions

The following eccentricity distributions are listed in the utils::SampleEccentricityDistribution() function:

    ECCENTRICITY_DISTRIBUTION::FIXED:              // FIXED - all systems have same initial eccentricity - not implemented
    ECCENTRICITY_DISTRIBUTION::IMPORTANCE  // IMPORTANCE - not implemented

Both return the default value of 0.0 (because neither of them are implemented).

This issue opened to decide what to do - remove them, or implement them.

Also, the distributions "THERMAL" and "THERMALISED" are treated as the same thing in the sampling function.  Can we settle on just one name?

Remove both, or implement both (I vote for remove both).

I think "FIXED" is superseded by the new --eccentricity option so is probably no longer necessary (that's just a guess from the name - I don't know what was actually planned).

Leave them be - a poor alternative...

**Additional context**

None - just looking for input.  If there are no objections over the next week to removing both I will do that.

Same deal for THERMAL and THERMALISED.  Knowing no better, I'll settle on THERMAL if I get no objections.

jeffriley:
(1) Added metallicity-distribution option: available distributions are ZSOLAR and LOGUNIFORM (see documentation)
(1.1) Added metallicity-min and metallicity-max options (for metallicity-distribution option)
(1.2) Metallicity is sampled if not explicitly specified via the --metallicity option - this was existing functionality, but no distribution was implemented: sampling always returned ZSOLAR.  This change adds the LOGUNIFORM distribution, and 'formalises' the ZSOLAR 'distribution'.
(2) Added MASS to default SSE_SYSTEM_PARAMETERS_REC
(3) Removed AIS code
(4) Removed variable 'alpha' from BinaryCEDetails struct - use OPTIONS->CommonEnvelopeAlpha()
(4.1) Removed BINARY_PROPERTY::COMMON_ENVELOPE_ALPHA - use PROGRAM_OPTION::COMMON_ENVELOPE_ALPHA
(5) Issue #443: removed eccentricity distribution options FIXED, IMPORTANCE & THERMALISE (THERMALISE = THERMAL, which remains)

Included changes for Issue #443  - earlier than I anticipated so it hasn't been the week I suggested I'd wait in the issue...

Also updated documentation per Issue #417 

Close Issues #411, #417,  and #443  when merged


jeffriley:
I included the ZSOLAR metallicity 'distribution' purely for backwards compatibility (that's what users get now if metallicity is sampled) - with the expectation that it will be removed when we finally move to a single 'convenience' distribution only for each of the initial attributes.

ilyamandel (commented):
Out of curiosity, why do you need to pass min and max metallicity into SampleMetallicity -- can't they be obtained from OPTIONS on the fly?

ilyamandel (approved):
Approving, but consider the question above.

jeffriley:
> Out of curiosity, why do you need to pass min and max metallicity into SampleMetallicity -- can't they be obtained from OPTIONS on the fly?

Good question.  The answer to the first part is because I just used the template of the sample eccentricity function without thinking too much about it.  The answer to the second part is yes, they can - as can the min and max for eccentricity if sampling eccentricity.  Passing them in as parameters technically allows more flexibility (we don't have to use the option values), but that's not a good argument - the options are there for this purpose so why wouldn't we use them?  So, you're absolutely right - I'll change that, and look at the other sampling functions to see if I can do the same.

jeffriley:
Ok, now I remember why the values are passed and not retrieved from the options: functions in utils.cpp don't have access to the OPTIONS object.  Some (not all) of the functions in utils.cpp are required before the OPTIONS object is constructed.  We could fix this specific issue by putting the sampling functions in a separate .cpp file, but I'm not sure we need to - I don't think passing the option values as parameters is egregious enough to do that.

I did make a few tweaks to the sampling functions while I was looking at the code though.

PR #452, opened 2020-11-16
Fixes to Schneider remnant mass

reinhold-willcox:
- Added HeHG to list of Case C MT types (if it is the only MT)
- Fixed call to exponentiate logremnantmass in schneider prescription. The parameter logRemnantMass is base 10, so the inverse call should be 10^(logRM) not e^(logRM).

### DIFF
--- a/src/GiantBranch.cpp
+++ b/src/GiantBranch.cpp
@@ -1074,8 +1074,9 @@ double GiantBranch::CalculateRemnantMassBySchneider2020(const double p_COCoreMas
                                                        STELLAR_TYPE::THERMALLY_PULSING_ASYMPTOTIC_GIANT_BRANCH, })) { // CASE C Mass Transfer - from EAGB or TPAGB 
             schneiderMassTransferCase = MT_CASE::C;
         }
-        else if ((mtHist.size() == 1) && 
-                 (utils::IsOneOf(mostRecentDonorType, { STELLAR_TYPE::NAKED_HELIUM_STAR_HERTZSPRUNG_GAP })))  {       // CASE C Mass Transfer from HeHG star - lost its envelope through winds before first MT 
+
+        // subtle corner case - mostly isolated star lost its envelope through winds before first MT on HeHG
+        else if ((mtHist.size() == 1) && (utils::IsOneOf(mostRecentDonorType, { STELLAR_TYPE::NAKED_HELIUM_STAR_HERTZSPRUNG_GAP })))  {       // CASE C Mass Transfer from HeHG star 
             schneiderMassTransferCase = MT_CASE::C;
         }
     }
@@ -1139,7 +1140,7 @@ double GiantBranch::CalculateRemnantMassBySchneider2020(const double p_COCoreMas
             logRemnantMass = 0.096910013; // gives MassRemnant = 1.25  
     }
 
-    return std::pow(logRemnantMass, 10.0);
+    return std::pow(10.0, logRemnantMass);
 
 }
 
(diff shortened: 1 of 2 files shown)


jeffriley:
Changelog & version?

Also, I think it reads better, and might be more flexible, if this:

    else if ((mtHist.size() == 1) && 
                 (utils::IsOneOf(mostRecentDonorType, { STELLAR_TYPE::NAKED_HELIUM_STAR_HERTZSPRUNG_GAP })))  {       // CASE C 
    Mass Transfer from HeHG star - lost its envelope through winds before first MT 
            schneiderMassTransferCase = MT_CASE::C;
        }

is broken into two and inserted between the existing:

    if (mtHist.size() == 0) {                                                           // No history of MT - effectively single star
        schneiderMassTransferCase = MT_CASE::NONE;
    }
    else { // (mtHist.size() > 0)  

So you would have:

    if (mtHist.size() == 0) {                                                           // No history of MT - effectively single star
        schneiderMassTransferCase = MT_CASE::NONE;
    }
    else if (mtHist.size() == 1) {
        if (utils::IsOneOf(mostRecentDonorType, { STELLAR_TYPE::NAKED_HELIUM_STAR_HERTZSPRUNG_GAP }))  {       // CASE C 
    Mass Transfer from HeHG star - lost its envelope through winds before first MT 
            schneiderMassTransferCase = MT_CASE::C;
        }
    }
    else { // (mtHist.size() > 1)  

That way the ifs for mtHist.size() could convert to a switch(mtHist.size()) if/when we want more granularity, and the '(mtHist.size() == 1)' could have additional code if required in the future.

reinhold-willcox:
@jeffriley That's not quite the exact logic, but I agree it could be clearer.

The logic is that if there is any MT history (size != 0), then look at the last element in the array, *unless* the array has size 1 AND its element is HeHG. It's a funny corner case, but it doesn't work to just use the last element throughout, because I don't want to include cases where there are 2+ episodes and the final episode is HeHG. 

I've left the order in, but added a comment to better explain why we want to include this corner case. 

I also fixed how I revert the log remnant mass - I was doing logRM^10 instead of 10^logRM. Should probably get more sleep...

jeffriley:
> unless the array has size 1 AND its element is HeHG

Ah - so you only want to treat mtHist.size() == 1 differently when the only entry is HeHG...?  Ok , got it.

jeffriley (approved):

ilyamandel:
@reinhold-willcox -- you weren't doing  logRemnantMass^10, you were doing exp(logRemnantMass),  i.e., e^logRemnantMass, which is a much more reasonable interpretation of log as ln.  

More seriously, I am puzzled by the decision to call an ultra-stripped supernova case C...

reinhold-willcox:
@ilyamandel ~~agreed about the base(e) base(10) confusion, it was a simple oversight.~~ I misread your comment. At first I confused base(e) with base(10), but then when I went to fix the mistake, I confused the order of arguments in std::pow, leading to the 10^x vs x^10 confusion.

The corner case with HeHG stars is a bit funny, but these are not ultrastripped stars. The requirement that mtHist.size() == 1 means that there was only ever one episode of MT. If it was a stellar type 8 at this episode, this means that the star went nearly its whole life without interacting, but after blowing off its outer layers due to winds it has naturally become a HeHG star. If this star then goes on to engage in MT with its companion, then it is almost a "post Case C" MT event. But since this really just means the core was unaffected before the end of CHeB, then I treat this as a case C star.

But now having written it down here, I'm second guessing myself. These stars have their He layer stripped off, so maybe they are ultra-stripped...

ilyamandel:
@reinhold-willcox --  I agree with what you wrote in the final paragraph: I think these *are* ultra-stripped stars.  Stripped stars lose their H envelope, ultra-stripped stars lose their He as well, which is exactly what happens here.

reinhold-willcox:
@ilyamandel Fair enough. I'll make another PR.

PR #454, opened 2020-11-16
fix typo in synchronisation timescale

### LINKED ISSUE #444: Mistake in the prefactor for the tidal synchronization for the dynamical (radiative) tide
**Describe the bug**
There is a typo in the prefactor for the timescale for tidal synchronization for the dynamical (radiative) tide.  The prefactor should be 5 * 2 ^ (5/3) but it is implemented as 52^ (5/3).

[
image
](url)

**Context**
Quite understandable where this typo came from. We can blame the typesetters of the original Zahn paper / our laziness to not rederive the results.  (Zero judgment, I made exactly the same mistake in my 2009 paper).  The typesetting in Hurley's paper is misleading. The space between the 5 and the 2 is too small.  But if you go back to the original paper by Zahn 1977 you can see that the correct version.  

The original paper by Zahn 1977 eq 5.7 states the following. Note the very tiny white space between the 5 and 2, which makes it look like 52. If you rederive it you see it should be 5 * 2 ^...
image

Confusingly in  Hurley et al. 2002 eq 44 (which was copied into compas) the factor 5 * 2^... became 52

image

**Label the issue**
`urgency_moderate` - Fix is very simple, we should correct this. 
`severity_minor`    - This probably does not significantly affect any of our results so far.

TomWagg:
Resolves #444 

(this is the smallest change ever I realise, but want to practice doing this properly). I created a new branch on my forked copy of the repo and now am trying to merge with dev in real COMPAS. Is that correct?

### DIFF
--- a/src/BinaryConstituentStar.cpp
+++ b/src/BinaryConstituentStar.cpp
@@ -384,7 +384,7 @@ double BinaryConstituentStar::CalculateSynchronisationTimescale(const double p_S
 
         case ENVELOPE::RADIATIVE: {                                             // solve for stars with radiative envelope (see Hurley et al. 2002, subsection 2.3.2)
 
-            double coeff2          = PPOW(52.0, 5.0 / 3.0);                      // JR: todo: replace this with a constant (calculated) value?
+            double coeff2          = 5.0 * PPOW(2.0, 5.0 / 3.0);                      // JR: todo: replace this with a constant (calculated) value?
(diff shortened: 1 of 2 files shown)


jeffriley (changes requested):
Code change is ok.  Need to update changelog.h and increment version number.

jeffriley (approved):
Approved.

PR #456, opened 2020-11-17
udpated grid files to new format

reinhold-willcox:
Updated grid files in the default folder to align with new syntax.

The new versions work, but are pretty bland, so please anyone feel free to change them to show more of the new functionality.


jeffriley (approved):
Bland's fine - they work as examples.  maybe late when we have more time we can provide more extensive documentation, examples, and demos.

Issue #431, opened 2020-11-06
Add Hurley LBV wind prescription


ilyamandel:
The currently used Belczynski LBV wind prescription allows high-mass stars to go beyond the Eddington limit.  The original Hurley+ (2000) model should avoid this.

jeffriley:
Implemented in v02.17.07

PR #458, opened 2020-11-18
Add option to change LBV wind prescription

TomWagg:
Resolve Issue #431 

I have created a new option to select the LBV wind mass loss prescription. Options are: NONE - no LBV winds, HURLEY_ADD - add Hurley LBV winds to other winds, HURLEY_REPLACE - replace other winds with Hurley LBV winds and BELCZYNSKI - replace other winds with Belcyznski winds.

I also made some slight cleanup changes to the other mass loss code:
- Replace a bunch of numbers with constants
- Consolidate checking of luminosity into NJ mass loss function (this also makes it so you check the luminosity on the MS which was not previously done, please confirm this wasn't a purposeful choice made previously)

Here's a plot comparing things at Z=0.001
image


jeffriley (changes requested):
Nice work Tom!  The changes wrt the added "physics functionality" look good to me, but perhaps @ilyamandel could also comment on that.

Here are my comments regarding the coding:

(1) So HURLEY_REPLACE just means use Hurley's prescription?  And BELCZYNSKI means use Belczynski's prescription?  Have you appended the "_REPLACE" to HURLEY to distinguish it from HURLEY_ADD?  That means HURLEY_REPLACE and BELCZYNSKI do the same thing (with different prescriptions), but they have different names.  Can we make them consistent with each other (so maybe just HURLEY and BELCZYNSKI - I don't think you need the "_REPLACE" to distinguish from HURLEY_ADD).


(2) Your changes to double BaseStar::CalculateMassLossRateNieuwenhuijzenDeJager():

Generally (there are always exceptions...) it is desirable to have a single entry point, and a single exit point, for functions (C++ doesn't allow multiple entry points, so we don't have to worry about that).  Occasionally, where it makes sense, it's not so bad to do some sanity checking at the start of the function and return there if you decide that you really shouldn't be there, but in this case I think it would be better to do something like:


double masslossRate = 0.0;

if (luminoisty > MINIMUM) {
    masslossRate = calculated value
}
else {
    masslossRate = 0.0;
}

return masslossRate;


It's not a huge issue in this function, but in longer functions, as a reader of the code, you don't really want to be looking for places in the function it might return unexpectedly (to you, the reader) - as a reader of the code it is better if you can assume that the function enters at the top, and exits at the bottom (with the exception I mention above - a possible immediate exit at the top, which should be obvious to the reader).


(3) BaseStar::CalculateMassLossRateLBV(const LBV_PRESCRIPTION p_LBV_prescription) - same comment regarding multiple exit points


(4) Your comment at BaseStar.cpp line 1624: "this should never happen given how function is called - remove?" - is p_Teff guaranteed to be >= VINK_MASS_LOSS_MINIMUM_TEMP?  What if someone who doesn't know that restriction exists adds some code and calls the function with p_Teff < VINK_MASS_LOSS_MINIMUM_TEMP?  There's a balance somewhere between performance and defensive code, but here I think defensive wins.


(5) BaseStar::CalculateMassLossRateVink() - same comment regarding multiple exit points

TomWagg:
Thanks Jeff! That all makes a lot of sense to me, I'll make those changes now. Thanks for explaining the multiple exit point thing, I hadn't thought about that but I completely agree your way is much clearer :)

**EDIT:** All changes made, ready for review again :)

jeffriley (approved):
Changes good - approved.

Just need to resolve merge conflicts before merging - they will be in changelog.h because Reinhold snuck in his merge :-) (you'll need to make sure your change description is in the right order, and the version number is incremented correctly (don't forget the version number on your comments as well as the actual version string)).

You can make the changes via github (it's a small edit, so I often just do it at the github webpage), or you can make it locally and push a modified changelog.h

(We're looking at automating the incrementing of the version number so we reduce merge conflicts - but really (from my perspective anyway) it doesn't happen often enough for it to be really annoying)

Once you've resolved the conflicts, feel free to merge.

jeffriley:
Ah - just occurred to me that we didn't update the documentation for the new option.  I'll do that - I'm updating the docs anyway.

PR #464, opened 2020-11-20
Schneider refix

reinhold-willcox:
Removed MT hist = 8 corner case (these stars should be considered ultra-stripped)


jeffriley (changes requested):
The version number on your comment in changelog needs to be incremented

jeffriley (approved):
I fixed it :-)  Approved.

PR #467, opened 2020-11-21
Enhancements, fixes and cleanup for mass loss

### LINKED ISSUE #463: Make the variable names in the mass loss routine, dms and dml, less headache-inducing.
The variable names in the mass loss routine dms and dml are headache-inducing.   dml changes meaning from line to line.  Would be great if we clean this and give more understandable names. 

Here is an example, but this occurs in several places. 
image

**Additional context**
This will be very helpful for further debugging and improving of the mass loss routines.

### LINKED ISSUE #465: WR-like winds being used for massive giant stars
**Describe the bug**
WR-like winds are being applied to massive giant stars, this is both non-physical and leads to 2-10x more mass loss than there should be. The reason is that the old formulation of WR-like winds are being used which didn't have a metallicity dependence.

**Label the issue**
`urgency_low`, `severity_minor`

**Expected behavior**
Stars in this range should really have NJ mass loss (or Reimers in some masses/Zs)

**Screenshots**
See masses ~10-30 Msun, purple lines are used and these should never appear (should be NJ or Reimers)
image

### LINKED ISSUE #466: Add variable for tracking dominant mass loss rate being used
I'd like to implement in COMPAS a feature that stores the dominant mass loss rate at each timestep (similar to stellar type at each type). This is useful for seeing which winds are used at different phases.

I've implemented this on a branch on my fork but wanted to check whether this is something that other people would find useful being making a pull request. @jeffriley this would add a class variable, do you try to keep those to a minimum or would this be okay?

**Additional context**
E.g. I have used this to make plots like this
image

TomWagg:
1. Resolves #463 - change variable names to more sensible names and make sure we don't overload them
2. Resolves #465 - This _especially_ should be reviewed. I changed the code to only use the metallicity dependent Hamann & Koesterke recipe to solve the issue I reference. This should not have any major effects on any results but it is also not the original prescription from Hurley+00. My argument is that Hurley+00 uses the original H&K equation but the updated one is the same equation just with some metallicity dependence from de Koter 2005. Additionally, COMPAS doesn't follow Hurley+00 precisely for the HURLEY prescription _anyway_...so I think this is a sensible change to remove non-physical behaviour. Thoughts? @ilyamandel @SimonStevenson maybe?
3. Add an option to check the photon tiring limit


jeffriley (dismissed):
First, coding-related comments:

(a) I think the new option should be in the m_RangeExcluded vector in Options.h.
(b) I'm not a fan of underscores in variable names - I don't think you need them (e.g. rate_NJ, rate_KR etc.), but that's personal preference.  It'd be better if we stay consistent (just makes things easier on the eyes if they're consistent), but that doesn't mean it has to be consistent my way :-)
(c) Your comment: "TW - shouldn't this be true by default? It says so in options.h" for "m_UseMassLoss = false" in Options.cpp.  Yes, it would seem one of them should change :-) I think though, but I'd have to go back and check, that the original default (from legacy) is false - I remember being surprised that we wouldn't use mass loss by default.  But I could be completely wrong - and in the end I have no strong preference for one over the other.  And just because it was false in the legacy code doesn't mean it should stay false - just that I imagine there was a reason (that I am not aware of) for it being false by default.

The rest of  the changes I'll leave for @ilyamandel or @SimonStevenson to review.

The variable names "dms", "dml" etc. came from legacy code.  I imagine they were given those names for a reason - and they meant something to whoever wrote that code originally.  I don't know what the reason is - maybe @SimonStevenson knows.  I'm not tied to those names - I'm all for giving variables names that describe their use/function - but I don't know why they had their original names.

And as far as adding class member variables - that's fine as long as they are useful.  As @ilyamandel points out, the more class member variables we have the more opportunity there is for something to go out of sync.  But if adding a variable to track dominant mass loss is useful, then by all means add it.

TomWagg:
Okay sounds good. Haha yeah I definitely prefer underscores to spread out the words but you're right we should keep everything consistent, I'll remove those :)

Regarding the default mass loss flag, it's also True by default in the default pythonSubmit.py so I just figured this was left over from something older? Let me know if I should change it or just leave it as is.

Okay cool, I think it would be useful so I'll add it to this pull request

TomWagg:
Okay that should be everything for the dominant mass loss rate stuff. For example, you can now produce a plot like this (attached). I tried to do the functions with the minimal number of comparisons so I don't affect the speed. Let me know if anything isn't clear :)

Resolves #466!

ilyamandel (commented):
m_DMLR                                     = MLR_TYPE::NONE;

I think DMLR and MLR will create the same problems down the line that some of the fixes in the PR are trying to address -- we won't be able to parse these names.

ilyamandel (commented):
In BaseStar::CalculateMassLossRateWolfRayet(), should there not also be 
if (utils::Compare(p_Mu, 1.0) < 0) 
to match Hurley?


I personally like how the max rate is computed in GiantBranch::CalculateMassLossRateHurley() better than in EAGB::CalculateMassLossRateHurley() [fewer extra variables to cause confusion], but that's my aesthetics.

TomWagg:
Very fair point about the variable names! I've expanded all of my acronyms now so they should be clearer (e.g. `m_DMLR` -> `m_DominantMassLossRate`)

You're exactly right about the condition in `BaseStar::CalculateMassLossRateWolfRayet()`, sorry for missing that, I don't use it in the code any more (only the Z dependent one) so didn't check as carefully. **Also, was it a purposeful choice that `f_WR` doesn't affect that function. I.e. Wolf-Rayet stars made with Hurley won't have their winds changed but ones with the Vink version will?** Although with this change they now will since both use the Vink function.

I agree that the GiantBranch function is less confusing than EAGB, though I think that is just because there are only 3 rates in GiantBranch and 4 in EAGB so it's like a special case of the other function. The brute force approach definitely looks cleaner but will have more comparisons each time. Happy to change to that, not sure whether this is really ever going to have an effect on the runtime just figured better safe than sorry.

ilyamandel:
Same number of comparisons for 4 options: doesn't matter if I compare A vs B, C vs D, and the winner of the two (three comparisons total), or A vs B, winner vs C, winner vs D (3 comparisons total).  To find a maximum of N variables requires N-1 comparisons -- there are no shortcuts.  Now, to find the maximum and second highest among N variables is a more interesting proposition: I leave it as a take-home problem. :)

ilyamandel:
Otherwise, happy with the changes in principle, but I think we need to settle on the best scheme to use going forward, and after Jon Sundqvist's presentation, I think it's none of the above...  I'll try to compile my thoughts and run them by @TomWagg  and @demink, but given how slow I am, they will likely beat me to it. :)

So I'd prefer to hold off on pulling this until we have our current "best suggested choice" sorted.

TomWagg:
Absolutely, makes sense! @ilyamandel - since this PR contains some other useful things other than the WR winds, how about I revert the WR changes and make those comparisons look neater, could we then merge this? (It would be convenient to have the dominant mass loss stuff and the other cleanups)

SimonStevenson:
Yeah, I agree with @TomWagg that it would be useful to separate the fixes/convenience changes (better variable names, dominant mass loss rate) from the "physics" changes that may require more discussions/decisions. This question of updated/improved mass-loss rates are something that I've been thinking about a bit as well.

ilyamandel:
There are already versions 02.17.10 and 02.17.11, so you need to git merge dev into your branch and update the changelog before resubmitting.

TomWagg:
Ah sorry about that @ilyamandel, I'm still finding my way around all of this git stuff! Hopefully should be all set now?

ilyamandel (approved):
All good, thanks @TomWagg !

PR #471, opened 2020-11-24
Updating fast post processing 

ilyamandel:
Updating fast post-processing to clean up a few residual inconsistencies, also fixing a couple of issues in other post-processing classes.


jeffriley (approved):
Changes look good, and fix the problem I was seeing.

Fyi, I evolved 10 million binaries, and sampled metallicity over the range 0.0001 - 0.015, so my .h5 file had ~6.5 million unique metallicities - so the COMPAS class variable metallicityGrid had a length of ~6.5 million.

The function COMPAS.setGridAndMassEvolved() calls retrieveMassEvolvedPerZ() from totalMassEvolvedPerZ.py, and that iterates over the COMPAS.metallicityGrid array, and that takes hours for ~6.5 million metallicities...

So, in my case the answer was to not call COMPAS.setGridAndMassEvolved() - Ilya's fast cosmic integration algorithm doesn't need the call to be made.  But...

I wonder whether we should have a sanity check in the COMPAS class on the length of COMPAS.metallicityGrid?  Should we reject any attempt to have a grid of metallicities that is "too big" (whatever too big is - a few hundred maybe)?  We could just assume people won't do it (they'll only do it once...), but maybe we could put a sanity check in and warn the user if the metallicityGrid is > a few hundred.  We probably didn't need it before because we didn't sample metallicity - we always used a small-ish grid.  But now we can sample metallicity and could end up with a really big metallicityGrid, so maybe a sanity check is needed.

ilyamandel:
@jeffriley : I agree, we should probably add that check and at least print a warning.  I’ll pull what I had committed for now since you approved it and it fixes a bug, but please feel free to add that.

PR #472, opened 2020-11-24
added in cap for Schneider remnant mass at the HeCore mass

reinhold-willcox:


jeffriley (changes requested):
changelog & version number?

And I should have noticed this last time - sorry:

    double remnantMass = std::pow(10.0, logRemnantMass);

We are using the macro PPOW() instead of std::pow() so we can profile calls to std::pow(), so this should be:

    double remnantMass = PPOW(10.0, logRemnantMass);

Profiling is disabled by default, so no performance degradation in the production build.

Also, instead of this:

    // Cap out at the pre-SN He Core mass
    if (remnantMass > m_SupernovaDetails.HeCoreMassAtCOFormation ) {
        remnantMass = m_SupernovaDetails.HeCoreMassAtCOFormation;
    }

You could just do this:

    return std::min(PPOW(10.0, logRemnantMass), m_SupernovaDetails.HeCoreMassAtCOFormation);    // Cap out at the pre-SN He Core mass

reinhold-willcox:
@jeffriley I had forgotten about PPOW, thanks.

jeffriley (approved):
Approved

PR #473, opened 2020-11-25
Adding some documentation to Ilya's wonderful fast cosmic integration

TomWagg:
Whilst going through Ilya's awesome new code I made some notes to help myself understand and then those notes became comments and then I decided to make the code look pretty. My hope is that this code is now understandable to a newcomer but is still just as fast and accurate as Ilya's original code.

In order to run the cosmic integration now you'll need to import `FastCosmicIntegration.py` and run `find_detection_rate`. An example call might look like
```
import FastCosmicIntegration as fci
detection_rate, formation_rate, merger_rate, redshifts, COMPAS = fci.find_detection_rate(
               path="/path/to/your/file/", filename="my_COMPAS_Output_file.h5")
```
but there are now a bunch of optional arguments for `find_detection_rate` where you can change things to your own needs.

The cosmetic changes that I've made are:

- Add a _bunch_ (maybe too many?) of comments
- Create functions for the different sections to help describe what is happening and let you grab certain parts
- Give some of the magic numbers variable names (e.g. 300 becomes Mc_max)

The actual substantive changes that I've made are:

- Add an optional `weight_column` arg for you to specify the h5 column that contains the AIS weights if you have a weighted sampled
- Fix the pDrawZ issue (I think Ilya actually already fixed this now?) with not having all the metallicities
- Made quite a few changes to how ClassCOMPAS works to remove some overhead/old unused things - I'm not sure if this is used elsewhere but couldn't see anywhere in the COMPAS repo so assumed it was okay to change this. Can also revert if you like
- Instead of just setting `sfrMassPerBinary=115`, now use `m1_min, m1_max, m2_min, f_bin` to calculate star forming mass per binary (see `find_star_forming_mass_per_binary_sampling` in ClassCOMPAS). You can choose custom distributions for m1 and q or adjust the Kroupa IMF as you like. I intend to add another version of this function that does it with integrals rather than sampling as well :)

Oh and Lieke pointed out I did everything in snake_case not camelCase so I can change that too if you prefer, old habit you know ¯\\\_(ツ)\_/¯

### DIFF
--- a/postProcessing/Folders/CosmicIntegration/PythonScripts/ClassCOMPAS.py
+++ b/postProcessing/Folders/CosmicIntegration/PythonScripts/ClassCOMPAS.py
@@ -3,6 +3,7 @@ import numpy as np
 import h5py as h5
 import os
 import totalMassEvolvedPerZ as MPZ
+import astropy.units as u
 
 class COMPASData(object):
     def __init__(
--- a/postProcessing/Folders/CosmicIntegration/PythonScripts/ClassCOMPAS.py
+++ b/postProcessing/Folders/CosmicIntegration/PythonScripts/ClassCOMPAS.py
@@ -68,8 +68,7 @@ class COMPASData(object):
             else:
                 return [compas_file[hdf5_file][var_name][...].squeeze() for var_name in var_names]
 
-    def set_DCO_masks(self, dco_types=["ALL", "BHBH", "BHNS", "NSNS"], merges_in_hubble_time=True, no_RLOF_after_CEE=True, pessimistic_CEE=True,
-                            rlof_mask=None, pessimistic_mask=None):
+    def set_DCO_masks(self, dco_types=["ALL", "BHBH", "BHNS", "NSNS"], merges_in_hubble_time=True, no_RLOF_after_CEE=True, pessimistic_CEE=True):
         """
             Create masks for different DCO types
 
@@ -78,8 +77,6 @@ class COMPASData(object):
                 merges_in_hubble_time --> whether to mask binaries that don't merge in a Hubble time
                 no_RLOF_after_CEE     --> whether to mask binaries that have immediate RLOF after a CCE
                 pessimistic_CEE       --> whether to mask binaries that go through Optimistic CE scenario
-                rlof_mask             --> a mask for RLOF after CEE (default is taken from CommonEnvelopes file, this is for if you don't have CE file)
-                pessimistic_mask      --> a mask for Optimstic CE scenarios (default is taken from CommonEnvelopes file, this is for if you don't have CE file)
         """
         # get the appropriate variables from the COMPAS file
         stellar_type_1, stellar_type_2, hubble_flag, dco_seeds = \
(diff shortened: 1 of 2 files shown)


jeffriley:
Nice job Tom.

> Made quite a few changes to how ClassCOMPAS works to remove some overhead/old unused things - I'm not sure if this is used elsewhere but couldn't see anywhere in the COMPAS repo so assumed it was okay to change this

Ilya made some changes to ClassCOMPAS, but I thought his philosophy was to leave existing functionality unchanged so that people could still use it as they were.  I don't know whether your changes are going to break other people's code - I just don't have the background with the post-processing code.  I'll add a few other people as reviewers that I think use the post-processing scripts more than me.

> Lieke pointed out I did everything in snake_case not camelCase so I can change that too if you prefer

Well, I always prefer camelCase over snake-case (where we have mixed-case, SNAKE_CASE works if you only have upper or lower case), but we should never change things based on one person's preference - apart from the obvious question of what makes one person's preference better than another's, if we did we'd spend all our time switching things between individual preferences.  We should go with the prevailing consensus (the COMPAS C++ code is predominantly (if not completely) camelCase(-ish) because I refactored it in isolation, so getting a consensus was easy :-)).  The only thing I'd ask is that we are consistent - if we decide on (e.g.) snake_case, then let's stick to snake_case throughout.

jeffriley:
@TomWagg Did you test the new classCOMPAS with an HDF5 file that includes  both DCO and CE files?  I'm seeing some inconsistencies, particularly in your set_DCO_masks() function.

TomWagg:
@jeffriley (just seeing your first comment too). Okay in that case I think I did slightly change the functionality of ClassCOMPAS in some cases so you may just prefer to revert it if you want. I _did_ check an HDF5 file with DCO and CE files but it could just be that the one I was looking at is weird. Could you please share the one that you tested with/the inconsistencies and I can figure out what I messed up? Thanks!

jeffriley:
Thanks @TomWagg.  Ok, let me look at the file I am using (it's a big file with CHE enabled and multiple WR wind factors, so I have to do some filtering) - if you tested your code then it may be that my file is the problem.  I'll see if I can reproduce the problem with a small, "standard" file.

jeffriley:
@TomWagg You might want to change this PR to "draft" while you're working on it

ilyamandel (changes requested):
Hi @TomWagg .  Sorry for the delayed review.  Very nice work on making the code modular and more readable!

- COMPASclass: I think these changes may break the Jupyter notebooks that we provide as post-processing examples;  if you are changing the functionality here, you need to make sure that all of the other code in the COMPAS repo that may use this functionality won't be broken.  [You are not, however, responsible for breaking anyone's private code that's not in the repo -- this is one way of encouraging people to submit code to the repo instead of keeping it private]

Otherwise, I did not check this class carefully, and all of my comments are on the integrator: 

- There's no need for max_logZ and min_logZ arguments to find_metallicity_distribution, you only need to compute metallicities between min_logZ_COMPAS and max_logZ_COMPAS [yes, I needlessly used arbitrary upper limits because I was initially planning to account for Z outside the range of COMPAS simulations, but I thought better of it]

- max_redshift           --> [float]          Maximum redshift to use in array
  max_redshift_detection --> [float]          Maximum redshift to calculate detection rates
-- add a comment (if not a check) that max_redshift_detection <= max_redshift

- In find_fomation_and_merger_rates, you don't need n_redshifts and redshift_step as arguments -- you can trivially get these from the redshifts array.  Should COMPAS_weights have a default value of all ones set in the argument list, so users can skip providing this input?

- "n_formed            --> [float]          Number of stars formed per yer per cubic Gpc"
should be "per year"; I'd prefer "binary formation rate (number of binaries formed per year per cubic Gpc) represented by each simulated COMPAS binary"

- "we have only calculated formation rate up to z=10, so we need to only find merger rates for formation times at z<10" -- replace 10 with max(redshifts) in the comments, since you are allowing users to change this

- "if first_too_early_index > 0:" shouldn't it be >= 0 in your notation?

- I would add a comment to the compute_snr... function that explains that this is only needed as a substitution for the following lines of code which are too slow:
for i in range(nrBinaries):
    pDetection=selection_effects.detection_probability(COMPAS.mass1[i],COMPAS.mass2[i], \
        redshifts, distances,  GWdetector_snrThreshold, GWdetector_sensitivity)

- "n_binaries                 --> [int]            Number of binaries in the COMPAS file" -- "number of merging binaries COMPAS binaries" (to distinguish from all simulated binaries)?
- "patience is a virtue my friend" -- probably shouldn't include that in public code comments :)

- eta_index = np.round(eta[i].value / eta_step).astype(int) - 1
and similar for snrs, Mc: these rely on the arrays starting at eta_step (snr_step, etc.), so you can't allow the user to freely specify the minimum values as you've done above; only the step size and the maximum value are allowed free parameters

- find_detection_rate: don't need min_logZ, max_logZ, Mc_min, snr_min, eta_min arguments.

- maybe we shouldn't talk about detection_probability_grid, but rather detection_probability_for_snr_list or similar [it's a 1-d array]?

ilyamandel:
@TomWagg -- are you still working on this?  I know I sent you a rather long list of comments two weeks ago, but your changes *are* very nice and useful, so it would be great to get them in.

TomWagg:
Ah my apologies Ilya, I completely forgot about this (Floor recently gave me some tremendous new data so I've been distracted!). I'll go through your comments tonight :)

Also hahaha I did _not_ mean to leave the "patience is a virtue my friend" comment in, that must be left over from when I was just commenting for myself oops!

TomWagg:
@ilyamandel when you say make sure ClassCOMPAS doesn't break the other code in the repo do you mean the Jupyter notebooks in the CosmicIntegration folder? If so, they seem to explain how the cosmic integration works but the older one that uses bins and so I wonder if we just need to rewrite those too to match this newer "fast cosmic integration"?

Otherwise, I _think_ I've now addressed your comments and also hopefully fixed ClassCOMPAS (thanks @jeffriley for helping me work that out!). But I also don't have an up to date big grid file from newer COMPAS (the one I was using from Lieke doesn't work anymore because the files/keys have changed) so haven't tested it 😨. Happy to do the testing though if someone could send me a file? :)

TomWagg:
Okay @jeffriley kindly uploaded a test file and so I've now tested that it works. I'm pretty sure minus the one typo I fixed it is otherwise okay. E.g. here is the grid of plots from Jeff's test file
image

@ilyamandel what should I do about the Jupyter notebooks? I can just reset ClassCOMPAS to the way it was (but I think it had a couple of things left behind that weren't useful any more/missed some functionality)? Or we can rewrite them to match the new method? EDIT: have spoken with Ilya, will make sure ClassCOMPAS is backwards compatible

TomWagg:
Okay I've fixed ClassCOMPAS so the notebooks aren't broken now. However those notebooks are getting a little outdated with the COMPAS column headers and stuff so we may want to update them also eventually. I added a couple of functions to ClassCOMPAS mostly to add new functionality (except get_COMPAS_variables() which I just added for convenience) but it should be completely backwards compatible, every outward-facing variable (I'm sure there's a proper term for that??) is unchanged.

Therefore should now be all set for review again!

ilyamandel (changes requested):
Nice work, @TomWagg , but I spotted one issue:    
    
    # set minimum values
     Mc_min = 0.1
     eta_min = 0.01
     snr_min = 0.1
     
     
This will give you grief if the user passes in non-default values of Mc_step, etc.  You should either set Mc_min=Mc_step, etc., or better yet not use these at all (e.g., use Mc_step in lieu of Mc_min when calling Mc_array = np.arrange(Mc_min, ...)
     
Did you test the code with non-default values of function arguments?

TomWagg:
Good point! Fixed that one. I did test some of the function arguments but only the ones that I directly changed, didn't think about how changes to ClassCOMPAS would affect most of them though sorry.

I've now tried changing _all_ of the default function arguments and none of them seem to cause issues as long as you don't do weird things like make maxes less than mins or steps. Do you think it is worth it to add assertions those or assume user doesn't do silly things? (e.g. m1_min <= m1_max)

jeffriley:
> Do you think it is worth it to add assertions those or assume user doesn't do silly things? (e.g. m1_min <= m1_max)

I think in the postprocessing code where we're not evolving possibly millions of objects, doing a bit of defensive programming is prudent - especially if not checking is going to (silently) produce bad results.

TomWagg:
Very fair point Jeff, I've added a series of assertions and warnings to the code that should prevent any silent errors. Also, I wasn't sure at what point to start warning the user about step size being too large, let me know if those values aren't sensible.

ilyamandel (approved):
Thanks a lot of the great work, @TomWagg !  Let's merge this in now, it's a clear improvement on what we had, and we can continue testing and tweaking as needed.

PR #474, opened 2020-11-28
Added option of overall Wind factor

LiekeVanSon:
Add an option to change the overall wind mass loss with a multiplication factor

### DIFF
--- a/defaults/pythonSubmit.py
+++ b/defaults/pythonSubmit.py
@@ -140,6 +140,7 @@ class pythonProgramOptions:
     mass_loss_prescription = 'VINK'
     luminous_blue_variable_prescription = 'BELCZYNSKI'
     luminous_blue_variable_multiplier = 1.5
+    overall_wind_mass_loss_multiplier = 1.0
     wolf_rayet_multiplier = 1.0
 
     circularise_binary_during_mass_transfer = False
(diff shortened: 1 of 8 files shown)


jeffriley:
Lieke, the reason you're getting the conflict for changelog.h is that you have built your branch on an earlier version of COMPAS (v02.17.09 - it was probably the latest version when you pulled it and started work on your change, but there was a change pushed after that (Reinhold's fix to the Schneider remnant mass function) which incremented the version to v02.17.10 - hence your conflict).  You could pull the latest dev from the teamcompas repo and merge it into your branch and fix the merge conflict there, or you could fix it after the review - either way works.

jeffriley:
The overall mass loss rate multiplier added by the PR is applied in BaseStar::CalculateMassLossRate():

    mDot = mDot * OPTIONS->OverallWindMassLossMultiplier();                       // Apply overall wind mass loss multiplier

*after* the LBV and WR winds multipliers have been applied.  Should we document the interaction of the multipliers?

LiekeVanSon:
Ok I think I resolved the changelog issue, please let me know if that worked! 

w.r.t. multiple multiplication factors, I think the behaviour is quite straight forward, but yes it might be good to put a warning somewhere that if you set wolf-rayet-multiplier = 2, but then overall-wind-mass-loss-multiplier = 0.5, you are going to end up with the same effect in the WR regime as setting wolf-rayet-multiplier = 1

I could add this as a warning in the Latex?

jeffriley:
> I could add this as a warning in the Latex?

Yes - I think the best place to put it for now is in the description of each of the options (LBV factor, WR factor, and overall factor) just make a note that there is an interaction (albeit, as you say, a straightforward one)

ilyamandel (approved):
Changes look good, thank you, @LiekeVanSon !

PR #477, opened 2020-12-03
Finishing touches on demo and a minor tweak to ClassCosmicIntegrator

### LINKED ISSUE #476: pythonSubmitDemo not updated with default pythonSubmit
**Describe the bug**

pythonSubmitDemo  (in docs/demo folder) not updated to reflect changes to default pythonSubmit file (new options etc.)

**Label the issue**

**To Reproduce**

Use pythonSubmitDemo woth COMPAS v02.17.10 or later - value of mass_ratio_min in pythonSubmitDemo is 0.01 - should be 0.01.  COMPAS terminates with an error indicating --mass-ratio-min cannot be <= 0.0

**Expected behavior**

COMPAS should work with pythonSubmitDemo.  Fix is to bring pythonSubmitDemo up to date with default pythonSubmit

**Screenshots**

 - OS: Ubuntu 18.04
 - COMPAS v02.17.10

**Additional context**

themikelau:
- Should now be able to run demo_chirpmass_dist.ipynb out of the box to reproduce Fig 10 in the methods paper
- Header names in demo_single_sys_plotter.py updated to work with recent changes in header names (e.g. denoting primary with "(1)" instead of "_1").
- Update pythonSubmitDemo.py to close issue #476 
- Allow adding the attribute "fileName" to ClassCosmicIntegrator, which is passed to ClassCOMPAS it contains. This allows ClassCosmicIntegrator to work with a user-specified filename rather than the default, "COMPAS_output.h5".


docs/demo/demo_single_sys_plotter.py
```diff
@@ -44,10 +44,10 @@ def main():
     
     ### Plot radius attributes 
     axes[0][1].plot(df['Time'], df['SemiMajorAxis'], linestyle='-', c='k', label='Semi-Major Axis')
-    axes[0][1].plot(df['Time'], df['Radius_1']/df['Radius_1/RL'], linestyle='--', alpha=0.5, c='r', label='Roche Radius 1')
-    axes[0][1].plot(df['Time'], df['Radius_2']/df['Radius_2/RL'], linestyle='--', alpha=0.5, c='b', label='Roche Radius 2')
-    axes[0][1].plot(df['Time'], df['Radius_1'], linestyle='-', c='r', label='Stellar Radius 1')
-    axes[0][1].plot(df['Time'], df['Radius_2'], linestyle='-', c='b', label='Stellar Radius 2')
+    axes[0][1].plot(df['Time'], df['Radius(1)']/df['Radius_1/RL'], linestyle='--', alpha=0.5, c='r', label='Roche Radius 1')
```
reinhold-willcox, line 47: Does Radius_1/RL need to be changed? If not here, then in COMPAS?

themikelau: No to your first question, and yes to the second question. But changing COMPAS headers is beyond the scope of this pull request.

ilyamandel: Why does it need to be changed?  If you need the RL radius, you can get it as 
'RocheLobe_1/a' * 'Separation' * (1-'Eccentricity'),
[or, as @themikelau has done, as 'Radius_1' / 'Radius_1/RL' ]
so there are already two ways of retrieving it in post-processing...

themikelau: @ilyamandel: I think @reinhold-willcox was referring to whether the header name 'Radius_1/RL' has to be similarly changed to 'Radius(1)/RL' as I have done with the other header names, to reflect the updated convention in primary/secondary designation. I was then saying it is not necessary because the previous commits that made these changes in header names forgot to change 'Radius_1/RL' (and 'Radius_2/RL')

reinhold-willcox: Yes, I'll bring in @jeffriley as well, since I noticed a few weeks ago that the new suffix changes, e.g (1), have not been applied everywhere. Jeff said he would fix those in a future PR: Jeff will that include the changes to this variable as well?

jeffriley:
Indeed, it is true the suffix changes haven't been applied everywhere - sorry about that.  The change I made to the code changed the suffix when it is automatically added to a header, but some of the headers (for binary properties) are hard-coded with "_1" and "_2" (in the header table in constants.h).  I forgot about those, and hadn't noticed (clearly I don't use those properties much).  I will open an issue and push a fix soon.

docs/demo/pythonSubmitDemo.py
```diff
@@ -77,24 +84,41 @@ class pythonProgramOptions:
...
     mode = 'BSE'                                                # evolving single stars (SSE) or binaries (BSE)?
 
-    grid_filename = 'Grid_demo.txt'
+    grid_filename = 'Grid_demo.txt'                                        # grid file name (e.g. 'mygrid.txt')
+
+    if grid_filename != None:
+        if compas_input_path_override == None:
+            grid_filename = os.getcwd() + '/' + grid_filename
+        else:
+            grid_filename = compas_input_path_override + '/' + grid_filename
```
reinhold-willcox, line 93: @themikelau In this section, it looks like you've added the possibility to use env var $COMPAS_INPUT_DIR_PATH to override the cwd path. This env var is new to me, is it something that's in use somewhere else? 

If not, the user already has to enter in the grid_filename, so I would think it would just be easier for the user to just enter in the path here, either as a relative path or a full path.

reinhold-willcox: Similar comment applies to logfile_defitions

jeffriley:
@reinhold-willcox  I added $COMPAS_INPUT_DIR_PATH a (short) while ago.  Note the comments in the file just above.  It's there for docker (though available to use outside docker) - so a user can run many docker instances and just vary where COMPAS looks its input files and puts its output files (the directory has to be mapped with the 'docker run' command).

reinhold-willcox (changes requested):
@themikelau I am not as familiar with the cosmic integration package, but the demo files look good. 
- I added a few comments to the pythonSubmit file asking for clarification. 
- Since you modified the pythonSubmit for the demo, could you make the same changes to the default pythonSubmit? 
- Also, I think you need to pull in the latest changes from dev, and add a line into the changelog. 
- The jupyter notebook is hitting a file not found error, so I can't really check this file effectively. Is the `COMPAS_CHE_30xZ_3000000.h5` file supposed to be included in the PR?

docs/demo/pythonSubmitDemo.py
themikelau: Hi @reinhold-willcox . Everything in pythonSubmitDemo.py was simply copied over from pythonSubmitDefault.py, and aren't changes I have personally made.

I am not familiar with `COMPAS_INPUT_DIR_PATH` either, but if it's not defined, then we will always have `        grid_filename = os.getcwd() + '/' + grid_filename`, which is fine, right? It doesn't seem like the `COMPAS_INPUT_DIR_PATH` functionality does anything bad.

themikelau: Oh sorry I have just missed Jeff's comment

postProcessing/Folders/CosmicIntegration/PythonScripts/ClassCosmicIntegrator.py
```diff
@@ -87,9 +88,16 @@ def __init__(self,pathCOMPAS=None, Cosmology='WMAP',hubbleConstant = 67.8,\
         if self.verbose:
             print("Creating instance COMPAS class User has to still set DCO and Data")
         #setting Mlower/Mupper etc to None to force warning for user
```
reinhold-willcox, line 90: @themikelau You can do the "if fileName is None" part more succinctly. In the __init__ call, use `fileName=COMPAS_output.h5` or whatever the default is.

themikelau:
@reinhold-willcox  To repeat my previous comment, there is nothing original in the changes made to pythonSubmitDemo—they are just copied over from pythonSubmitDefault.

As for the changelog, there has been no changes made to the main code, just demo and post-processing. I don't think a changelog is necessary for demo-related development, and I am unaware that a post-processing changelog exists.

As for the missing file, I plan to instruct the users to download it themselves from Zenodo (I will update the instructions). It would be bad to put large data files (1.78 GB in this case) in the repo of the main code.

reinhold-willcox:
> @reinhold-willcox To repeat my previous comment, there is nothing original in the changes made to pythonSubmitDemo—they are just copied over from pythonSubmitDefault.
> 
Ah, I missed that, thanks.

> As for the changelog, there has been no changes made to the main code, just demo and post-processing. I don't think a changelog is necessary for demo-related development, and I am unaware that a post-processing changelog exists.
> 
That's valid, no need to worry about that then.

> As for the missing file, I plan to instruct the users to download it themselves from Zenodo (I will update the instructions). It would be bad to put large data files (1.78 GB in this case) in the repo of the main code.
>
That's true, good call. A link to the download would also be very useful.

postProcessing/Folders/CosmicIntegration/PythonScripts/ClassCosmicIntegrator.py
themikelau: The idea here is that ClassCosmicIntegrator will use the default filename in ClassCOMPAS if no filename is given. This has the advantage of not having to maintain and keep consistent the two default filenames in the two classes, but indeed at the cost of succinctness. I can try to think about how to shorten this code.

ilyamandel:
@themikelau  @reinhold-willcox -- do I understand correctly that this is good to go and waiting for @reinhold-willcox 's approval?

themikelau:
> @themikelau @reinhold-willcox -- do I understand correctly that this is good to go and waiting for @reinhold-willcox 's approval?

Not yet, Reinhold has requested a few changes.

themikelau:
Hey @reinhold-willcox , I have made the requested change about instructing the user to download the data file from Zenodo (although the file hasn't actually been uploaded yet—still figuring out the details of a Zenodo upload). Let me know if you want the data file—I can send you the link to download it

Regarding your comments about changes to ClassCosmicIntegrator, I think it is best to leave it as it is.

If there are no further concerns, can you please merge the pull request? Thanks.

reinhold-willcox (approved):

postProcessing/Folders/CosmicIntegration/PythonScripts/ClassCosmicIntegrator.py
reinhold-willcox: I understand, but python has the functionality you're describing built-in.

```
myFunc(fileName=None):
	if fileName is None:
		doSomething(defaultFileName)
	else:
		doSomething(fileName)
```
Is the same as 
```
myFunc(fileName=defaultFileName):
	doSomething(fileName)
```

themikelau: The correct pseudo-code that describes what is currently implemented is
```
myFunc(fileName=None):
	if fileName is None:
		doSomething()
	else:
		doSomething(fileName)
```
which is different from 
```
myFunc(fileName=defaultFileName):
	doSomething(fileName)
```
because in the first case, when no fileName is given to myFunc, the filename that ends up getting used is the default filename in doSomething, whereas in the second case, the filename getting used is the default filename in myFunc.

PR #480, opened 2020-12-11
Defect repair -- uninitialised variables in Options class causing unpredictable behaviour

jeffriley:
Defect repair
(1) uncomment initialisations of mass transfer critical mass ratios in Options.cpp (erroneously commented in v02.16.00)

This defect may not manifest all the time, and when it does it may not be obvious.  The underlying problem is that some program option variables are not initialised in Options.cpp.  The behaviour of uninitialised variables in C++ is not defined - the memory location assigned to the variable is not initialised when the program loads, so whatever value happens to be in that memory location is used.  The contents of uninitialised memory can vary depending upon the OS and compiler used, and may even vary based on the memory load on the system when the program is loaded.

The specific problem I was seeing was that for v02.17.11, because the program option variable m_MassTransferCriticalMassRatioHeliumHG was not initialised to FALSE in Options.cpp, the call to OPTIONS->MassTransferCriticalMassRatioHeliumHG() in HeHG::IsMassRatioUnstable() was erroneously returning TRUE, causing HeHG::IsMassRatioUnstable() to return TRUE instead of FALSE: 

    bool HeHG::IsMassRatioUnstable(const double p_AccretorMass, const bool p_AccretorIsDegenerate) {

        bool result = false;                                                                                                    // default is stable

        if (OPTIONS->MassTransferCriticalMassRatioHeliumHG()) {
            result = p_AccretorIsDegenerate
                        ? (p_AccretorMass / m_Mass) < OPTIONS->MassTransferCriticalMassRatioHeliumHGDegenerateAccretor()        // degenerate accretor
                        : (p_AccretorMass / m_Mass) < OPTIONS->MassTransferCriticalMassRatioHeliumHGNonDegenerateAccretor();    // non-degenerate accretor
        }

        return result;
    }

This was causing this check in BaseBinaryStar::CalculateMassTransfer():

    if (qCritFlag && m_Donor->IsMassRatioUnstable(m_Accretor->Mass(), m_Accretor->IsDegenerate()) ) {
        m_CEDetails.CEEnow = true;
    }

to erroneously set m_CEDetails.CEEnow  to TRUE, causing the stars to merge prematurely.

The problem may not have been apparent in earlier versions through sheer luck - the memory locations occupied by the unititialised variables just happened to be 0 (actually, since we compile with -O3 only the least significant bit (LSB) of the variable is checked (0 = FALSE, 1 = TRUE), so the memory locations only needed to have even numbers in them for the defect to not manifest.

I did not see the problem (that I know of) in v02.17.10, but did see it in v02.17.11 (it manifested for me as a severe lack of DCOs).  I suspect the difference (and reason for the defect to manifest) was the addition of a new program option (so new variable declared in Options.h and initialised in Options.cpp) - this may have caused the memory locations of the uninitialised variables to shift sufficiently for the problem to occur.  (The addition of the new program option in v02.17.11 was entirely correct - no problem there - it just caused the actual defect to appear).

The problem was introduced in v02.16.00 when I rewrote the Options code - so completely my fault.  I was a bit too eager in commenting out some code that was (I thought - or didn't check well enough) not being used.  I have reinstated that code in this defect repair.  I have verified that the problem is fixed.

As an aside, I commented the code out in v02.16.00 because @avigna had commented out the mass transfer critical mass ratios options in Options.cpp. @avigna - can we look at either reinstating those options, or removing the code that uses them (or should use them - at the moment they just all default to FALSE)?  To be clear, this was my mistake, not @avigna's - I should have been more careful - I just want to know if we can clean up this part of the code.

### DIFF
--- a/src/Options.cpp
+++ b/src/Options.cpp
@@ -357,13 +357,12 @@ void Options::OptionValues::Initialise() {
     m_MassTransferRejuvenationPrescription.type                     = MT_REJUVENATION_PRESCRIPTION::NONE;
     m_MassTransferRejuvenationPrescription.typeString               = MT_REJUVENATION_PRESCRIPTION_LABEL.at(m_MassTransferRejuvenationPrescription.type);
 
+    // AVG
     // Mass transfer critical mass ratios
     m_MassTransferCriticalMassRatioMSLowMass                        = false;
     m_MassTransferCriticalMassRatioMSLowMassNonDegenerateAccretor   = 1.44;                                                 // Claeys+ 2014 = 1.44
     m_MassTransferCriticalMassRatioMSLowMassDegenerateAccretor      = 1.0;                                                  // Claeys+ 2014 = 1.0
 
-    // AVG
-    /*
     m_MassTransferCriticalMassRatioMSHighMass                       = false;
     m_MassTransferCriticalMassRatioMSHighMassNonDegenerateAccretor  = 0.625;                                                // Claeys+ 2014 = 0.625
     m_MassTransferCriticalMassRatioMSHighMassDegenerateAccretor     = 0.0;
@@ -391,7 +390,6 @@ void Options::OptionValues::Initialise() {
     m_MassTransferCriticalMassRatioWhiteDwarf                       = false;
 	m_MassTransferCriticalMassRatioWhiteDwarfNonDegenerateAccretor  = 0.0;
     m_MassTransferCriticalMassRatioWhiteDwarfDegenerateAccretor     = 1.6;                                                  // Claeys+ 2014 = 1.6
-    */
 
     // Common Envelope options
     m_CommonEnvelopeAlpha                                           = 1.0;
--- a/src/Options.cpp
+++ b/src/Options.cpp
@@ -357,7 +357,6 @@ void Options::OptionValues::Initialise() {
     m_MassTransferRejuvenationPrescription.type                     = MT_REJUVENATION_PRESCRIPTION::NONE;
     m_MassTransferRejuvenationPrescription.typeString               = MT_REJUVENATION_PRESCRIPTION_LABEL.at(m_MassTransferRejuvenationPrescription.type);
 
-    // AVG
     // Mass transfer critical mass ratios
     m_MassTransferCriticalMassRatioMSLowMass                        = false;
     m_MassTransferCriticalMassRatioMSLowMassNonDegenerateAccretor   = 1.44;                                                 // Claeys+ 2014 = 1.44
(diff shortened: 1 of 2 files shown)


avigna:
Hi Jeff! I'm sure we can share the blame :) 

I removed those options because we wanted to make the code cleaner and, more importantly, not to give the impression that the code can do something it can't/won't. 

I think it is as good time as any to properly re-introduce the critical mass ratios, so we can reinstate them. My suggestion is we resintate them following the numerical values from Hurley+2002 and Claeys+2014 (i.e. following Hurley+2000 stellar types) but default the options to FALSE. If, for whatever reason you think it is easier to clean them now, so be it; we can rewrite them later more carefully.

P.S.: I am left mildly concerned about other variables in this situation... is this something easy to check? I still don't get how you find this defect.

jeffriley:
Ok, well, let's implement them correctly.  For now, the reinstated code just defaults the options to FALSE, so the mass ratios are not used - and since the actual options are commented in Options.cpp users can't set the options TRUE, so for now all is good.  

Completing the implementation can be done later, so doesn't affect this PR.

jeffriley:
> I am left mildly concerned about other variables in this situation... is this something easy to check? 

There are tools (e.g. Valgrind for Linux, Purify for Windows - there are others, but those two are good).  I'm also a little surprised that -Wuninitialized didn't pick up the problem (by default we compile with -Wall, so we enable all warnings).  Ordinarily -Wuninitialized is annoying because it warns about possible instances of uninitialised variables that aren't really...  I may look into that when I get time (it is possible to confuse it - hence the false positives from it. Maybe it's possible to confuse it the other way too).

I'm not too concerned - I think we've mostly been pretty careful (just my mistake here...).  We could, though, run Valgrind across COMPAS occasionally...


> I still don't get how you find this defect.

I was fortunate here - v02.17.10 did not exhibit the problem, but v02.17.11 did (for me anyway).  It was just a matter of tracking down where the difference occurred.  I actually opined in the code hack earlier this week that it was probably an uninitialised variable - Ilya pulled the v02.17.11 code, compiled and ran with the same pythonSubmit file as me and didn't see the problem.  I couldn't see how that could happen unless the underlying problem was some sort of silent memory corruption (e.g. writing past the bounds of an array), or an uninitialised variable - I guessed the uninitialised variable (the memory corruption possibility, I thought, probably wouldn't have produced such a surgical result), so I had something in mind to look for as I was troubleshooting.

Also, one of my jobs at HP (for about 5 years) was troubleshooting/debugging HP's proprietary Unix OS (HP-UX), so I have some experience :-)  About 6 million lines of C code I think when I was working on it, written by hundreds of programmers over about 15 years.  COMPAS is ~30,000 lines of code, young enough and written by only a few programmers, so still consistent in style etc.  Troubleshooting COMPAS isn't that hard by comparison...

ilyamandel (approved):
Thank you for the bug fix, @jeffriley !

ilyamandel:
@avigna -- issue #482 created based on what you wrote above.

Issue #483, opened 2020-12-14
Mass transfer case labelling wrong (most notably, Case C mass transfer being labelled as Case B)


TomWagg:
**Describe the bug**
_tldr: I think HG::DetermineMassTransferCase is overwriting GiantBranch::DetermineMassTransferCase such that case C mass transfer is always labelled as case B. But there seem to be more issues with the labelling other than this..._

@LiekeVanSon presented some work to us today and pointed out that COMPAS seems to be mislabelling mass transfer cases. Here are some lovely plots made by her that illustrate the point.
<img width="50%" src="https://user-images.githubusercontent.com/21990332/102116671-4b128780-3e0b-11eb-86e8-e5966e9f916e.png">
In this plot each binary is coloured based on the mass transfer case of each star in the binary (i.e. A B, means primary is case Case, secondary is case B). As you can see, there are no pink/purple points. The unique cases that show up in the results are `[‘- -’ ‘- A’ ‘- B’ ‘A B’ ‘B -’ ‘B B’]`. To further illustrate this, here is another of Lieke's plots that shows the detailed evolution of a binary labelled ["-, B"] (none, case B)
<img width="50%" src="https://user-images.githubusercontent.com/21990332/102116971-b65c5980-3e0b-11eb-91a3-d4279f6dc4e1.png">
It seems from this plot that the mass transfer for each star seems to be occurring after CHeB starts and so should actually be case CC.

We looked into this further together and it seems that due to the way the inheritance works in COMPAS, the HG::DetermineMassTransferCase function is overwriting the GiantBranch::DetermineMassTransferCase function and leading to all case C being labelled case B. I think this can be fixed by moving GiantBranch::DetermineMassTransferCase to CHeB::DetermineMassTransferCase!

This does **not** explain why in the second plot, case C mass transfer is being labelled as None. So there may be further issues.

Lieke and I were also puzzled as to why HeMS and HeHG are referred to as Case A and Case B mass transfer?

**Label the issue**
`urgency_low`,`severity_moderate`

**To Reproduce**
Run a grid of binaries and check the unique mass transfer cases, case C never shows up

**Expected behavior**
Stars of CHeB and beyond should be case C.

**Versioning (please complete the following information):**
 - COMPAS - v02.17.11

PR #484, opened 2020-12-14
Make sure Case C mass transfer is labelled as Case C not Case B

TomWagg:
This addresses one of the problems in #483 but probably not all. I moved the DetermineMassTransferCase definition from GiantBranch.h to CHeB.h so that stellar types 4+ are labelled as case C (previously HG took priority and everything 2+ got labelled case B).

Additionally, why are HeMS and HeHG labelled as case A and B, Lieke and I had assumed those should also just be case C?

### DIFF
--- a/src/CHeB.h
+++ b/src/CHeB.h
@@ -109,6 +109,8 @@ protected:
 
     ENVELOPE        DetermineEnvelopeType();
 
+    MT_CASE         DetermineMassTransferCase()                                  { return MT_CASE::C; }                                                         // Mass Transfer Case C for CHeB stars and beyond
+
     STELLAR_TYPE    EvolveToNextPhase();
 
     bool            IsEndOfPhase()                                               { return !ShouldEvolveOnPhase(); }                                             // Phase ends when age at or after He Burning
--- a/src/CHeB.h
+++ b/src/CHeB.h
@@ -109,7 +109,7 @@ protected:
 
     ENVELOPE        DetermineEnvelopeType();
 
-    MT_CASE         DetermineMassTransferCase()                                  { return MT_CASE::C; }                                                         // Mass Transfer Case C for CHeB stars and beyond
+    MT_CASE         DetermineMassTransferCase()                                  { return GiantBranch::DetermineMassTransferCase(); }                           // Mass Transfer Case C for CHeB stars and beyond
 
     STELLAR_TYPE    EvolveToNextPhase();
 
--- a/src/CHeB.h
+++ b/src/CHeB.h
@@ -109,8 +109,6 @@ protected:
 
     ENVELOPE        DetermineEnvelopeType();
 
-    MT_CASE         DetermineMassTransferCase()                                  { return GiantBranch::DetermineMassTransferCase(); }                           // Mass Transfer Case C for CHeB stars and beyond
-
     STELLAR_TYPE    EvolveToNextPhase();
 
     bool            IsEndOfPhase()                                               { return !ShouldEvolveOnPhase(); }                                             // Phase ends when age at or after He Burning
--- a/src/FGB.h
+++ b/src/FGB.h
@@ -69,6 +69,8 @@ protected:
 
     ENVELOPE        DetermineEnvelopeType()                                                     { return ENVELOPE::CONVECTIVE; }                                                                // Always CONVECTIVE
 
+    MT_CASE         DetermineMassTransferCase()                                                 { return GiantBranch::DetermineMassTransferCase(); }                                            // Mass Transfer Case C for stars after HG
+
     STELLAR_TYPE    EvolveToNextPhase();
 
     bool            IsEndOfPhase()                                                              { return !ShouldEvolveOnPhase(); }                                                              // Phase ends when age at or after He ignition timescale
--- a/src/GiantBranch.h
+++ b/src/GiantBranch.h
@@ -114,8 +114,6 @@ protected:
 
     virtual double          CalculateZeta(ZETA_PRESCRIPTION p_ZetaPrescription)			    { return 0.0; }
 
-            MT_CASE         DetermineMassTransferCase()                                             { return MT_CASE::C; }                                              // Mass Transfer Case C for GiamtBranch stars
-
             bool            IsMassRatioUnstable(const double p_AccretorMass, const bool p_AccretorIsDegenerate);
 
     virtual void            PerturbLuminosityAndRadius();
--- a/src/GiantBranch.h
+++ b/src/GiantBranch.h
@@ -114,6 +114,8 @@ protected:
 
     virtual double          CalculateZeta(ZETA_PRESCRIPTION p_ZetaPrescription)			    { return 0.0; }
 
+            MT_CASE         DetermineMassTransferCase()                                             { return MT_CASE::C; }                                              // Mass Transfer Case C for CHeB stars and beyond
+
             bool            IsMassRatioUnstable(const double p_AccretorMass, const bool p_AccretorIsDegenerate);
 
     virtual void            PerturbLuminosityAndRadius();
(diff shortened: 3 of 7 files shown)


jeffriley (changes requested):
Though moving the DetermineMassTransferCase() function from GiantBranch to CHeB will achieve the desired result in this case, it is not the correct long-term solution.

The correct solution is to leave the DetermineMassTransferCase() function in GiantBranch unchanged, and add a DetermineMassTransferCase() function to CHeB.h that calls the GiantBranch function.  So reinstate the GiantBranch function and change your CHeB function to:

MT_CASE         DetermineMassTransferCase()                                  { return GiantBranch::DetermineMassTransferCase(); }

So it skips over HG in the inheritance chain.

I think this is just another case that indicates HG probably shouldn't inherit from GiantBranch.  The main reason it does is that HG uses the GBParams.  Maybe we can fix that one day.

jeffriley:
>  why are HeMS and HeHG labelled as case A and B, Lieke and I had assumed those should also just be case C?

That you'll have to take up with @SimonStevenson, @avigna, or @ilyamandel.  Here is the function from the legacy code:

    int massTransferCase(int stellarType){
	    // Function to determine if the star which initiated mass transfer will go through case A, B or C.
        if(stellarType == MS_MORE_THAN_07 or stellarType == MS_LESS_THAN_07 or stellarType == NAKED_HELIUM_STAR_MS){
            return CASE_A;
        }
        else if(stellarType == HERTZSPRUNG_GAP or stellarType == NAKED_HELIUM_STAR_HETZSPRUNG_GAP){
            return CASE_B;
        }
        else if((stellarType >= FIRST_GIANT_BRANCH and stellarType <= THERMALLY_PULSING_ASYMPTOTIC_GIANT_BRANCH) or NAKED_HELIUM_STAR_GIANT_BRANCH){
            return CASE_C;
        }
        else{
            return NO_MASS_TRANSFER;
        }
    }

It looks to me that HeMS is CASE_A because it is treated as a MS star, and similarly HeHG is treated as a HG star - but someone with a better understanding might need to explain why :-)

TomWagg:
Ah okay, I've changed it to do what you said now @jeffriley. For future reference, why is this the correct solution? Does there always need to be a definition in GiantBranch.h?

And I see, good to know, definitely intrigued to hear from the others why this is the case :)

jeffriley:
And based on the legacy function (now that I understand better what is going on...), we probably should change the DetermineMassTransferCase() functions for all stellar types to actually inherit rather than just set the mass transfer case.

From the legacy function the inheritance should be:

    MainSequence sets CASE_A
        MS_lte_07 inherits from MainSequence, so needs no DetermineMassTransferCase() function
        MS_gt_07 inherits from MainSequence, so needs no DetermineMassTransferCase() function
        HeMS inherits from TPAGB , so needs a DetermineMassTransferCase() function that calls MainSequence ::DetermineMassTransferCase()

    GiantBranch sets CASE_C
        FGB inherits from HG, so needs a DetermineMassTransferCase() function that calls GiantBranch::DetermineMassTransferCase()
        CHeB inherits from FGB , so needs no DetermineMassTransferCase() function
        HeGB inherits from HeHG, so needs a DetermineMassTransferCase() function that calls GiantBranch::DetermineMassTransferCase()

    HG sets CASE_B
        HeHG inherits from HeMS, so needs a DetermineMassTransferCase() function that calls HG::DetermineMassTransferCase()


*Unless* we just want to set the mass transfer case for each stellar type individually, and not use inheritance for mass transfer case (e.g. envelope type - I think we set envelope type for each stellar type individually - we don't use inheritance).

I have to say, though, that from the legacy function mass transfer case definitely looks like it should be inheritable.

jeffriley:
> For future reference, why is this the correct solution? 

It's the correct solution if we want to use inheritance to set mass transfer case.  If someone adds a new stellar type (as we did with CH) then they may be expecting to be able to inherit from GiantBranch.  If we don't want to use inheritance then we should just se the mass transfer case for each stellar type individually - but even then you'd probably want to fallback to the inheritance chain if someone forgot...

TomWagg:
Thanks for explaining Jeff :)

I am happy to change the functions to do as you say above. I just want to check this is the definition of mass transfer cases that everyone uses or at least understand why it is different from what I expect. I'm still new to all of this but I had learnt from the Pols notes ([See 6.2.3](https://www.astro.ru.nl/~onnop/education/binaries_utrecht_notes/Binaries_ch6-8.pdf)) that they are defined

- Case A: A binary that fills its Roche lobe for the first time as it expands on the main sequence
- Case B: A binary that fills its Roche lobe for the first time as it expands after hydrogen exhaustion
- Case C: A binary that fills its Roche lobe for the first time as it expands after helium exhaustion

Which I had always interpreted as A=MS-HG, B=HG-CHeB, C=Else. I realise that Pols notes can sometimes be out of date so perhaps these definitions are more exact now?

jeffriley:
> I just want to check this is the definition of mass transfer cases that everyone uses 

I'll leave that to others to answer - actually I think there was a discussion in slack somewhere about this recently.  From memory there may not be consensus on the definitions (I think @demink was involved in that discussion).

TomWagg:
Ah yeah I see you're right, @reinhold-willcox has already had these same questions [here](https://compashq.slack.com/archives/CMCEQ0RS8/p1603847418103500) :D

Based on that thread it seems that this variable going to become obsolete anyway so probably not worth worrying about this all too much

jeffriley:
Oh, and HeWD should set NONE - and the the other remnants inherit from that

jeffriley:
> Based on that thread it seems that this variable going to become obsolete anyway so probably not worth worrying about this all too much

Possibly not, but until it becomes obsolete we should probably make the code consistent with legacy.  You can do that with this PR if you want, or I can do it later - whatever works for you.

TomWagg:
Yes very fair! I think this latest commit should follow what you said above?

ilyamandel:
@TomWagg -- why do we need this functionality at all? :)  Seriously, I'd prefer to get rid of it.  We have functionality for recording the evolutionary stages of the donor and accretor in every mass transfer episode.  This allows the user to determine for themselves what the type of MT was, rather than debating the type of MT that should be associated with, say, a HeMS donor (I would call it case B, for what my opinion is worth, but that's not the point).

jeffriley:
@ilyamandel @TomWagg Happy for the functionality to come out now (since, as Ilya says, we have the history now), but if we're not going to pull it out now we should at least make it not wrong.

TomWagg:
Yes completely agree Ilya, did Reinhold already implement the MT history array as you mentioned in the slack thread? If so I figure we can just remove all of this mass transfer case stuff.

If not, I agree with Jeff that we might as well merge these fixes in the meantime for anyone currently using the code (e.g. Lieke and I didn't realise the MT_CASE stuff was off and were trying to analyse things based on it)

jeffriley:
> did Reinhold already implement the MT history array

Yes, in v02.17.01.  Look for MT_DONOR_HIST in the documentation

TomWagg:
Ah nice, that looks awesome! In that case, I would vote we just remove this other mass transfer stuff as it can just be confusing and the history has all of the information.

reinhold-willcox:
Just catching up on all these comments - yes the MassTransferDonorHist variable is working, though it doesn't include information about the Accretor (which may or may not be useful, depending on how you interpret the "secondary" in Lieke's plots above). 

Also, note that it gives just the stellar types of the donor, so you have the freedom to choose which definition of the Mass Transfer Case to use.

avigna:
> Just catching up on all these comments - yes the MassTransferDonorHist variable is working, though it doesn't include information about the Accretor (which may or may not be useful, depending on how you interpret the "secondary" in Lieke's plots above).

I'd say accretor is important.

> Also, note that it gives just the stellar types of the donor, so you have the freedom to choose which definition of the Mass Transfer Case to use.

Agreed.

jeffriley:
Ok, so what's the consensus?  Do we leave this functionality in (and fix it), or do we take it out?

ilyamandel:
I vote for getting rid of the MT case functionality (and yes, I agree we should actually get rid of it, not just keep something that's wrong).

PR #487, opened 2020-12-16
Code and architecture cleanup

jeffriley:
The architecture changes here implement the architecture as documented in the COMPAS Methods Paper.

Code and architecture cleanup
(1) Architecture changes:
    - Added Remnants class    - inherits from HeGB class
    - Added WhiteDwarfs class - inherits from Remnants class; most of the WD code moved from HeWD, COWD and ONeWD to WhiteDwarfs class
    - Changed HeWD class      - inherits from WhiteDwarfs class (COWD still inherits from HeWD; ONeWD from COWD)
    - Change NS class         - inherits from Remnants class; code added/moved as necessary
    - Change BH class         - inherits from Remnants class; code added/moved as necessary
    - Change MR class         - inherits from Remnants class; code added/moved as necessary
(2) Code cleanup:
    - added "const" to many functions (mostly SSE code) that dont modify class variables ("this") (still much to do, but this is a start)
    - added "virtual" to GiantBranch::CalculateCoreMassAtBAGB() and BaseStar::CalculateTemperatureAtPhaseEnd()
        - will have no impact given where they are called, but the keyword should be there (in case of future changes)
    - changed hard-coded header suffixes from _1 -> (1), _2 -> (2)


jeffriley:
Tested: runs of 25000 binaries using default pythonSubmit file for v02.17.13 and this new code (v02.17.14 if merged) produced the following (identical results for both runs):

    SEED COUNT          : 25000 , UNIQUE: 25000
    SURVIVED SEED COUNT : 25000 , UNIQUE: 25000
    
    Z COUNT             : 25000 , UNIQUE: 1 , Z[0] = 0.0142
    
    DCO SEED COUNT      : 157
    MERGE IN HUBBLE TIME: 53
    BBH                 : 18
    DNS                 : 12
    BHNS                : 23
    
    Minimum ZAMS e      : 0.0
    Maximum ZAMS e      : 0.0
    Minimum ZAMS Mass(1): 2.694898
    Maximum ZAMS Mass(1): 149.7777
    Minimum ZAMS Mass(2): 0.1006393
    Maximum ZAMS Mass(2): 137.5675
    Minimum ZAMS a      : 0.0169233
    Maximum ZAMS a      : 999.3991

reinhold-willcox:
Without reading the code, these changes sound great Jeff! 

~~What's the difference between this PR and the one with the identical name that was just merged, #486?~~ Nevermind, I see it was a merge into the new remnants branch, which is now the base branch of this PR

ilyamandel (approved):
I can't claim to have checked 70 files line by line [though I did find and fix one typo in a comment :) ], but this looks very nice indeed.  Thank you, @jeffriley !  I trust that the impact has been checked on a single-system basis for at least a few systems (diffs of BSE_Supernovae,  BSE_CommonEnvelope, and BSE_DoubleCompactObject files for ~100 binaries), not just population properties.

jeffriley:
Thanks for reviewing @ilyamandel!

> I trust that the impact has been checked on a single-system basis

I did that (to some extent) on some systems as I was testing, but just to be sure I will do some more systematic testing today before I merge

ilyamandel:
@jeffriley -- note -- I just pulled in #489 (it was a bug fix), so you'll need to merge in dev and update the changelog.

jeffriley:
Additional changes after testing:

(3) Added call to main() to seed random number generator with seed = 0 before options are processed (and user-specified seed is known).  Ensures repeatability.
(4) Changed \"timestep below minimum\" warnings in Star.cpp to be displayed only if --enable-warnings is specified

(4) is just fixing up some warnings that should have been fixed up earlier - unrelated to this PR

(3) on the other hand...

I ran tests of 100 binaries with v02.17.13 and the code for this PR (which will now be v02.17.15 if merged):

- Some header strings were different (changes in his PR - "_1" -> "(1)", "_2" -> "(2)")
- A few numerical values were different - and really shouldn't have been.

It turns out that the changes to the way options are processed in v02.16.00 meant that the default values for SN kick parameters were being set before the random number generator was seeded with the user-specified seed, but only for the first star/binary evolved.  This was not producing incorrect results, just non-repeatable results (the SN kick parameters that are generated from a random number were different in each run for the first star/binary because different random numbers were drawn). The fix is to seed the random number generator with a fixed seed (I went for 0) before the options are initialised and the user-specified random-seed is known - the tests are then repeatable because the same random numbers are drawn in the same sequence.

So, good call to do detailed diffs of the output files @ilyamandel - there was no problem with the code changes for this PR, but it did expose a defect in a previous version.

Now ready for review and merge.

ilyamandel (approved):
Looks great, @jeffriley , and thank you for the extra check and the nice catch!

PR #489, opened 2020-12-16
Fix bug in LBV winds for fLBV=0.0

TomWagg:
Fixing a bug that Ilya pointed out where if we set fLBV=0 the default results are different from before my LBV updates. I had been including the other winds if LBV winds were 0, rather than just always ignoring them.

Z = 0.02, MZAMS=150Msun and fLBV=0 fWR=0 now gives a remnant of mass 39 Msun rather than the previous 27 Msun.

Also, and this just make me a me thing that everyone else realises, but we may want to consider adding some warnings to fLBV. Because my understanding is that in the LBV regime they _are_ other winds, they are just dominated by the LBV winds. But if you then set fLBV to 0, you are essentially also setting these other winds to 0 possibly without realising. Basically wind multipliers + assumptions about which mass loss is dominant = weird effects.


ilyamandel (approved):
Looks good -- thank you for the quick fix, @TomWagg !

PR #491, opened 2020-12-17
Removed functions related to MassTransferCase, in favor of the MassTransferDonorHist param

### LINKED ISSUE #483: Mass transfer case labelling wrong (most notably, Case C mass transfer being labelled as Case B)

reinhold-willcox:
I think this addresses the conversation in PR #484.


jeffriley (changes requested):
Nice job Reinhold.

I think there are references in the documentation to MASS_TRANSFER_CASE_INITIAL  that need to be removed.

TomWagg (commented):
All compiles and runs properly, Mt_Case_Initial is gone from the output. Looks good to merge after what Jeff said about docs.

Also, to address what @avigna  was saying in the other PR, is there a reason to _not_ include the accretor information in the history? If it is useful then perhaps you could list the history as something like "1-1,2-1" where it's MS->MS and HG->MS mass transfer?

jeffriley:
> something like "1-1,2-1" where it's MS->MS and HG->MS mass transfer

Just need to make sure we don't use a character that could possibly be a delimiter for fields in the file (currently comma, tab, and space).  So maybe "1-1:2-1".

ilyamandel:
I would vote against including accretor information in the history without an obvious use case.   We already have that information available in the RLOF output file.  The only reason to keep it in the code is we are making "live" decisions about future evolution based on past evolution.  I don't currently see us doing that based on accretor history...  and every extra line of code is an extra line to maintain.

reinhold-willcox:
@ilyamandel sounds good, I'll just fix the documentation then.

reinhold-willcox:
@jeffriley Should be all set, let me know if I need to do anything else.

jeffriley (approved):
All good

PR #496, opened 2020-12-17
added option to run postprocessing in an alternative directory

reinhold-willcox:
Small update to the postProcessing file that allows the file to be run in a different directory to where it is currently located, with the directory specified as an option on the command line. E.g:

python $COMPAS_ROOT_DIR/defaults/postProcessing.py path/to/data/dir is now valid, removing the need to copy the postProcessing around.

This will also enable stroopwafel to run the postProcessing immediately after finishing a run.


ilyamandel (approved):
Looks good to me.

PR #498, opened 2020-12-18
Switch log fix

jeffriley:
Typo in options code for option --switch-log: "switchlog" was incorrectly used instead of "switch-log"


ilyamandel (approved):
Looks good, thanks, Jeff!

PR #499, opened 2020-12-24
Cool wind factor

LiekeVanSon:
Added an option to vary all winds on the cool side of the VINK_MASS_LOSS_MINIMUM_TEMP  (default =  12500 K) when using the "VINK" mass loss prescription. It's a really minor change that I had lying around for a bit.

If you are using the VINK mass loss prescription, and  Teff < 12500, this factor applies to the following 3 wind mass loss prescriptions:

- Kudritzki, R. P. & Reimers, D. (1978)
- Vassiliadis & Wood (1993)
- Nieuwenhuijzen & de Jager (1990)

Note that Kudritzki already contains a parameter "MASS_LOSS_ETA" to vary the winds, however this eta is currently not added as an option.﻿

This factor does NOT influence LBV winds (regardless of their temperature)

The effect for single stars is mostly to create more WR stars (I'm currently looking into the effect on binaries!  :)
CoolWindMassLoss


jeffriley (changes requested):
Looks good, other than typo on line 619 of Options.cpp ("Multiplicitive")

jeffriley (approved):
All good

PR #504, opened 2021-01-08
hdf5 - added support for HDF5 files

jeffriley:
Enhancement:
(1) Added support for HDF5 logfiles (see notes at top of log.h)
    - Added 'logfile-type' option; allowed values are HDF5, CSV, TSV, TXT; default is HDF5
    - Added 'hdf5-chunk-size' option - specifies the HDF5 chunk size (number of dataset entries)
    - Added 'hdf5-buffer-size' option - specifies the HDF5 IO buffer size (number of chunks)
    - Removed 'logfile-delimiter' option - delimiter now set by logfile type (--logfile-type option described above)
    - Changed header strings containing '/' character: '/' replaced by '|' (header strings become dataset names in HDF5 files, and '/' is a path delimiter...)

I have also provided a Python script, h5copy.py, in the postprocessing/Folders/H5/PythonScripts directory, that can be used to join multiple HDF5 files into a single file.  The script will copy one or more HDF5 files to an output HDF5 file – if the output file already exists the user can choose to erase it or append to it, so multiple runs can be consolidated into a single file over time if required.

There is copious documentation at the top of Log.h, and h5copy.py.  The discussion of HDF5 chunking is important – chunk size selection has a significant impact on performance, and memory and storage media usage.

This code requires the HDF5 C++ libraries – I have updated the “Getting Started” document to indicate what needs to be installed.  The COMPAS documentation (latex files) has also been updated.

There are changes to the Makefile, so users who use a custom Makefile will need to update their Makefile.

This code changes the default logfile type from CSV files to HDF5 files – if we do merge this update then we should announce that in the appropriate Slack channel.

I have built this version on both Linux (Ubuntu 18.04) and OS X (Catalina, v10.15.7).  A small test run using the default pythonSubmit on the Mac produced the correct output files.  A larger test run on Linux – 32 individual runs of 1 million binaries each – produced the expected output files.  Joining the 32 HDF5 log files into a single HDF5 file containing 32 million binaries using the h5copy.py script took about 3 minutes.

I have run multiple tests, but I would be happy for others to try to beat it up a bit…


ilyamandel (changes requested):
from ProgramOptions.tex:

\programOption{cool-wind-mass-loss-multiplier}{}{Multiplicative constant for wind mass loss of cool stars, i.e. those with temperatures below the VINK\_MASS\_LOSS\_MINIMUM\_TEMP.  \\ Only applicable when mass-loss-prescription is set to VINK.}{12500K}

As far as I can tell, the last (default) value in figure brackets should remain 1.0, not be changed to 12500K.

jeffriley:
> As far as I can tell, the last (default) value in figure brackets should remain 1.0, not be changed to 12500K.

Ah - you're absolutely right.  1.0 is the default.  Why did I pull out 12500K?  I think I must have been tired..

Here's the original text:

"Multiplicative constant for wind mass loss of cool stars, i.e. those with temperatures below the VINK_MASS_LOSS_MINIMUM_TEMP  (default 12500 K). Affects   \\ Only applicable when mass-loss-prescription is set to VINK."

I see what I did.  I was cleaning up the stray "Affects", and I must have keyed off "default is 12500K" without thinking about it.  That default (12500K) applies to VINK_MASS_LOSS_MINIMUM_TEMP.

I'll fix it.

ilyamandel:
I was going to do it, but you beat me to it, @jeffriley ! :) The code looks great, and I trust your tests, so will merge this in.

ilyamandel (approved):
All good!

Issue #506, opened 2021-01-11
Correctly evolve disrupted binaries 


SimonStevenson:
**Describe the bug**
I am looking at evolving wide binaries (which are disrupted by the first supernova) in the context of pulsars.

In binaries in which both components form neutron stars, I am finding that the evolution ceases immediately after the formation of the second supernova. I would like it to continue, to follow the pulsar evolution of the second neutron star. 

The binary stops with  evolutionStatus = EVOLUTION_STATUS::STOPPED  because of the line
`else if (IsDCO() && m_Time>(m_DCOFormationTime + m_TimeToCoalescence)){` in BaseBinaryStar.cpp
both m_DCOFormationTime and m_TimeToCoalescence are 0, so the second part of this statement is always true. The first part becomes true once the second neutron star forms. 

I think that we should only use this condition for bound binaries (i.e. we should add a check if the binary is bound here). i.e. adding a check for !IsUnbound()

**Label the issue**
`urgency_moderate` - This is a moderately urgent issue
`severity_moderate` - This is a moderately severe bug

**To Reproduce**
My example system has
`    initial_mass_1  = 17.0                                      # primary initial mass for BSE
    initial_mass_2  = 14.0                                      # secondary initial mass for BSE
    semi_major_axis = 1000.0                                    # semi-major axis for BSE
`
though I think this is a general issue for the reasons described above. My pythonSubmit is attached.

[pythonSubmit.txt](https://github.com/TeamCOMPAS/COMPAS/files/5793990/pythonSubmit.txt)

**Expected behavior**
The evolution would continue until one of the other stopping conditions (maximum time or number of timesteps) is reached.

**Versioning (please complete the following information):**
 - OSX [10.14.6]
 - COMPAS [v02.18.00]

**Additional context**

SimonStevenson:
Added PR #507 which resolves this issue

SimonStevenson:
PR merged

PR #507, opened 2021-01-11
Added check for unbound binary

### LINKED ISSUE #517: 2nd Supernova not printing
**Describe the bug**
The 2nd SN is not printing if the first one is disruptive. This is a new change following PR #507 (sorry Simon to throw you under the bus!). 

The issue can be fixed by reverting the line in BaseBinaryStar.cpp in that commit, from:

```
else if (IsDCO() && m_Time>(m_DCOFormationTime + m_TimeToCoalescence) && !IsUnbound()){
                            evolutionStatus = EVOLUTION_STATUS::STOPPED; 
                        } 
```

to 

```
else if (IsDCO() && m_Time>(m_DCOFormationTime + m_TimeToCoalescence)){
                            evolutionStatus = EVOLUTION_STATUS::STOPPED; 
                        } 
```

However, that just opens up Simon's issue again, so this may require more thought. Also, I see I was tagged in the other PR, but I must have overlooked it, that's on me. 

**Label the issue**

**To Reproduce**
```
COMPAS -n 1 --random-seed 1253348 --initial-mass-1 7.774703670821885 --semi-major-axis 693.9547537158253 --initial-mass-2 6.628395253709512 --metallicity 0.0142 --eccentricity 0 --kick-theta-1 1.121298905520773 --kick-theta-2 -0.11028283259069305 --kick-phi-1 3.2789799698707145 --kick-phi-2 0.07861287414621784 --kick-mean-anomaly-1 5.386188554554972 --kick-mean-anomaly-2 5.452499790241414 --evolve-unbound TRUE --remnant-mass-prescription FRYER2012 --kick-magnitude-1 100 --kick-magnitude-2 100
```
This system results in 2 NSs (disrupted at first SN), but if run on the current version of dev, only one SN shows up in the output. (If the line above is corrected, both SNe appear).

**Expected behavior**
Both SNe show up. 

 - COMPAS v02.18.03

SimonStevenson:
This addresses issue #506

### DIFF
--- a/src/BaseBinaryStar.cpp
+++ b/src/BaseBinaryStar.cpp
@@ -2330,8 +2330,8 @@ EVOLUTION_STATUS BaseBinaryStar::Evolve() {
 
     if (OPTIONS->PopulationDataPrinting()) {
         SAY("\nGenerating a new binary - " << m_Id);
-        SAY("Binary has masses " << m_Star1->Mass() << " & " << m_Star2->Mass());
-        SAY("Binary has initial semiMajorAxis" << m_SemiMajorAxis);
+        SAY("Binary has masses " << m_Star1->Mass() << " & " << m_Star2->Mass() << " Msol");
+        SAY("Binary has initial semiMajorAxis " << m_SemiMajorAxis << " AU");
         SAY("RandomSeed " << m_RandomSeed);
     }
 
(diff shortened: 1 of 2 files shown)


jeffriley (commented):
Couple of things:

(i) need to update changelog and version string

(ii) These lines in BaseBinaryStar.cpp:

    ...
    2378    else if (IsUnbound() && !OPTIONS->EvolveUnboundSystems()) {                                                                 // binary is unbound and we don't want unbound systems?
    2379        evolutionStatus = EVOLUTION_STATUS::UNBOUND;                                                                            // stop evolution
    2380    }
    ...

should prevent line 2406 from being executed if the user chose not to evolve unbound systems.

(Updated comment because I think I misread some code...)

We check for !IsUnbound() in line 2390 - presumably because even if the user wants to evolve unbound systems we don't want to if we have a DCO?  Or is that a mistake?  Ah no - looking at that again that's just to resolve coalescence...

So, the addition if '&& !IsUnbound()' on line 2406 ignores OPTIONS->EvolveUnboundSystems(), doesn't it?

Do we need to revisit the whole "evolve unbound systems" check?  

I changed this code recently (I think) - to take out what I thought at the time was a check that had become redundant after Ilya made a change.  I hope that change didn't break this...

I'll leave this for @reinhold-willcox to comment - I think he  was working on the "evolve unbound binaries" option - or did some work on it - a while back.

SimonStevenson:
i) sure can be done once this is ready to go
ii) I do want to evolve unbound binaries, so for me OPTIONS->EvolveUnboundSystems() is true. This problem only occurs for unbound binaries. Adding the extra check whether the binary is still bound in 2406 resolves the issue for me. 

I'm not sure if/when any recent changes may have caused this issue

jeffriley:
Oh... I see - yes.  That works - it's effectively the same reason the check is at 2390.

Ok, that means my change to clean up the code probably didn't break this - we never used to evolve unbound systems, so the check on line 2406 was valid then - but now we need to check for an unbound system there (no point checking DCO coalescence time for an unbound system).

Sorry -   I got there in the end :-)

Ok, so I'm ok to approve the PR once changelog is updated.

SimonStevenson:
Thanks Jeff! Change log updated and version bumped. I noticed the last version update went back in time a year :)

jeffriley (approved):
Approved

jeffriley:
> I noticed the last version update went back in time a year :)

oops...

Issue #513, opened 2021-01-19
Collapse neutron star to black hole after accreting above maximum neutron star mass


SimonStevenson:
**Is your feature request related to a problem? Please describe.**
We should ensure that neutron stars which accrete mass and become more massive than the user-specified maximum neutron star mass collapse into black holes.

**Describe the solution you'd like**
Check if neutron star mass is greater than maximum neutron star mass (OPTIONS->MaximumNeutronStarMass()); if so, collapse star to a black hole.

This could be implemented in the ShouldEvolveOnPhase function in NS.h, similar to how it works for WDs

**Describe alternatives you've considered**
NA

**Additional context**
May arise during accretion in common envelopes or in NS-LMXB phases.

SimonStevenson:
I've had a go at implementing this, but ShouldEvolveOnPhase doesn't seem to be being called for NSs ...

SimonStevenson:
Implemented in PR #514

PR #514, opened 2021-01-19
Added a check of neutron star mass against maximum neutron star mass 

SimonStevenson:
Added a check of neutron star mass against maximum neutron star mass to ShouldEvolveOnPhase in NS.h

This should resolve Issue #513

### DIFF
--- a/src/NS.h
+++ b/src/NS.h
@@ -7,6 +7,7 @@
 #include "utils.h"
 
 #include "Remnants.h"
+#include "BH.h"
 
 
 class BaseStar;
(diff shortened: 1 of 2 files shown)


jeffriley (commented):
Looks good.  Happy to approve once changelog.h and version string are updated.

SimonStevenson:
Thanks Jeff! Version updated now.

jeffriley (approved):
All good

ilyamandel:
@SimonStevenson -- I am not convinced by this line:

m_Mass = BH::CalculateNeutrinoMassLoss_Static(m_Mass); 

Does an accretion induced collapse of an NS into a BH involve significant mass loss in neutrinos (at the same level as normal core collapse)?  I wouldn't have thought so, but I don't know the answer...  Will ask Bernhard for his thoughts.

ilyamandel:
Following up, here's a response from Bernhard.  The summary is that mass accretion efficiency for nearly-maximal-mass NS may be reduced (separate story from this -- it's about how we handle accretion), but that there is no appreciable neutrino mass loss during the collapse from NS to BH.  I'll put together a new PR to make this change.

"this should be easier to answer than in the case of BH formation inside a CCSN.
In the CCSN problem, the key issue is that the diffusion time scale of the PNS interior is longer
than the accretion time scale. In this case, you shouldn't have that problem. You have steady state
accretion up to the point of instability, and then the NS will collapse on a dynamical time scale, too fast
for neutrinos to do anything.

But you still get to answers, depending on whether you consider the BH mass of the total system mass.
The collapsing NS must have a mass somewhat *above* the maximum mass limit for non-rotating
NSs (depending on the degree of differential rotation). No matter what happened before, you must have got to the stability limit.

The total system mass should have decreased by O(1/2 GM/R \Delta M)  because the accreted stuff has to get rid
of its energy some way. However, if the NS is close to mass shedding, the efficiency will go down considerably.
Anyway, if you don't accrete too much, the change in mass of the total system shouldn't be overwhelming in the first place."

SimonStevenson:
Cool, I honestly wasn't sure about that so thanks for checking!

PR #515, opened 2021-01-20
Fix for neutrino mass loss during BH formation Issue #453

### LINKED ISSUE #453: Make neutrino mass loss during BH formation an user specified option
Currently, during BH formation we lose 10% of mass in neutrinos (following Fryer+2012). This is an excesive amount that should be reduced for high mass stars.

We should add an user specified maximum/fixed fraction or mass loss that can be lost in neutrinos.

SimonStevenson:
Fix for neutrino mass loss during BH formation Issue #453 details in the issue

### DIFF
--- a/defaults/pythonSubmit.py
+++ b/defaults/pythonSubmit.py
@@ -210,6 +210,9 @@ class pythonProgramOptions:
 
     neutron_star_equation_of_state = 'SSE'
 
+    neutrino_mass_loss_BH_formation = "FIXED_FRACTION"          # "FIXED_MASS"
+    neutrino_mass_loss_BH_formation_value = 0.1                 # Either fraction or mass (Msol) to lose
+
     remnant_mass_prescription   = 'FRYER2012'
     fryer_supernova_engine      = 'DELAYED'
     black_hole_kicks            = 'FALLBACK'
@@ -436,7 +439,8 @@ class pythonProgramOptions:
             self.log_level,
             self.debug_level,
             self.hdf5_chunk_size,
-            self.hdf5_buffer_size
+            self.hdf5_buffer_size,
+            self.neutrino_mass_loss_BH_formation_value
         ]
 
         return numericalChoices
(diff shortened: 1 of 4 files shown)


SimonStevenson:
(I know that I need to update the changelog + version number, but will wait until people are happy with the changes and then we can add those in)

reinhold-willcox:
Leaving to @avigna who I think already has tests prepared for this.

ilyamandel:
Two questions:
- Should we consider changing the pythonSubmit default to NEUTRINO_MASS_LOSS_PRESCRIPTION::FIXED_MASS rather than FIXED_FRACTION, with 0.1 solar masses of mass loss?
- Should we change the default behaviour in  BH::CalculateNeutrinoMassLoss_Static() when the mass loss prescription is unknown to gravitationalMass = p_BaryonicMass rather than gravitationalMass = 0?

avigna:
Thanks for the questions Ilya. I don't feel strongly about the first one, if we have a reference for that lets make it the default. The second one I strongly suggest we do it, will add it to the review (as well as changing the log and version number).

ilyamandel:
@avigna : For a reference, see footnote 7 in Stevenson+, 2019, courtesy of Bernhard Mueller: 
https://ui.adsabs.harvard.edu/abs/2019ApJ...882..121S

defaults/pythonSubmit.py
```diff
@@ -210,6 +210,9 @@ class pythonProgramOptions:
 
     neutron_star_equation_of_state = 'SSE'
 
+    neutrino_mass_loss_BH_formation = "FIXED_FRACTION"          # "FIXED_MASS"
```
avigna: Ilya suggest to change to FIXED_MASS.

defaults/pythonSubmit.py
```diff
@@ -210,6 +210,9 @@ class pythonProgramOptions:
...
+    neutrino_mass_loss_BH_formation_value = 0.1                 # Either fraction or mass (Msol) to lose
```
avigna, line 214: Ilya suggest to change to 0.1 solar masses of mass loss.

avigna (changes requested):
In file src/GiantBranch.cpp in line 1472 add "gravitationalMass = p_BaryonicMass" for default/unknown SN engine (see Ilya's comment on that on the conversation).
Update changelog and version number.
After that, good to go.

SimonStevenson:
I've incorporated the requested changes @avigna . Sorry it took me a while to get around to it!

avigna (approved):
All good! Thanks Simon!

PR #516, opened 2021-01-24
Postprocessing mask

TomWagg:
Thanks to Neha Singh from Warsaw University for making us aware of this issue and @LiekeVanSon for locating the exact cause!

The `hubble_mask` in the post-processing script was taking the mask directly from the Merges_Hubble_Time variable which is a list of integers rather than booleans. This leads to the mask indexing the DCOs rather than masking them and getting only 2 DCOs no matter the true total. Fixed this on line 76 with `hubble_mask.astype(bool)`.

Other small things are just updated names from newer COMPAS version.


ilyamandel (approved):
Looks good -- thank you, @TomWagg !

BTW, if you have time and haven't done it yet, could you check that the other python/jupyter notebooks that Coen provided as examples still work OK with the post-processing and other changes (including variable name changes)?  Thank you!

PR #526, opened 2021-02-18
Updates to Demo Plot

### LINKED ISSUE #523: Demo plotting file not updated to take h5 output
Title says it all - the plotting file in docs/demo/ which is used to generate the plot in the COMPAS paper, reads in the old csv output format, but we now output hdf5 by default. 

The file should be updated to reflect this. 

If a user wants to use the plotter, we could require that they do a run with the output type set to csv (non-default).

reinhold-willcox:
Moved demo/ folder to root level, and updated the plotting script to read in h5 data.


reinhold-willcox:
@themikelau I didn't touch the chirpmass jupyter notebook, but that may also be out of sync.

themikelau:
@reinhold-willcox : The getting_started instructions for the demo needs to be updated accordingly. I have already made some changes. Can you check these changes and also update Section 2.2?

I've checked that the single_sys_plotter.py script works and the pythonSubmitDemo.py works.

themikelau (changes requested):
See above.

themikelau:
What is gw151226evol.eps? Was it committed by mistake and can be deleted?

themikelau:
I don't think the chirp mass Jupyter notebook needs to be updated, because it is designed to read a H5 file that will be made available from Zenodo.

reinhold-willcox:
@themikelau The eps file is the image generated by the script - eps files are vectorized so they scale better. It is the same one used in the paper, so I think it makes sense to include it.

reinhold-willcox:
@themikelau Are there other issues with this PR you would like me to address, or is it safe to pull this in?

themikelau:
> @reinhold-willcox : The getting_started instructions for the demo needs to be updated accordingly. I have already made some changes. Can you check these changes and also update Section 2.2?

Have you done this?

themikelau (approved):
Happy with everything here.

PR #528, opened 2021-02-19
Enhancement - add options to set initial rotational frequencies

jeffriley:
(1) Added 'rotational-frequency' option so users can specify initial rotational frequency of SSE star
(2) Added 'rotational-frequency-1' and 'rotational-frequency-2' options so users can specify initial rotational frequency of both BSE stars
(3) Changed units of rotational frequencies written to logfiles (omega, omega_break, omega_ZAMS) from rotations per year to Hz
(4) Changed program option header strings containing '_1' and '_2' to '(1)' and '(2)' for consistency


ilyamandel (approved):
Excellent!  This PR is so clever, you can pin a tail on it and call it a fox. :) [almost Blackadder]

PR #534, opened 2021-02-25
Defect repairs: Issue #530 - some commandline options ignored when a grid file is used

### LINKED ISSUE #530: Command line option ignored when a grid is used
**Describe the bug**
The issue described here is very specific to a single command line option, but may be a more general trend (I haven't tested other flags).

I want to do a run with ESCN kicks set to 0, while also using a grid file. When both flags are set, the ECSN kicks are not set to 0. However, when the same grid arguments are included on the command line, the ECSN kicks are properly zero'd. 

**Label the issue**

**To Reproduce**

For the purposes of a minimum working example, the following single lined gridfile was used to produce a system which will ECSN.
```
 --random-seed 6 --initial-mass-1 7.926799769578849 --semi-major-axis 31.43376315763422 --initial-mass-2 1.540423527044136 --metallicity 0.0142 --eccentricity 0    
```

Then the following commands are run:
```
$ COMPAS --kick-magnitude-sigma-ECSN 0 --grid grid.txt
$ COMPAS --kick-magnitude-sigma-ECSN 0  $(< grid.txt) -n 1
```
where the `-n 1` is required because without the grid Compas will run the same binary 10x. 

A useful python script to quickly see the drawn ECSN kick of this system is:
```
import h5py as h5    
import sys     
     
myf = h5.File(sys.argv[1], 'r')    
SNe = myf['BSE_Supernovae']    
sntype = SNe['SN_Type(SN)'][()]    
drnKick= SNe['Drawn_Kick_Magnitude(SN)'][()]    
     
print("The ECSN kick is {}.".format(drnKick[sntype == 2])) 
```

With commands and output:
```
$ py test_ecsn.py COMPAS_Output/COMPAS_Output.h5 
The ECSN kick is [36.6312464].
$ py test_ecsn.py COMPAS_Output_1/COMPAS_Output.h5 
The ECSN kick is [0.].
```

It is apparent that the use of a grid file negates the use of the ECSN flag. I'm not sure if this applies to other flags as well. 

**Expected behavior**
The existence of the grid file should not override the command line args.

 - COMPAS [e.g. v02.18.06]

jeffriley:
(1) Remove stray diagnostic print from BaseStar constructor in BaseStar.cpp
(2) Fix for issue #530 - some commandline options ignored when a grid file is used
    - the issue here was case-sensitive vs case-insensitive matches (asking BOOST to do case-insensitive matches for option names doesn't propagate to all matches BOOST does...)
        - the options affected were all options that have mixed-case names:
            - case-BB-stability-prescription
            - kick-magnitude-sigma-CCSN-BH
            - kick-magnitude-sigma-CCSN-NS
            - kick-magnitude-sigma-ECSN
            - kick-magnitude-sigma-USSN
            - mass-transfer-thermal-limit-C
            - muller-mandel-kick-multiplier-BH
            - muller-mandel-kick-multiplier-NS
            - neutrino-mass-loss-BH-formation
            - neutrino-mass-loss-BH-formation-value
            - PISN-lower-limit
            - PISN-upper-limit
            - PPI-lower-limit
            - PPI-upper-limit


jeffriley:
Close issue #530 when merged

reinhold-willcox (approved):
Looks good, thanks Jeff!

PR #537, opened 2021-03-22
fix bugs in EAGB and TPAGB Nanjing lambda calculation

themikelau:
I have found a possible mistake in the Nanjing lambda calculation for EAGBs and TPAGBs. There are instances where we are using log lambda as lambda, and instances where we are using 1/lambda as lambda. This is inconsistent with StarTrack’s implementation of Nanjing lambda (and also the implementation described by the Xu & Li paper). I have made a pull request of a fix that matches the lambda calculation to StarTrack’s—it is open to anyone for review.

### DIFF
--- a/src/EAGB.cpp
+++ b/src/EAGB.cpp
@@ -347,7 +347,8 @@ double EAGB::CalculateLambdaNanjing() const {
     }
 
     if (lambdaBG.empty()) {                                                 // calculate lambda B & G - not approximated by hand
-        if (utils::Compare(m_Metallicity, LAMBDA_NANJING_ZLIMIT) > 0 && utils::Compare(m_MZAMS, 1.5) < 0) {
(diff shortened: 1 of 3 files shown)


ilyamandel:
@themikelau — great catch!  Can you please show the impact of these changes (compare the old vs corrected lambda implementations for EAGB stars — just the lambdas, no need to propagate to a full COMPAS run)?

themikelau:
Here are some examples of where log lambda was used instead of lambda.
Screenshot 2021-03-22 at 12 46 09 PM
Screenshot 2021-03-22 at 12 46 32 PM

ilyamandel (approved):
This fix brings our implementation in line with the StarTrack implementation.  The affected mass range probably doesn't impact DCO formation.  

There are clearly issues with the Nanjing lambdas in this range (discontinuity at 36 < R/Rsun < 53, very large values of lambda >> 1) which will need to be addressed, probably with a new lambda prescription rather than patches to this model.

PR #540, opened 2021-04-06
Ppisn farmer

LiekeVanSon:
Added Farmer et al. 2019 (http://dx.doi.org/10.3847/1538-4357/ab518b, appendix A, eq. 1) as a PPISN prescription.

Ran simple test grid to inspect output, and COMPAS output seems to match analytical expectation :) (see plot below, '+' markers are CCSN, 'o' markers experienced PPISN)

image


jeffriley (changes requested):
Just a couple of comments:

In Giantbranch.cpp:

1773: 'et al' -> 'et al.'

Consider putting constants (esp cutoffs like 38.0, 60.0, 140.0) in constants.h - not always useful, but often (at least it gives the a symbolic name so everybody knows what they represent)

1795: we don't put calls to std::cerr in the code.  Use SHOW_ERROR() or SHOW_WARN() - you'l see examples throughout the code

Otherwise, looks good (with the caveat that I'm doing this on my phone from a hospital bed, so may well have missed something).

jeffriley (changes requested):
Ok, now that I've looked in more detail

Giantbranch.cpp

1780: delete the clause 'm_COCoreMass >=38.0 and'

1788: delete the clause 'm_COCoreMass >= 60.0 and'

Neither is required - they're just overhead that does nothing.  We already know those conditions are true - otherwise we wouldn't be executing that line of code.
.

and change:

            else if (m_COCoreMass >= 140.0){                                                            // BH mass becomes CO-core mass above the PISN gap
                m_Mass = m_COCoreMass;
              }
            else{
                std::cerr << "\tError in Farmer PPISN function. Shouldn't get here." << std::endl;
            }

to just

            else {                                                            // BH mass becomes CO-core mass above the PISN gap
                m_Mass = m_COCoreMass;
              }


If the CPU is so broken it gets any of those things wrong we're going to have more problems than just not performing the correct PPI prescription.

jeffriley (changes requested):
Oh - and the documentation needs to be updated to reflect the new option value.

LiekeVanSon:
from a hospital bed?! I hope all is OK?

Anyways, thanks for the helpful review! I've left the quadratic fit constants within Giantbranch.cpp since I'd say they are intrinsic to this prescription, but moved the boundaries to constants.h.

I think I added all the other comments, but let me know if I missed something!

jeffriley (approved):
Yeah, all under control - back home now.

Code changes look good.  Approved.

ilyamandel:
@LiekeVanSon -- looking at the code (haven't actually tried it), it seems that it is possible for the post-PPISN mass to exceed the pre-PPISN mass with this prescription (e.g., if as a result of stripping through winds or mass transfer, the amount of mass on top of a CO core is less than 4 solar masses in the linear regime).  That's clearly non-physical, however, so I think we should always cut the maximum remnant mass to be no larger than the total pre-SN mass.  Can you please confirm and, if necessary, make this modification?

LiekeVanSon:
Hi @ilyamandel , You are absolutely right! Thanks for catching this. 

I've just added an extra std::min() to compare with M_mass and opened a new pull request. 
Not entirely sure if that is conform standard procedure.. ! So let me know if I should alter anything.

PR #542, opened 2021-04-20
Enhancements and Defect Repairs

jeffriley:
Enhancements and Defect Repairs:
- Enhancements:
   - Added option to enable users to add program options values to BSE/SSE system parameters files
      - option is '--add-options-to-sysparms', allowed values are {ALWAYS, GRID, NEVER}.  See docs for details.
   - Included "Run_Details" file in HDF5 output file if logfile type = HDF5.  The text Run_Details file still exists
     so users can still easily look at the contents of the Run_Details file - this enhancement adds a copy of the
    Run_Details file to the HDF5 output file.

- Defect Repairs:
   - fixed a few previously unnoticed typos in PROGRAM_OPTION map in constants.h, and in Options::OptionValue() function.
     Fairly benign since they hadn't been noticed, but needed to be fixed.

Modified h5copy.py (in postProcessing/Folders/H5/PythonScripts) so that groups (COMPAS files) will not be copied
if the group exists in the destination file but has a different number of datasets (columns) from the group in
the source file.

Also provided h5view.py - an HDF5 file viewer for COMPAS HDF5 files (in postProcessing/Folders/H5/PythonScripts).  See
documentation at top of source file for details.

The benefits of having the Run_Details file inside the HDF5 file are:

(a) it can be queried programmatically when postprocessing is being done
(b) when multiple COMPAS HDF5 files are joined together, the Run_Details file from each is preserved in the resultant file.  So the number of entries in the Run_Details in the combined file indicates how many smaller files were joined to make the combined file, and since each entry in the Run_Details file in the combined file relates to a constituent HDF5 file, each entry contains the random seed and the number of systems - so the data in the Run_Details entries can be matched to individual systems (by seed).

The benefit of adding the program options to the sysparms file (BSE/SSE) is that, along with the Run_Details file (and the benefits listed above), if a grid file was used, or ranges or sets were specified for option values, the actual values of the program options may change for individual systems (i.e. they may differ from the values recorded in the Run_Details file(s)).  Having the as columns in the sysparms file allows the user to determine the actual values of all program options for each individual system in the HDF5 file.

### DIFF
--- a/docs/COMPAS_LaTex/sections/Appendices/LogFileRecordSpecs-Options.tex
+++ b/docs/COMPAS_LaTex/sections/Appendices/LogFileRecordSpecs-Options.tex
@@ -12,6 +12,8 @@ Following is the list of program option properties available for inclusion in lo
 
 \subsection{Program Option Properties}\label{sec:ProgramOptionProperties}
     
+\optionProperty{ADD\_OPTIONS\_TO\_SYSPARMS}{INT}{Options::m\_AddOptionsToSysParms}{Value of program option \textit{\texttt{-{}-}add-options-to-sysparms}}{Add\_Options\_To\_SysParms}{}
+    
 \optionProperty{ALLOW\_MS\_STAR\_TO\_SURVIVE\_COMMON\_ENVELOPE}{BOOL}{Options::m\_AllowMainSequenceStarToSurviveCommonEnvelope}{Value of program option \textit{\texttt{-{}-}common-envelope-allow-main-sequence-survive}}{Allow\_MS\_To\_Survive\_CE}{}
     
 \optionProperty{ALLOW\_RLOF\_AT\_BIRTH}{BOOL}{Options::m\_AllowRLOFAtBirth}{Value of program option \textit{\texttt{-{}-}allow-rlof-at-birth}}{Allow\_RLOF@Birth}{}
--- a/docs/COMPAS_LaTex/sections/Appendices/ProgramOptions.tex
+++ b/docs/COMPAS_LaTex/sections/Appendices/ProgramOptions.tex
@@ -7,7 +7,7 @@ Following is the list of program options that can be specified at the command li
 
 \programOption{version}{v}{Prints COMPAS version string.}{}
 
-\programOption{add-options-to-sysparms}{}{Add columns for program options to SSE\_System\_Parameters/BSE\_System\_Parameters file (mode dependent). \\ Options: \lcb\ ALWAYS, GRID, NEVER\ \rcb}{NEVER \\ \\ ALWAYS indicates that the program options should be added to the sysparms file \\ GRID indicates that the program options should be added to the sysparms file only if a GRID file is specified, or RANGEs or SETs are specified for options \\ NEVER indicates that the program options should not be added to the sysparms file}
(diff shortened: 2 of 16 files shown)


ilyamandel:
@jeffriley -- why would we not add the options to the system parameter files (at least when a grid was used, but perhaps always)?  It seems like we should keep this information, and the overhead is sufficiently small that we should just do this as a default, not make it optional....

ilyamandel (commented):
In Log.cpp, there's a comment about FHDF5 files.  What are these?  Or is this a typo for HDF5?

ilyamandel:
Other than comments above, I have not spotted any issues, so happy to approve.

jeffriley:
> why would we not add the options to the system parameter files (at least when a grid was used, but perhaps always)? 
> It seems like we should keep this information, and the overhead is sufficiently small that we should just do this as a 
> default, not make it optional

Adding the options to the system parameters file increases the number of columns from ~30 to ~180 (I think).  I took the most benign option - don't change anything unless the user asks for it.  I don't think we should do it always (unless that's what the user wants) - if no grid file is used (or ranges or sets specified), the options are the same for every system - we could be printing 150 columns of the same data for millions of systems.

I'd be happy to make the default for the '--add-options-to-sysparms' option 'GRID' - then the user can change it to 'ALWAYS' or 'NEVER' at their discretion.  But I'm also happy to take advice - if the consensus is that the default should be 'ALWAYS', I'll do that.

> In Log.cpp, there's a comment about FHDF5 files. What are these? Or is this a typo for HDF5?

A typo :-)

jeffriley:
So, I propose changing the default for  '--add-options-to-sysparms' option to 'GRID'.  If no grid file (or range or set) is used, then the options are as they appear in the Run_Details file - which is now in the HDF5 file.

stellarphysicsandpopulations:
Amusing that this was created so similar in time to Floris & my realisation.   Good that you were already doing this; thanks. 

I now see that the option you're creating to solve the problem I was worrying about is indeed the "add all the options to the record of every system" option.   I agree that that's an improvement, and does solve the problem.  Although, in principle, I still think an intermediate way which saves all the invariant options in Run_Details would probably be cleaner for the user, and remove noise from the system parameters files. (But would maybe be too messy to code to be worth it.)  

(In my imagination, most of the time most of the options will be global/invariant, even when using a grid file.  But my imagination for how GRID files might be used is probably blinkered by how I've been using them. My imagined worries aren't important anyway, as long as the documentation is clear.)

stellarphysicsandpopulations:
Slightly reformulating one of the ideas from Slack: 
    If, e.g., COMPAS accepted a string on the command line to mean "we only want to specify this parameter from the GRID file", and this was recorded in the Run_Details, then I think that would solve my worries/use case whilst still preserving the ability to keep the invariant global parameters cleanly in a Run_Details file. 
    
Okay, I shut up now.  Sorry if I used a lot of words to unclearly say a simple thing.

ilyamandel (approved):
Thank you for addressing my request, @jeffriley !  

@stellarphysicsandpopulations -- I see your point that there could be another option, to only place system-unique information that is strictly necessary in the params file.  This would save disk space and I/O time.  On the other hand, it would take extra time to code up (both for writing the information and for reading it out).  I propose we wait to see if there is a strong use case for this option before spending that time.  If there is, we can always add another feature request. :)

PR #546, opened 2021-04-26
Fix to error handling in h5view.py

jeffriley:
Fixed error handling in h5view.py in a few places - was printing wrong variable for filename.

No changes to COMPAS C++ code - no version change.  Change is to post-processing tool.  We probably should version the post-processing code too...


ilyamandel (approved):
Thank you, @jeffriley !

PR #547, opened 2021-04-27
Ppisn farmer

LiekeVanSon:
Fixed issue in PPISN Farmer where the mass could exceed the stellar mass in the linear regime

### DIFF
--- a/src/GiantBranch.cpp
+++ b/src/GiantBranch.cpp
@@ -1773,7 +1773,7 @@ STELLAR_TYPE GiantBranch::ResolvePulsationalPairInstabilitySN() {
         case PPI_PRESCRIPTION::FARMER: {                                                                // Farmer et al. 2019 http://dx.doi.org/10.3847/1538-4357/ab518b
                                                                                                         // Three cases:
             if (m_COCoreMass < FARMER_PPISN_UPP_LIM_LIN_REGIME){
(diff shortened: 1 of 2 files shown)


ilyamandel:
Thank you, @LiekeVanSon , but I don't think this covers all bases.  For example, suppose the CO core mass is 38.01 solar masses (and there's no envelope on top at the time of PPISN, as an extreme case), and the metallicity is log Z = - 4.  Then, following the logic in
 else if (m_COCoreMass < FARMER_PPISN_UPP_LIM_QUAD_REGIME)
I believe you will get a remnant mass of 42.13 solar masses, which exceeds the total mass.  

So I think the correct solution has to be to check whether the mass exceeds the total mass at the very end of the calculation, not inside the individual cases.

LiekeVanSon:
Hi @ilyamandel, sorry for my really slow response, it's been a bit of a hectic week! 
You're right, thanks!  I moved the mass check to the end of the `case PPI_PRESCRIPTION::FARMER`:, let me know if this is what you had in mind (and sorry if I've still misunderstood and you meant to putting the check at the end of `ResolvePulsationalPairInstabilitySN()`

ilyamandel (changes requested):
Looks good, thank you, @LiekeVanSon !  But given that we have made other changes in the meantime, I think it's best for transparency to record this in the changelog and assign a new version number.

ilyamandel (approved):
Thank you, @LiekeVanSon !

PR #550, opened 2021-04-30
Issue548 +

### LINKED ISSUE #548: Detailed output not printing when multiple gridlines specify a random seed
**Describe the bug**
This is a pretty specific corner case, but I think I've reduced this to a MWE.

Detailed output is not working correctly, precisely when a gridfile is used which has the --random-seed parameter specified *more than once*. 

If detailed output is turned off, or if the gridfile specifies --random-seed on only one line (or none), the output seems to work fine. However, if there is more than one line with --random-seed specified, the following error is output:

```
ERROR: Unable to create HDF5 detailed file with file name /home/rwillcox/astro/compas/COMPAS/src/COMPAS_Output_7/Detailed_Output/BSE_Detailed_Output_0.h5
Logging disabled
Error writing to file - data not written while writing to logfile BSE_System_Parameters
Segmentation fault (core dumped)
```

This occurs on the current most up to date `dev` branch, as well as quite a few past ones (all the ones I tested, though at some point far enough back the code/tests don't work anymore). It also occurred for me on my desktop (Ubuntu 18.04) and on OzStar, so I don't think this is an OS issue.

**Label the issue**

**To Reproduce**
Create grid.txt:
```
--random-seed 0 --initial-mass-1 10    
--random-seed 1 --initial-mass-1 10   
```
And run `COMPAS --detailed --grid grid.txt`. You should see the error. Then edit grid.txt:
```
--random-seed 0 --initial-mass-1 10    
--initial-mass-1 10   
```
and rerun, there should be no error.

**Expected behavior**
The detailed output should be produced for all the systems, regardless of the random-seed. 

 - Ubuntu 18.04
 - COMPAS v02.19.00

jeffriley:
Enhancements and Defect Repairs:
- Enhancements:
   - changed chunk size for HDF5 files to HDF5_MINIMUM_CHUNK_SIZE for Run_Details group in COMPAS_Output and for detailed output files.
      - Run_Details is a small file, and detailed output files are generally a few thousand records rather than hundreds of thousands, 
        so a smaller chunk size wastes less space and doesn't impact performance significantly

- Defect Repairs:
   - fixed issue #548 - HDF5 detailed output files not created when random-seed specified in a grid file
   - fixed defect where records in HDF5 output files would be duplicated if the number of systems exceeded the HDF5 chunk size
     being used (the default chunk size is 100000 - that might explain why this problem hasn't been reported)

 Modified h5view.py (in postProcessing/Folders/H5/PythonScripts) to handle detailed ouput files

Close issue #548 when merged

### DIFF
--- a/postProcessing/Folders/H5/PythonScripts/h5view.py
+++ b/postProcessing/Folders/H5/PythonScripts/h5view.py
@@ -159,20 +159,35 @@ def printSummary(h5name = None, h5file = None, excludeList = ''):
         print('\nFile size    :', strFileSize.strip(), 'GB')
         print('Last modified:', lastModified)
 
-        # get widths for columns to be displayed - it's a bit more
-        # work, and a bit redundant, but it is neater... (and we don't have thousands of files...)
+        # get widths for columns to be displayed - it's a bit more work, 
+        # and a bit redundant, but it is neater... (and we don't have thousands of files...)
         maxFilenameLen  = -1
         maxColumns      = -1
         groupMaxEntries = []
-        for gIdx, group in enumerate(h5file.keys()):
+
+        keyList = list(h5file.keys())
+        detailedOutput = isinstance(h5file[keyList[0]], h5.Dataset)                                 # processing detailed output file?
+
+        if detailedOutput:                                                                          # detailed output file
+
+            maxColumns = len(keyList)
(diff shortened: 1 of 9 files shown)


ilyamandel (approved):

PR #551, opened 2021-05-06
FastCosmicIntegrator fix

ilyamandel:
Fixing poor behaviour of FastCosmicIntegrator when Mc_max is below maximum binary chirp mass * (1+maximum redshift for detectability calculation)
Also fixing incorrect behaviour of detectability calculator when SNR is below minimal SNR step.

More details:

There was a recently discovered issue, where when @jeffriley used max_redshift_detection=10, he got unreasonable detections coming from very high redshifts.

The reason for this was that Mc_max was still set to 300, so when a binary with Mc of 35 was placed at z=10, it had a redshifted chirp mass above Mc_max.  This was exacerbated by the fact that when the SNR was very low, we would incorrectly report a high detection probability.
The solution is:
add a new warning if max(chirp_masses)*(1+max_redshift_detection) < Mc_max
when the SNR is so low that its index into the SNR array is negative, record a detection probability of zero.


jeffriley (approved):
Tested - looks good - thanks @ilyamandel !

PR #553, opened 2021-05-09
Enhancement: Add lower bound for secondary mass to totalMassEvolvedPerZ()

jeffriley:
Add lower bound for secondary mass to totalMassEvolvedPerZ() in totalMassEvolvedPerZ.py (post-processing script).

More accurately reflects range of masses in COMPAS population.

### DIFF
--- a/postProcessing/Folders/CosmicIntegration/PythonScripts/totalMassEvolvedPerZ.py
+++ b/postProcessing/Folders/CosmicIntegration/PythonScripts/totalMassEvolvedPerZ.py
@@ -159,26 +159,27 @@ def retrieveMassEvolvedPerZ(path, fileName):
 
 
 
-def totalMassEvolvedPerZ(path=None, fileName=None, Mlower=None, Mupper=None, binaryFraction=0.7, \
+def totalMassEvolvedPerZ(path=None, fileName=None, M1lower=None, M1upper=None, M2lower=None, binaryFraction=0.7, \
                          x1=0.01, x2=0.08, x3=0.5, x4=200., a1=-0.3, a2=-1.3, a3=-2.3, C1=1., Mmax=200):
 
     #the default values assume a Kroupa IMF for M1
     if path is None:
         raise TypeError("\n Need to give path to directory COMPASOutput.h5")
-    if Mlower is None:
+    if M1lower is None:
         raise TypeError("\n Need to give lower limit M1 of pythonSubmit")
-    if Mupper is None:
+    if M1upper is None:
         raise TypeError("\n Need to give upper limit M1 of pythonSubmit")
+    if M2lower is None:
+        raise TypeError("\n Need to give lower limit M2 of pythonSubmit")
     
-
     M1, M2 = createSampleUniverse(binaryFraction=binaryFraction, x1=x1, x2=x2, x3=x3, x4=x4, \
                                   a1=a1, a2=a2, a3=a3, C1=C1, Mmax=Mmax)
 
     totalMassInStarFormation = np.sum(M1) + np.sum(M2)
 
     #Now mask M1 and M2 to see what lies in the range of COMPAS
-    maskM1 = (M1>=Mlower) & (M1<=Mupper)
-    maskBinaries = (M2!=0)
+    maskM1 = (M1 >= M1lower) & (M1 <= M1upper)
+    maskBinaries = (M2 >= M2lower)
     mask = maskM1 & maskBinaries
 
     totalMassEvolvedCOMPAS = np.sum(M1[mask]) + np.sum(M2[mask])
--- a/postProcessing/Folders/CosmicIntegration/PythonScripts/ClassCOMPAS.py
+++ b/postProcessing/Folders/CosmicIntegration/PythonScripts/ClassCOMPAS.py
@@ -128,8 +128,9 @@ class COMPASData(object):
         _, self.totalMassEvolvedPerZ = MPZ.totalMassEvolvedPerZ(
             path=self.path,
             fileName=self.fileName,
-            Mlower=self.Mlower,
-            Mupper=self.Mupper,
+            M1lower=self.Mlower,
+            M1upper=self.Mupper,
+            M2lower=self.m2_min,
             binaryFraction=self.binaryFraction,
         )
         # Want to recover entire metallicity grid, assume that every metallicity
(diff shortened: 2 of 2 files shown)


ilyamandel (approved):
That looks fine, thanks, @jeffriley ! 
self.Mlower vs. self.m2_min is perhaps not the clearest / most consistent notation, but I imagine it's required for backward consistency (i.e., to keep the Jupyter notebooks working -- are they? :) ).  Perhaps a comment would be helpful (though I am not sure where it should go, so approving without insisting on that :) ).

jeffriley:
Hmmm.  I don't know if the nomenclature is required for the Jupyter notebooks - I don't use them :-)

I chose M2lower for totalMassEvolvedPerZ() because it already used Mlower and Mupper as parameters (as does class COMPAS).  I don't know why class COMPAS uses m2_min instead of M2lower, but it does...  

Given your comment re the Jupyter notebooks, maybe I should redo the parameter names for totalMassEvolvedPerZ() just to be safe.

I'll do that - I'll push a change.

ilyamandel:
I don't use the Jupyter notebooks either, I just meant that we should make sure we don't break them when we update the core python classes.  @themikelau , @reinhold-willcox , could you confirm?

reinhold-willcox:
@jeffriley @ilyamandel I don't use the notebooks much either, but this does seem to break at least the 1_ClassCOMPAS notebook. I don't have time to dig into this right now, but potentially in the next day or so I can take a closer look.

reinhold-willcox (changes requested):
Breaks some of the functionality in the jupyter notebooks.

themikelau:
Interesting—I haven't really thought about the minimum secondary mass in the pythonSubmit. I have always thought that I have been drawing primary masses from 5 to 150 Msun and mass ratios uniformly from 0.01 to 1. But the minimum secondary mass of 0.1 Msun implies that the actual birth mass ratio distribution is uniform from 0.02 to 1.

reinhold-willcox:
@themikelau I think that's not quite right, your birth mass ratio distribution is only U(.02, 1) when M1 is 5. More generally, it's a conditional distribution, so I believe there's not a simple one line formula to capture it. Same problem with the Moe+DiStefano prescriptions.

themikelau:
@reinhold-willcox Yeah I think you are right. I think the correct statement is that for any fixed M1, q will be uniformly distributed in (0.1/M1, 1). But M1 is itself drawn from the IMF, so the q distribution is more complicated than that.

reinhold-willcox:
While we're working out the best solution to how to integrate over the initial sampling distributions, I would like to advocate that we change these parameter names consistently, e.g switch to m1Lower, m1Upper, m2Lower everywhere (doesn't have to be those names, just the same name for the same variable).

ilyamandel (commented):
There's a "miluplu" in lieu of multiply in a comment in totalMassEvolved...

ilyamandel:
@TomWagg , @jeffriley :

line 107 in totalMassEvolvedPerZ.py

drawM2 = np.random.uniform(0,1,sampleSize)

does not take into account the possibility that the user may wish to draw from a different range of mass ratios; e.g., in defaults/pythonSubmit.py :

mass_ratio_distribution = 'FLAT'
  mass_ratio_min = 0.01
  mass_ratio_max = 1.0
  
It should.

postProcessing/Folders/CosmicIntegration/PythonScripts/totalMassEvolvedPerZ.py
```diff
@@ -159,26 +159,27 @@ def retrieveMassEvolvedPerZ(path, fileName):
 
 
 
-def totalMassEvolvedPerZ(path=None, fileName=None, Mlower=None, Mupper=None, binaryFraction=0.7, \
+def totalMassEvolvedPerZ(path=None, fileName=None, M1lower=None, M1upper=None, M2lower=None, binaryFraction=0.7, \
                          x1=0.01, x2=0.08, x3=0.5, x4=200., a1=-0.3, a2=-1.3, a3=-2.3, C1=1., Mmax=200):
```
TomWagg, line 163: Is it clear to anyone else what the difference between `M1upper` and `Mmax` is? In the same vein, we don't seem to pass `M1lower` to `createSampleUniverse` (leaving `M1min=0.01` as default)

ilyamandel: Is that a rhetorical question? :)  From the code, I assume that Mmax is the maximum stellar mass in the Universe (the upper cutoff of the Kroupa IMF), while Mupper or M1upper is the maximum mass simulated in a COMPAS run.  I think you should feel free to use more descriptive names, @TomWagg , as long you keep all the functions, classes, and notebooks mutually consistent.

ilyamandel:
@TomWagg -- I just noticed that FastCosmicIntegration uses COMPAS.mass_evolved_per_binary for normalisation, while ClassCosmicIntegrator uses COMPAS.totalMassEvolvedPerZ, which is based on totalMassEvolvedPerZ.  I don’t know why, but they are different code paths.  @jeffriley seems to find that the two normalisations are not consistent (I haven't checked).  Should they move to using the same code path?

postProcessing/Folders/CosmicIntegration/PythonScripts/totalMassEvolvedPerZ.py
TomWagg: Ah my apologies @ilyamandel, not a rhetorical question but perhaps a misguided one sorry 😬 

I got confused because `x4` also seems to be the upper cutoff the Kroupa IMF but I guess the `x4` just says when Kroupa stops and `Mmax` is used to tweak this in some way? I honestly still don't quite understand what the difference between the `x4` and `Mmax` is (which is why I wondered if `Mmax` is related to `M1upper`)

Sorry for my ignorance!

TomWagg:
@ilyamandel yes this is what I was referring to in the Slack thread. I made the `COMPAS.mass_evolved_per_binary` variable and associated functions to replace the `totalMassEvolvedPerZ.py` when I was working on your fast cosmic integration code but didn't change everything in the last PR as it was pointed out that it would break the notebooks. So my plan is to adjust everything to use this one instead and fix the notebooks at the same time now. Sorry for the delay with this, I've been at a conference for multiband GW detections this week so haven't had much time to code.

But more worrying that the normalisations are not consistent, could you elaborate @jeffriley?

jeffriley:
Moving to draft - this has been here so long I've forgotten where we're up to - I'll have to review...

ilyamandel:
@TomWagg , @jeffriley, @themikelau , @reinhold-willcox  -- a ping on this one.

jeffriley:
Will review asap

ilyamandel:
@jeffriley, this PR is 2.5 years old.  I don't know if it's relevant any more.  Is it time to kill it?

ilyamandel:
See above. :-)

PR #554, opened 2021-05-10
Demo plot updates

reinhold-willcox:
Small changes to the demo system and plotting routine to improve the plot used in the paper.

Specifically, added reduced kicks so that the post-SN eccentricity wasn't so high (previously was >.9), and then fixed the bounds of the eccentricity subplot to be between 0 and 1.


reinhold-willcox:
Sorry, was not on dev when I made these changes, ignore this PR.

PR #555, opened 2021-05-10
updates to the demo plotting script

reinhold-willcox:
Updated the demo plot with the latest version of Compas. The post-SN eccentricity is now lower, so I've added limits to the eccentricity subplot to fix the boundaries at 0 and 1. The new plot has now been uploaded to the methods paper.


themikelau (approved):
Good

PR #557, opened 2021-05-10
Issue #556: Fixes to DOMINANT_MASS_LOSS_RATE description

### LINKED ISSUE #543: Documentation inconsistency
At the moment, we have COMPAS_Doc.pdf, COMPAS_CHE.docx, COMPAS_CHE.pdf, and a latex directory which does not automatically produce PDF outputs. The instructions in getting_started.md point to COMPAS_Doc.pdf, which is out of date relative to the latex files.  A non-expert user may not know where to look.  

I propose we delete COMPAS_Doc.pdf, COMPAS_CHE.docx, COMPAS_CHE.pdf (any information contained there should be available via the latex documentation) to avoid confusion, and update the getting_started.md documentation to point to it.   It would be great if we could set up an automatic re-compilation of a pdf from the .tex files, but that's less urgent as long as we have clear instructions.

### LINKED ISSUE #556: Documentation minor issues
I can't search the PDF compiled from latex documentation for entries with an underscore -- the PDF file seems to think it's a space.

There is something funny in the DOMINANT MASS LOSS RATE definition:
"Note that this property has the same header string as
. It is expected that one or the other is printed in any file, but not both. If both are printed then the file will
contain two columns with the same header string."

I could not find where the meanings of (SN), (CP), etc. were defined.  We do say the following at the top of Log File Record Specification: Stellar Properties, but I am not sure I would be able to guess that (CP) stands for companion, for example.
 STAR PROPERTY for SSE
 STAR 1 PROPERTY for the primary star of a binary for BSE
 STAR 2 PROPERTY for the secondary star of a binary for BSE
 SUPERNOVA PROPERTY for the exploding star in a supernova event for BSE
 COMPANION PROPERTY for the companion star in a supernova event for BSE

The documentation does not seem to say that Mass(SN) is the mass after the SN.

jeffriley:
For issue #556:

Added missing header information for DOMINANT_MASS_LOSS_RATE
Also fixed formatting of the itemised list

For issue #543:

Deleted COMPAS_CHE.docx and COMPAS_CHE.pdf
Generated updated COMPAS.pdf

Close issue #556 when merged.  Issue #543?


ilyamandel (approved):
Thank you for the quick fix!

PR #561, opened 2021-05-17
Update cosmic post processing notebooks

### LINKED ISSUE #559: Jupyter notebook to use latest FastCosmicIntegration
The relevant Jupyter notebook should be switched to use the now standard FastCosmicIntegration code; once that is done, we can then phase out the no longer maintained (?) older cosmic integration code.

FloorBroekgaarden:
I updated the first of the post-processing scripts: 2_MSSFR/ SFRD script, that shows users of compas a demo of how to use and interpretate the SFRD scripts (to construct the SFRD). This is a jupyter notebook that is independent of a COMPAS simulation, i.e. it can be run and used without having COMPAS data and is meant to give insights in some of the functions behind the post-processing. 

It also nicely recreates Figures 11, 13 and 14 in the COMPAS methods paper.. 

partially resolves #559 (first of several scripts that needed updating)


ilyamandel (approved):
Looks good to me!

PR #563, opened 2021-05-18
Directory reorganization

reinhold-willcox:
- Renamed defaults folder to preProcessing, to avoid confusion with compas "defaults". This folder should now contain the most generic versions of pythonSubmit, stroopwafel_interface, SSE/BSE grids, and the like.
- Renamed demo to examples.
- Created subdir examples/methods_paper_plots to hold all plots (and the scripts that generate them) that we use in the methods paper. As we update Compas versions on the way to publishing, we can simply rerun the scripts in these folders to regenerate the plots and ensure they are (i) consistent and (ii) up to date. Currently in that folder we have chirpmass_distribution and detailed_evolution, but others should put their plots in there too.


themikelau (changes requested):
Need to update the "Getting Started" tutorial and other tutorials to reflect these changes in directory names.

reinhold-willcox:
Good catch @themikelau . I've pushed those changes, but we should probably review all the .md files at some point, they're a bit outdated.

ilyamandel:
Looks good to me, but @themikelau has looked at this more carefully, so he should review.

themikelau (approved):
All good!

PR #566, opened 2021-05-18
Change default LBV wind prescription 💨 

TomWagg:
@demink asked me to make this PR to change the default LBV wind prescription to HURLEY_ADD instead of BELCZYNSKI based on the discussion in the methods paper meeting. I *think* I changed it everywhere but hopefully the ever wise and knowledgeable @jeffriley can set me straight if not 😁 

If it's helpful, I changed it:

- in pythonSubmit.py
- in baseStar.cpp if an unknown prescription is given
- in options.cpp
- in the LaTeX file


ilyamandel (commented):
Looks good to me, but will leave this for @jeffriley to approve.

jeffriley:
Happy to approve, but @ilyamandel this reminds me that we have another level of defaults in the code.  Note in BaseStar.cpp the default case for the switch specifies a default LBV prescription.  This is purely defensive programming - the switch needs a default case.  When I ported the legacy code I chose to make the default case the default LBV prescription - one alternative would be to show a warning and abandon processing of the binary.  The point is, the way the code is at the moment, here and in other places, switch statement default cases may be hard-coded to use a particular, default, option value.  The question arising is whether we should endeavour to keep those in sync with the defaults in Options.cpp and pythonSubmitDefault?

jeffriley:
@ilyamandel Also note this bumps the version number to v02.19.03 - a magic number wrt the methods paper...

jeffriley (approved):
Thanks @TomWagg!  Good job - and yes, I think you changed it everywhere it should be changed :-)

ilyamandel:
@jeffriley -- well, we'll leave the methods paper number until we have pulled defaults over from pythonSubmit into Options.cpp :)

ilyamandel:
@jeffriley -- Yes, I noticed the change in the fallback option, too.  Since this should never be encountered in practice (as you say, it's just defensive programming), I don't think it really matters what we put there.  That said, I agree that an error + abandoning the binary would not be terrible (the user should probably know that the code didn't go down the expected path).

PR #567, opened 2021-05-19
Add methods paper plots/code (Figures 5-7)

TomWagg:
Here is the code to reproduce the figures that I made for the methods paper.

I've added three sub-directories (one for each plot) that contain:

- A README.md file explaining how to get the plot
- A Jupyter notebook that makes the plot
- A python file that creates the necessary grid file
- A python submit file with slightly altered defaults
- A PDF of the plot (in case they don't want to run the code)

I also added the `stellar_types.py` that just contains the colours I use for the plots.

**For testing I recommend pulling and just following the README for each folder and it should hopefully give you some plots :)**

Let me know if you need to me add anything! 😄


reinhold-willcox (approved):
Looks really good Tom, nice work! I ran it as is following the README and found no problems.

PR #571, opened 2021-05-24
Small update thermal timescale

### LINKED ISSUE #570: change pre-factor in thermal (kelvin-helmholtz) timescale calculation
The function `CalculateThermalTimescale` uses a pre-factor of GMsun^2 / Rsun Lsun = 30.0 Myr, but when you calculate this it should actually be 31.4 Myr (see attachment) so why not update this? I also updated the documentation

I am changing the 30.0 to 31.4 Myr. Since both numbers have the same precision , and hence the same computational cost, I do not see a (big) problem. 

Screen Shot 2021-05-24 at 10 43 29 AM

FloorBroekgaarden:
closes #570 

I updated the pre-factor and added documentation. I updated the log file accordingly

### DIFF
--- a/src/GiantBranch.cpp
+++ b/src/GiantBranch.cpp
@@ -986,7 +986,8 @@ double GiantBranch::CalculateLifetimeToHeIgnition(const double p_Mass, const dou
 /*
  * Calculate thermal timescale
  *
- * Kalogera & Webbink 1996, eq 2 [note that (61) of BSE proposes a value a factor of 3 smaller]
+ * pre-factor from Kalogera & Webbink 1996 (https://arxiv.org/abs/astro-ph/9508072), equation 2, combined with p_Mass * p_EnvMass case from equation 61 from https://arxiv.org/abs/astro-ph/0201220 for k in {2,3,4,5,6,8,9}
+ * [note that equation 61 of BSE (https://arxiv.org/abs/astro-ph/0201220) approximates this with a value a factor of 3 smaller]
  *
  *
  * double CalculateThermalTimescale(const double p_Mass, const double p_Radius, const double p_Luminosity, const double p_EnvMass) const
@@ -998,7 +999,7 @@ double GiantBranch::CalculateLifetimeToHeIgnition(const double p_Mass, const dou
  * @return                                      Thermal timescale in Myr
 */
 double GiantBranch::CalculateThermalTimescale(const double p_Mass, const double p_Radius, const double p_Luminosity, const double p_EnvMass) const {
-    return 30.0 * p_Mass * p_EnvMass / (p_Radius * p_Luminosity);       // G*Msol^2/(Lsol*Rsol) ~ 30 Myr
+    return 31.4 * p_Mass * p_EnvMass / (p_Radius * p_Luminosity);       // G*Msol^2/(Lsol*Rsol) ~ 31.4 Myr (~ 30 Myr in Kalogera & Webbink)
 }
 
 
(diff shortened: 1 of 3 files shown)


ilyamandel:
While looking at this, I started wondering if HeHG.h correctly overwrites the "no-envelope" behaviour of HeMS.h for CalculateThermalTimescale(const double p_Mass,
                                                      const double p_Radius,
                                                      const double p_Luminosity).  
@jeffriley , would you mind taking a quick look?

jeffriley:
HeHG inherits CalculateThermalTimescale() from HeMS.  HeMS::CalculateThermalTimescale() calls MainSequence::CalculateThermalTimescale() which ignores the p_EnvMass parameter.

So, I think from your question, HeHG should implement it's own version of CalculateThermalTimescale(), calling GiantBranch::CalculateThermalTimescale() which does not ignore the p_EnvMass parameter?

HeGB inherits from HeHG (so currently ignores p_EnvMass) - so if we change HeHG we'll need to decide what to do with HeGB

FloorBroekgaarden:
good point @jeffriley @ilyamandel , it actually also made me wonder whether the MainSequence::CalculateThermalTimescale() needs to pass the variable p_EnvMass? It currently does, but there's no need to? Or is it preferred to keep them the same?

jeffriley:
> it actually also made me wonder whether the MainSequence::CalculateThermalTimescale() needs to pass the variable p_EnvMass?
> It currently does, but there's no need to? Or is it preferred to keep them the same?

The signature needs to be the same for inheritance - I think that's why I did it that way

jeffriley:
(And because p_EnvMass has a default value in MainSequence.h, callers don't need to pass the value - it's just there to satisfy the signature requirement)

ilyamandel:
Cleaner alternative: since each star knows whether it has an envelope (whether m_Mass == m_CoreMass), why not just have a single function in the base class that will use m_Mass if m_Mass == m_CoreMass, and  the envelope mass, m_Mass - m_CoreMass, otherwise?  That way, we won't need multiple overrides in the inheritance tree.

jeffriley:
@FloorBroekgaarden  Is this still a WIP?  Should we mark it as a draft?

FloorBroekgaarden:
I'm not sure what we decided? My part (changing the pre factor) is good to go. But @ilyamandel do we want to then implement your suggestion?

jeffriley:
@ilyamandel I was just looking at implementing your suggestion, but noticed that Remnants.h has CalculateThermalTimescale() call CalculateDynamicalTimescale(), which ultimately calls BaseStar::CalculateDynamicalTimescale_Static().

So, just to be sure, implement your suggestion for everything except remnants, and leave the override in Remnants.h?

ilyamandel:
Hi @jeffriley , @FloorBroekgaarden -- sorry for the silence.  Yes to @jeffriley 's last suggestion.  @jeffriley , I leave it to you whether it's easier to accept this change first or implement the more general one we discussed above. :)

jeffriley:
Ok, I'll take a look.  What makes this a little more problematic is the docs - we need to update the online docs, but we don't yet have them in the COMPAS repo.  I can do that separately.  I'll look at how to get all this done with minimum fuss :-)

ilyamandel:
Ping for @jeffriley  :)

jeffriley:
That was 2 weeks ago?  Time files...  Ok, I will get to this - promise

jeffriley:
Closed in favour of PR #646

PR #574, opened 2021-06-01
Ensure that no file paths in pythonSubmit.py contain spaces

### LINKED ISSUE #573: If grid file is in a folder that has spaces in the name it breaks
**Describe the bug**
If the name of the folder has spaces in the name (for example if someone was in a rush and wrote "helium stars" as the name of the directory instead of "helium_star") then you get an error that says

```
Commandline Options error: option is ambiguous and matches 'PISN-lower-limit', 'PISN-upper-limit', 'PPI-lower-limit', 'PPI-upper-limit', 'add-options-to-sysparms', 'allow-rlof-at-birth', 'allow-touching-at-birth', 'angular-momentum-conservation-during-circularisation', 'black-hole-kicks', 'case-BB-stability-prescription', 'check-photon-tiring-limit', 'chemically-homogeneous-evolution', 'circularise-binary-during-mass-transfer', 'common-envelope-allow-main-sequence-survive', 'common-envelope-alpha', 'common-envelope-alpha-thermal', 'common-envelope-lambda', 'common-envelope-lambda-multiplier', 'common-envelope-lambda-prescription', 'common-envelope-mass-accretion-constant', 'common-envelope-mass-accretion-max', 'common-envelope-mass-accretion-min', 'common-envelope-mass-accretion-prescription', 'common-envelope-recombination-energy-density', 'common-envelope-slope-kruckow', 'cool-wind-mass-loss-multiplier', 'debug-classes', 'debug-level', 'debug-to-file', 'detailed-output', 'eccentricity-distribution', 'eccentricity-max', 'eccentricity-min', 'eddington-accretion-factor', 'enable-warnings', 'envelope-state-prescription', 'errors-to-file', 'evolve-pulsars', 'evolve-unbound-systems', 'fix-dimensionless-kick-magnitude', 'fryer-supernova-engine', 'grid', 'hdf5-buffer-size', 'hdf5-chunk-size', 'initial-mass', 'initial-mass-1', 'initial-mass-2', 'initial-mass-max', 'initial-mass-min', 'initial-mass-power', 'kick-direction', 'kick-direction-power', 'kick-magnitude', 'kick-magnitude-1', 'kick-magnitude-2', 'kick-magnitude-distribution', 'kick-magnitude-max', 'kick-magnitude-random', 'kick-magnitude-random-1', 'kick-magnitude-random-2', 'kick-magnitude-sigma-CCSN-BH', 'kick-magnitude-sigma-CCSN-NS', 'kick-magnitude-sigma-ECSN', 'kick-magnitude-sigma-USSN', 'kick-mean-anomaly-1', 'kick-mean-anomaly-2', 'kick-phi-1', 'kick-phi-2', 'kick-scaling-factor', 'kick-theta-1', 'kick-theta-2', 'log-classes', 'log-level', 'logfile-common-envelopes', 'logfile-definitions', 'logfile-detailed-output', 'logfile-double-compact-objects', 'logfile-name-prefix', 'logfile-pulsar-evolution', 'logfile-rlof-parameters', 'logfile-supernovae', 'logfile-switch-log', 'logfile-system-parameters', 'logfile-type', 'luminous-blue-variable-multiplier', 'luminous-blue-variable-prescription', 'mass-loss-prescription', 'mass-ratio-distribution', 'mass-ratio-max', 'mass-ratio-min', 'mass-transfer', 'mass-transfer-accretion-efficiency-prescription', 'mass-transfer-angular-momentum-loss-prescription', 'mass-transfer-fa', 'mass-transfer-jloss', 'mass-transfer-rejuvenation-prescription', 'mass-transfer-thermal-limit-C', 'mass-transfer-thermal-limit-accretor', 'maximum-evolution-time', 'maximum-mass-donor-nandez-ivanova', 'maximum-neutron-star-mass', 'maximum-number-timestep-iterations', 'mcbur1', 'metallicity-distribution', 'metallicity-max', 'metallicity-min', 'minimum-secondary-mass', 'mode', 'muller-mandel-kick-multiplier-BH', 'muller-mandel-kick-multiplier-NS', 'neutrino-mass-loss-BH-formation', 'neutrino-mass-loss-BH-formation-value', 'neutron-star-equation-of-state', 'orbital-period', 'orbital-period-distribution', 'orbital-period-max', 'orbital-period-min', 'overall-wind-mass-loss-multiplier', 'pair-instability-supernovae', 'population-data-printing', 'print-bool-as-string', 'pulsar-birth-magnetic-field-distribution', 'pulsar-birth-magnetic-field-distribution-max', 'pulsar-birth-magnetic-field-distribution-min', 'pulsar-birth-spin-period-distribution', 'pulsar-birth-spin-period-distribution-max', 'pulsar-birth-spin-period-distribution-min', 'pulsar-magnetic-field-decay-massscale', 'pulsar-magnetic-field-decay-timescale', 'pulsar-minimum-magnetic-field', 'pulsational-pair-instability', 'pulsational-pair-instability-prescription', 'quiet', 'random-seed', 'remnant-mass-prescription', 'revised-energy-formalism-nandez-ivanova', 'rlof-printing', 'rotational-frequency', 'rotational-frequency-1', 'rotational-frequency-2', 'rotational-velocity-distribution', 'semi-major-axis-distribution', 'semi-major-axis-max', 'semi-major-axis-min', 'stellar-zeta-prescription', 'switch-log', 'timestep-multiplier', 'use-mass-loss', 'wolf-rayet-multiplier', 'zeta-adiabatic-arbitrary', 'zeta-main-sequence', and 'zeta-radiative-envelope-giant'`
```

which seems very difficult for people to find.

(Disclaimer: I realise it is crazy to be putting spaces in directory names, but I figured since the error message seems to be completely unrelated it might be good to document this somewhere/throw a better error)

**Label the issue**
`urgency_low`, `severity_minor`

**To Reproduce**
1. Use usual pythonSubmit in `preProcessing`
2. Change directory name to be "pre Processing"
3. Run python submit

**Expected behavior**
Usual run occurs.

 - OS: MacOS 10.15.7 Catalina
 - COMPAS v.02.19.04

TomWagg:
This PR fixes #573.

It checks the input and output filepaths plus the logfile definitions ones. If it finds a space, it throws an error and points out where in the filepath the user left a space and exits.

image

### DIFF
--- a/preProcessing/pythonSubmit.py
+++ b/preProcessing/pythonSubmit.py
@@ -94,6 +94,14 @@ class pythonProgramOptions:
         else:
             logfile_definitions = compas_input_path_override + '/' + logfile_definitions
 
+    # ensure that no file paths have spaces that could confuse Boost
+    for path in [output, grid_filename, logfile_definitions]:
+        if path is not None and path.find(" ") >= 0:
+            print("ERROR: File path contains spaces, replace with underscores or another character.")
+            print("\t" + path)
+            print("\t" + " " * path.find(" ") + "^")
+            exit(1)
+
     initial_mass    = None                                      # initial mass for SSE
     initial_mass_1  = None                                      # primary initial mass for BSE
     initial_mass_2  = None                                      # secondary initial mass for BSE
--- a/preProcessing/pythonSubmit.py
+++ b/preProcessing/pythonSubmit.py
@@ -25,7 +25,7 @@ class pythonProgramOptions:
     # in the bin directory (rather than the src directory)
     compas_executable_override = os.environ.get('COMPAS_EXECUTABLE_PATH')
     print('compas_executable_override', compas_executable_override)
-    
+
     if (compas_executable_override is None):
         git_directory = os.environ.get('COMPAS_ROOT_DIR')
         compas_executable = os.path.join(git_directory, 'src/COMPAS')
@@ -50,7 +50,7 @@ class pythonProgramOptions:
     # if COMPAS_LOGS_OUTPUT_DIR_PATH is not set (== None) the current working directory
     # is used as the value for the --output-path option
     compas_logs_output_override = os.environ.get('COMPAS_LOGS_OUTPUT_DIR_PATH')
-    
+
     if (compas_logs_output_override is None):
         output = os.getcwd()
         output_container = None                 # names the directory to be created and in which log files are created.  Default in COMPAS is "COMPAS_Output"
@@ -210,7 +210,7 @@ class pythonProgramOptions:
 
     neutrino_mass_loss_BH_formation = "FIXED_MASS"              # "FIXED_FRACTION"
     neutrino_mass_loss_BH_formation_value = 0.1                 # Either fraction or mass (Msol) to lose
-    
+
     remnant_mass_prescription   = 'FRYER2012'                   #
     fryer_supernova_engine      = 'DELAYED'
     black_hole_kicks            = 'FALLBACK'
@@ -298,17 +298,6 @@ class pythonProgramOptions:
 
     debug_to_file  = False
     errors_to_file = False
-
-    # ensure that no file paths have unescaped spaces that could confuse Boost
-    paths = [output, grid_filename, logfile_definitions, logfile_common_envelopes, logfile_detailed_output,
-             logfile_double_compact_objects, logfile_rlof_parameters, logfile_pulsar_evolution,
-             logfile_supernovae, logfile_switch_log, logfile_system_parameters]
-    for i in range(len(paths)):
-        if paths[i] is not None:
-            paths[i] = re.sub(r"(?<!\\) ", r"\ ", paths[i])
-    output, grid_filename, logfile_definitions, logfile_common_envelopes, logfile_detailed_output,\
-        logfile_double_compact_objects, logfile_rlof_parameters, logfile_pulsar_evolution,\
-        logfile_supernovae, logfile_switch_log, logfile_system_parameters = paths
 
     def booleanChoices(self):
         booleanChoices = [
@@ -668,12 +657,12 @@ class pythonProgramOptions:
         and run directly as a terminal command, or passed to the stroopwafel interface
         where some of them may be overwritten. Options not to be included in the command 
         line should be set to pythons None (except booleans, which should be set to False)
-    
+
         Parameters
         -----------
         self : pythonProgramOptions
             Contains program options
-    
+
         Returns
         --------
         commands : str or list of strs
@@ -682,17 +671,17 @@ class pythonProgramOptions:
         booleanCommands = self.booleanCommands()
         nBoolean = len(booleanChoices)
         assert len(booleanCommands) == nBoolean
-    
+
         numericalChoices = self.numericalChoices()
         numericalCommands = self.numericalCommands()
         nNumerical = len(numericalChoices)
         assert len(numericalCommands) == nNumerical
-    
+
         stringChoices = self.stringChoices()
         stringCommands = self.stringCommands()
         nString = len(stringChoices)
         assert len(stringCommands) == nString
-    
+
         listChoices = self.listChoices()
         listCommands = self.listCommands()
         nList = len(listChoices)
@@ -702,23 +691,23 @@ class pythonProgramOptions:
         ### Collect all options into a dictionary mapping option name to option value
 
         command = {'compas_executable' : self.compas_executable}
-    
+
         for i in range(nBoolean):
             if booleanChoices[i] == True:
                 command.update({booleanCommands[i] : ''})
-    
+
         for i in range(nNumerical):
             if not numericalChoices[i] == None:
                 command.update({numericalCommands[i] : str(numericalChoices[i])})
-    
+
         for i in range(nString):
             if not stringChoices[i] == None:
-                command.update({stringCommands[i] : stringChoices[i]})
-    
+                command.update({stringCommands[i] : cleanStringParameter(stringChoices[i])})
+
         for i in range(nList):
             if listChoices[i]:
                 command.update({listCommands[i] : ' '.join(map(str,listChoices[i]))})
-    
+
         return command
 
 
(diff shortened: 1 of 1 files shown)


jeffriley (changes requested):
Nice job Tom.  Just a couple of comments:

First, a small detail: there are a number of filenames you don't check:

    logfile_common_envelopes
    logfile_detailed_output
    logfile_double_compact_objects
    logfile_rlof_parameters
    logfile_pulsar_evolution
    logfile_supernovae
    logfile_switch_log
    logfile_system_parameters

Next, and more importantly (for me), I'm a big fan of not arbitrarily constraining the user's choices.

Your changes prevent the user from specifying a directory name or file name that has embedded spaces - a logical thing to do since embedded spaces cause problems because the shell parser uses space as a delimiter.

However, directory and file names with embedded spaces are allowed by the OS, and they're not completely crazy :-)  In fact, Windows users are (or should be) quite used to directory names with spaces (e.g. 'c:/Program Files' and 'c:/Program Files (x86)').

I think we should not prevent users from specifying (and using) directories and/or files with embedded spaces in their names - we just need to pass the parameters through to the COMPAS executable correctly.

If I specify the grid file name as:

grid_filename = '"My Grid File"'

in pythonSubmit (not your modfied version - your modified version objects to the spaces), then executing

python3 pythonSubmit.py

works just fine - I have a grid file named 'My Grid File", and pythonSubmit passes the option through as

--grid ./"My Grid File"

and the COMPAS executable happily uses that grid file.

So, I would prefer that pythonSubmit allows embedded spaces in path names if the relevant part of the path name is enclosed in quotes - just a small change to your check code.

When you make that change you'll need to allow for either of these constructs (both work):

grid_filename = '"My Grid File"'  (hard to see, but that's outer single quotes and inner double quotes)

grid_filename = "'My Grid File'"  (outer double quotes and inner single quotes)

I should add that this is just my preference, but I am not the sole arbiter of what goes into COMPAS code - others may disagree, and if the consensus is that we should prevent embedded spaces always, that's fine.

TomWagg:
Great points @jeffriley 🙂

My solution to this has been to use a regex to find any spaces that are not immediately preceded by backslashes (which would escape them) and insert backslashes. This works for me if I change the `BSE_grid.txt` file in preProcessing to `BSE grid.txt`.

**Also a question:** I'm not really sure how to pass variables by reference in Python (and my google searches today didn't help). So I currently create a list of variables, change the ones in the list with a loop. But this doesn't change the original variables so I then reset those variables using the list afterwards. This seems a bit dodgy to me, is there are a better way of doing this? This would be easy in c++ I think, but not sure how to do it in Python?

jeffriley (changes requested):
Again, nice work Tom - an elegant solution.  Just one comment this time: 

An astute user might know that spaces in a pathname might confuse the shell parser, so might declare (e.g.) the grid file name as:

grid_filename = '"My Grid File"'

If they do that, your solution passes the option through to the commandline as:

--grid ./"My\ Grid\ File"

and that breaks.  So you should check whether any space is inside enclosing quotes.

However, maybe my answer to your question is a reasonable solution:

> I'm not really sure how to pass variables by reference in Python

Ok, first, a (very) brief explanation of python variables.  Variables in python are not what you might be used to if you are familiar with c, c++, Fortran, etc.  In python, a variable is really an object.  Consider the following two statements:

a = 1
a = 2

In a programming language like c, or c++, the statement

a = 1

puts the integer value 1 in a piece of memory allocated and reserved for the variable 'a' (I've glossed-over a lot of detail).  If you interrogate that piece of memory at runtime with a debugging tool you will see the integer value 1.  Then the statement

a = 2

puts the integer value 2 into the memory location reserved for the variable 'a', and if you interrogate that piece of memory at runtime you will now see the integer value 2. 

That's not what happens with Python.  In Python, the statement

a = 1

does pretty-much what it does in c++ - at puts the value 1 into the memory location allocated to 'a'.  But then, the statement

a = 2

allocates a new piece of memory, puts the value 2 in it, and points the variable (object) 'a' at the new memory location.  So, when you pass 'a' as a parameter to a function, Python allocates a new piece of memory for the parameter, so there is no way to reference the variable passed in - the function doesn't have a reference to that memory location.

Python uses 'pass by assignment' - you can't really pass by reference.  You can kind-of do something that works if you use a class and refer to the self.variable, and pass it to a function in a list/array, but that would mean a fairly big change to pythonSubmit (and I'm not sure it would work with an array of variables)

The bottom line is that I don't think you can't really do what you want to do - the workaround you have implemented is as good as any.

I wrote the following code that might be a bit cleaner though (up to you if you want to use it).  It requires users to use a new class to declare pathnames - actually, I think we could use it for all string options so that no string option would have a stray space in it.  If we document the class inside pythonSubmit and explain how to use it, then you probably don't need to worry about checking for enclosing quotes.

    import re

    class StringParameter():
        def __init__(self, value = None):

            if value is not None:
                value = re.sub(r"(?<!\\) ", r"\ ", value)

            self.value = value

    path1 = StringParameter('This is my path').value
    path2 = StringParameter('Another path').value
    grid_filename = StringParameter('My grid file').value
    logfile_definitions = StringParameter('logdefs.txt').value

    print(path1)
    print(path2)
    print(grid_filename)
    print(logfile_definitions)

Running that code gives:

    This\ is\ my\ path
    Another\ path
    My\ grid\ file
    logdefs.txt

*Afterthought: you don't really need to implement StringParameter as a class - it could just be a function that does the fixup and returns the new string (and then the user doesn't need the '.value' on the end of the declaration)

TomWagg:
I love asking question during these PRs with you @jeffriley, I always learn so much! Thanks for explaining all of that ☺️

I think it's much cleaner now that I do the same thing for every string parameter and change them just before updating the command dict.

My solution for the quotes has been to strip any quotes from the ends of the string but _escape_ any that are within in (just in case someone decides to name their file something like "Tom's Grid.txt").

This seems to work for the following test cases if I change the filename of `BSE_Grid.txt` in the `preProcessing` folder

- BSE_Grid.txt
- BSE Grid.txt
- 'BSE Grid.txt'
- "BSE Grid.txt"
- Tom's Grid.txt
- "Tom's Grid.txt"
- 'Tom's Grid.txt'

Let me know if there are any issues with doing it this way! :)

ilyamandel:
Thank you for the educational discussion, @TomWagg and @jeffriley !

jeffriley (approved):
Looks good - seems to cover all the bases - good work, thanks Tom!

PR #575, opened 2021-06-10
Compas compilation test

reinhold-willcox:
Added github action to automatically test whether the compas code compiles.

This will be combined with a requirement (on the github webpage) that the test be run at the time of pull_request.

At a later date, compas doc latex and postprocessing python scripts should also be added, to check that everything still compiles as expected.


reinhold-willcox:
Actually, the test actually ran already, so there's conveniently nothing to add on the webpage!

jeffriley (approved):
Looks good!

PR #578, opened 2021-06-11
Latex CI

reinhold-willcox:
Several updates related to CI and the COMPAS documentation.

- Added script to github workflow script (.yml file) to test that *.tex documentation files compile, to be run any time a new PR is submitted (if test fails, PR should not be approved).
- If the PR is approved, the script publishes the COMPAS_Documentation.pdf file to a separate github branch, Documentation. 
- This branch is entirely independent of all other COMPAS code, and so should never be PR'd or merged, but it will be a public location that always contains the most up to date version of the documentation. We should point to this from the Compas website (TODO), and any other locations we think people might look for the documentation.
- Removed the current documentation pdf, and added appropriate hyperlinks into the README.md file
- Moved latex files from COMPAS_LaTex to COMPAS_LaTeX - not really a big deal, but it was bugging me.  
- Cleaned up the other workflow script which tests if compas src/ compiles properly. Now the github action test names are clearer.
- Added .pdf and removed .github/ from .gitignore, since we don't want to do versioning on a compiled file, and we want to keep the .yml files around in all the different repos & branches.

This PR might be a bit tricky to test, but I can help anyone who's keen to try it before accepting and merging.


jeffriley (approved):
Looks good to me

PR #579, opened 2021-06-11
Updates to ci

LiekeVanSon:
I added some updates to FastCosmicIntegration.py and created a new notebook that explains it step by step (basically the same as the notebook that Tom made earlier to break down fast cosmic integrator :)

Main changes:

- I added an ArgumentParser to run FastCosmicIntegration from the terminal
- I updated find_metallicity_distribution to a skew-log-normal distribution which is the general form of the distribution, a log-normal distribution (such as the one in Neijssel + 2019) can be retrieved by setting alpha =0
- I added an append_rates function to append you CI rates to your COMPAS_output.hdf5 file when you're done

I'm not sure everyone would be happy with all the functionality changes that I made, so feedback is very welcome!


reinhold-willcox:
@cneijssel @FloorBroekgaarden @TomWagg 

Do any of you feel comfortable reviewing this?

TomWagg:
@reinhold-willcox sure I'll take a look at this tomorrow/Tuesday :)

postProcessing/Folders/CosmicIntegration/PythonScripts/FastCosmicIntegration.py
```diff
@@ -44,66 +52,109 @@ def calculate_redshift_related_params(max_redshift=10.0, max_redshift_detection=
...
+        metallicities      --> [list of floats] Metallicities at which dPdlogZ is evaluated
+        p_draw_metallicity --> float            Probability of drawing a certain metallicity in COMPAS (float because assuming uniform)
     """
-    # use Neijssel+19 Eq. 9 to find means
-    mean_metallicities = Z0 * 10**(alpha * redshifts)
-    mu_metallicities = np.log(mean_metallicities) - sigma**2/2
-
-    # create a range of metallicities
-    log_metallicities = np.arange(min_logZ_COMPAS, max_logZ_COMPAS + step_logZ, step_logZ)
+    import scipy
```
TomWagg: I would avoid putting imports in the middle of a file

TomWagg: In fact, it's actually already at the top so I'd just delete this

postProcessing/Folders/CosmicIntegration/PythonScripts/FastCosmicIntegration.py
```diff
@@ -44,66 +52,109 @@ def calculate_redshift_related_params(max_redshift=10.0, max_redshift_detection=
...
+
+    ##################################
+    # the PDF of a standard normal distrtibution
+    def normal_PDF(x):
+        return 1./(np.sqrt(2* np.pi)) * np.exp(-(1./2) * (x)**2 )
```
TomWagg: I don't think you need this. You can just use `scipy.stats.norm.pdf` instead :)

postProcessing/Folders/CosmicIntegration/PythonScripts/FastCosmicIntegration.py
```diff
@@ -44,66 +52,109 @@ def calculate_redshift_related_params(max_redshift=10.0, max_redshift_detection=
...
+
+    ##################################
+    # the CDF of a standard normal distrtibution
+    def normal_CDF(x):
+        return 1./2. * (1 + scipy.special.erf(x/np.sqrt(2)) )
```
TomWagg: Similar to the above, why not instead use `scipy.stats.norm.cdf`?

postProcessing/Folders/CosmicIntegration/PythonScripts/FastCosmicIntegration.py
```diff
@@ -44,66 +52,109 @@ def calculate_redshift_related_params(max_redshift=10.0, max_redshift_detection=
...
+        max_redshift       --> [float]          max redshift for calculation
+        redshift_step      --> [float]          step used in redshift calculation
+        min_logZ_COMPAS    --> [float]          Minimum logZ value that COMPAS samples
+        max_logZ_COMPAS    --> [float]          Maximum logZ value that COMPAS samples
+        
+        mu0    = 0.025    --> [float]           location (mean in normal) at redshift 0
+        muz    = -0.05    --> [float]           redshift scaling/evolution of the location
+        sigma_0 = 1.25     --> [float]          Scale (variance in normal) at redshift 0
+        sigma_z = 0.05     --> [float]          redshift scaling of the scale (variance in normal)
+        alpha   = -1.77    --> [float]          shape (skewness, alpha = 0 retrieves normal dist)
```
TomWagg: Do you have a citation for where these defaults come from? Maybe put it in the docstring?

postProcessing/Folders/CosmicIntegration/PythonScripts/FastCosmicIntegration.py
```diff
@@ -406,23 +474,162 @@ def find_detection_rate(path, filename="COMPAS_Output.h5", dco_type="BBH", weigh
...
+            path                   --> [string] Path to the COMPAS file that contains the output
+            filename               --> [string] Name of the COMPAS file
+            detection_rate         --> [2D float array] Detection rate for each binary at each redshift in 1/yr
+            formation_rate         --> [2D float array] Formation rate for each binary at each redshift in 1/yr/Gpc^3
+            merger_rate            --> [2D float array] Merger rate for each binary at each redshift in 1/yr/Gpc^3
+            redshifts              --> [list of floats] List of redshifts
+            COMPAS                 --> [Object]         Relevant COMPAS data in COMPASData Class
+            n_redshifts_detection  --> [int]            Number of redshifts in list that should be used to calculate detection rates
+
+            maxz                   --> [float] Maximum redshhift up to where we would like to store the data
```
TomWagg, line 485: typo with "redshhift"

postProcessing/Folders/CosmicIntegration/PythonScripts/FastCosmicIntegration.py
```diff
@@ -406,23 +474,162 @@ def find_detection_rate(path, filename="COMPAS_Output.h5", dco_type="BBH", weigh
...
+            sensitivity            --> [string] Which detector sensitivity you used to calculate rates 
+            dco_type               --> [string] Which DCO type you used to calculate rates 
+            mu0                    --> [float] Parameter used for calculating metallicity density dist
+            muz                    --> [float] Parameter used for calculating metallicity density dist
+            sigma0                 --> [float] Parameter used for calculating metallicity density dist
+            sigmaz                 --> [float] Parameter used for calculating metallicity density dist
+            alpha                  --> [float] Parameter used for calculating metallicity density dist
```
TomWagg: Copy L367-371 down for more specific comments

postProcessing/Folders/CosmicIntegration/PythonScripts/FastCosmicIntegration.py
```diff
@@ -406,23 +474,162 @@ def find_detection_rate(path, filename="COMPAS_Output.h5", dco_type="BBH", weigh
...
+    #Open hdf5 file that we will write on
+    print('filename', filename)
+    with h5.File(path +'/'+ filename, 'r+') as h_new:
+        # The rate info is shaped as BSE_Double_Compact_Objects[COMPAS.DCOmask] , len(redshifts)
+        DCO             = h_new['BSE_Double_Compact_Objects']#
+        print('shape DCO[SEED]', np.shape(DCO['SEED'][()]) )
+
+        #################################################
+        # Create a new group where we will store data
+        new_rate_group  = 'Rates_mu0'+str(mu0)+'_muz'+str(muz)+'_alpha'+str(alpha)+'_sigma0'+str(sigma0)+'_sigmaz'+str(sigmaz)
```
TomWagg: Nothing wrong with this, but I think it's easier to read/write these things using `.format()`

e.g.
`new_rate_group = 'Rates_mu0{}_muz{}_alpha{}_sigma0{}_sigmaz{}'.format(mu0, muz, alpha, sigma0, sigmaz)`

So that you can easily see what the final string looks like and you only need one set of quotes and no need to wrap in `str()`.

postProcessing/Folders/CosmicIntegration/PythonScripts/FastCosmicIntegration.py
```diff
@@ -406,23 +474,162 @@ def find_detection_rate(path, filename="COMPAS_Output.h5", dco_type="BBH", weigh
...
+        else:
+            print(new_rate_group, 'exists, we will overrwrite the data')
+
+
+        ######################s###########################
+        # If you just want to get rid of this group, remove and exit
+        if remove_group:
+            print('You want to remove this group, %s, from the hdf5 file, removing now..'%(new_rate_group))
+            del h_new[new_rate_group]
+            return
```
TomWagg: Am I correct in understanding this is for stripping rates out of hdf5 files? If so, I personally think this should be a separate function rather than a parameter for clarity

postProcessing/Folders/CosmicIntegration/PythonScripts/FastCosmicIntegration.py
```diff
@@ -406,23 +474,162 @@ def find_detection_rate(path, filename="COMPAS_Output.h5", dco_type="BBH", weigh
...
+        
+
+        #################################################
+        # Bin rates by redshifts
+        #################################################
+        if append_binned_by_z:
+            print('IN append_binned_by_z')
```
TomWagg: Is this a debugging print statement or do you want to leave it in?

postProcessing/Folders/CosmicIntegration/PythonScripts/FastCosmicIntegration.py
```diff
@@ -406,23 +474,162 @@ def find_detection_rate(path, filename="COMPAS_Output.h5", dco_type="BBH", weigh
...
+            print('Adding rate info of shape', np.shape(data))
+            # Check if dataset exists, if so, just delete it
+            if rate_list_names[i] in h_new[new_rate_group].keys():
+                del h_new[new_rate_group][rate_list_names[i]]
+            # write rates as a new data set
+            dataNew     = h_new[new_rate_group].create_dataset(rate_list_names[i], data=data)
+
+    #Always close your files again ;)
+    h_new.close()
+    print('Done with append_rates :) your new files are here: ', path + '/' + filename )
```
TomWagg: I think the '/' here is extra since `path` already includes it.

image

postProcessing/Folders/CosmicIntegration/PythonScripts/FastCosmicIntegration.py
```diff
@@ -456,7 +666,100 @@ def plot_rates(formation_rate, merger_rate, detection_rate, redshifts, chirp_mas
...
+
+
+
+
+##################################################################
+### 
+### Run it!
+###
+##################################################################
+if __name__ == "__main__":
```
TomWagg, line 719: I **LOVE** this command line stuff! ❤️

postProcessing/Folders/CosmicIntegration/PythonScripts/FastCosmicIntegration.py
```diff
@@ -456,7 +666,100 @@ def plot_rates(formation_rate, merger_rate, detection_rate, redshifts, chirp_mas
...
+    parser.add_argument("--m1max", dest= 'm1_max',  help="Maximum primary mass sampled by COMPAS",type=float, default=150.) 
+    parser.add_argument("--m2min", dest= 'm2_min',  help="Minimum secondary mass sampled by COMPAS",type=float, default=0.1) 
+    parser.add_argument("--fbin", dest= 'fbin',  help="Binary fraction used by COMPAS",type=float, default=0.7) 
+
+    # Parameters determining dP/dZ and SFR(z), default options from Neijssel 2019
+    parser.add_argument("--mu0", dest= 'mu0',  help="mean metallicity at redshhift 0",type=float, default=0.035)
+    parser.add_argument("--muz", dest= 'muz',  help="redshift evolution of mean metallicity, dPdlogZ",type=float, default=-0.23)
+    parser.add_argument("--sigma0", dest= 'sigma0',  help="variance in metallicity density distribution, dPdlogZ",type=float, default=0.39)
+    parser.add_argument("--sigmaz", dest= 'sigmaz',  help="redshift evolution of variance, dPdlogZ",type=float, default=0.0)
+    parser.add_argument("--alpha", dest= 'alpha',  help="skewness of mtallicity density distribution, dPdlogZ",type=float, default=0.0)
```
TomWagg, line 749: These are different from the defaults in the function right? I think you should make the default the same in both cases.

TomWagg (changes requested):
This is awesome @LiekeVanSon!

I tested it out and the notebook works perfectly as does the command line version, very cool!

I've left some line by line comments of changes that I think you need. One other general thought I had is that in the notebook it might be cool to first show the One Function to Rule Them All (`find_detection_rates`) before showing how all of the individual parts work as I think it is less intimidating that way. But that's just personal preference on my part.

@reinhold-willcox @FloorBroekgaarden I should also say that I tested this with the same hdf5 file that Lieke used so it may be smart to also test this with a different run based on the latest version of COMPAS if either of you have one. I tried using one of the test runs in the pinned issues but I think the file headers have changed since then so it just crashed.

postProcessing/Folders/CosmicIntegration/PythonScripts/FastCosmicIntegration.py
reinhold-willcox: Tom, is it always the case that the path will have a '/' at the end? If not, this could be a safe choice.

There are also python libraries that will handle correct joining of path names, which may be preferable.

reinhold-willcox:
@TomWagg thanks for looking closely at this. Following your suggestion, I reran with an updated COMPAS dataset. I didn't check that the results were scientifically valid, just that it compiled with the newer version. That's good to know that the pinned issue test data is outdated. @jeffriley Could you double check this? Is it worthwhile to setup a github action to automatically upload a new one with pushes to dev. 

@LiekeVanSon Overall, looks great and works really well. Some minor changes/requests follow, but should be straightforward to implement. 

- The default filename for ClassCOMPAS.COMPASData() is not the default output by COMPAS. This should be updated. Also, ideally, a user could specify the full path (with data file included) and COMPASData should be able to parse out the path from the filename (though this can be a TODO for a future PR). 
- The link to 2_MSSFR-or... in the root ipynb file 0_CosmicInt... is broken and should be updated with the new file name.
- In 2_MSSFR... we run some plots. We should include the jupyter notebook magic command `%matplotlib inline` at the top of the file, with all the import statements, to keep plotting in the same notebook. This is not _quite_ a bug, but errored for me because my xWindow setup is weird and so jupyter notebooks cannot connect to localhost. Perhaps a small corner case, but it has an easy solution. We should probably include this command in all notebooks which produce plots.
- In 5_Mergers... the final code-running cell has a call to `CI.COMPAS.setCOMPASDCOmask(types=t, optimistic=True)` which gives me an error "setCOMPASDCOmask() got an unexpected keyword argument 'optimistic'". At some point, we changed this to 'pessimistic', so this should be updated to reflect the change.

Otherwise looks great, notebooks all run on the latest COMPAS output.

reinhold-willcox (changes requested):
see above comment

reinhold-willcox:
Oh also, please merge in the latest version of dev

jeffriley:
> That's good to know that the pinned issue test data is outdated.  @jeffriley Could you double check this? 

I'm sure it has the older header strings - that'll happen whenever we change header strings, but I think we're probably done with header string changes for a while, so maybe a good time to create a new sample file.

> Is it worthwhile to setup a github action to automatically upload a new one with pushes to dev.

Automatically upload from where - where would the file come from?  By running COMPAS with default pythonSubmit on a github hosted runner?  Can you do that?  I don't know enough about github... 

But yes, if we change header strings we should probably provide a new sample output file that has the new header strings.  Not sure that we need to upload a new file for every change to the source.

reinhold-willcox:
@jeffriley I've looked around for this, but I can't find a clear answer. The usage limits are explained in the link below, but there's nothing about CPU hours. Just raw clock time... I'll play around with it a bit and let you know.

https://docs.github.com/en/actions/reference/usage-limits-billing-and-administration

postProcessing/Folders/CosmicIntegration/PythonScripts/FastCosmicIntegration.py
LiekeVanSon: They come from my paper in prep (which I'm hoping to get out this summer!) For now, I will put Neijssel 2019 as the default :)

TomWagg: Ooh fancy, that's exciting! 😎

postProcessing/Folders/CosmicIntegration/PythonScripts/FastCosmicIntegration.py
LiekeVanSon: agreed!

postProcessing/Folders/CosmicIntegration/PythonScripts/FastCosmicIntegration.py
LiekeVanSon: in my experience an extra '/' never hurts, while one too little could hurt! So I would argue to keep it

postProcessing/Folders/CosmicIntegration/PythonScripts/FastCosmicIntegration.py
LiekeVanSon: I've set all defaults to reflect the phenomenological model from Neijssel + 2019

postProcessing/Folders/CosmicIntegration/PythonScripts/FastCosmicIntegration.py
TomWagg: Definitely a fair point! I think you could maybe get around this by doing `(path + '/' + filename).replace('//', '/')` (since I don't *think* paths should ever had a double slash).

But either way agreed that one too many is better than one too few!

postProcessing/Folders/CosmicIntegration/PythonScripts/FastCosmicIntegration.py
LiekeVanSon: I thought about this and I think you're right! I've added an extra function "delete_rates" in the main function, and I've added 2 arguments to the args parser to make sure people can choose if they want to append their rates or not, or just delete them

LiekeVanSon:
Hi all, thanks for the fast and thorough checks!! :D
@TomWagg I think I've addressed all your comments, thanks for being so careful! 


@reinhold-willcox same goes for you, see below for my responses to the points you addressed! 

>- The default filename for ClassCOMPAS.COMPASData() is not the default output by COMPAS. This should be updated. Also, ideally, a user could specify the full path (with data file included) and COMPASData should be able to parse out the path from the filename (though this can be a TODO for a future PR). 

Just double checking, but you meant to change COMPAS_output.h5 to COMPAS_Output.h5, correct?

>- The link to 2_MSSFR-or... in the root ipynb file 0_CosmicInt... is broken and should be updated with the new file name.

check

>- In 2_MSSFR... we run some plots. We should include the jupyter notebook magic command %matplotlib inline at the top of the file, with all the import statements, to keep plotting in the same notebook. This is not quite a bug, but errored for me because my xWindow setup is weird and so jupyter notebooks cannot connect to localhost. Perhaps a small corner case, but it has an easy solution. We should probably include this command in all notebooks which produce plots.

I never have this issue, but that's good to know! I've added the magic command to all notebooks in CI that do plotting :)

>- In 5_Mergers... the final code-running cell has a call to CI.COMPAS.setCOMPASDCOmask(types=t, optimistic=True) which gives me an error "setCOMPASDCOmask() got an unexpected keyword argument 'optimistic'". At some point, we changed this to 'pessimistic', so this should be updated to reflect the change.

check!

postProcessing/Folders/CosmicIntegration/PythonScripts/FastCosmicIntegration.py
LiekeVanSon: fair point! I've cleaned it up with this suggestion! :)

TomWagg (commented):
All looks good to me! Thanks @LiekeVanSon 😄

@reinhold-willcox I'll leave you to approve once you've seen it :)

reinhold-willcox:
Made a small change to section 4, to change COMPAS_output for COMPAS_Output. Otherwise, looks great, thanks Lieke!

reinhold-willcox (approved):

reinhold-willcox:
@TomWagg Can you change your review to be approved, then merge the PR? thanks

TomWagg (approved):

Issue #562, opened 2021-05-18
Default options in C++


ilyamandel:
Migrate defaults from defaults/pythonSubmit.py to C++ code in order to have a single consistent model to refer to as "default".

ilyamandel:
Closing -- see PR #581 .

PR #581, opened 2021-06-14
Moving defaults to Options.cpp

ilyamandel:
- Port defaults from preProcessing/pythonSubmit.py to options.cpp
- Minor fixes (e.g., documentation)

### DIFF
--- a/docs/COMPAS_LaTeX/sections/Appendices/ProgramOptions.tex
+++ b/docs/COMPAS_LaTeX/sections/Appendices/ProgramOptions.tex
@@ -99,7 +99,7 @@ Following is the list of program options that can be specified at the command li
 
 \programOption{initial-mass-min}{}{Minimum mass to generate using given IMF~(\Msun).}{8.0}
 
-\programOption{initial-mass-power}{}{Single power law power to generate primary mass using given IMF.}{\minus{2.3}}
+\programOption{initial-mass-power}{}{Single power law power to generate primary mass using POWERLAW IMF.}{\minus{2.3}}
 
 \programOption{kick-direction}{}{Natal kick direction distribution. \\ Options: \lcb\ ISOTROPIC, INPLANE, PERPENDICULAR, POWERLAW, WEDGE, POLES\ \rcb}{ISOTROPIC}
 
--- a/examples/methods_paper_plots/detailed_evolution/pythonSubmitDemo.py
+++ b/examples/methods_paper_plots/detailed_evolution/pythonSubmitDemo.py
@@ -144,14 +144,14 @@ class pythonProgramOptions:
     envelope_state_prescription = 'LEGACY'
 
     mass_loss_prescription = 'VINK'
-    luminous_blue_variable_prescription = 'BELCZYNSKI'
+    luminous_blue_variable_prescription = 'HURLEY_ADD'
     luminous_blue_variable_multiplier = 1.5
     overall_wind_mass_loss_multiplier = 1.0
     wolf_rayet_multiplier = 1.0
     cool_wind_mass_loss_multiplier = 1.0
     check_photon_tiring_limit = False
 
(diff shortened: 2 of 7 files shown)


ilyamandel:
[Setting to draft because, contrary to expectations, I do not get an identical output when running ./COMPAS from the command line vs. from pythonSubmit]

ilyamandel:
OK, I figured out what the difference is, but not why there is a difference.  Even though I explicitly set m_Metallicity = 0.0142, I get binaries with metallicities of 0.02 (unless I again set --metallicity 0.0142 at the command line).  Now, that's weird, because my runDetails are identical when ./COMPAS --random-seed 0 from the command line or running via pythonSubmit :

< metallicity = 0.014200, DEFAULT_USED, DOUBLE
< metallicity-distribution = 'ZSOLAR', DEFAULT_USED, STRING
< metallicity-max = 0.030000, DEFAULT_USED, DOUBLE
< metallicity-min = 0.000100, DEFAULT_USED, DOUBLE
---
> metallicity = 0.014200, USER_SUPPLIED, DOUBLE
> metallicity-distribution = 'ZSOLAR', USER_SUPPLIED, STRING
> metallicity-max = 0.030000, USER_SUPPLIED, DOUBLE
> metallicity-min = 0.000100, USER_SUPPLIED, DOUBLE

Which actually means that something is wrong in my book.  @jeffriley , can you help out? :)

jeffriley:
> Which actually means that something is wrong in my book. @jeffriley , can you help out? :)

Not with your notebook (...book :-)), but I think I can explain why you get Z = 0.02 

Look in BaseBinaryStar.cpp where we initialise metallicity for the binary (line 110 in v02.19.04).  If the user specified metallicity we use it, otherwise we sample it - we don't actually use the default value (I noticed this the other day - we do the same with secondary mass - we could think a little more about what we do here (there are several options - use what's specified, use the default, use a distribution - what's the best way to decide which?  Should we only sample if the user specified a distribution? When I say "we" I'm using the royal "we" - these choices were mine when I coded this part, so all my fault :-))).  

So, back to metallicity.  If the user doesn't specify metallicity we'll sample it, using utils::SampleMetallicity().  The default metallicity distribution is ZSOLAR, and that just returns ZSOL.  So to default to 0.0142 you should change that.  Which brings up...

I notice you've just used hard-coded constants in Options.cpp (you replace a couple of symbolic names defined in constants.h with numbers).  I ran out of time and energy when I ported the legacy code, but I had always had in the back of my mind that we would pull the hard-coded constants out of the code and put them in constants.h with symbolic names.  For the program option defaults it really doesn't matter where they are I suppose (constants used throughout the code - i.e. not program option defaults -  should be in constants.h), as long as they are all in one place, but I think replacing numbers with symbolic names (that are associated with numbers as in constants.h) might make it easier for people to find the values (I know whenever I have to find the default value for a program option I go to Options.h first to find the getter so I can see what the m_* variable name is, then I go to Options.cpp to find where the variable is initialised - I think if we had a list of "option constants" somewhere (whether that be in constants.h or (say) at the top of Options.h with symbolic names that people could recognise, it might make life easier).

jeffriley:
Oh, and one more thing.  After this PR is merged into the code, we need to go through the methods paper to make sure any defaults we describe agree with the code (I think in my section I used C++ code defaults, whereas in other places we might have use pythonSubmit defaults)

ilyamandel:
@jeffriley -- thank you.  As we discussed offline, I was trying to ensure that calling 
./COMPAS --random-seed 0
from the command line would produce identical results to calling default pythonSubmit.  That's now been achieved, and a new ZSOL_ASPLUND constant has been defined. :)
I'll also propose a new feature to improve documentation (for later).

jeffriley (approved):
All good!

ilyamandel:
This will close #562 .

PR #584, opened 2021-06-15
Updated methods paper figures

TomWagg:
Nothing special here, the figures stayed pretty much identical but I needed to update the pythonSubmit.py file to match v.02.20.00 🙂


reinhold-willcox (approved):
Thanks for your patience on this, changes look good!

PR #587, opened 2021-06-21
Defect repair: Fix issue 585

### LINKED ISSUE #585: Inconsistent columns/headers in SSE_Supernova
**Describe the bug**
When running COMPAS in SSE mode, I find a mismatch between the headers and the values listed in SSE_Supernova (command line to reproduce below).

The output header has 15 columns, but there are only 14 values.  This leads to some variables being matched to the wrong header. For example, I have Stellar_Type_Prev = 7.11 (stellar type should be an integer, I believe this is a core mass).

**Label the issue**

**To Reproduce**
Using the latest COMPAS version

Run:
./COMPAS --initial-mass 20.0 --mode SSE -z 0.01 -n 1 --logfile-type CSV

and then inspect

COMPAS_Output/SSE_Supernovae.csv 

**Expected behavior**
The number of columns and headers should match, values should match their header.

 - OSX 10.14.6
- COMPAS v02.20.00

**Additional context**

jeffriley:
Fix for issue #585: add formatted value and delimiter to logrecord string in Log.h (defect introduced in v02.18.00; only affected SSE_Supernovae logfile)

Close issue #585 when merged

### DIFF
--- a/src/Log.h
+++ b/src/Log.h
@@ -599,6 +599,7 @@ private:
 
                 value    = p_SpecifiedPropertyValue;
                 valueStr = boost::apply_visitor(FormatVariantValue(), value, fmtStr);       // format value
+                logRecord += valueStr + delimiter;                                          // add value string to log record - with delimiter
             }
             else {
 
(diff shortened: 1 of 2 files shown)


SimonStevenson (approved):
Thanks so much Jeff, as always! I just checked this out and reran and can confirm the issue is resolved.

PR #598, opened 2021-07-26
Defect repair: Add HDF5 support to logging code for SSE/BSE switch log files.

jeffriley:
Add HDF5 support to logging code for SSE/BSE switch log files.  Support for HDF5 files was inadvertently not added when HDF5 file support as added in v02.18.00 for all standard log files.  Switch log files are 'special' (they have extra columns, not part of the 'standard' log file functionality), and that was missed.

Workaround is to use CSV files instead of HDF5 files if the switch logs are required.

We probably should prioritise some regression testing...


ilyamandel (approved):
Look good -- thank you, @jeffriley !

Issue #601, opened 2021-07-28
pythonSubmit.py doesn't allow users to specify "False" for boolean program options


jeffriley:
**Describe the bug**
The current pythonSubmit.py doesn't allow users to specify "False" for boolean program options if the options are set "True" by default in the C++ code.

Background:

Boolean program options were not handled correctly in the legacy code - they were processed thus:

    if (vm.count("quiet")) {
        options.quiet = true;
    }

So if the program option name was present on the commandline, the option value was set "True" regardless of the value specified.  That didn't allow users to set boolean program options to "False".

When I refactored from legacy to v2, I wrote the program option processing code to allow users to specify "False" for boolean program options - so users could turn off any boolean program options that were on by default.

Setting program options to "False" (i.e. '--quiet False') works if COMPAS is run from the commandline, but it can't be done with the current version of pythonSubmit.py, because of this code:

    for i in range(nBoolean):
        if booleanChoices[i] == True:
            command.update({booleanCommands[i] : ''})

which only puts the program option on the commandline to be executed if the value of the program option is "True".

The fix is to change pythonSubmit.py to put all boolean parameters on the commandline, with "True" or "False" as their value.

**Label the issue**
`urgency_moderate` - This is a moderately urgent issue
`severity_moderate` - This is a moderately severe bug

**To Reproduce**
Change the value of any boolean program option whose default value in the C++ code is "TRUE" (e.g. RLOFPrinting) to "False" in pythonSubmit.py - it has no effect (the RLOF printing file is produced)

**Expected behavior**
Honour the user's specification of "False" for boolean program options

**Screenshots**

**Versioning (please complete the following information):**
 - OS: Ubuntu 20.04 LTS
 - COMPAS v02.20.02

**Additional context**

jeffriley:
Fixed in PR #602 v02.21.00

After this change, boolean options set to either "True" or "False" in pythonSubmit.py will be listed on the commandline constructed by pythonSubmit.py: options whose value in pythonSubmit.py is "True" will be listed with no argument (e.g. '--quiet'), whereas options whose value is "False" in pythonSubmit.py will be listed with the argument "False" (e.g. '--quiet False'). (We could list the argument "True" explicitly if that is preferred by most people). Boolean options whose value in pythonSubmit.py is "None" will not be listed on the commandline (and therefore will take the default C++ value).

Issue #600, opened 2021-07-28
pythonSubmit.py does not treat fully-qualified grid filenames and fully-qualified logfile-definitions filenames correctly


jeffriley:
**Describe the bug**
pythonSubmit.py does not treat fully-qualified grid filenames and fully-qualified logfile-definitions filenames correctly - it adds CWD even though the filename is already fully-qualified.

**Label the issue**
`urgency_low`      - This issue is not urgent
`severity_minor`    - This is a minor bug with minimal impact

**To Reproduce**
Specify 'grid = /dirname/gridfilename.txt' in pythonSubmit.py

pythonSubmit.py will prepend the grid filename with the CWD, which then causes COMPAS to fail (file not found)

**Expected behavior**
Don't add CWD to grid filename or logfile-definitions file if already fully-qualified

**Screenshots**

**Versioning (please complete the following information):**
 - OS: Ubuntu 20.04 LTS
 - COMPAS v02.20.02

**Additional context**

jeffriley:
Fixed in PR #602 v02.21.00

After this change, any grid file or logfile-definitions file that is specified as a fully-qualified filename (i.e. the string given as the value of the 'grid' variable or 'logfile_definitions' variable in pythonSubmit.py includes both a path and a base filename) will be passed as given to the commandline - no path will be added by pythonSubmit.py (either CWD or compas_input_path_override). The assumption is that if the user specifies a path then that's the path that should be used.

PR #603, opened 2021-07-28
Copy input files (grid, logfile-definitions) to output container + defect repairs

jeffriley:
Enhancement and Defect Repairs:

- Added code to copy any grid file and/or logfile-definitions file specified to output container.

The motivation is to ensure that all the input data required to reproduce the run is available.  The initial conditions/evolutionary parameter values are in the Run_Details file but, prior to this change, any grid file or logfile-definitions file was not included in the output container, so could be inadvertently lost.

I see the COMPAS output container (directory created to contain the output files) as the main repository for output files, and this is definitely true if the user specifies a logfile type of anything other than HDF5 (e.g. CSV, TSV, etc.).  If the user does specify a logfile type of HDF5 (probably the most common use case now that HDF5 is available), then a case could be made to consider the HDF5 output file the main repository, and maybe the grid file and logfile-definition file should be included in t he HDF5 file.  I don't think users need access to either of those files during post-run analysis (but maybe I'm missing a use case), so the only benefit in adding them to the HDF5 file is to make that file (the HDF5 file) the repository for all data required to reproduce the run (and I think a case could be made for that - if HDF5 is specified, then the only file the user really needs is the HDF5 file, but if anything other than HDF5 file is specified (CSV, TSV, etc.), then the containing directory is required because multiple output files are produced and need to be maintained as a collection).

Bottom line - we could extend this code to add the grid and logfile-definitions file to the HDF5 output file if enough people think that would be useful.

- Copying a large grid file could take time, and take up much space, so added new program option '--store-input-files' which is TRUE by default.  If FALSE, neither the grid file (if specified) nor the logfile-definitions file (if specified) will be copied to the output container (if TRUE, both will be copied (if specified)).

The user should have control over whether the files are copied to the output container.

- Fixed issue #600: changed pythonSubmit.py to treat fully-qualified grid filenames and fully-qualified logfile-definitions filenames correctly (i.e. don't add CWD if the filename is already fully-qualified).

After this change, any grid file or logfile-definitions file that is specified as a fully-qualified filename (i.e. the string given as the value of the 'grid' variable or 'logfile_definitions' variable in pythonSubmit.py  includes both a path and a base filename) will be passed as given to the commandline - no path will be added by pythonSubmit.py (either CWD or compas_input_path_override).  The assumption is that if the user specifies a path then that's the path that should be used.

- Fixed issue #601: changed pythonSubmit.py to put all boolean parameters on the commandline, with True or False value.

After this change, boolean options set to either "True" or "False" in pythonSubmit.py will be listed on  the commandline constructed by pythonSubmit.py: options whose value in pythonSubmit.py is "True" will be listed with no argument (e.g. '--quiet'), whereas options whose value is "False" in pythonSubmit.py will be listed with the argument "False" (e.g. '--quiet False').  (We could list the argument "True" explicitly if that is preferred by most people).  Boolean options whose value in pythonSubmit.py is "None" will not be listed on the commandline (and therefore will take the default C++ value).

Close issue #600 when merged
Close issue #601  when merged

### DIFF
--- a/docs/COMPAS_LaTeX/sections/Appendices/ProgramOptions.tex
+++ b/docs/COMPAS_LaTeX/sections/Appendices/ProgramOptions.tex
@@ -307,6 +307,8 @@ In both cases the number of objects evolved will be the number specified by the 
 
 \programOption{stellar-zeta-prescription}{}{Prescription for stellar zeta. \\ Options: \lcb\ STARTRACK, SOBERMAN, HURLEY, ARBITRARY \rcb}{SOBERMAN}
 
+\programOption{store-input-files}{}{Enables copying of any specified grid file and/or logfile-definitios file to the COMPAS output container}{TRUE}
+
 \programOption{switch-log}{}{Enables printing of the Switch Log logfile}{FALSE}
 
 \programOption{timestep-multiplier}{}{Multiplicative factor for timestep duration}{1.0}
--- a/docs/COMPAS_LaTeX/sections/RevisionHistory.tex
+++ b/docs/COMPAS_LaTeX/sections/RevisionHistory.tex
@@ -73,6 +73,19 @@
 
 \revisionHistoryRow{18 December 2020}{2.12}{Removed references to Mass\_Transfer\_Case\_Initial parameter}{Reinhold Willcox}
 
+\end{tabularx}  % why is this too wide?
+
+
+\begin{tabularx}{\linewidth}{
+    |>{\hsize=0.8\hsize}X
+    |>{\hsize=0.3\hsize}X
+    |>{\hsize=2.1\hsize}X
+    |>{\hsize=0.8\hsize}X
+    |
+}
+
+\hline                                                                      % 
+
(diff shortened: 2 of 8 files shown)


reinhold-willcox:
@jeffriley I couldn't get it to work. Using a small grid (250 lines) I ran:

```
COMPAS --grid grid_0.csv 
COMPAS --grid grid_0.csv --store-input-files 
COMPAS --grid grid_0.csv --store-input-files FALSE
```

I confirmed that the Run_Details for all three correctly show that the parameter was set (or not) and what its value is. But the size of the respective COMPAS_Output.h5 files is the same for all 3, and a quick check of the contents shows that none contains the grid file. Am I doing something wrong?

jeffriley:
> Added code to copy any grid file and/or logfile-definitions file specified to output container.

The files are copied to the output container (by default, the COMPAS_Output directory).  Note the text above discussed whether we should put the files inside the HDF5 file, but currently we don't.

reinhold-willcox:
Right, ok, I should have read more closely. In that case, yes, it works!

jeffriley:
Great!  So, merge?

reinhold-willcox (approved):

PR #606, opened 2021-08-18
RLOF printing bug fix

### LINKED ISSUE #604: BSE_RLOF output skips a stellar type
**Describe the bug**
In some (rare) cases, the `BSE_RLOF` output shows a donor star with previous stellar type 1 (MS) become stellar type 7 (HeMS). This generally should not happen, since MS stars do not yet have He cores. Upon closer inspection of the detailed output, the donor is in fact a HG star for a few timesteps before it becomes a HeMS star, but this is not correctly reflected in the `BSE_RLOF` output. 

The attached screenshot shows a zoom in of the detailed evolution around the transition, and in particular that the HG phase is indeed reached. 

**Label the issue**

**To Reproduce**
```
COMPAS -n1 --random-seed 142952 --initial-mass-1 20.572500791714123 --semi-major-axis 0.15567910245419636 --initial-mass-2 11.78490823947635 --metallicity 0.0142 --eccentricity 0
```
Edit: added `-n1` 

**Expected behavior**
The RLOF output should only ever show a MS star become a MS star. If it evolves into a HG in the middle of an otherwise typical Case A MT phase, the next line in the output should show the star as HG before stripping into a HeMS star. This seems like an issue in how the RLOF parameters are stored.

**Screenshots**
rlof_skips_HG

 - OS: [e.g. Ubuntu 18.04]
 - COMPAS v02.21.00

reinhold-willcox:
Address issue #604 by adding a stashRLOFproperties call immediately before EvaluateBinary. 

This means that the "previous" values in the BSE_RLOF output will refer to the moment immediately after the SSE phase, but before the BSE phase, in a given timestep. 

For cases where there are 2 stellar type changes in one timestep (e.g a MS star engaged in RLOF moves off the MS into the HG during SSE, and then due to the RLOF in the BSE immediately transitions from HG into HeMS, see #604), this means that the previous stellar type will show HG, not MS (which is more sensible). 

For all other cases, the output will be very slightly different to what it is currently, since both stars will have evolved via SSE, but this is expected to be a very modest difference.


jeffriley:
@reinhold-willcox I'm not sure that having both calls to StashRLOFProperties() is doing what you want - let's chat on Zoom again.

reinhold-willcox:
### TODO: check the MT event counter

reinhold-willcox:
Note for posterity: Jeff noted above the 2 calls to StashRLOFProperties were confusing (though achieved the desired functionality). 

Now the function takes an argument to specify whether we are stashing the pre-MT values or the post-MT values, and the calls have the appropriate argument specified, which should hopefully clarify any confusion for future devs.

jeffriley (commented):
Looks good @reinhold-willcox .  A suggestion:

You have the parameter to StashRLOFProperties() as a boolean - which works, but is a bit obscure.  As a reader of the code, when I see the call:

StashRLOFProperties(true);

I have to go and look at the function to see what a parameter of "true" means.

Instead, if you declare an enum something like:

enum class MASS_TRANSFER_TIMING: int { PRE_MT, POST_MT };

then instead of

StashRLOFProperties(true);

and

StashRLOFProperties(false);

you can use

StashRLOFProperties(MASS_TRANSFER_TIMING::POST_MT);

and 

StashRLOFProperties(MASS_TRANSFER_TIMING::PRE_MT);

and then someone reading the code can tell what you're doing without having to inspect the function to work out what the parameter means.

I think it is clearer code, but it's up to you.

jeffriley:
Oh, and you could use a ternary operator as shorthand for the 'if' statement:

For the boolean argument:

rlofPropertiesToReset = p_StashPostMassTransfer ? m_RLOFDetails.propsPostMT : m_RLOFDetails.propsPreMT;

and for the enum argument if you go that way:

rlofPropertiesToReset = p_Which ==  MASS_TRANSFER_TIMING::POST_MT ? m_RLOFDetails.propsPostMT : m_RLOFDetails.propsPreMT;

jeffriley (changes requested):
@reinhold-willcox See my comments.

One required change is to update changelog and version

reinhold-willcox:
@jeffriley Implemented all your fixes, and also removed the parameter EventCounter_Prev. Previously it was always just one less than Event_Counter, and now it is always just 0. Not much use for it.

jeffriley (dismissed):
Approved

jeffriley:
Ok, I approved the changes and was about to merge - then realised there should be more changes to the code, and as a result to the documentation.  

Properties such as 

BINARY_PROPERTY::RLOF_CURRENT_STAR1_STELLAR_TYPE
BINARY_PROPERTY::RLOF_PREVIOUS_STAR1_STELLAR_TYPE

should really be renamed to reflect the new functionality - they are not really "previous" and "current", they are "pre-MT" and "post-MT", so maybe should be changed to

BINARY_PROPERTY::RLOF_POST_MT_STAR1_STELLAR_TYPE
BINARY_PROPERTY::RLOF_PRE_MT_STAR1_STELLAR_TYPE

(Those are just examples - there are more than just two properties affected)

Also should change the column header names for those properties too - what were the "previous" values have a "_prev" suffix - maybe that should change to "_PRE_MT", and what were the "current" values should have a "_POST_MT" suffix (or something similar).

Then the  following sections of the documentation would need to be updated to reflect the new names:

Section "Log File Record Specification: Binary Properties" (LogFileRecordSpecs-BinaryProps.tex)
Section: "Default Log File Record Specifications" (DefaultLogFileRecordSpecs.tex)

I tried to rescind the approval (dismiss the review), but apparently only the repository owner can do that...

jeffriley:
@reinhold-willcox see comments

jeffriley (approved):
Looks good - thanks @reinhold-willcox, I know those last-minute requested changes were a pain, but I think the product is better for them

reinhold-willcox:
@jeffriley easier to change variables in the documentation than try to invent new routines!

PR #616, opened 2021-09-01
updated plot for methods paper, now it uses the correct version

reinhold-willcox:


ilyamandel (approved):
Thank you, @reinhold-willcox !

Issue #615, opened 2021-09-01
random-seed incorrect implementation


reinhold-willcox:
**Describe the bug**
The random seed that is input on the command line or in a grid file is mis-aligned with the one that is used for a given run for the SN kick random parameters, and the first grid file line always uses the default seed of 0 for the kick parameters. This makes it  difficult to reproduce single systems of interest. I've spoken to Jeff about this and he knows the solution, so I won't go into more detail here.

**Label the issue**
`urgency_low`      - This issue is not urgent
`severity_moderate` - This is a moderately severe bug

 - COMPAS v02.21.00

jeffriley:
The problem here is that the random seed is set in main() after the program options are parsed and processed.  Some program options have default values that are drawn from distributions, set from random numbers, or are calculated from other option values that may have been drawn or set from random numbers.  The fix is to calculate those default option values after the random seed is set for the system being evolved.


In the case described in the issue, the run is reproducible, but extracting a specific system from the grid file and running it with the random seed reported would not generate the same output - the calculated option defaults (kick values) would not be calculated using the same random seed.

Fixed in PR #619 COMPAS v02.22.01

PR #618, opened 2021-09-11
Defect repair: Fix for issue #615

jeffriley:
Defaults for calculated/drawn program options now calculated after random seed is set for the system being evolved.

See issue #615 for details


reinhold-willcox:
thanks!

PR #619, opened 2021-09-11
Defect repair: Fix for Issue 615

### LINKED ISSUE #615: random-seed incorrect implementation

jeffriley:
@reinhold-willcox I believe this change fixes the problem you describe in issue #615 - please check and merge as appropriate.

Defaults for calculated/drawn program options now calculated after random seed is set for the system being evolved.

See issue #615 for details

Close issue#615 when PR merged


reinhold-willcox:
@jeffriley I haven't looked closely at the PRs, but why is this different to #618?

jeffriley:
@reinhold-willcox PR#618 was where I pulled the branch in from my fork - I always create a branch in the COMPAS repo rather tan merge directly into dev from my fork - that way if the reviewer wants to fix a typo or something, they can do it (I don't remember if I've granted write access to my fork to all possible reviewers,so creating a local branch here avoids issues).

reinhold-willcox:
Got it! Thanks for clarifying, I didn't notice the target branch for 618. I'll review this now.

reinhold-willcox:
Why is the vscode settings.json file included here? It shows it is now deleted, was this accidentally included in a previous commit?

reinhold-willcox:
@jeffriley I'm not seeing a difference in the output, unfortunately. I recompiled and ran with a simple test case, and the randomly generated values don't align as they should. 

For my test, I used 2 grid files, where the 2nd line of the first was taken as the only line in the second.
grid1.txt
> --initial-mass-1 1 --initial-mass-2 1 -a 1000
> --initial-mass-1 2 --initial-mass-2 1 -a 1000
grid2.txt
> --initial-mass-1 2 --initial-mass-2 1 -a 1000

I ran
```
COMPAS --grid grid1.txt --random-seed 10
COMPAS --grid grid2.txt --random-seed 11
```

In principle, this should mean that the seed of the second line in the first grid would be 11, to align with the seed of the second grid. But here is a screenshot of the SystemParams output, showing that the random numbers for the different kick params are not equal. Apologies that the header is cutoff, but the columns correspond to [units, grid1line1, grid1line2, units, grid2line1]. The column headers correctly show that the 'SEED' params are 10, 11, and 11 as expected, but actual random numbers seem to indicate that the first lines of each grid are getting the same random seed.

jeffPR_seedbug

jeffriley:
> Why is the vscode settings.json file included here? It shows it is now deleted, was this accidentally included in a previous commit?

Yeah, somehow it crept into my commit, so I deleted it


>  I'm not seeing a difference in the output

Let me take a look - my tests worked, but maybe my tests were too simple :-)

jeffriley:
Absolutely right @reinhold-willcox .  Poor testing on my part...  I just tested that the seed was being set in the right place correctly - I didn't check that the kick values were being set correctly.  I believe I have fixed it now, but you should test (my testing clearly can't be trusted ;-)

reinhold-willcox:
Sorry for the delay, this looks good. Thanks Jeff!

reinhold-willcox (approved):

PR #622, opened 2021-09-17
Changes to --help descriptions and output parameter names (previously: Some cosmetic changes)

### LINKED ISSUE #617: units in the --help
**Describe the bug**
In the COMPAS --help output, the description for --semi-major-axis does not say what units are used. I haven't checked the rest, but they should also be verified. This is an easy fix that I'm happy to handle, just reporting here to remind myself later.

**Label the issue**

### LINKED ISSUE #621: BSE_RLOF output SEED parameter name bug
**Describe the bug**
Another really small, almost cosmetic bug (sorry for being so picky). In the BSE_RLOF output, the `SEED` parameter has recently been renamed to `SEED>MT`. Many of the other parameters have been renamed similarly to more clearly signify if they apply >MT or <MT, but the SEED parameter doesn't really fit into that mold, and should be renamed back to just `SEED`. 

Jeff, I'm happy to fix this, though I may ping you later if I need some guidance.

**Label the issue**

 - COMPAS v02.22.00

reinhold-willcox:
- cleaned up units in the --help options, 
- changed the SEED>MT output param back to SEED for BSE_RLOF
- other typo fixes


jeffriley (changes requested):
If you are removing BINARY_PROPERTY::RLOF_POST_MT_RANDOM_SEED completely, you might as well remove the random seed from the RLOFProperties struct in typedefs, and remove the references to it in BaseBinaryStar::StashRLOFProperties() - no need to stash it away if we're never going to use it.

And, for completeness, why not, for the same reasons, give the same treatment to the random seed in the BeBinaryProperties struct in typedefs.h, BaseBinaryStar::StashBeBinaryProperties(), and the corresponding property in constants.h (BE_BINARY_CURRENT_RANDOM_SEED) - and replace that property in BSE_BE_BINARIES_REC with the binary random seed?

Need to update changelog and version string.

reinhold-willcox:
I'll remove all those things, but I don't think this warrants a change in version number - we're not changing code functionality.

jeffriley:
> I'll remove all those things, but I don't think this warrants a change in version number - we're not changing code functionality.

Sure we are - the code will behave differently after the change from what it did before the change.  Even the default output generated for the BSE_RLOF file will be different.

But even if you just, for example, refactored a function, and that refactoring was purely to tidy-up the code and wasn't meant to introduce any change in functionality, you should increment the version number.  If the refactoring inadvertently introduced a change (maybe even a defect - just because we don't intend there to be a functionality change doesn't mean it is guaranteed there won't be)), we need to be able to track the introduction of that change (or defect) to a version number, then look at what changed in that version.  Any code change requires a change to the version number.

jeffriley:
Also, it just occurred to me that you'll need to change the documentation - the section headed "Log File Record Specification: Binary Properties" - since you've removed properties that are listed as available to be printed.

Also, further rationale for recording your changes in the changelog and incrementing the version string:

1. You've removed variables from structs in typedefs.h
2. We are no longer storing values that we were storing prior to the change
3. We are no longer printing values that we were printing prior to the change, and furthermore we cannot print those values because they have been removed
4. The column name of the seed in the BSE_RLOF file has changed

If I had a copy of the current version of COMPAS, and I had a python script that relied on the header string for the seed in the BSE_RLOF file, and then I pulled the latest version after you merged your changes, my script would fail - and there would be no indication of why the BSE_RLOF file had changed.  I could fix my script, but I shouldn't have to change anything if the only change to COMPAS was a cosmetic change.  And if I wanted to print the values we now no longer store or print (or can print), I couldn't - all with no explanation.  So if I kept the existing executable I had before pulling the latest version, I could be in the position of having two COMPAS executables, both indicating the same version number, but behaving differently even for the same input values.  That's not ok.  These changes are not just cosmetic.

reinhold-willcox:
Some new additions to this PR:

- removed the random seed sub-parameter from RLOF and BeBinaries Details, since this parameter is already included in the overall binary.
- changed JLoss to AccretorMass in the header of CalculateGammaAngularMomentumLoss (cosmetic change that aligns with the input parameter name).
- Cleaned up output parameter names "CE_Event_Counter" and "MT_Event_Counter"
- Added Mass(1)>CE to CE output (and same for 2)

reinhold-willcox:
As I discussed with @jeffriley, I can't update the documentation for these parameter changes yet because the documentation source code isn't in the Compas repo. It might be best to wait until the documentation is in the repo and we can update everything at once, but I'll defer to Jeff's judgement.

In the meantime, there are a couple other issues I noticed that may be worth fixing here, but may also require further investigation.

- In the RLOF output, we have parameters `Time<MT` and `Time>MT`, but these appear to always be the same (at least to within the precision of the output). If they are always the same, we should scrap the preMT one. If they are not, then the difference between them is probably important, in which case we should still scrap the preMT one, but then add in a parameter for `Duration` or something which calculates the difference internally, to avoid the precision issues.
- The parameter m_Mass1Final is ambiguous and misleading. As far as I can figure out, it refers to the postCEMass of a star, but then this is getting printed as "CoreMass" which is incorrect. If the star prior to CE was MS or HeMS, it would not have a core but the core mass parameter was showing the full mass. This parameter is used in a few places in the CE code though, so I'm hesitant to start changing things ad hoc without more careful thought. I am happy to make a new issue for this if need be.

ilyamandel (changes requested):
changelog update missing

Changes to comments suggest that the kick sigma is used exclusively for Maxwellian distributions, but I am not sure that's strictly true -- see, e.g., BaseStar::DrawSNKickMagnitude().

reinhold-willcox:
Note: added *.eps to gitignore 

> changelog update missing

I can see it on mine, do you have a file filter in place? 

> Changes to comments suggest that the kick sigma is used exclusively for Maxwellian distributions, but I am not sure that's strictly true -- see, e.g., BaseStar::DrawSNKickMagnitude().

Fair point, I'll fix that back

ilyamandel (approved):
Looks good!  Thanks, @reinhold-willcox !

reinhold-willcox:
@jeffriley Are you happy with your requested changes?

jeffriley:
Oh, sorry @reinhold-willcox - I didn't realise you were waiting on me.  I was waiting for the PR to change state from draft.  I'll take a look.

jeffriley (commented):
@reinhold-willcox Changes look good.  Just a couple of comments:

(a) you have removed MASS_1_FINAL and MASS_2_FINAL from BSE_COMMON_ENVELOPES_REC.  This was the only place these variables were referenced - they were only there to match the legacy code CE file.  Is it likely they'll ever be needed elsewhere?  That is, should we actually remove the variables from the code completely?  If you think someone might want to print those values, with those headers, then leave them, but otherwise we probably should avoid having unused and unhelpful variables in the code.

Ok, I just noticed that you note the underlying class member variables (m_Mass1Final and m_Mass2Final) are used in a number of places.  We could still remove the properties MASS_1_FINAL and MASS_2_FINAL, but that would mean users couldn't print the variables m_Mass1Final and m_Mass2Final - maybe that's a good thing if we don't really know what they refer to...  Maybe leave them there until we spend some more time rationalising those variables)

(b) I noticed some inconsistencies in the options help (in Options.cpp):

(i) Option 'mcbur1'
"MCBUR1: Min core mass at BAGB, in Msol, to avoid fully degenerate CO core"

This seems to be the only option that has it's name ("MCBUR1:") in the help string - remove it?  Also, to be consistent we probably should spell out "minimum".

(ii) Option 'neutrino-mass-loss-BH-formation-value'
"Value (fraction or fixed mass) corresponding to neutrino mass loss assumption"

I know you and Simon discussed this - I'm not sure that text is clear, but we don't have a lot of space to expand I suppose.  Maybe we need to expand the description in the documentation.  Just a note...

(iii) We're not consistent with these:

"semi-major-axis,a": Initial semi-major axis a, in AU
"semi-major-axis-max": Maximum semi major axis, in AU, to generate
"semi-major-axis-min": Minimum semi major axis, in AU, to generate

The first mentions "a" (i.e. "semi-major axis a"), but the other two don't.
The first hyphenates "semi-major", but the other two don't

Let me know your thoughts.  If you prefer we can merge this PR and submit another to make the changes noted above.

(p.s. I can make the necessary changes to the online docs - when I'm done I'll let you know so that you can check them)

reinhold-willcox:
> (a) you have removed MASS_1_FINAL and MASS_2_FINAL from BSE_COMMON_ENVELOPES_REC. This was the only place these variables were referenced - they were only there to match the legacy code CE file. Is it likely they'll ever be needed elsewhere? That is, should we actually remove the variables from the code completely? If you think someone might want to print those values, with those headers, then leave them, but otherwise we probably should avoid having unused and unhelpful variables in the code.

> Ok, I just noticed that you note the underlying class member variables (m_Mass1Final and m_Mass2Final) are used in a number of places. We could still remove the properties MASS_1_FINAL and MASS_2_FINAL, but that would mean users couldn't print the variables m_Mass1Final and m_Mass2Final - maybe that's a good thing if we don't really know what they refer to... Maybe leave them there until we spend some more time rationalising those variables)

I will remove the MASS_X_FINAL properties, I replaced them with MASS_X_POST_COMMON_ENVELOPE in the logs which is clearer for anyone who wants to get that output. I am happy to hold off on changing the member variables until we look closer at the code that uses them. But ultimately, they should be swapped out for something clearer. 

> (b) I noticed some inconsistencies in the options help (in Options.cpp):

> (i) Option 'mcbur1'
> "MCBUR1: Min core mass at BAGB, in Msol, to avoid fully degenerate CO core"

> This seems to be the only option that has it's name ("MCBUR1:") in the help string - remove it? Also, to be consistent we probably should spell out "minimum".

This is an interesting point. There's both a MCBUR1 and MCBUR2 in the code - arguably both should be included in the input parameters (I recently ran some models where I had to recompile to change MCBUR2). These parameters come from Hurley, so my instinct is to leave them as is if it will help other Hurley afficionados. But looking at the original paper, he doesn't use "M_{c,BUR1}" or anything like that, so I don't where these particular parameter names come from or what they mean, and I'm not sure if they are actually helpful. 

I can remove the parameter name from the help string (and spell out "minimum"). We should probably discuss keeping MCBUR1/2 as is or renaming, and I'll open an issue to add MCBUR2 to the Options.

> (ii) Option 'neutrino-mass-loss-BH-formation-value'
> "Value (fraction or fixed mass) corresponding to neutrino mass loss assumption"

> I know you and Simon discussed this - I'm not sure that text is clear, but we don't have a lot of space to expand I suppose. Maybe we need to expand the description in the documentation. Just a note...

I agree it's unclear. How about "Amount of BH mass lost due to neutrinos (either fraction or fixed value, depending on neutrino-mass-loss-BH-formation)". I don't think length is a concern, there are other long parameter help strings which wrap nicely.

> (iii) We're not consistent with these:
> "semi-major-axis,a": Initial semi-major axis a, in AU
> "semi-major-axis-max": Maximum semi major axis, in AU, to generate
> "semi-major-axis-min": Minimum semi major axis, in AU, to generate

> The first mentions "a" (i.e. "semi-major axis a"), but the other two don't.
> The first hyphenates "semi-major", but the other two don't

Good catch, I'll fix those and repush.

jeffriley:
> (i) Option 'mcbur1'
> "MCBUR1: Min core mass at BAGB, in Msol, to avoid fully degenerate CO core"
>
> This seems to be the only option that has it's name ("MCBUR1:") in the help string - remove it? Also, to be consistent we
> probably should spell out "minimum".
>
> This is an interesting point. There's both a MCBUR1 and MCBUR2 in the code - arguably both should be included in the input
> parameters (I recently ran some models where I had to recompile to change MCBUR2). These parameters come from Hurley,
> so my instinct is to leave them as is if it will help other Hurley afficionados. But looking at the original paper, he doesn't use
> "M_{c,BUR1}" or anything like that, so I don't where these particular parameter names come from or what they mean, and I'm
> not sure if they are actually helpful.

I was only advocating removing the "MCBUR1:" string from the help string.  The constants were added in v02.11.02 by @ilyamandel , and the option for MCBUR1 was added in v02.13.07 by Ilya.  Ilya has put comments describing them in constants.h

> How about "Amount of BH mass lost due to neutrinos (either fraction or fixed value, depending on 
> neutrino-mass-loss-BH-formation)". 

Works for me.

ilyamandel:
MCBUR1 and 2 come from Fryer et al. -- see section 4.4.  More later, when done with homeschooling.

jeffriley (approved):
All good - approved

ilyamandel (commented):
@reinhold-willcox , @jeffriley -- has the documentation been changed to match the new output variable names?

reinhold-willcox:
@ilyamandel thanks for the reference. Jeff has said he will handle documentation updates separately until the website source code is added to the repo.

PR #625, opened 2021-09-18
Updating paper plot

LiekeVanSon:
Uploaded scripts needed to reproduce Fig. 10 in the Method paper. 

Also includes a two of bug fixes in the Cosmic integration:
1) in Class COMPAS I noticed there were flags read as integers/indexing in stead of bools
2) in FastCosmicIntegrator there was a buggy version of the append_binned_by_z that caused issues at the edges of bins and did not properly sum rates


FloorBroekgaarden (approved):
Looks fine to me. I would only add maybe a reference somewhere to the STROOPWAFEL package (maybe at the top of strroopwafel_interface.py)?

FloorBroekgaarden:
I added myself the stroopwafel comment and merged branch into dev

PR #627, opened 2021-09-23
updated readme with the COMPAS methods paper

FloorBroekgaarden:
I included a bibtex file to get the right (and preferred) way of citing this paper with Jeff included


ilyamandel (approved):
Great, thank you, @FloorBroekgaarden!

Issue #572, opened 2021-05-27
FastCosmicIntegration documentation


ilyamandel:
@jeffriley  pointed out two things about FastCosmicIntegration which I am afraid aren’t sufficiently clearly stated in the documentation (mea culpa) but explain some differences with the old cosmic integration that Jeff encountered:

1. FastCosmicIntegration, as currently written, expects metallicities to be distributed uniformly in the log, whether binned or stochastically sampled metallicities are used.  (This is easy to extend, but that’s the current choice which should be documented.)

2. When the metallicity distribution extends beyond the range of COMPAS-simulated metallicities, we need to make a choice.  Coen’s cosmic integration code folded the metallicities outside the simulated range into the edge bins.  I am renormalising the metallicity distribution so that dP/dZ integrates to 1 between min_logZ_COMPAS & max_logZ_COMPAS — effectively, this means I am actually using a slightly different metallicity distribution than the input one.  Or one could use the initial distribution without re-normalisation and simply assume that metallicities outside the COMPAS range have yields of zero.  All choices can be justified, we again just need to be more explicit in documenting them (in code and paper).

@TomWagg — maybe you could include the documentation in your updates.  @stellarphysicsandpopulations  — just drawing your attention to this, because I know you have thought about this. ;)

ilyamandel:
Included in #628 .

PR #628, opened 2021-09-27
Merger output

### LINKED ISSUE #591: Backwards check in FastCosmicIntegrator
@jeffriley points out that this check in FastCosmicIntegrator appears to be backward.  @TomWagg , can you please check?

    # another warning on poor input
    if max(chirp_masses)*(1+max_redshift_detection) < Mc_max:
        warnings.warn("Maximum chirp mass used for detectability calculation is below maximum binary chirp mass * (1+maximum redshift for detectability calculation)", stacklevel=2)

ilyamandel:
1. Added an option to run with just one metallicity, corresponding to the assumption that all of the star formation in the universe happened at exactly that metallicity (this is an option we discussed at a recent meeting with Jan’s group).  

2. I added an option to create an output text file storing the mergers in the way we discussed with Jan et al.

3. I find it very risky and a bit unnecessary to set the mass cutoffs for sampling when we have the actual samples.  In the longer term, I think we should find a better way to output these in some auxiliary file (we discussed this with @jeffriley  already).  For now, I introduced a new optional parameter use_sampled_mass_ranges.  

Also fixed #591 and #572 .


ilyamandel:
@FloorBroekgaarden , @TomWagg : if/when you review this, ignore generate_frame_file.py -- that's a work in progress for now, for the next step.

PR #634, opened 2021-10-01
added mean anomaly to grid

reinhold-willcox:


reinhold-willcox:
@ryosuke-hirai I added you as a reviewer in case you'd like to see how it's done. If you're not bothered, anyone else can do it.

Issue #612, opened 2021-08-30
Eddington mass accretion limit


ilyamandel:
The current implementation of the Eddington mass accretion limit onto BHs corresponds to an accretion efficiency of about 0.5, which is about a factor of 5 higher than expected, reducing the mass limit by a factor of 5.  Probably doesn't have much impact in the end, but would be good to fix.

ilyamandel:
Resolved in #639 , closing.

PR #639, opened 2021-10-04
Eddington limit fix to issue #612

ilyamandel:
Corrected Eddington mass accretion limits, issue #612 (very minor change for WDs and NSs, factor of a few increase for BHs)


ilyamandel:
Moving the .eps file was intentional -- otherwise, I keep over-writing it every time I re-create it as a test. :)

jeffriley (approved):
Approved

reinhold-willcox:
@ilyamandel should we put *.eps files into the gitignore? We can force add them if we ever need to update new plots, but that way you can run the tests and not accidentally include them in future PRs.

ilyamandel:
@reinhold-willcox -- I agree, gitignore would probably have been a more elegant solution. ;-)

reinhold-willcox:
Added to #622

PR #620, opened 2021-09-16
added the figure numbers to two of  the folders to be consistent with…

FloorBroekgaarden:
… the other plotting folders

Just a naming shift; nothing big. This will make the folder hierarchy clearer for readers.

PR #641, opened 2021-10-06
Revert "added the figure numbers to two of  the folders to be consistent with…"

ilyamandel:
Reverts TeamCOMPAS/COMPAS#620

@FloorBroekgaarden -- I should have spotted this earlier, but you moved a bunch of files which are specifically referenced in section 7 of the paper, with URLs.  This is not good — the paper is under review, and we don’t want to avoid a reviewer who might decide to check the links in the paper.  At the moment, links to Grid_demo.txt , detailed_evol_plotter.py, chirpmass_dist.ipynb and possibly others are broken.  I am therefore reverting this pull request -- all links in the paper should work (at a minimum, via soft links).


reinhold-willcox (approved):
Looks safe -> most files are simply renamed back to align with methods paper, some svg files are now deleted (but they are just copies of existing images in different formats) and a change in version of a notebook.

PR #644, opened 2021-10-09
Brief evolutionary history printing

ilyamandel:
Added a function to print a synopsys of the evolutionary history to the command line; it can eventually include cartoons as well

### DIFF
--- a/examples/methods_paper_plots/detailed_evolution/detailed_evol_plotter.py
+++ b/examples/methods_paper_plots/detailed_evolution/detailed_evol_plotter.py
@@ -153,7 +153,9 @@ def printEvolutionaryHistory(Data):
         if Data['Eccentricity'][i]>1 or Data['SemiMajorAxis'][i]<0:     #unbound
             print(Data['Time'][i], '   Unbound binary')
 
-    if Data['Time'][i]<14000 and not ((Data['Stellar_Type(1)'][i]==13 or Data['Stellar_Type(1)'][i]==14) and (Data['Stellar_Type(2)'][i]==13 or Data['Stellar_Type(2)'][i]==14)) and not (Data['Eccentricity'][i]>1 or Data['SemiMajorAxis'][i]<0):     #proxy for unrecorded meregr
+    isDCO = (Data['Stellar_Type(1)'][i]==13 or Data['Stellar_Type(1)'][i]==14) and (Data['Stellar_Type(2)'][i]==13 or Data['Stellar_Type(2)'][i]==14)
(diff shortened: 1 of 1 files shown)


ilyamandel:
@reinhold-willcox -- ping on this one (no rush). :)

reinhold-willcox:
@ilyamandel Couple of questions:

- Can you walk me through the logic of the stellar merger block of the printEvolutionaryHistory function (reproduced below)? I understand that i here takes on the index of the final timestep, so it looks like the condition is "If (evolution ends before a Hubble Time) and if not (a binary composed of some combination of BHs and NSs)". Why is this indicative of a stellar merger, as opposed to (say) a wide double white dwarf binary? Similarly, why does the time come into play? Couldn't we have a binary of two very low mass MS stars which do not evolve significantly within a Hubble time, but which will still ultimately merge?
```
if Data['Time'][i]<14000 and not((Data['Stellar_Type(1)'][i]==13 or Data['Stellar_Type(1)'][i]==14) and (Data['Stellar_Type(2)'][i]==13 or Data['Stellar_Type(2)'][i]==14)):     #proxy for unrecorded merger
         print(Data['Time'][i], '   Stellar merger')
```

- In the final block, you run through the Peters EQs for GW insprial time. Is this actually from your recent paper on this? If so, could you link the paper (I can't find it on arxiv or ads)? Secondly, this calculation is done internally in the code for DCO mergers, but is recorded in an output file distinct from the detailed output file used here. We could either keep this as is, or import the DCO output file and use the internally calculated value. Keeping as is means we effectively have the same code written twice (and hopefully consistently!), but importing the DCO output could get cumbersome if someone moves the files around separately or renames them. I'm happy either way, since neither is a particularly elegant approach.

ilyamandel:
@reinhold-willcox:

1. Since the detailed evolution output doesn't flag mergers, I am using a simple proxy: if we stop the evolution in less than 14 Gyr and it's not a disruption (searched for separately, now added to this line) nor a DCO, then merger is the only other explanation. :)

2. I added a reference to https://iopscience.iop.org/article/10.3847/2515-5172/ac2d35 in the comments.  It is indeed the same calculation that we now use in the COMPAS C++ code, so I agree that doing it twice may appear suboptimal -- but this is easiest for the purposes of producing a quick story / quick set of plots for the run-a-binary-online project.

reinhold-willcox:
Don't we also stop evolution for WD+WD binaries though? Or are you counting that as an (eventual) merger if it's not unbound?

ilyamandel:
Great catch, @reinhold-willcox -- I'd forgotten about our stopping on DWDs!  Check added.

reinhold-willcox:
@ilyamandel The check looks good. I was getting a bit confused though, trying to interpret which parameters fell under which headers, so I've added some code that formats the output into columns. Let me know what you think.

ilyamandel:
Looks great, @reinhold-willcox !  Thank you!  I can't approve, since I started the PR.   Can you?  If not, we'll ask someone else. :)

reinhold-willcox (approved):

PR #645, opened 2021-10-10
Enhancement: update to thermal timescale calculation

jeffriley:
(a) updated kelvin-helmholtz (thermal) timescale calculation with more accurate pre-factor and updated documentation.
(b) rationalised parameters of, and calls to, CalculateThermalTimescale()


ilyamandel (commented):
Looks good!

ilyamandel:
For some reason, I can only comment, not approve...

PR #646, opened 2021-10-10
Enhancement: Thermal timescale update

jeffriley:
(a) updated kelvin-helmholtz (thermal) timescale calculation with more accurate pre-factor and updated documentation.
(b) rationalised parameters of, and calls to, CalculateThermalTimescale()

This PR opened to replace PR #571 (I didn't want to have to deal with potential merge conflicts, so I started from a clean copy of dev).  See PR #571 for details.

Produced methods paper detailed output plot as check - identical (that I could see...).  No other tests performed.


ilyamandel (approved):
OK, never mind. :)

PR #648, opened 2021-10-11
Minor code cleanup

jeffriley:
(a) Typo fixed in version for changes made on October 11, 2021
(b) Changed KROUPA_POWER to SALPETER_POWER in utils:SampleInitialMass(); Removed KROUPA_POWER from constants.h
(c) Removed p_Id parameter from SSE/BSE switchlog functions - leftover from debugging
(d) Added CHEMICALLY_HOMOGENEOUS_MAIN_SEQUENCE property to SSE_SYSTEM_PARAMETERS_REC and BSE_SYSTEM_PARAMETERS_REC (both stars)
(e) Tidied up some parameters etc. to better comply with COMPAS coding guidelines
(f) Typo fixed in preProcessing/COMPAS_Output_Definitions.txt

- Online documentation updated to reflect new default sysparms record (SSE and BSE)
- Methods paper detailed plot produced - identical (no other tests performed)

### DIFF
--- a/preProcessing/COMPAS_Output_Definitions.txt
+++ b/preProcessing/COMPAS_Output_Definitions.txt
@@ -48,7 +48,7 @@ bse_sysparms_rec -= {                                                   # SUBTRA
 }
 
 
-# BSE Double Compas Objects
+# BSE Double Compact Objects
 
 BSE_DCO_Rec = {STAR_1_PROPERTY::MZAMS,STAR_2_PROPERTY::MZAMS}           # set the BSE Double Compact Objects specifivation to MZAMS for Star1, and MZAMS for Star2
 BSE_DCO_Rec = {}                                                        # set the BSE Double Compact Objects specification to empty - nothing will be printed (file will not be created)
--- a/src/.vscode/settings.json
+++ b/src/.vscode/settings.json
@@ -1,5 +1,0 @@
-{
-    "files.associations": {
-        "random": "cpp"
-    }
-}
--- a/src/.vscode/settings.json
+++ b/src/.vscode/settings.json
@@ -1,0 +1,5 @@
+{
+    "files.associations": {
+        "random": "cpp"
+    }
+}
--- a/src/BaseBinaryStar.h
+++ b/src/BaseBinaryStar.h
@@ -271,7 +271,7 @@ public:
 
             EVOLUTION_STATUS    Evolve();
 
-            bool                PrintSwitchLog(const long int p_Id, const bool p_PrimarySwitching) { return OPTIONS->SwitchLog() ? LOGGING->LogBSESwitchLog(this, p_Id, p_PrimarySwitching) : true; }
+            bool                PrintSwitchLog(const bool p_PrimarySwitching) { return OPTIONS->SwitchLog() ? LOGGING->LogBSESwitchLog(this, p_PrimarySwitching) : true; }
 
             COMPAS_VARIABLE     PropertyValue(const T_ANY_PROPERTY p_Property) const;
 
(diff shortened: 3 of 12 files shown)


ilyamandel (changes requested):
I think KROUPA_MINIMUM should then be changed to SALPETER_MINIMUM and ditto for MAXIMUM for consistency.

jeffriley:
@Ilya Mandel ***@***.***> I thought of that when I was making
the changes, but I don't think those constants exist. Is it worth creating
them for this purpose, or should we have more generic constants, e.g.
IMF_DEFAULT*

ilyamandel:
But we do have KROUPA_MINIMUM and KROUPA_MAXIMUM defined, and the only place we use them is in the default (unknown IMF) catch of SampleInitialMass():

 default:                                                                                                // unknown IMF
                thisMass = utils::InverseSampleFromPowerLaw(KROUPA_POWER, KROUPA_MAXIMUM, KROUPA_MINIMUM);          // calculate mass using power law with default values
                
You've now correctly renamed the first argument KROUPA_POWER->SALPETER_IMF; I am just suggesting renaming the last two, for consistency.

jeffriley:
Huh!  I just assumed KROUPA_MINIMUM and MAXIMUM were used elsewhere...  Sorry - I wasn't near my desk, so wasn't looking at the code.

Ok, we have this in Options.cpp:

    m_InitialMassFunction.type                                      = INITIAL_MASS_FUNCTION::KROUPA;
    m_InitialMassFunction.typeString                                = INITIAL_MASS_FUNCTION_LABEL.at(m_InitialMassFunction.type);
    m_InitialMassFunctionMin                                        = 5.0;
    m_InitialMassFunctionMax                                        = 150.0;

@ilyamandel Should the default minimum and maximum be the same as (what will now be) SALPETER_MINIMUM and MAXIMUM (though they don't actually match - KROUPA_MINIMUM is currently 0.5 in constants.h, and max is 100.0)?  Or should we define other constants for those so they have meaningful names and we know where the numbers came from (I have no idea what those names would be)?

ilyamandel:
Why not just use SALPETER_MINIMUM=0.5, SALPETER_MAXIMUM=100 ?  It doesn't really matter too much -- we are just picking some values for the fallback in defensive programming, but we expect to never use it, since we should be defining an IMF choice...

jeffriley:
Ok, @ilyamandel, I've changed the KROUPA_MINIMUM and KROUPA_MAXIMUM in constants.h and utils.cpp to SALPETER_MINIMUM and SALPETER_MAXIMUM respectively.

I wasn't clear in my previous comment.  We now still have the following in Options.cpp:


    m_InitialMassFunction.type       = INITIAL_MASS_FUNCTION::KROUPA;
    m_InitialMassFunction.typeString = INITIAL_MASS_FUNCTION_LABEL.at(m_InitialMassFunction.type);
    m_InitialMassFunctionMin         = 5.0;
    m_InitialMassFunctionMax         = 150.0;

These defaults are used in main.cpp when initialising a single star, and in BaseBinaryStar.cpp when initialising the constituent stars of the binary.

Should we:

(a) have separate named constants for the 5.0 and 150.0 for m_InitialMassFunctionMin and m_InitialMassFunctionMax respectively
(b) use SALPETER_MINUMUM and SALPETER_MAXIMUM (i think that would look odd since the default function is KROUPA)
(c) leave them as 5.0 and 150.0 in Options.cpp (my least preferred option - I think we should eventually have all (such) constants defined in one place (constants.h), and as just numbers instead of meaningful names we don't know how there were derived)

ilyamandel (approved):
Looks great!

ilyamandel:
@jeffriley -- I rather like (c), but I can tell you don't. :-)

jeffriley:
> @jeffriley -- I rather like (c), but I can tell you don't. :-)

Indeed @ilyamandel, but I suppose I could be convinced that option defaults could be in Options.cpp - but we still lose something by just having numbers without reference to how we decided upon them.

PR #651, opened 2021-10-11
Minor enhancements/optimisations

### LINKED ISSUE #538: Simplify and clean `CalculateTimeToCoalescence` function
**Describe the bug**
The `CalculateTimeToCoalescence` currently handles exceptions individually; however, there is no need for this following Peters 1964. We need to remove the if statements and leave an unique integral. Moreover, we need to use an integrator to solve for that integral instead of an ad-hoc Rieman sum solver.

**Label the issue**
`urgency_low`
`severity_minor`

 - COMPAS v02.18.09

### LINKED ISSUE #595: Mass transfer donor/acceptance rate should be a property of the object star
COMPAS v02.20.01 includes a non-ideal treatment of mass transfer, as the mass donor and acceptance rates are calculated manually and then passed as arguments to a function (see `CalculateMassTransfer()` function in `BaseBinaryStar.cpp`):
```
		// Begin Mass Transfer
        double thermalRateDonor    = m_Donor->CalculateThermalMassLossRate();
        double thermalRateAccretor = OPTIONS->MassTransferThermallyLimitedVariation() == MT_THERMALLY_LIMITED_VARIATION::RADIUS_TO_ROCHELOBE
                                        ? (m_Accretor->Mass() - m_Accretor->CoreMass()) / m_Accretor->CalculateThermalTimescale(m_Accretor->Mass(), CalculateRocheLobeRadius_Static(m_Accretor->Mass(), m_Donor->Mass()) * AU_TO_RSOL, m_Accretor->Luminosity(), m_Accretor->Mass() - m_Accretor->CoreMass())                                                  // assume accretor radius = accretor Roche Lobe radius
                                        : m_Accretor->CalculateThermalMassLossRate();
                
        std::tie(std::ignore, m_FractionAccreted) = m_Accretor->CalculateMassAcceptanceRate(thermalRateDonor, thermalRateAccretor);
``` 
This behaviour is likely poor practice from COMPAS Legacy. We should adapt and allow each object to know it's mass donor and acceptance rate in order to have something like, e.g.,:
```
std::tie(std::ignore, m_FractionAccreted) = m_Accretor->CalculateMassAcceptanceRate(m_Donor->massDonatingRate(), m_Accretor->massAccretionRate());
```

jeffriley:
(a) Added BaseStar::CalculateThermalMassAcceptanceRate() as a first-pass to address issue #595 - can be changed/expanded as required
(b) Changed BaseBinaryStar::CalculateTimeToCoalescence() to use Mandel 2021 https://iopscience.iop.org/article/10.3847/2515-5172/ac2d35, eq 5 to address issue #538

Close issue #595 when merged
Close issue #538 when merged

### DIFF
--- a/src/BaseBinaryStar.cpp
+++ b/src/BaseBinaryStar.cpp
@@ -1010,6 +1010,8 @@ double BaseBinaryStar::CalculateTimeToCoalescence(const double p_SemiMajorAxis,
                                                   const double p_Mass1,
                                                   const double p_Mass2) const {
 
+    if (p_Eccentricity >= 1.0) return 0.0;                                                              // save some cpu cycles...
+
     // pow() is slow - use multiplication where possible
 
     // calculate time for a circular binary to merge - Mandel 2021, eq 2
@@ -1024,9 +1026,10 @@ double BaseBinaryStar::CalculateTimeToCoalescence(const double p_SemiMajorAxis,
     double e0_20  = e0_10 * e0_10;
     double e0_100 = e0_10 * e0_10 * e0_10 * e0_10 * e0_10 * e0_10 * e0_10 * e0_10 * e0_10 * e0_10;
     double f      = 1.0 - (e0 * e0);
-    double f_3    = f <= 0.0 ? 0.0 : f * f * f;
+    double f_3    = f * f * f;
 
-    return f <= 0.0 ? 0.0 : tC * (1.0 + 0.27 * e0_10 + 0.33 * e0_20 + 0.2 * e0_100) * f_3 * sqrt(f);     // time for an eccentric binary to merge
+    // return time for an eccentric binary to merge
+    return f <= 0.0 ? 0.0 : tC * (1.0 + 0.27 * e0_10 + 0.33 * e0_20 + 0.2 * e0_100) * f_3 * sqrt(f);    // check f <= 0.0 just in case a rounding error hurts us
 }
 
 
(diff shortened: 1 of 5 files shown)


jeffriley:
Methods paper detailed output plot duplicated ok - no other tests performed.

jeffriley:
@ilyamandel  I ran a test comparing this change with COMPAS v02.23.01 - calculated coalescence time difference as a % of v02.23.01 number.  Option values were COMPAS defaults.

    pop count = 100,000
    dco count = 336

    count % diff < 0 = 315
    count % diff = 0 = 0
    count % diff > 0 = 21

    min % diff = -5.14013776
    max % diff = 11.38863153
    mean %diff = -0.76298961
    stddev % diff = 1.22384021

ilyamandel:
@jeffriley -- thank you for the extra check.  What was the eccentricity on the large (11%) difference.  Was it above 0.9?

ilyamandel (commented):
@jeffriley -- it's still a bit quirky that BaseStar::CalculateMassAcceptanceRate(), which is called on the accretor, needs to have the accretor's thermal mass acceptance rate as an input, but I think this is good enough for now.

ilyamandel (approved):

jeffriley:
@ilyamandel For the 11% difference, e was 0.99089, and for the -5% difference e was 0.984237

I agree re BaseStar::CalculateMassAcceptanceRate(), but it needs the Roche lobe radius which is unknown to the BaseStar class.  That's precisely why I wrote "as a first-pass" - we should look at how we can fix that.

PR #661, opened 2021-10-26
Proper printing for stellar mergers

### LINKED ISSUE #660: Stellar Merger classification
**Describe the bug**
We do not currently have a consistent way to output the final parameters of a system that undergoes stellar merger, without using the detailed printing. This is generally desirable (I won't list all the use cases here) even without an internal prescription for how the post-merger remnant should look - this can be done in post-processing. I believe this should be possible with the current output files we have, without the need for a new one. 

Here is a rundown of the stellar merger-related parameters across all the output files:
- In the `BSE_System_Parameters` output, we have the parameters `Merger` and `Merger_at_birth` which appear to work as advertised. I can get all the seeds of non-immediate mergers by masking for `'Merger' and not 'Merger_at_birth'`. 
- In the `BSE_RLOF` output, we have the radii of both stars and the separation before and after each MT event (which we can use to calculate if the post-MT system has merged). We also have `CEE>MT` which is potentially useful, though I have found that it occasionally gives false negatives.
- In the `BSE_Common_Envelopes` output, we have a parameter specifically for `Merger`, which works as advertised but requires that a Common Envelope event be triggered.

I have attached an output file that contains several example systems which undergo stellar merger, but are not flagged correctly in the RLOF or CEE outputs, and thus cannot be analyzed. The output was created prior to the seed bug fix, so it's not straightforward to reproduce the systems on the seed alone, but the seeds in question include:
[ 106277 148043 159355 159374 159850 160013 ]

Edit: The datafile, even zipped, is too large for github. I'll try to slim it down, but it may be easier to send to interested parties through ozstar...

Interestingly, all of these systems undergo a 1st SN USSN, where the companion is a very low mass (<1 Msol) star. The timing of the SN also coincides with the 2nd MT event, and the separation and radius parameters help to illuminate what (I think) is going on. Taking values specifically from seed 106277 at time 18.664696 Myr:

| Output file | Param | pre-Event | post-Event |
| ------------- | --------- | --------------| ---------------|
| BSE_RLOF | SemiMajorAxis |  5.720709 | 1.296268 |
| BSE_RLOF | Radius(1) | 3.307674 | 0.000014 |
| BSE_RLOF | Radius(2) | 0.655627 | 0.655701 |
| BSE_Supernovae | SemiMajorAxis | 0.454136 | 1.296268 |

The post separations from RLOF and SN agree, but the pre separations do not - I attribute this to the SN output sampling the preSN separation in the middle of the timestep ( _after_ the RLOF). If this is the case, the true separation after RLOF is less than the sum of the post-RLOF radii, and thus this should be considered a merger. The System_Parameters flags this as such, so internally Compas is recognizing this correctly. But it then continues and evaluates the SN, widening the separation and resulting in a post-SN and post-RLOF separation that allows for the two stars to fit. It is worth noting that the 2nd MT isn't classified as a CE, so there is no associated CE output event.

A simple fix for this specific bug, which would also solve the larger issue of inconsistent printing of stellar mergers, would be to trigger the CE output any time there is a merger, i.e when the separation is less than the sum of the radii (with tests added at the end of the timestep and in the middle - between the SSE and BSE components). From there, the relevent parameters of interest for mergers can be added as desired to the BSE_COMMON_ENVELOPES_REC and stellar mergers will be appropriately recorded. 

 - COMPAS v02.22.00 for the provided data file, but the bug exists in the current v02.24.02 as well.

reinhold-willcox:
First PR includes initial changes to allow for stellar mergers to be printed, by considering them as CE events.
- BSE_RLOF now correctly reports CEE>MT for all CE events, including stellar mergers
- boolean donorMS is now isDonorMS
- boolean StellarMerger() now includes a call to HasStarsTouching(), and various functions which check `if HasStarsTouching() { m_Flags.stellarMerger = true;}` have been removed as they are redundant.
- After ResolveMassChanges() call, if there is a StellarMerger(), then ResolveCommonEnvelopeEvent() is called. This allows stellar mergers to be printed as CE events. It needs to be done here because some mergers only happen after ResolveMassChanges(), but if you hit the end of the loop, the binary quits out because of the merger (without outputting anything)
- In the line which prints extra detailed output (sidebar: should we rename 'extra' to 'additional'?), I removed the condition ` && !StellarMerger()`, so now the final timestep in the detailed output shows that a stellar merger has happened. 
- I removed this as well at the end of the main Evolve() loop.
- I removed the main loop checks for HasStarsTouching(). This is conceptually similar to StellarMerger() and so somewhat confusing, and by redefining StellarMerger() to include the call to HasStarsTouching(), this becomes superfluous. Admittedly, this is not hugely important, but hopefully easier for new developers to follow the logic.
- some typo fixes

This is a draft PR because I haven't checked it thoroughly and I'm sure there are other improvements before we add this functionality to the dev branch.
- In particular, we should check if the stellarMerger check at the beginning of CalculateMassTransfer() can have the CHE bits removed (it seems safe to me, but we should check).
- TODO: remove EVOLUTION_STATUS::STARS_TOUCHING flag - it is superceeded by STELLAR_MERGER


reinhold-willcox:
Closing as very outdated

PR #667, opened 2021-10-31
 Adding online docs in preparation for move to compas.readthedocs.io

jeffriley:
Will test, then push a readme with instructions on how to update the online docs.


ilyamandel (approved):
Thank you very much for putting this together, @jeffriley !

Issue #637, opened 2021-10-04
Change default treatment of immediateRLOFPostCEE and optimisticCE


ilyamandel:
At the moment, on immediateRLOFPostCEE (RLOF immediately after CE) and optimisticCE (CE from radiative-envelope donor), we just record the appropriate flags and continue with evolution as if the binary survived the CE.  However, this has two problems:

1. If we want to implement the treatment of mergers (see #209 ), this is no longer a flag issue, but an actual bifurcation in the binary's evolution, and we have to make a decision on whether to treat this as a merger event or a surviving binary event.  

2. Many beginning users or users of our zenodo outputs don't realise that they need to take out binaries with such flags (I've had to explain this to a number of people recently).

I therefore propose adding two additional flags to pythonSubmit / command-line calls, AllowImmediateRLOFPostCE and AllowRadiativeDonorsSurviveCE .  Both flags would be set to false by default, which would treat such systems as mergers and discontinue future evolution.  If the flags are set to true, we would follow the current behaviour (still recording the flags as now).  

Since this is a change in default behaviour, I'm opening a one-week comment period before implementing this.

ashruiter:
This seems completely reasonable!

ilyamandel:
Closed with PR #674 .

PR #674, opened 2021-11-01
Change default treatment of immediateRLOFPostCEE and optimisticCE

ilyamandel:
As described in issue #637 , this PR introduces two new options, common-envelope-allow-radiative-envelope-surive and common-envelope-allow-immediate-rlof-post-ce-survive .  Both flags would be set to false by default, which would treat such systems as mergers and discontinue future evolution. If the flags are set to true, we would follow the current behaviour (still recording the flags as now).  Tested that figure 15 is reproduced correctly.


jeffriley (approved):
All ok.

I fixed detailed_evol_plotter.py inline.

Noticed you have "Allow_Immediate_RLOF_Post_CE_To_Survive_CE" as a new column header string.  We use ">CE" to indicated "post CE" in other headers.  Do you want to use that here, or ok to spell out "Post_CE"?

ilyamandel:
Thanks, @jeffriley  -- changed Allow_Immediate_RLOF_Post_CE_To_Survive_CE to Allow_Immediate_RLOF>CE_To_Survive_CE to inline. :)

PR #679, opened 2021-11-01
Minor fixes

jeffriley:
(a) fixed typo in Options.cpp for option --common-envelope-allow-immediate-RLOF-post-CE-survive (was typed common-envelope-allow-immediate-RLOF-post-CE_survive)
(b) fixed typo in Options.cpp for option --common-envelope-allow-radiative-envelope-survive (was typed common-envelope-allow-radiative-envelope-surive)

(neither of these caused problems because Boost matches only as many characters as necessary to determine the option name - would have if the names were not unique up to the typos)

Updated all pythonSubmit files - all are tested functioning.

Also restored examples/methods_paper_plots/detailed_evolution/gw151226evol.png (which I somehow managed to remove with the last PR)


ilyamandel (approved):
Thank you for catching my bugs, @jeffriley -- very embarrassing! :)

PR #692, opened 2021-11-08
Update mchirp notebook

LiekeVanSon:
Minor changes to Chirpmass distribution notebook

- Figure is now consistent with new dataset (ran with correct COMPAS version and uploaded to Zenodo)
- names and layout made more consistent with other plotting notebooks :)


jeffriley (changes requested):
Hi Lieke - I think pythonSubmit has regressed - you may want to check which version you used as your base.

ilyamandel:
@LiekeVanSon , @jeffriley -- why do we have pythonSubmit in chirpmass_distribution at all?  As far as I understand it, this pythonSubmit file will not produce any DCOs -- or at least not enough for a chirp mass distribution.  Should we delete the submit file and instead just explain to the reader where to download the hdf5 file for this example?

LiekeVanSon:
I specifically used the PythonSubmit to match the version that the data was run with. In this way, the data is most reproducible: all anyone would have to do to reproduce this set is clone the right COMPAS version, and run the files as attached to this folder. I _also_ link to the hdf5 file. 

Regarding this point, I think the other example folders also contain a PythonSubmit file, but ones that have been updated. I think including updated pythonSubmit files might not make much sense (they won't in any way reproduce the data linked because it was run with another version). 

I would argue to either keep 'old' python submit files (as used to create the dataset) attached to the folder. Ór no reproduction files whatsoever and only link to Zenodo. (I personally prefer the former :)

jeffriley:
Hmmm.  Ok, I recently updated the pythonSubmit files - some code that we have to locate the grid and/or logfile-definitions files was changed.  While I did that I thought I might as well add the new options Ilya put in the code - just to bringthem all up-to-date.  It didn't occur to me that people would use older versions of the code to run the examples, so would therefore need older versions of pythonSubmit.  In fact I think part of the motivation for updating them was because Heloise was using the latest version of the code to run the examples as part of the JOSS review.  @LiekeVanSon, you raise a good point. I don't know what the right answer is.  If we want people to actually reproduce the data (and not just use the data we've already created as @ilyamandel suggests), then we either keep the old pythonSubmit and let the user know what version of COMPAS we used to produce the data, or we tell them to use the latest and update pythonSubmit for the latest version of COMPAS (while making sure the option values are correct to reproduce the example data) if we need to.  I think I might prefer the former - then we don't have to update multiple copies of pythonSubmit every time we make a change.

jeffriley:
Do all examples actually need to have the data generated, or, as @ilyamandel suggests, can we just have the users download the data from Zenodo?  I think I agree with @LiekeVanSon though - downloading from Zenodo could be a pain for large datasets, and that might discourage users.  So, maybe best to revert to the old versions of pythonSubmit (though they should have defects repaired - we just need to make sure we don't change any of the option values), and have the users clone the correct version of COMPAS.  But we need to tell them what version, and how to clone an old version (I don't think that's anywhere in the documentation, is it?  I don't remember).

@ilyamandel, @LiekeVanSon - thoughts?

ilyamandel:
@LiekeVanSon , @jeffriley -- this is a very useful discussion!  My thoughts below:

The root folder is called examples, so I was interpreting these files as examples for how to use COMPAS, not necessarily as tools to specifically reproduce plots in the methods paper.  I imagine the set of people wanting an example is much larger than the set of people who want to reproduce an already existing plot. :)

Also, it's rather awkward to have to pull and compile one version of the C++ code but a different version of the python scripts.  I'd rather accept that things didn't work perfectly in the past (including in the release branch), but will work self-consistently now.

I do agree that needing to update pythonSubmit.py in lots of places is a pain.  But, at the moment, we don't have a clean alternative.  Though for large data sets, I think it's fine to just point people to zenodo results.

LiekeVanSon:
Just re-opening this discussion: sorry if I missed this, but did we reach a consensus on what to place where?
 I think it would be useful to have the updated figure notebook in the COMPAS main branch rather sooner than later!

jeffriley:
Hi @LiekeVanSon,

Our resolution to the question of multiple pythonSubmit files is here:

https://github.com/TeamCOMPAS/COMPAS/pull/720#issuecomment-992986652

I think if you create a README file as @ilyamandel  suggests in that post we can merge this.

jeffriley:
Also, you might want to look at PR #718 just in case there's any overlap there

ilyamandel:
@LiekeVanSon -- I think there's now a conflict in pythonSubmit.py because of other merges.   If you can resolve that, we can bring in your changes and just mark the latest version this will run with in a README (can add that later).

LiekeVanSon:
Thanks @jeffriley and @ilyamandel , sorry for taking so long to get back to this, but I the merge conflicts are solved and I added a README.md , so I think it should be good to go. 
But first: 🎄 Happy holidays!! 🎄

jeffriley (approved):
Looks good to me.

Happy holidays to you too @LiekeVanSon :-)

PR #697, opened 2021-11-15
Defect repair: boost::bad_get error when writing HDF5 SSE_Supernovae files

jeffriley:
(a) Fixed error introduced in v02.25.00: Added HDF5 support to GetLogStandardRecord().
    Defect introduced was omission of code for HDF5 file support if a specified property is supplied to GetLogStandardRecord(), causing a boost::bad_get error.
    The defect only affected HDF5 SSE_Supernovae files.  This fix adds the omitted code.
(b) Changed Options::PrintOptionHelp() to print help (-h/--h) to stdout instead of stderr.

### DIFF
--- a/src/Log.h
+++ b/src/Log.h
@@ -600,7 +600,7 @@ private:
 
     void ClearEntry(const int p_LogfileId) {
         if (IsValidId(p_LogfileId)) {
-            m_Logfiles[p_LogfileId].active          = false;                       // not active
+            m_Logfiles[p_LogfileId].active          = false;                        // not active
             m_Logfiles[p_LogfileId].logfiletype     = LOGFILE::NONE;
             m_Logfiles[p_LogfileId].filetype        = LOGFILETYPE::NONE;
             m_Logfiles[p_LogfileId].name            = "";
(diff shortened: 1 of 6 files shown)


src/Log.h
```diff
@@ -771,8 +771,13 @@ class Log {
 
                 if (ok && p_UseSpecifiedValue && (thisProperty == p_SpecifiedProperty)) {                                       // replace specified property?
                     value    = p_SpecifiedPropertyValue;                                                                        // yes - use value passed as parameter
-                    valueStr = boost::apply_visitor(FormatVariantValue(), value, fmtStr);                                       // format value
-                    logRecord += valueStr + delimiter;                                                                          // add value string to log record - with delimiter
+                    if (hdf5) {                                                                                             // yes - HDF5 file?
```
TomWagg: Not sure how much you care about these sorts of things but the comments here aren't aligned properly

TomWagg (approved):
Recompiled new COMPAS version and can confirm that everything runs smoothly, nice! One small comment about a comment but approved 😄

src/Log.h
jeffriley: I do care :-)  Fixed.  Good pickup - thanks

PR #707, opened 2021-11-19
Issue 704

### LINKED ISSUE #704: Timestep calculation in SSE code could result in timesteps smaller than expected
**Describe the bug**
The BaseStar function CalculateTimestep() is used to calculate the duration of the timestep for stellar evolution:

    double BaseStar::CalculateTimestep() {
    
        // the GBParams and Timescale calculations need to be done
        // before the timestep calculation - since the binary code
        // calls this functiom, the GBParams and Timescale functions
        // are called here
    
        CalculateGBParams();                                                                // calculate giant branch parameters
        CalculateTimescales();                                                              // calculate timescales
    
        double dt = ChooseTimestep(m_Age);
    
        return LimitTimestep(dt);
    }

The virtual function ChooseTimestep() returns a timestep clamped to a minimum value of NUCLEAR_MINIMUM_TIMESTEP (currently set to 0.0001 in constants.h).

The BaseStar::LimitTimestep() function was added to limit the mass change in any single timestep to 1% of the star's mass:

    /*
     * Limit timestep to 1% mass change
     *
     *
     * double LimitTimestep()
     *
     * @return                                      Timestep
     */
    double BaseStar::LimitTimestep(const double p_Dt) {
    
        double dt = p_Dt;
    
        // cap timestep to maximum of MAXIMUM_MASS_LOSS_FRACTION change in mass star.
        if (OPTIONS->UseMassLoss()) {
            double mDot     = CalculateMassLossRate();                                          // First, calculate mass loss rate
            double massLoss = CalculateMassLoss_Static(m_Mass, mDot, dt);                       // Next, calculate mass loss - limited to (mass * MAXIMUM_MASS_LOSS_FRACTION)
    
            if (utils::Compare(massLoss, 0.0) > 0) {                                            // No change if no mass loss
                double dtWind = massLoss / (mDot * 1.0E6);                                      // Calculate timestep to match (possibly limited) mass loss
                       dt     = min(dt, dtWind);                                                // choose dt
            }
        }
    
        return dt;
    }

The issue is that LimitTimestep() is called after ChooseTimestep(), so the timestep returned by CalculateTimestep() could be lower than NUCLEAR_MINIMUM_TIMESTEP.

Is that correct behaviour, or should we clamp the value returned by BaseStar::CalculateTimestep() to a minimum NUCLEAR_MINIMUM_TIMESTEP by changing the return statement thus:

    return std::max(LimitTimestep(dt), NUCLEAR_MINIMUM_TIMESTEP);

And if we make that change, should we remove the clamps in the ChooseTimestep() functions, or leave them there for safety?  The only place in the code ChooseTimestep() is called is BaseStar::CalculateTimestep(), but I don't know that we can guarantee that will always be the case, so maybe leaving the clamps in ChooseTimestep() is prudent, though redundant as the code now stands.

**Label the issue**

**To Reproduce**
Run COMPAS in SSE mode and check timestep durations

**Expected behavior**
The SSE timestep calculation should be consistent with expectations (i.e if we are expecting the timestep duration to be clamped to some minimum value, it should be).

**Screenshots**

 - OS: Ubuntu 20.04
 - COMPAS v02.25.09

**Additional context**

jeffriley:
Defect repairs:
(a) clamp timestep returned in BaseStar::CalculateTimestep() to NUCLEAR_MINIMUM_TIMESTEP
(b) change NUCLEAR_MINIMUM_TIMESTEP to 1 year (from 100 years) in constants.h

Close issue #704 when merged.

I chose to leave the NUCLEAR_MINIMUM_TIMESTEP clamps in the ChooseTimestep() functions - just in case people decide to call them in a future change to the code.

Example detailed plot reproduced as test - all good.


ilyamandel (commented):
@jeffriley : Will this produce similar SSE and BSE tracks -- and do those tracks appear converged?

jeffriley:
> Will this produce similar SSE and BSE tracks -- and do those tracks appear converged?

I have not run any tests to check that.

All the first part does is clamp the timestep returned by BaseStar::CalculateTimestep() to NUCLEAR_MINIMUM_TIMESTEP.  That was, I believe, an omission (BaseStar::CalculateTimestep() could (potentially) return a timestep smaller than NUCLEAR_MINIMUM_TIMESTEP if there was mass loss).  That will have no effect on the BSE timestep - because the BSE code clamps its timestep to a minimum of NUCLEAR_MINIMUM_TIMESTEP anyway.

The second part changes NUCLEAR_MINIMUM_TIMESTEP to 1 year (from 100 years).  That could affect both SSE and BSE evolution, depending upon the value of the timestep calculated - if it is less than 100 years, it now won't be clamped to 100 years, but if it drops below 1 year, it will be clamped there.  The change should affect SSE and BSE in the same way.

These changes could make the BSE tracks look more similar to the SSE tracks if the calculated timestep drops below 100 years.  But the radial change check is unique to the SSE code, so if the radial change is greater than the threshold, the SSE track will look different from the BSE track - these changes don't affect that.

ilyamandel:
Right, it's just that 1 yr was a guess as to what would be sufficient for convergence, not a guarantee.  Anyway, I'll approve, pull, and experiment. :-)

ilyamandel (approved):

PR #708, opened 2021-11-23
Output parameter name changes

reinhold-willcox:
Simple parameter name cleanup.

Changed RLOF(X)>MT to just RLOF(X) in the BSE_RLOF output. This is much clearer, since it refers to the state immediately during the MT, not after it.


jeffriley (commented):
Update online docs?

https://compas.readthedocs.io/en/latest/pages/User%20guide/COMPAS%20output/standard-logfiles-record-specification-binary.html#binary-props-r

jeffriley:
Also, you might want to expand the comment in changelog to indicate that it is the header string in the BSE_RLOF file that has changed - just so when someone runs an old script against a file created by this version of COMPAS, and the script fails (because the header has changed), and they look in changelog the change is clearly described.

ilyamandel (changes requested):
Actually, I am not sure we should do this.  As we get a larger user base, people are developing their own tools for processing outputs.  Every change to the output "column names" runs the risk of breaking those processing codes.  I think we should err on the side of caution in  output updates (make updates only if something is really wrong).  Or at least package such updates so that we make multiple changes at once, rather than piecemeal (users can probably stomach updating processing scripts once a year, but not on a continuing basis).

reinhold-willcox:
@ilyamandel I agree with taking caution on regularly updating parameter names. I'll make this a draft PR, and we can incrementally add these small output changes until we hit some critical mass and decide to pull it in. Does that suit?

jeffriley:
> changed references to 'Event' in constants.h to 'Bool' for clarity

The datatype header string for booleans will be printed as 'BOOL'.  What you are changing is the "units" header string - I don't think that's a useful change - I don't think listing "Bool" twice in the headers strings for a column adds anything.  For most boolean columns we had either "Event" or "State" as the units string to indicate the value doesn't have units, but instead the boolean value indicates either an event happened, or a state exists.  If those don't apply to the columns you're changing, can you think of another word to help describe what the value actually means?

reinhold-willcox:
How do I get the datatype header string in the output? I've only seen the 'units' col, which as you say contains only 'Event' or 'State'. I'm happy to revert back, I'm only changing things for clarity in the postprocessing, but I didn't realize bool was already included as an attribute in the output.

jeffriley:
> How do I get the datatype header string in the output?

For .csv, .tsv, and .txt files, the datatype string is printed just above the units (between the header string (column name) and the units strings.

For hdf5 files, take a look at h5view.py, particularly printHeaders() and getDataType()

reinhold-willcox:
Closing as outdated

PR #712, opened 2021-11-24
space out van den heuval event letters

reinhold-willcox:
added function to space out event letters when they would overlap


themikelau:
@reinhold-willcox I've added a few more tweaks. Feel free to merge if you think they're good.

themikelau (approved):

Issue #630, opened 2021-10-01
Plotting / diagrams


reinhold-willcox:
General place for discussion on improvements to the Compas output plots, as we prepare for generalized van den heuval output plots.

As a first TODO: we should add metallicity to the output plots like the one we used in the methods paper.

ilyamandel:
Implemented metallicity printing in #717.  Closing -- specific issues with plotting should be raised separately.

PR #717, opened 2021-11-25
Fix issue #630

ilyamandel:


themikelau (approved):
The updated script works and solves #630

PR #719, opened 2021-11-30
Resolve sudden mass drop at end of HG

ilyamandel:
- only decrease effective initial mass for HG and HeHG stars on mass loss when this decrease would not drive an unphysical decrease in the core mass
- change mass comparisons (e.g., mass vs. He flash mass threshold) to compare effective initial mass rather than current mass
- minor code and comment cleanup


jeffriley (approved):
Fixed a typo (line 920 HG.cpp) (and then made a type in my comment, but fortunately that won't persist beyond this PR :-)

Approved

ilyamandel:
Thank you for catching the typo, @jeffriley !

Issue #715, opened 2021-11-24
Refactoring pythonSubmit.py


HeloiseS:
_This issue is submitted as part of the JOSS review [#3838](https://github.com/openjournals/joss-reviews/issues/3838)_

When it comes to the pythonSumbit.py file refactor you can go one of two routes - the one you choose will depend on whether you *ever* use the pythonProgramOption class anywhere else. 

If you do not use that class anywhere else, there it is not a use case for a class. Remove the class, just have the  giant list of options listed as it is, your comments are detailed already, and maybe instead of having two separate lists for the commands and the variable names, you can put those in a dictionary directly. You do not need a whole bunch of functions and methods (e.g. the ones whose sole purpose is to make and return the list) if they are only called one time in the __main__ portion of the file. 

If you DO want this to be a class to be imported in other scripts and feel OO is necessary, then the main thing is to remove the methods without purpose (e.g. creating lists doesn't need to be done in a method) and have an __init__ method. importantly this method would then be where you parse the configuration for COMPAS so that you don't have to go in the class everytime you want a different simulation. **if you are going to keep it a class, your user shouldn't have to modify the class itself to change the COMPAS options**, that is behaviour you expect from a script, not a class. You can put your options in a setup file in a format like yaml. This makes the actual code in pythonSubmit much shorter. Here is a quick refactor i did early on (note github won't let me have .py or .yaml so  you'll need to amend the extension after downloading) that you can use as inspiration or to build on. Note I have not tested this refactor extensively.

[compas_config.txt](https://github.com/TeamCOMPAS/COMPAS/files/7598005/compas_config.txt)
[pythonSubmitDemo.txt](https://github.com/TeamCOMPAS/COMPAS/files/7598006/pythonSubmitDemo.txt)

reinhold-willcox:
If I can add my two cents to this, since we need a major refactor of the pythonSubmit code anyway, I propose we work with the stroopwafel folks to put together a single script where users can specify all COMPAS parameters, as well as whether to use stroopwafel and/or HPC (and the related parameters there), instead of having many disparate scripts that will sometimes need to be combined in some convoluted way.

HeloiseS:
yes please - was there a stroopwafel tutorial i missed? i haven't used that yet

reinhold-willcox:
[...]

Issue #680, opened 2021-11-03
DOCKER - Logfile record specifications error &  command typos in the example pythonSubmit files


HeloiseS:
_This issue is submitted as part of the JOSS review [#3838](https://github.com/openjournals/joss-reviews/issues/3838)_

### The typo
so I was testing running compas in a docker image (it all installed beautifully, very easy) and i got the following error [Abridged] `Commandline Options error: unrecognised option '--common-envelope-allow-radiative-envelope-surive'`

I used one of your original pythonSubmit files to run this (the one in examples/\*/chirpmass_distribution), and when i go look in the file it does seem that the commands and the python variables associated with `--common-envelope-allow-radiative-envelope-survive` all have that typo in them (surive instead of survive). Same for the pythonSubmitDemo from looking at the github. 

### The log file
After correcting the typo in the command I ran it again and got the following [Abridged]
```
Logfile record specifications error
File does not exist: /app/COMPAS/COMPAS_Output_Definitions.txt
```
I couldn't figure out why it wasn't happy, the COMPAS_Output_Definitions.txt is in the directory where i got the pythonSubmit file :shrug:  - I'll need your help to fix this one :)

### The Command I ran
```
sudo docker run                                                      
--rm                                                       
-it                                                         
-v $COMPAS_ROOT_DIR/compas-logs:/app/COMPAS/logs                      
-v $COMPAS_ROOT_DIR/examples/methods_paper_plots/chirpmass_distribution/pythonSubmit.py:/app/starts/pythonSubmit.py       
-e COMPAS_EXECUTABLE_PATH=/app/COMPAS/bin/COMPAS            
-e COMPAS_LOGS_OUTPUT_DIR_PATH=/app/COMPAS/logs             
teamcompas/compas                                           
python3 /app/starts/pythonSubmit.py
```

### Full Output Bug 1 
```
python_version = 3
[...]

PR #720, opened 2021-12-01
Joss python submit

avigna:
This addressed the JOSS review of the pythonSubmit.py file as described in #715 .

The main changes are:
- Renamed `pythonSubmit.py` to `runSubmit.py`, and cleaned it's contents.
- Created `compasConfigDefault.yaml`, a file that is read by runSubmit.py and contains the user-defined choices for COMPAS.

My main comments are:
- I ran the `Grid_demo.txt` to verify nothing broke. It seems to be working fine.
- @jeffriley can you check if the way I deal with the grids works out for the docker runs? (lines 56-63 from `runSubmit.py`). I think it is fine, but would be good to double check.
- Heloise suggests to rename the pythonProgramOptions class to something else (line 14 from `runSubmit.py`). No strong feelings, feel free to suggest so I can do the renaming (if needed).
- I haven't tested the stroopwaffel compatibility, nor edited `stroopwafelInterface.py`. Given that there were no major changes, it should be pretty straight forward to adapt by renaming a few variables. Once this PR is approved, I will open an issue to test this with Floor, Tom, or whomever is using stroopwaffel.
- Once the PR is approved, I will open an enhancement to automatically generate the `compasConfigDefault.yaml` from the COMPAS src defaults.


ilyamandel (changes requested):
Thank you very much for this work, @avigna !

A quick scan through the code looks OK, modulo removing comments like "# TODO: rename? | HFS oct 31 2021", but @jeffriley may have further comments.

Main request: please make sure that all documentation is updated as part of the same PR.

jeffriley:
> @jeffriley can you check if the way I deal with the grids works out for the docker runs? (lines 56-63 from runSubmit.py). I think it is fine, but would be good to double check.

Sorry for the delay - looks good to me.  Thanks @avigna!

avigna:
Thanks both. Will remove "# TODO: rename? | HFS oct 31 2021", update the documentation, and get back to you asap.

reinhold-willcox:
Is there a benefit to having a randomSeed.txt file over supplying the random seeds to a gridfile?

jeffriley:
> Is there a benefit to having a randomSeed.txt file over supplying the random seeds to a gridfile?

How would it be used?  Would you expect something (pythonSubmit? stroopwafel?) to populate the grid file with the seeds from the text file, or are you thinking the COMPAS executable would read both a random seeds text file and a grid file?  I don't see any benefit in the latter, and I can't immediately think of a use case for the former.  Are you thinking of something?

reinhold-willcox:
No, I don't see a use case for it. I think the randomSeed.txt file is a relic from older compas versions, it seems to me that if you only wanted to supply random seeds, you could do so just as easily in a grid, so I would propose getting rid of it. But I haven't really used it, so maybe someone else knows better.

avigna:
> Main request: please make sure that all documentation is updated as part of the same PR.
Starting going through this, and realized this is potentially problematic. The less problematic bit is that the we should change `pythonSubmit.py` to `runSubmit.py` everywhere, which its somewhat straight forward, but could lead to missing things our here and there. I could either:
1. Bite the bullet and do it, or
2. Revert the name to pythonSubmit.py and avoid this issue.
In any case, I would need to add some explanations about the `compasConfigDefault.yaml` here and there, but that should be straight forward.

A more concerning thing is that we currently have many pythonSubmits all over the place, with different options. E.g., in /COMPAS/examples/methods_paper_plots :

methods_paper_plots alejandro$ ls ./*/
./chirpmass_distribution/:
COMPAS_Output_Definitions.txt	pythonSubmit.py
chirpmass_dist.ipynb		stroopwafel_interface.py

./detailed_evolution/:
Grid_demo.txt			gw151226evol.png		gw151226evol_1.png
detailed_evol_plotter.py	gw151226evol_1.eps		pythonSubmitDemo.py

./fig_5_HR_diagram/:
README.md		hrd.pdf			pythonSubmit.py
create_fig_5_grid.py	make_fig_5.ipynb

./fig_6_max_R/:
README.md		grid.txt		maximum_radius.pdf
create_fig_6_grid.py	make_fig_6.ipynb	pythonSubmit.py

./fig_7_MS_lifetime/:
README.md			main_sequence_lifetime.pdf	make_fig_7.ipynb

./fig_8_initial_core_final_mass_relations/:
MM20_grid.txt				make_fig_8.ipynb
README.md				pythonSubmitDefaults.py
create_fig_8_grids.py			pythonSubmitFryerRapid.py
grid.txt				pythonSubmitMandelMueller.py
initial_core_final_mass_relations.pdf	rapid_grid.txt
logfile_defs.txt

Which in principle become outdated by updating to the new python+yaml configuration. In principle, one could update each one of them, but that would require to (manually?) update and check each one of them, and whichever other pythonSubmits are in the repo.

I am uncertain what is the best way to proceed with this, any thoughts or suggestions?

The "easy way out" is to keep the `pythonSubmit.py` name and add some comments about the `compasConfigDefault.yaml` file, but unfortunately, that doesn't solve that most scripts become outdated with this PR.

jeffriley:
> A more concerning thing is that we currently have many pythonSubmits all over the place, with different options.

Indeed.  Discussed in Issue #680 and PR #692.  I was hoping we could address it with this PR, but I don't know what the solution is.  It may be, as I think @ilyamandel suggested, we just provide the "extras" as part of the examples, but that they are not living documents - they worked at the time to produce the example plots or whatever, but are not guaranteed to continue working in perpetuity (at least I think that's what Ilya was suggesting).

ilyamandel:
1. I like having runSubmit.py with a different name to avoid confusion.  Note that, since the C++ code is unchanged, a  user could still run with pythonSubmit.py if they still wanted to (or even through a command line, as @jeffriley does).
2. I do not think we need to update all examples.  What I propose is to include a one-line README in each example stating which code version it was last tested to work with.  That way, if a user really wants to run that example, they can just check out the relevant code version and things should run smoothly.  And if one of us ever decides to update an example (or even happens to rerun it and confirms that it works with the current version), we'd just need to update the README file to point to the current version.

avigna:
I like that plan @ilyamandel . Will proceed that way.

avigna:
I have done the following:

- Removed: "# TODO: rename? | HFS oct 31 2021"
- Adapted the documentation, particularly getting_started.md and docker.md. In the former, I added a whole new section `2.1 Running COMPAS out-of-the-box` in which I briefly comment on the `runSubmit.py`and `compasConfigDefault.yaml` combo superseding the `pythonSubmit.py`.
- Finally, I added the following disclaimer in all the `pyhtonSubmit.py` from $COMPAS_ROOT_DIR/examples/methods_paper_plots/: 
```
#### DISCLAIMER: This script uses the `pythonSubmit.py` format
#### that has been replaced by the `runSubmit.py` and
#### `compasConfigDefault.yaml` combo as of v02.25.10.
#### The `pythonSubmit.py` format will eventually become deprecated.
```

Let me know if I missed anything.

jeffriley:
@avigna The changes look good.  What about the online documentation?

https://compashq.slack.com/archives/CMCUK1G69/p1635754297004900

ilyamandel:
Right, @avigna , I was thinking of updates to documentation at https://compas.science/docs/ -- though it's good that you got the ones at github!

That said, I'll accept and merge this to prevent the PR from getting too large and creating conflicts -- and create another issue for you to address for the online docs.

ilyamandel (approved):

PR #731, opened 2021-12-16
Nan radius fix

### LINKED ISSUE #711: NaN radius
**Describe the bug**
(copied from slack) The attached system goes through 2 MT episodes, the second of which is a merger (though not very clearly recorded). The first MT is the weird one though. It's stable, very conservative MT, from CHeB to CHeB. But after the first MT, the radius of the companion becomes a NaN. Then, by the time of the second MT, the secondary radius is back to a normal value! Sometimes there will be NaN's in the event of a merger (as is the case for the second MT, where the postMT semi-major axis is a NaN), but it doesn't make sense why the radius would be a NaN when there is no merger, then convert back.

image

**Label the issue**

**To Reproduce**
The output and detailed output are included below (it's just a single system, so the files aren't too big), which also give you the input parameters if you want to rerun. 
*Note:* github is annoying about file extensions, so these have both been "converted" to .txt files, just mv them back to h5 and they'll run.

 - COMPAS v02.25.9
[COMPAS_Output.h5.txt](https://github.com/TeamCOMPAS/COMPAS/files/7593429/COMPAS_Output.h5.txt)
[BSE_Detailed_Output_0.h5.txt](https://github.com/TeamCOMPAS/COMPAS/files/7593432/BSE_Detailed_Output_0.h5.txt)

reinhold-willcox:
Changed calls to PPOW(base, exponent) where exponent was something like 1/2 or 1/3 to sqrt and cbrt. In the current functionality, the 1/3 power was unable to handle negative arguments, even though mathematically this is fine. Cbrt has no issue with negative arguments.


jeffriley (changes requested):
Looks good.  For consistency we should probably used 'std::cbrt' and 'std::sqrt' always (or 'cbrt' and 'sqrt' always), but that's a small point.

Changelog?

EDIT: I just noticed you missed a semi-colon in WhiteDwarfs.cpp, and the compile fails.

And our minimum test is the detailed plot in the examples.  Can you run that and check your output against the standard?

reinhold-willcox:
Woops, that's what I get for trying to code before bed. Yup, on it.

jeffriley (approved):
Did you produce the detailed plot?  I'll approve and merge, but I think you should produce the plot just to be sure.

reinhold-willcox:
Where is the one to compare to?

reinhold-willcox:
Figure_1

Looks good to me!

PR #732, opened 2021-12-17
Option to specify initial stellar type

reinhold-willcox:

This is still a work in progress. Remaining TODOs:
- Add in a check that the supplied type is one of the allowed types listed above
- Fix the changelog
- Remove temporarily commented out lines
- Update the docs
- Test more robustly (the code compiles and runs on a few test cases)


reinhold-willcox:
@jeffriley I made this a draft PR due to the remaining TODOs, but could you take a look and see if the syntax / functionality looks good or if I inadvertently broke something subtle?

jeffriley:
@reinhold-willcox I haven't looked at functionality yet, but if you are doing it the same way we did when we tested this a while ago, it should be ok.  I'll look at functionality later, but for now just a comment about the way you have implemented the initial stellar type options.

Stellar types as numbers - the Hurley types represented as integers.  I don't think this is documented very well, if at all, so I'll put something here then transfer it to the documentation later (to the Developer Guide).

We don't refer to stellar types as integers in the code - we refer to them using symbolic names.  That was a design decision we took early on so that we didn't rely on:

   (a) the Hurley types being in any particular order
   (b) an ordered list of Hurley types (by integer number) remaining the same (i.e. we might want to insert a new type, or remove an existing type - maybe unlikely, but we have the flexibility to do that)
   (c) being able to iterate over a range of Hurley types by integer number

In short, though the Hurley types in the code have associated integers that are currently the same as the integers as listed in Hurley+2000, they don't need to be, and we don't guarantee that they will be.

We do print the integer values to the log files, but that was really just to be compatible with the legacy code.  There's no reason that we shouldn't print symbolic names to the logfiles instead of integers - Python scripts can just as easily check "CHeB" instead of "4".  But for now we print integers, and we (should if we don't) document the mapping we use (integer to stellar type).

The stellar type symbolic names are in enum class STELLAR_TYPE in constants.h.  Strings associated with the stellar types are in the STELLAR_TYPE_LABEL map following the symbolic names in constants.h.  

There are also a number of initializer_lists there that you can use as sets of stellar types (i.e. you can check in the code if a stellar type belongs to any of those lists).  You could create a new initializer_list for the stellar types that you want to allow to be set as initial stellar types.

In the Options code you have declared:

            int             m_InitialStellarTypeNum;           // Initial stellar type number of single star (SSE)
            int             m_InitialStellarTypeNum1;          // Initial stellar type number of primary (BSE)
            int             m_InitialStellarTypeNum2;          // Initial stellar type number of secondary (BSE)
            STELLAR_TYPE    m_InitialStellarType;              // Initial stellar type of single star (SSE)
            STELLAR_TYPE    m_InitialStellarType1;             // Initial stellar type of primary (BSE)
            STELLAR_TYPE    m_InitialStellarType2;             // Initial stellar type of secondary (BSE)

You don't need the first three (the integers).  We shouldn't have the users specify stellar type with an integer (for the reasons above).

And you have initialised them as:

    m_InitialStellarTypeNum     = 1;    // Default Main Sequence
    m_InitialStellarTypeNum1    = 1;    // Default Main Sequence
    m_InitialStellarTypeNum2    = 1;    // Default Main Sequence

(Btw, is there a reason you have initialised them to MS_GT_07 rather than MS_LTE_07?  If so, it'd be good to document that reason.  If you want MS, shouldn't the initial mass decide which MS type?)

You should instead declare (say) m_InitialStellarType in Options.h as an ENUM_OPT with type STELLAR_TYPE:

    ENUM_OPT<STELLAR_TYPE>    m_InitialStellarType;

And then initialise it in Options.cpp as:

    m_InitialStellarType.type       = STELLAR_TYPE::MS_GT_07;
    m_InitialStellarType.typeString = STELLAR_TYPE_LABEL.at(m_InitialStellarType.type);


When you set the option value, here:

    (
        "initial-stellar-type",                                            
        po::value<int>(&p_Options->m_InitialStellarTypeNum)->default_value(p_Options->m_InitialStellarTypeNum),                                                                          
        ("Initial stellar type (Hurley type) for the star (SSE) (default = " + std::to_string(p_Options->m_InitialStellarTypeNum) + ")").c_str()
    )

you should do this instead:

    (
        "initial-stellar-type",                                            
        po::value<std::string>(&p_Options->m_InitialStellarType.typeString)->default_value(p_Options->m_InitialStellarType.typeString),                                                                          
        ("Initial stellar type (Hurley type) for the star (SSE) (default = " + m_InitialStellarType.typeString + ")").c_str()
    )

That would require users to type in the strings shown in the STELLAR_TYPE_LABEL map as option values.  If you think that is a little onerous ("Thermally_Pulsing_Asymptotic_Giant_Branch" is a lot to type...), we could create a new map (don't change the existing one - it is used elsewhere in the code to print pretty names for the stellar type), that just uses short names like "HG", "TPAGB", "HeWD" etc.  So the users could just type the option as "--initial-stellar-type TPAGB" etc.  I think that is more intuitive than using numbers, and it protects against future changes in the code that might result in remapping the integers that are associated with stellar types.

reinhold-willcox:
Good call, thanks Jeff! The hurley numbers have been swapped out for stellar types, so now users can submit any of the tau=0 stellar types as string names. If an incorrect name is submitted, the following is printed:

```Grid File Options error: Initial stellar type (--initial-stellar-type) must be one of { MS HeMS HeWD COWD ONeWD NS BH }```

Hopefully this is clear enough for users. I still need to update the docs so don't merge this in just yet, but let me know if the code is suitable or needs more tweaking.

reinhold-willcox:
All fixed up, @jeffriley let me know if you see any remaining TODOs.

jeffriley (commented):
@reinhold-willcox  At first glance the functionality looks ok.  I haven't tested it - have you?  I will run a few tests in the next few days.

One thing: you have used bitwise operators for booleans in a number of places (e.g. '|' rather than '||').  The boolean operators ('||', '&&', etc.) are guaranteed to return 'true' or 'false' and nothing else.  The bitwise operators ('|', '&') return numbers, and while they may well be '1' or '0' if working with booleans, they are not guaranteed to be if someone inadvertently changes something.

Booleans in C++ are guaranteed to be 'true' or 'false' (which convert to '1' or '0'), but not all languages guarantee that - using bitwise operators in place of boolean operators is not a good habit to get into.  Where the domain of numbers the operators are working on is '1' and '0', the boolean and bitwise operators will return the same results, but only for that domain.  But there is another difference in C++ that matters: bitwise operators do not short-circuit (because they don't return booleans).

Short-circuit evaluation is where the clauses in an 'if' statement are evaluated left to right, but as soon as the result is known evaluation stops (i.e. is short-circuited).  This is important for the following construct:

    if (i < arrSize && arr[i] == 0) { ... }

The 'if' statement above will short-circuit, and not evaluate the right-hand clause, if i >= arrSize, and so prevent a bounds violation (i.e. addressing outside the bounds of an array), whereas

    if (i < arrSize & arr[i] == 0) { ... }

will not short-circuit, so both clauses will always be evaluated, and bad things could happen as a result.

Is there a reason you use bitwise operators?  While they might work in the specific places you've used them, it's not great programming practice.  We should follow the principle of least surprise - code that does unusual things is harder to read and understand - and code is read far more often than it is written.  It is way better to be more verbose than less - fewer characters doesn't help the compiler, and a little bit of verbosity can help make the code more understandable (things like adding parentheses for clarity for the reader doesn't hurt the compiler, nor does adding white-space around operators etc.  Not putting things like that in doesn't make the compile (or execution) any faster, but putting them in helps other readers understand the code more easily).  We should always code for future readers - not for the compiler, even if it looks inefficient (most compilers are optmising compilers these days, so what you might think are small inefficiencies will be optimised away anyway (though an optimising compiler can't help an essentially bad algorithm...)).

So, unless there is a really good reason for using bitwise operators where you have (and I can't think of one - but that doesn't mean there isn't one), could you change them to boolean operators please?

reinhold-willcox:
Thanks Jeff, wasn't aware of the distinction. Should be all fixed now.

jeffriley:
Code review looks ok, but I haven't done any testing.  @reinhold-willcox what testing have you done?

- can you reproduce the detailed example plot?
- does a run NOT using the new options produce the same results as a run from the previous version?
- does a run using the new options produce the results you expect?
- have you checked that CHE still works as expected, both using and not using the new options (because your new code touches the CHE code)?
- have you checked corner cases?  What happens, for example, if you specify initial stellar type = BH for both the primary and the secondary?  Are the results expected?  What about combos that you mightn't expect but can be specified?

reinhold-willcox:
Thanks @jeffriley for the reminder! 

The detailed example plot is identical to what is currently on dev, but different to what is in the methods paper. I noted this in the devel_bugs slack channel. I'll look into why that might be the case, but since the plots are identical between dev as it is currently and my changes, I think that means this PR is safe. 

The runs with the new options work as expected. I checked all pair-wise combos and things look good. Notably, starting two COs together (of any kind) leads to an immediate end of evolution, as expected (these are trivial cases but good that they work). Also, initializing stars that don't belong in a certain mass range leads to them immediately switching types, which is also good. But generally, I assume someone testing with an alternative stellar type will also know to set the mass explicitly. 

How do I test the CHE code? I can run with CHE turned on, but are there other specific settings I should implement?

reinhold-willcox:
Regarding the detailed plot: thanks to @themikelau for pointing me to the new default settings needed to reproduce the figure. I've updated the pythonSubmitDemo.py file with these new settings. The figure is now successfully reproduced.

jeffriley:
> How do I test the CHE code?

Did you do a run of a reasonably large population with your options turned OFF and check against the previous version of COMPAS?  That should produce identical results.  CHE is enabled by default, so if that test passed, you're good (provided the population was large enough that some CH stars were produced - check 'Stallar_Type(1)@ZAMS' and 'Stellar_Type(2)@ZAMS' for type = 16)

reinhold-willcox:
@jeffriley Thanks for your suggestions. Following the discovery of a bug where only the initial stellar type was being set in the first timestep (and not the corresponding new radius or luminosity), I've cleaned this up so that those values are set in the FastForward() function you suggested, following the stellar type change. 

For others who read this, we are effectively initializing the star as we always do (with the usual parameters that apply to MS stars), then immediately fast-forwarding its evolution to the later stellar type, and updating the radius and luminosity to the appropriate new values. This also sets the framework for future implementation of initializing non-trivial stellar types, i.e those which require tau (or other) values to be set explicitly. 

Much of the point of this was to make sure that the stellar values were correct from the first timestep. Previously, these would be updated after the first timestep to the correct values, but it prevented, for example, the creation of HeMS+NS binaries which were so tight that they did not fit within the (anomalous) MS radii of the stars. Now, these tight binaries can be initialized correctly.

Additionally, the demo figure is correctly reproduced, with some tweaks to the pythonSubmit to correct a bug that previously snuck in, and in a run of 4000 binaries (using the same grid file & random-seeds), this version agrees with dev in all outputs. I verified that this is a large enough sample to produce CHE stars, so these changes do not impact CH evolution. 

I believe I have addressed all the concerns listed above, but let me know if I am missing anything else.

ilyamandel:
@jeffriley -- since you've already reviewed this, are you happy to sign off?

jeffriley:
> @jeffriley -- since you've already reviewed this, are you happy to sign off?

Not yet.  I actually started testing it this morning.  I have some questions and comments that I'll put here after I've completed my testing.

jeffriley (changes requested):
I have run some rudimentary tests and, to a good approximation, I think the new functionality produces the expected results.  I assume @reinhold-willcox ran more extensive tests.  The functionality is pretty-much as expected, but I have a few comments and questions.


(1) Star::SwitchTo()

What's the rationale for changing

    if (p_StellarType != m_Star->StellarType()) {

to

    if ((p_StellarType != m_Star->StellarType()) || p_SetInitialType) {


The code change not only changes the functionality of Star::SwitchTo(), it also changes the meaning of the parameter p_SetInitialType.

The existing functionality switches stellar type iff the current stellar type is not already the required stellar type, so the function can be called frequently (e.g. at every timestep) but will only switch stellar type if it is actually necessary.  Switching stellar type requires the creation of a new object, copying the attributes of the existing object to the new object, and the destruction of the old object - there's a lot of potential for negative impact to performance if we do that a lot.  The existing meaning of the p_SetInitialType parameter is to set the m_InitialStellarType class variable.  The proposed code change changes the meaning of that parameter to effect the stellar type switch regardless of the current stellar type.  That means, with the proposed code change, if we call Star::SwitchTo() inside a loop (e.g. at every timestep) with p_SetInitialType set TRUE, we will switch stellar types at every execution of the loop whether we really need to or not - that could have a significant impact on performance.  Also, every time we actually switch stellar types we potentially write an entry to the SSE/BSE switch log file - we don't need to do that if we're not really switching stellar type (i.e. the new stellar type is the current stellar type).  If we accept this code change then at a minimum we need to document the changed functionality of the function, and the changed meaning of the parameter - future maintainers shouldn't need to read the details of the code to discern the functionality of the function and/or the meaning of the parameters.

Have you checked what impact this change will have on the existing calls to Star::SwitchTo() in the code?  Those calls are expecting the switch to only occur if the current stellar type is different from the requested stellar type, and expect that the p_SetInitialType only causes the m_InitialStellarType class variable to be set, and even then only if the switch actually occurs.  The proposed functionality is very different.

Granted, nobody should be setting the initial stellar type in a loop anyway (really should only be done once), but the change makes me wonder what it is you're actually trying to achieve.  Is the real aim of this code change to just update the m_InitialStellarType class variable?  If so, then you could call BaseStar::SetInitialType() for the star in question (see the call to m_Star->SetInitialType(p_StellarType) in Star::SwitchTo() - you can expose the BaseStar::SetInitialType() function in Star.h).


(2) comment in EvolveSingleStars() in main.cpp

    // the initial of the star is supplied - this is to allow a single star to initialise its own 
    // stellar type (rather than sample it).  Here we use the stellar type supplied by the user 
    // via the program options or, if no stellar type was supplied by the user, default to MS.

I think "stellar type " is missing from the first line.  Also, we don't sample stellar types - I think that needs rewording.


(3) Question for my own edification: do we not need to calculate the temperature in all variants of FastForward()?


(4) What's the rationale for removing MZAMS and MASS_0 from default BSE_Detailed_Output record?  And RZAMS and MASS_0 from the default SSE_Detailed_Output record?  Is that because they're not required, or because they might show inconsistencies wrt the new "INITIAL_*" variables?  Is it possible that if a user adds those columns manually (via the logfile-definitions file) they would show inconsistencies with the new "INITIAL_*" properties?

This change brings up another point - changing default logfile records could break existing scripts.  That's not to say we shouldn't do it, but we probably need to introduce a "What's new in this release" type document that gives users a place to quickly check what's changed and whether the changes will impact them and any of their existing scripts.  Not something for this PR, but something we should put on the todo list.  If the changelog (changelog.h) was detailed enough we could just publish that, but often the updates to changelog.h are a little cryptic/obscure and not sufficiently detailed to use as a "What's new in this release" document.


(5) New properties (e.g. INITIAL_RADIUS) have not been added to the documentation


(6) We need a section in the documentation explaining what's happening when these new options are used.  We need to explain the concept - what it is we're trying to achieve by setting the "initial" stellar types.  We need to carefully explain that the initial-stellar-type is not the ZAMS stellar type - rather it is the stellar type that we fast forward evolution to, then resume the normal pace of evolution.  If we are not going to change the nomenclature (see below), we need to be very clear about what the new options are and what they do, what the new properties (for printing) are, what the headers strings are, and what they mean.  And I think we need to make sure we haven't changed the meaning of any existing properties/column headers.  And think carefully about, and describe, any confusion that could arise...


General comment on nomenclature:  I don't like it, and I think we need to clean it up.  I think it has a high potential to confuse users and, at least as importantly, maintainers of the code.  I think the new option names are misleading, and I think using "initial" as the prefix for option names and code variable names is likely to cause confusion with "ZAMS" variables.

What about we use something like

    EVOLVE_[PRIMARY/SECONDARY]_FROM instead of INITIAL-STELLAR-TYPE-[1/2]

e.g. instead of

    ./compas --mode bse --initial-stellar-type-1 HeMS --initial-stellar-type-2 COWD

we use

    ./compas --mode bse --evolve-primary-from HeMS --evolve-secondary-from COWD

The SSE equivalent would just be --evolve-from

I think "evolve-from" is a better description of what we are doing than "initial-stellar-type".

As for the code variable names, maybe we could use something like "phase_initial" instead of "initial", just to indicate what the variable really is.  At the same time we could probably look at renaming, and disentangling the overloading of, the mass_0 variable.

I think if we don't look carefully at the nomenclature here, especially for code variables, we'll pay for it later.

As a case in point, look at the name of @reinhold-willcox's branch: generalized_zams_stellar_type.  Reinhold knows what he is doing, yet he slipped into calling the initial stellar type the ZAMS stellar type - that's not what it is, and I know Reinhold knows that.  Reinhold knows it, and we all know it because we're involved in the coding/reviewing of these changes.  Five years down the track when we're all off doing other things, new maintainers of the code could easily slip with the current nomenclature.

@ilyamandel @SimonStevenson @avigna

reinhold-willcox:
Thanks @jeffriley for the detailed feedback! I'll respond point by point.

> (1) Star::SwitchTo()

You are correct, this was a change from my first attempt at implementing the generalized stellar types, which I did not catch when I tried to clean up that code. It has now been removed.

> (2) comment in EvolveSingleStars() in main.cpp

I agree, and the comment has been fixed.

> (3) Question for my own edification: do we not need to calculate the temperature in all variants of FastForward()?

Most of the FastForward types do not have a relevant temperature that we wish to track. I suppose WDs do in reality, but we don't have existing functionality to calculate it anyway, so I opted to ignore those. I've added in a parameter for InitialTemperature, in case that value is ever desired, but similar to InitialLuminosity, it is not currently printed to any outputs (and so also not documented, though we may want to revisit that).

> (4) What's the rationale for removing MZAMS and MASS_0 from default BSE_Detailed_Output record? And RZAMS and MASS_0 from the default SSE_Detailed_Output record? 

It does not make sense to include in the detailed output any parameters which are constant throughout the evolution. That's redundant and takes up unnecessary disk space. For MZAMS parameter specifically, that value can be retrived by looking at the first entry of the MASS parameter. If we believe that there may be an inconsistency between MZAMS and MASS[0], that would qualify as a bug, and devs can re-add in MZAMS manually to check, but they shouldn't be included for standard runs. Same goes for the other ZAMS parameters. While MASS_0 could change over the course of the evolution, I would argue that that is an internal parameter, and one that would only be needed by devs trying to debug something. 

I would further argue that other constants of evolution and less commonly used parameters, such as random_seed, metallicity, the timescales and zetas, and pulsar parameters, should also be removed by default. With the logfile_definitions functionality, these parameters are very easy to add back in as necessary, but keeping them bloats the detailed output files. 

> This change brings up another point - changing default logfile records could break existing scripts. That's not to say we shouldn't do it, but we probably need to introduce a "What's new in this release" type document that gives users a place to quickly check what's changed and whether the changes will impact them and any of their existing scripts. Not something for this PR, but something we should put on the todo list. If the changelog (changelog.h) was detailed enough we could just publish that, but often the updates to changelog.h are a little cryptic/obscure and not sufficiently detailed to use as a "What's new in this release" document.

I agree with incorporating "What's new in this release", and with improving future changelogs with this in mind. I'm not sure where this should be announced to inform users, but it would certainly also help developers who cannot always keep up with many of the changes they are not involved in (I'm sure I'm not the only one!). 

> (5) New properties (e.g. INITIAL_RADIUS) have not been added to the documentation
Added.

> (6) We need a section in the documentation explaining what's happening when these new options are used. We need to explain the concept - what it is we're trying to achieve by setting the "initial" stellar types. We need to carefully explain that the initial-stellar-type is not the ZAMS stellar type - rather it is the stellar type that we fast forward evolution to, then resume the normal pace of evolution. If we are not going to change the nomenclature (see below), we need to be very clear about what the new options are and what they do, what the new properties (for printing) are, what the headers strings are, and what they mean. And I think we need to make sure we haven't changed the meaning of any existing properties/column headers. And think carefully about, and describe, any confusion that could arise...
> 
> General comment on nomenclature: I don't like it, and I think we need to clean it up. I think it has a high potential to confuse users and, at least as importantly, maintainers of the code. I think the new option names are misleading, and I think using "initial" as the prefix for option names and code variable names is likely to cause confusion with "ZAMS" variables.
> 

To this point, I disagree somewhat on the degree of confusion caused by the nomenclature. Absolutely, ZAMS should be reserved for stars initialized on the MS - it's in the name, and most people would rightly be confused if a ZAMS parameter implicitly referred to the HeZAMS (or BHZAMS, or something). However, I believe "--initial-stellar-type" is sufficiently clear in the context of evolving stellar binaries, especially given the list of available stellar types shown in the --help string and documentation. If a user implements an alternative initial stellar type, it also seems natural to expect that Initial_Radius, Initial_Luminosity, and Initial_Temperature refer to the initial values at the beginning of the simulation for the non-MS star. I am happy to explain this more fully in the documentation if desired, and also to hear counter-arguments.

jeffriley:
> To this point, I disagree somewhat on the degree of confusion caused by the nomenclature.

How should a user who is not familiar with the code (and has no interest in reading the code) decipher the SSE_Systems_Parameters file that has a column labelled "Mass@ZAMS" and another labelled "Initial_Radius", for a star that was evolved using "initial-stellar-type HeMS"?  The Mass@ZAMS column indicates the ZAMS mass, but the Initial_Radius column indicates the radius at the start of the HeMS phase.  You don't think that's inconsistent and liable to lead to confusion?  I think a user not familiar with the implementation of the code (and maybe even some users who are somewhat familiar with the code) could mistakenly believe that both those columns indicate "the first value".  What is printed in the "Initial_Radius" column is not "the first value" (unless evolving from ZAMS) - the star had a different radius at ZAMS, but we fast-forwarded the evolution to the start of the HeMS phase and now the radius is different.  I think calling it "Initial_Radius" and printing it along with "Mass@ZAMS" is just plain confusing.

reinhold-willcox:
That's true. I would opt to remove ZAMS values from the output, and consider these parameters to be internal to the code. And then ensure that Initial values are consistent with ZAMS values whenever the initial type is MS.

jeffriley:
I evolved a single star in SSE mode with the following command:

    ./compas --mode sse -n 1 --initial-mass 16.0 --logfile-type csv

And I got the following SSE_System_Parameters file

    SEED       Mass@ZAMS Initial_Radius Metallicity@ZAMS Omega@ZAMS Stellar_Type@ZAMS Stellar_Type ...
    1644200262  1.60E+01       4.97E+00         1.42E-02   0.00E+00                 1           13

I then evolved another single star with the command:

    ./compas --mode sse -n 1 --initial-mass 16.0 --initial-stellar-type HeMS --random-seed 1644200262 --logfile-type csv

and got the following SSE_System_Parameters file

    SEED       Mass@ZAMS Initial_Radius Metallicity@ZAMS Omega@ZAMS Stellar_Type@ZAMS Stellar_Type ...
    1644200262 1.60E+01       1.257E+00         1.42E-02   0.00E+00                 7           14

(1) How should I interpret the "Initial_Radius" column?  It's the same star evolving - all I did (or all I should have done) was effectively skip printing the MS phase in the detailed output.  Isn't "Initial_Radius" really "Initial_Radius_On_Phase"?

(2) "Stellar_Type@ZAMS" is wrong

(3) The final stellar type is different - I get that we will have minor difference in the evolution, but should the final stellar type be different?

jeffriley:
>That's true. I would opt to remove ZAMS values from the output, and consider these parameters to be internal to the code. And
>then ensure that Initial values are consistent with ZAMS values whenever the initial type is MS.

The you should do a similar thing with code variables - if ZAMS vs INITIAL is confusing in the output files then it will be confusing as variable names for future maintainers of the code.

jeffriley:
Actually, the commands I used above

    ./compas --mode sse -n 1 --initial-mass 16.0 --logfile-type csv
    ./compas --mode sse -n 1 --initial-mass 16.0 --initial-stellar-type HeMS --random-seed 1644200262 --logfile-type csv

raise another question:

Should the options --initial-mass* (e.g. --initial-mass, --initial-mass-max, --initial-mass-function, etc.) not be consistent in meaning with --initial-stellar-type (and also the "initial_" columns headers)?  The current "--initial*" options take "initial" to mean "ZAMS", but "--initial-stellar-type", and the column header "Initial_Radius" overload the meaning of "Initial".

reinhold-willcox:
> The you should do a similar thing with code variables - if ZAMS vs INITIAL is confusing in the output files then it will be confusing as variable names for future maintainers of the code.

I take your point that the internal parameters for ZAMS vs Initial may confuse future maintainers, in which case it makes sense to rename to something clearer. I'm not sure what that would be, but its probably worth bringing up at a group meeting. 

> Should the options --initial-mass* (e.g. --initial-mass, --initial-mass-max, --initial-mass-function, etc.) not be consistent in meaning with --initial-stellar-type (and also the "initial_" columns headers)? The current "--initial*" options take "initial" to mean "ZAMS", but "--initial-stellar-type", and the column header "Initial_Radius" overload the meaning of "Initial".

The current initial* options currently work with the non-MS stellar types in the expected way, where initial mass corresponds to the mass at the beginning of the simulation, regardless of the initialized stellar type. I'm not sure what you mean by 'The current "--initial*" options take "initial" to mean "ZAMS" ', could you clarify?

Regarding your sample output above, everything there looks good to me. The initial radius is, as I expected, the radius at the beginning of the simulation for the stellar type I specified. Without knowing the exact radius as a function of mass for MS and HeMS off the top of my head, those values look reasonable to me, and I expect them to be different to each other. The Stellar_Type@ZAMS looks correct to me, why do you say it's wrong? The final stellar type also looks good. A 16 Msun MS star is probably in the range where it could become a NS or a BH (depending on the prescriptions you use for remnant mass, winds, etc.), but a NS is within reason, and a 16 Msun He star will certainly become a BH in the absense of interactions.

jeffriley:
>  I'm not sure what you mean by 'The current "--initial*" options take "initial" to mean "ZAMS" ', could you clarify?

Sorry - I was confused, so part of my question doesn't apply - see how easy it is :-) But...

What happens if we sample mass using one of the IMFs (KROUPA, or SALPETER for example) for "--initial-stellar-type = HeMS"?  That wouldn't be consistent would it?

> The Stellar_Type@ZAMS looks correct to me, why do you say it's wrong? 

Again, highlights the confusion.  We are not suggesting the star was born as a HeMS star - how can the stellar type at ZAMS be HeMS?

> The final stellar type also looks good. A 16 Msun MS star is probably in the range where it could become a NS or a BH 
> (depending on the prescriptions you use for remnant mass, winds, etc.)

Yeah - that's because I was confused about initial-mass.  And speaking of initial mass, that means the "Mass@ZAMS" column is also wrong - for the same reason the Stellar-Type@ZAMS is wrong.

reinhold-willcox:
Ah you got me, Stellar_Type@ZAMS is wrong here. Like I said, I think we can remove all ZAMS references in the output files in favor of Initial's, since that conceptually captures what the user is expecting for any supplied stellar type. But it would be wise to plan ahead as far as naming the internal parameters to make it easier for future developers to follow the flow of the code. 

If a user wants to use the kroupa IMF for HeMS stars, they are probably a bit foolish, but I don't think it's something we need to be concerned with. Anyone doing a serious run will put some thought into the mass distribution they want to use, and for the non-MS stars, will likely use a fixed value or sample from a small list anyway. 

As far as responding to the requested changes, are these all the remaining TODO items, or am I missing any?
- remove ZAMS values from output files, replace with Initial values where needed
- decide on and implement a clearer naming scheme for the internal Initial variables
- clarify these decisions in the documentation

jeffriley:
> If a user wants to use the kroupa IMF for HeMS stars, they are probably a bit foolish

Well, sure, but the point isn't that it would be a foolish thing to do - the point is that we would be using the same terminology for our starting stellar type ("--initial-stellar-type") which we know doesn't mean ZAMS stellar type, and for our initial mass function ("--initial-mass-function") which we know does mean the mass function that models the ZAMS mass distribution.  We're almost inviting users to do something foolish.

I'm still not over the INITIAL vs ZAMS confusion..,  

I don't think "initial" is the right word to use for the options that describe the stellar type at which we want to start recording the evolution - it's *not* the initial stellar type (even in the code it's not the initial stellar type), it's the stellar type to which we fast forward evolution before settling in to record the evolution at the pace we want.  The column names, and the code variable, should be named (something like) "phase_initial_..." because that's what they represent - the initial values for the evolutionary phase.  I think (and maybe it's only me...) that "initial" on its own connotes the initial values for the star(s), and that's not what the values are.

So, I don't know whether removing all the ZAMS references is the right thing to do.  And (clearly) I don't think the option names should be "initial-stellar-type...".  Both might be expedient, but they'll come back to bite us.  But I am not the keeper of the code - how about we get input from @ilyamandel  @SimonStevenson  @avigna  and anyone else that wants to chip in?

avigna:
My 2 cents: I see @jeffriley's point, and almost by definition trust his judgement when it comes to coding and conventions. However, I do think of the problem more as @reinhold-willcox describes it. Let me ellaborate a bit.

What we (in COMPAS) have done until now has made sense because we could *only* begin the evolution of stars from ZAMS, a bit tied up to Hurley's approach. Therefore, initial distributions end up mapping to mass distributions at ZAMS. This resulted in the current nomenclature and coding convention, which I think we all agree on.

What Reinhold suggests in this PR is a very interesting tool that *maybe* confronts the foundations of COMPAS as a tool for population studies. As far as I understood, one can now create systems that do not necessarily arise from forward binary evolution (but I am actually uncertain if the FastForward() function allows for this). If this is the case, then the term initial is no longer interchangeable with ZAMS. This has broader implications. For example, an initialized naked He star of mass M might come from:

- a single star of ZAMS mass M1 that self-stripped, 
- an interacting binary of ZAMS mass M2 that accreted some mass at some point, or 
- a merger of component ZAMS masses M3 (not implemented in COMPAS, but conceptually). 

Additionally, the metallicity at the moment it is initialized as a naked He star will not be the metallicity at ZAMS, but a higher metallicity (because it has been burning H and He to heavier elements). Therefore, this binary is not necessarily fully specified by its ZAMS properties. This binary is then defined from a set of stellar tracks at certain metallicity (environment) and an initial mass of these set of stellar tracks. Note how these quantities are both metallicity@ZAMS and mass@ZAMS, but in a different context.

Therefore, I think if we want to be really precise, we need to depart from Hurley and disinguish between ZAMS and the moment we initialize the simulation. However, this would require some discussion on nomenclature, non-negligible amount of work/coding, and could potentially add a few bugs (hopefully not) when renaming.

I do think of "initial" as the set of initial conditions for the simulation and "ZAMS" as a physical initial condition of the system. So, for example, your initial distribution in mass does not need to follow Kroupa's IMF, it can be ad hoc. But your ZAMS temperature and luminosity do need to reflect the conditions of core hydrogen ignition for a star of mass M and metallicity Z.

Hope this was somewhat helpful...

jeffriley:
Thanks @avigna.

> As far as I understood, one can now create systems that do not necessarily arise from forward binary evolution

That's kind-of true, but only for selected stellar types at the moment.   The reason I say "kind-of" is because we're not suggesting that just because we start evolving a (e.g.) HeMS star for which we have specified a set of attributes at the start of the HeMS phase, the star was born that way.  All we are doing is specifying a set of attributes for a star that has some age and starting our evolution from that point.  We can do that because, for a particular set of stellar types, Hurley has told us how to do that.  But we are not suggesting these stars are born with the attributes we provide (they are not the initial attributes of the star - they are the attributes at the start of the phase at which we chose to begin evolution).

> your initial distribution in mass does not need to follow Kroupa's IMF

Indeed - but our option name "initial-mass-function" is a well-known term, isn't it?  Wouldn't most people assume the IMF models the mass distribution at ZAMS?  We can't overload IMF to mean something else - wouldn't that cause confusion?

> I do think of "initial" as the set of initial conditions for the simulation and "ZAMS" as a physical initial condition of the system

I think the unqualified use of "initial" has the potential to be confusing because some people would consider "initial" to mean the initial attributes of the star (at birth), not the initial attributes of the star on the phase that we start evolving.  (I know ZAMS isn't technically birth, but in COMPAS it is).

>Additionally, the metallicity at the moment it is initialized as a naked He star will not be the metallicity at ZAMS, but a 
> higher metallicity (because it has been burning H and He to heavier elements). 

I don't believe that is taken into consideration with this change.  Edit:  Or maybe we do - maybe it's inherent in Hurley's equations.  I don't know.  I don't believe we do anything explicit with metallicity when we set the phase initial values.


The bottom line for me is that the unqualified use of "initial" in option names, column headers, and code variable names, has the potential to cause confusion, and why would we do that when there are other options?  We can name these things anything we want.  The simplest option is to qualify the word "initial" - use "phase-initial" or something.

But... stick a fork in me - I'm done.  I'll just keep repeating myself if I continue to contribute to this discussion.  I'll let others have their say and then a decision can be made and we can move forward.

jeffriley:
A comment - maybe a question - on something @avigna said - unrelated to the nomeclature...

> For example, an initialized naked He star of mass M might come from:
>
> a single star of ZAMS mass M1 that self-stripped,
> an interacting binary of ZAMS mass M2 that accreted some mass at some point, or
> a merger of component ZAMS masses M3 (not implemented in COMPAS, but conceptually).

We just set the phase-initial values and evolve from there using Hurley's equations.  Is that valid in all cases - do Hurley's equations for the evolution of (say) the naked He stars assume anything about how they got to be naked He stars?

Edit:

@avigna's mention of metallicity raises a broader question - how do these proposed changes interact with existing options?  For example, if the use specifies "--initial-stellar-type*" (ine eithe SSE or BSE mode), how does that work with options such as  (not an exhaustive list):

    --metallicity
    --rotational-frequency

I did not think to test those - @reinhold-willcox have you tested the changes with combinations of options such as those listed above?  What *should* happen if the user specifies those options (and is that in the new documentation)?  (I don't think metallicity should make a difference with our current code, but in view of @avigna's comments, should it?)

reinhold-willcox:
@avigna to clarify, the FastForward() function is really just for convenience, it's not really running through any evolution (there are no winds or anything). So we can configure systems that might never occur evolutionarily, but that's sort of the point anyway, to focus our studies on these later stages without worrying about how they got there. 

I don't think metallicity is affected. My impression is that metallicity, as far as it applies to winds, is mostly due to the iron fraction, which doesn't change in the star until negligibly close to the end of its life. So we should be safe. I haven't looked at rotational-frequency, but we don't have any functionality for that anyway (aside from CHE). 

Regarding multiple evolutionary paths to get to a single He star - this is only relevant when we switch to a stellar type with a non-zero tau parameter. That's why we're holding off on implementing that slightly more complex feature. The stellar types included for initialization in this PR are those which reset their tau to 0, and thus evolve with no memory of their history.

avigna:
(Just to reafirm: none of the things I listed should be a problem or a bug, mostly used them as examples on how definitions and nomenclatures can be misleading)

jeffriley:
@reinhold-willcox:

- how do these changes interact with --allow-rlof-at-birth and -allow-touching-at-birth options?
- how is the "stellarMergerAtBirth" flag affected by these changes?
- how are these changes affected by the equilibration and circularisation code in the BaseBinaryStar constructor?
- how is the "massesEquilibratedAtBirth" flag affected by these changes?

reinhold-willcox:
@jeffriley I will look into all these and respond soon (hopefully by mid-next week!). 

Regarding the nomenclature, I am happy to rename parameters to improve clarity as you suggest - I would just like to have the functionality in the code, regardless of how it's named. If I replace all instances of "Initial[Param]" with "Initial[Param]OnPhase" will that be sufficient? Similarly, what would be a clearer name for the "--initial-stellar-type*" options, to avoid confusion with the initial-mass-distribution and similar options?

avigna:
(@reinhold-willcox, maybe you can also post very summarized version of this thread in the #coding channel in Slack. It will probably generate a ton of replies, but seems that is confusing enough that involving other might be helpful... and then you can come back here.)

jeffriley:
@reinhold-willcox  I had been thinking, and I think it's clear from the discussion here, that the changes brought about by this PR are more far-reaching than we had at first imagined way back when we first hacked COMPAS to get this working.  @avigna's comment "that maybe confronts the foundations of COMPAS as a tool for population studies" has solidified my thoughts.  I know @avigna was also thinking about different routes to the star at which we start evolution, but I think he is right to consider this PR something of a paradigm shift for the current version of COMPAS.

COMPAS in its current form evolves stars from ZAMS through to some end point.  This PR changes that to “take a star (or these two stars) with these attributes (including stellar type(s), and evolve it (or them, as a binary) from there”.  We don’t care how it got to have those attributes (and maybe we don’t really mind that maybe it shouldn’t have those attributes) – we just want to see what happens when we evolve it from that point.  At the moment we can only do that for some stellar types, so we only have the first half of the new paradigm – but more on that later.

When we implement this PR, we won’t have just added a neat new feature to COMPAS, we will have changed something fundamental about how it works.  That requires a lot of change to the code, and to the input parameters and output properties, and I think that (and in some ways, especially the changes to the input parameters and output properties) requires that we move the version to v03.00.00 – this will be a new release, fundamentally different from the existing release.

So this PR will be bigger than you intended, but I think it’s close to being done – or at least not too far away.  We just need to put some careful consideration into what else needs to be changed.

We need to look at input options, output properties, and code variables, and for anything that implies (or explicitly states) it is a ZAMS or “initial” value or attribute, we should look at changing its name – and its meaning.  We need to move away from thinking of a value as a ZAMS or “initial” value and instead think of it as “the attribute value of the star to be evolved”.  So, instead of the user specifying the “initial mass” of a star to be evolved, they should specify “the mass of the star to be evolved”.  It’s not the initial mass – that was a long time ago, and they (the users) are specifying the mass (and stellar type…) of an evolved star that they want to continue to evolve. 

So I think what you now have as “initial-stellar-type” should change:

    initial-stellar-type    -> 'stellar-type'
    initial-stellar-type-1  -> 'stellar-type-1' (or ‘primary-stellar-type’ if you prefer)
    initial-stellar-type-2  -> 'stellar-type-2' (or ‘secondary-stellar-type’ if you prefer)

If the “stellar-type’ option is not specified, then we default to ZAMS and choose one of the MS types based on mass.
The option ‘rotational-frequency’ could be left as it is, but instead of the ZAMS rotational frequency it would be the rotation frequency of the star we want to evolve.  Other options would change in the same way.  For example:

    initial-mass    -> 'mass', with short version '-m'
    initial-mass-1  -> 'mass-1' or 'primary-mass', with short version '-m1'
    initial-mass-2  -> 'mass-2' or 'secondary-mass', with short version '-m2'

where these should be taken to mean "mass of the star to be evolved" rather than "ZAMS" or "initial" mass.

I’d suggest we think about changing:
    initial-mass-function -> 'IMF'
    initial-mass-min      -> 'IMF-minimum-mass' (or some variant of that)
    initial-mass-max      -> 'IMF-maximum-mass' (or some variant of that)
    initial-mass-power    -> 'IMF-power'

just to get away from “initial” (even though IMF = initial mass function”). 

And then I think we just need to address output properties/column header strings and code variables in the same way – we want to try to avoid names that imply “ZAMS” or “initial” (unless that’s exactly what we mean) and instead use names that imply the the attribute of the star.  If we need to know the attribute at the start of the phase, we could use (something like) "phaseInitialMass" etc. (for variable names), and (something like) "Initial_Mass_On_Phase" as the output property/column header.

That’s the first half – it works for stellar types for which we don’t need tau.  For a first-pass at the second half – the stellar types that need tau – can’t we just let the user specify tau (and anything else required – radius, temperature?)?  That would let a user, say, evolve a bunch of stars (using COMPAS), and if we output the attributes required to continue evolution (e.g. tau, radius – whatever attributes we can’t calculate for these stellar types), they could at a later time use the output attributes to continue evolution of stars of interest.  So the user could take a couple of stars from their evolved population, use those attributes to specify the stars to evolve (using our new paradigm), and possibly pairs star together that COMPAS would otherwise not evolve in a binary.  Maybe in the future we could figure out how to calculate tau and the other attributes required without running a full simulation first, but wouldn’t that work as an interim step?  And who knows, maybe we could keep a database of stars evolved so that users could look for the evolved stars they want to experiment with, and take the attributes (tau, etc.) from that database.

I know you want to get the changes into the code base, but I don't think we can do this piecemeal - it needs to be done in one go (at least the first half - we could leave the second half - even the interim step - for later).  If you really need the code for a current project you could just use what you have now in your own branch - you know what you are doing and you know what the catches are (and what else might need to be changed just to get it to work), so you should be able to do some runs for your project.

These are my thoughts about what we’re trying to achieve with this PR, and the directions I think we should take.  But, as I’ve said before, I’m not the keeper of the code, neither am I the chief architect.  I have some experience in architecting software solutions, so I have some views about how it should be done – but I am always happy to hear and consider other views.

reinhold-willcox:
I'll respond to the above comments a bit later, but I've found a bug in the current approach here. 

If you initialize a He+NS binary and they are close enough to engage in mass equilibration at birth, the masses will reset to be equal, and for large enough masses, the NS will then collapse into a BH (with no SN), resulting in a He+BH binary that then evolves "like normal" for a binary initialized this way. But we should not have mass equilibration at birth if one of the stars is a CO.

jeffriley:
> But we should not have mass equilibration at birth if one of the stars is a CO.

Precisiely why I asked above "how are these changes affected by the equilibration and circularisation code in the BaseBinaryStar constructor?".  I think you should probably look at the other possible interactions I highlighted there too.

SimonStevenson:
I appreciate the discussion above, and think it is good that this has come up. I agree with Jeff's concern regarding the overloading of the term 'initial' to mean both 'start of the simulation' and 'start of the stars life', which used to coincide, but may no longer do so following this PR.

Some comments on Jeff's suggestion above:

> I’d suggest we think about changing:
> initial-mass-function -> 'IMF'
> initial-mass-min -> 'IMF-minimum-mass' (or some variant of that)
> initial-mass-max -> 'IMF-maximum-mass' (or some variant of that)
> initial-mass-power -> 'IMF-power'
> 
> just to get away from “initial” (even though IMF = initial mass function”).

I don't like the idea of using acronyms if we can avoid it (and plus, as you say, this really only hides the problem, not resolves it). My suggestion would be to switch to 'primary-mass-distribution', 'primary-mass-distribution-min', 'primary-mass-distribution-max', 'primary-mass-distribution-power'. Then we could consider renaming some of the exisiting options like 'KROUPA' to 'KROUPA_IMF' or similar, where I think it might make a bit more sense. Part of the motivation for this would be that a generic power-law distribution is not necessarily an IMF, but may represent e.g., the mass distribution of black holes or helium stars or whatever. This hopefully also avoids the 'initial' confusion described above.

(I also have a moderate preference for primary-stellar-type and secondary-stellar-type and primary-mass and secondary-mass but I think using 1 and 2 as a shorthand is probably fine)

This PR will also need to include associated documentation updates and updates to pre/post-processing to account for any major changes as well.

avigna:
> I think that (and in some ways, especially the changes to the input parameters and output properties) requires that we move the version to v03.00.00 – this will be a new release, fundamentally different from the existing release.

Yes, I agree.

> If you initialize a He+NS binary and they are close enough to engage in mass equilibration at birth, the masses will reset to be equal, and for large enough masses, the NS will then collapse into a BH (with no SN), resulting in a He+BH binary that then evolves "like normal" for a binary initialized this way. But we should not have mass equilibration at birth if one of the stars is a CO.

This should (at least partially) be resolved by using accretion limits, e.g., Eddignton accretion limit. A compact object wouldn't accrete all that mass that fast. This doesn't resolve HeMS+MS binary, which would likely lead to an explosive trantient... and definitely not to a vanilla equlibration.

reinhold-willcox:
Thanks all for your input. I appreciate that the simple "initial*" naming scheme could lead to confusion for users and developers and thus should be avoided. I also understand that the "best" solution involves some reworking of the internal code so that this ambiguity is avoided. 

Realistically, I don't have the time to implement this in the paradigm-shift style which it sounds like is the preferred approach. The funcionality is nearly there (pending the tests/fixes to the mass equilibration and other points Jeff brought up), so if others would like to use this, I'm happy to share what I have. But unless someone else is able to implement this properly, or we decide to forgo clarity for the sake of functionality, I think it will have to remain on a separate branch.

jeffriley:
I don't have time right now either.  I vote for a separate branch over forgoing clarity for the sake of functionality.  

I don't think there is too much left to do - the main thing is to be fairly meticulous in checking that all assumptions that we're evolving a ZAMS star are looked at and fixed where necessary, and the nomenclature is similarly looked at and fixed.

reinhold-willcox:
@jeffriley Re-opening this PR. I haven't made any changes since this was last touched (aside from merging in dev), so some of your above points may still be valid. That said, we may want to start a new PR just to avoid clutter. Let me know what you prefer.

jeffriley:
I'm happy to take this on, but I can't do it until after May.  Yes, I think start a new PR, but let's wait until then - unless you want to start it before then.

reinhold-willcox:
Closing this PR as out of date, and re-opening the same changes in PR #931

PR #734, opened 2022-01-09
Clean-up of CE treatment

ilyamandel:
Cleaned up treatment of HG donors having CONVECTIVE envelopes in LEGACY; fixes an issue with missed failed CEs introduced in 02.25.01

### DIFF
--- a/src/BaseBinaryStar.cpp
+++ b/src/BaseBinaryStar.cpp
@@ -1526,7 +1526,7 @@ void BaseBinaryStar::ResolveCommonEnvelopeEvent() {
             m_Flags.stellarMerger        = true;
         }
     }
-	else {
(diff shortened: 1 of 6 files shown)


ilyamandel:
Temporarily changing to draft while double-checking.

ilyamandel:
Turns out there was another logical flaw in 02.25.01, which I now fixed.  Note that the detailed_evolution example (figure 15) has MT from an HG donor which is dynamically unstable, so I had to change AllowRadiativeEnvelopeStarToSurviveCommonEnvelope to true in the submit file (i.e., allowing optimistic choice) to reproduce figure 15.

jeffriley (approved):
Approved.

Issue #735, opened 2022-01-13
FastCosmicIntegration.py filename argument causing an error


tcla0007:
**Describe the bug**
In the file FastCosmicIntegration.py we can input the path to the correct file as an argument. Previously the file name was a separate argument that could also be specified. In the current version it looks like we should include the file name in the path argument since there isn't any option to add a filename in the arguments list now. The bug is that the functions 'append_rates' and 'delete_rates' still include the file name argument when they are called, which raises an error since this argument is no longer in the function definition. 

**Label the issue**

`urgency_moderate` - This is a moderately urgent issue

`severity_minor`    - This is a minor bug with minimal impact

**To Reproduce**
running:
`python FastCosmicIntegration.py --path /path/to/COMPAS_Output/COMPAS_Output.h5 --m1min 10 --m2min 10`
gives 
`Traceback (most recent call last):
  File "FastCosmicIntegration.py", line 803, in <module>
    append_binned_by_z = False, redshift_binsize=0.05)
TypeError: append_rates() got multiple values for argument 'maxz'`

I was able to resolve this by removing args.fname from the following function call on line 794:

`append_rates(args.path, args.fname, detection_rate, formation_rate, merger_rate, redshifts, COMPAS, n_redshifts_detection,
            maxz=args.max_redshift_detection, sensitivity=args.sensitivity, dco_type=args.dco_type, mu0=args.mu0, muz=args.muz, sigma0=args.sigma0, sigmaz=args.sigmaz, alpha=args.alpha,
            append_binned_by_z = False, redshift_binsize=0.05)`

**Versioning**

 - COMPAS v02.26.03

reinhold-willcox:
Thanks for catching this! Submitting the fix now.

reinhold-willcox:
@tcla0007 Do you want to take a look at the PR and see if it solves your issue?

tcla0007:
@reinhold-willcox Thanks it all works now!

PR #737, opened 2022-01-14
Fastcosmicintegration bugfix

reinhold-willcox:
Removed references to `filename` in fastcosmicintegration.py (it was previously merged with `path`).

Addresses issue #735


jeffriley (approved):
Looks good.

PR #738, opened 2022-01-17
Readthedocs run submit

### LINKED ISSUE #728: Update online documentation to describe runSubmit.py
Follow up on #720 .

avigna:
I updated the read-the-docs following the upgrade from the `pythonSubmit.py` to the `runSubmit.py` and `compasConfigDefault.yaml` files (see #720 ).

This PR resolves #728 , a follow up to document the changes made as part of the edits of the JOSS paper referee report.

Most of the changes are just a replacement from "pythonSubmit" to either "runSubmit" or "compasConfigDefault.yaml", but sometimes additional notes or text were added.

Finally, this PR does not resolves the upgrade and updated documentation of the two following pending items:

1. Sampling (stroopwaffel) using the new `runSubmit.py` + `compasConfigDefault.yaml` tandem, nor
2. The use of a demo file to test COMPAS (a la `pythonSubmitDemo.py`).

Once this PR is approved and closed, I will open 2 issues for the aforementioned pending items.

### DIFF
--- a/online-docs/notebooks/CosmicIntegration.ipynb
+++ b/online-docs/notebooks/CosmicIntegration.ipynb
@@ -175,12 +175,12 @@
     "    \n",
     "Mlower:\n",
     "\n",
-    "    lower limit used for M1 in the pythonSubmit of the simulation. \n",
+    "    lower limit used for M1 in the user-defined options of the simulation. \n",
     "    Needed to recover `true` amount of Msun evolved (see step 4)\n",
     "    \n",
     "Mupper:\n",
     "\n",
-    "    upper limit used for M1 in the pythonSubmit of the simulation. \n",
+    "    upper limit used for M1 in the user-defined options of the simulation. \n",
     "    Needed to recover `true` amount of Msun evolved (see step 4)\n",
     "\n",
     "binaryFraction:\n",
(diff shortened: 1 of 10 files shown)


reinhold-willcox:
~~Not sure why, but recompiling the html for this required many new sphinx packages for me, even though I was able to compile previous versions of the docs. Not crucial for this PR, but in the future we may want to put together a requirements.txt for the online-docs so people can install them all in one go. Actually, looking at the pinned post in the documentation slack channel, we used to have exactly that, but I don't see it now...~~ Edit: I was not on the correct branch, so I'm not sure why I had package install issues, but they were not from this PR. At any rate, we should still try to reincorporate the requirements.txt for the online-docs, to avoid further headaches in the future.

Regarding the specifics of this PR, I agree with all the changes that Alejandro made and everything compiles. Looking at the docs closely, there are other areas that need to be fixed / improved, but we are already aware of those and this PR is a step in the right direction. So I'm happy to accept this, and then we can have a deeper discussion later about how to fix the other areas.

reinhold-willcox (approved):

ilyamandel (approved):
Fixed some typos, happy to approve -- but will immediately start a new issue to generate an example using runSubmit and update documentation accordingly. :)

PR #742, opened 2022-01-27
Fixes to the detailed plotting routine

reinhold-willcox:
- Readded the van_den_heuvel images to the repo (in new utils/media/ folder) and updated the path in the plotting script
- Added flexibility to the detailed output file name - an alternative file name can be optionally specified on the command line
- Fixed the eventImage functionality, now every event has default eventImage of None (is cleaner in other places)
- Fixed the vandenheuval plot width to 9, to hopefully always fit the text strings (not robustly tested)


jeffriley (commented):
@reinhold-willcox should 

utils/PythonScripts/plot_detailed_evolution.py

and

utils/example_plots/methods_paper_plots/detailed_evolution/detailed_evol_plotter.py

be in sync?

reinhold-willcox:
No, the methods paper version is a precursor to the general one, but includes details about sizing that only apply if you know exactly what you're plotting. It also doesn't make the van_den_heuvel plot.

jeffriley (approved):
Approved

PR #747, opened 2022-02-02
He Giant core mass growth exponents p and q

### LINKED ISSUE #744: Hurley gbParams p and q values for He giants
**Describe the bug**
In Hurley+ 00, section 6.1, the discussion on He giants states that the Luminosity is a piecewise function of two power laws (similar to equation 37 for giant stars). 

L = min(B* Mcore^q, D* Mcore^p)  (eq. 37)
L = min(B* Mcore^3, D* Mcore^5)  (eq. 84)

Below (37), p and q are defined as piecewise functions of the total mass, and these p and q values appear in the later equations calculating the GB core mass and timescales (39-42). We use analogs of these later equations for the He giants, with various values swapped out. Namely, A_H becomes A_He, t_BGB becomes t_HeMS, L_BGB becomes L_THe, and the B and D values defined just below equation 87 get swapped in for the values of the same letters in (39-42). However in COMPAS, we do not recalculate (p, q) = (3, 5) for He giants. We use them correctly in the luminosity calculation in (84), but we do not change them in gbParams for use in (39-42). 

If this is in fact a bug, it will affect core mass growth for He giants, and consequently the luminosity, radius, and transition time between HeHG and HeGB stars. 

This is easy enough to implement in the code, and I am happy to do so, but could someone who is more familiar with Hurley please check that this is the correct approach?

**Label the issue**

 - COMPAS v02.27.01

reinhold-willcox:
Added p and q as gbParams to HeHG (and also HeGB which inherits from HeHG). 

Previously, we were using the values set for GB stars, which do not apply to He giants.

Edit: Based on some quick testing, this does not affect the He evolution much. For HeGiants, (p, q) = (5, 3). For GB stars, it is mass dependent, but either (p, q) = (6, 3) or (5, 2) (or is interpolated between those endpoints), so there is not a huge difference. 
Also, it appears as though we still call the GB values as well, though they are not the ones getting used. I added a print statement to `GiantBranch::CalculateCoreMass_Luminosity_Mx_Static(...)` and saw both GB and HeGiant values, even though the star was initialized as a HeMS star. So there's probably some excessive calls being made that do not need to be, but I can't say if it would be worthwhile trying to track this down.


reinhold-willcox:
Reverting to draft for the following reason:

In HeHG.h, Initialise() calls CalculateTimescales() before CalculateGBParams. But the parameters in timescales rely on the GBParams being set correctly (namely, the p and q being set in this PR). So I've moved the CalculateGBParams() call to immediately before CalculateTimescales, but this is inconsistent with the other stellar type objects, and so may lead to incorrect functionality somewhere else in the pipeline. 

@jeffriley can you advise on the best way to ensure that the p and q are set correctly in the CalculateTimescales function() ?

jeffriley:
From memory the order of the calls is important - I think we rely on one being calculated before the other.  I remember it being a bit messy in the legacy code, and I think that was for some stellar types the order was reversed - but only reversed for some.  That is for most stellar types we'd calculate the timescales, then the gbparams (or the gbparams then the timescales - I don't remember the order off the top of my head), but for some stellar types we'd do something like calculate half the timescales, then some gbparams, then the rest of the timescales, then the rest of the gbparams - and some calculations were done before switching stellar type, and some after.  As I said, it was messy.  It may be that we need to refine some of the calculations here.  I'll take a look later.

ilyamandel:
What's the status of this PR?

reinhold-willcox:
@ilyamandel This is still a necessary change to align COMPAS with Hurley. But it's a low-impact change and very difficult to implement, hence the delay.

reinhold-willcox:
@jeffriley ping on this. We are technically not in agreement with Hurley, but if the changes wouldn't make a huge difference anyway, perhaps we're willing to overlook that.

ilyamandel:
@jeffriley -- maybe this will automatically be taken care of when you review the agreement with Hurley on SSE?

ilyamandel:
@jeffriley, @reinhold-willcox  -- did you say that this got fixed already, or just that you were planning to return to it?

jeffriley:
I had clearly forgotten this PR existed.  This was fixed in v02.46.05.

PR #749, opened 2022-02-03
Fixed condition for envelope type

ilyamandel:
Fixed condition for envelope type when using ENVELOPE_STATE_PRESCRIPTION::FIXED_TEMPERATURE (previously, almost all envelopes were incorrecctly declared radiative)


themikelau (changes requested):
The same fix needs to be applied to other stellar types: HG, HeHG, MS_gt_07,

ilyamandel:
@themikelau , @jeffriley -- should be fixed now.

themikelau (approved):
Okay now. I seem to get sensible results from a test run using the FIXED_TEMPERATURE envelope prescription.

PR #751, opened 2022-02-03
Fixed mass change on forced envelope loss in response to issue # 743

ilyamandel:


jeffriley (approved):
Approved.

@ilyamandel did the addition of "const" in HeWD.h cause the compiler warnings to go away (I didn't see them with the linux compiler so can't check)

ilyamandel:
@jeffriley -- yup, no more compiler warnings (@themikelau made that fix in the previous PR, but somehow it got highlighted again here, sorry).

PR #755, opened 2022-02-08
Defect repair: Fix for issue # 745

jeffriley:
- Fixed for issue # 745 - logfile definition records not updated correctly when using logfile-definitions file (see issue for details)


jeffriley:
Oops - needs a change.  I'll fix and re-push

Issue #750, opened 2022-02-03
Add yaml to documentation


ilyamandel:
https://compas.readthedocs.io/en/latest/pages/Getting%20started/COMPAS-dependencies-python.html does not mention yaml, but apparently yaml is required for runSubmit.py .  pip3 install yaml doesn’t work, though — one needs pip3 install pyyaml.  That’s sufficiently non-trivial that it definitely needs documenting.

PR #758, opened 2022-02-08
Added comment about pyyaml dependency

avigna:
Easy, quick fix addressing #750. HTML compiles and looks as it should.


reinhold-willcox (approved):
looks good!

PR #762, opened 2022-02-15
Fix for missing USSNe

### LINKED ISSUE #761: No USSNe
**Describe the bug**
We currently do not produce any USSNe in COMPAS. It is unclear why this is happening, or when it was introduced. We still have doubly-stripped systems which _should_ be flagged as USSNe, but as far as I can tell, they are all classified as ECSNe.

One possible source of the problem is that in `GiantBranch::ResolveSupernova()`, we identify ECSNe if  `(snMass < MCBUR2)` where MCBUR2 = 2.25 is defined as the "Core mass at base of the AGB above which the CO core is completely non-degenerate". So we should be comparing to the CO core mass not the total mass. 

Meanwhile, snMass comes from the function `CalculateInitialSupernovaMass()` which in GiantBranch.h just returns m_Mass, but this is overridden in some stellar type classes. Specifically, in HeHG.cpp, we have:
```
double HeHG::CalculateInitialSupernovaMass() const {
    if (utils::Compare(m_COCoreMass, m_Mass) == 0) {    // special case of ultra-stripped-star -- use current mass
        return std::max(m_Mass, m_GBParams[static_cast<int>(GBP::McBAGB)]);
    }   
    return GiantBranch::CalculateInitialSupernovaMass();
}
```

I checked for the cases of SNe which should be USSNe but are flagged as ECSNe, and the snMass is consistently 1.6 Msol (which is why the comparison with MCBUR2 always evaluates to true), even when m_Mass is larger than 1.6. I'll keep looking into this, but need to run now, so posting this issue and these first steps in case someone else has time/interest to look into this. 

**Label the issue**
`urgency_high`     - This is a very urgent issue and should be resolved as soon as possible
`severity_major`    - This is a severe bug

**To Reproduce**
Current default COMPAS with some of these systems. I have a larger grid, so these might not be representative, but I don't want to post the whole thing. 

```
 --initial-mass-1 9.322100263051011 --initial-mass-2 3.1587987252729217 -a 4.132760779522892 --random-seed 28
 --initial-mass-1 8.580491526089412 --initial-mass-2 8.104881912725139 -a 0.1390697759295803 --random-seed 332
 --initial-mass-1 11.240187803832582 --initial-mass-2 5.403709003224469 -a 0.09884499860753199 --random-seed 394
 --initial-mass-1 9.642267051178868 --initial-mass-2 5.904521565226383 -a 3.6341561146041745 --random-seed 415
 --initial-mass-1 10.024793009863979 --initial-mass-2 5.753485603089303 -a 4.894731749605503 --random-seed 485
 --initial-mass-1 7.7857720728077755 --initial-mass-2 6.643531797427209 -a 0.12856422984343968 --random-seed 500
 --initial-mass-1 11.447083709043914 --initial-mass-2 9.405316385429147 -a 0.08444318351785884 --random-seed 546
 --initial-mass-1 9.908471630123616 --initial-mass-2 8.43337510279692 -a 0.1548327098090108 --random-seed 580
 --initial-mass-1 9.236483400993375 --initial-mass-2 7.877727073242662 -a 0.2456129104998611 --random-seed 599
 --initial-mass-1 9.413357939845339 --initial-mass-2 8.522692050627962 -a 0.07610412563750199 --random-seed 757
 --initial-mass-1 12.249779968449845 --initial-mass-2 9.21058360481808 -a 0.09075176232472035 --random-seed 832
```

**Expected behavior**
Doubly stripped stars with masses > 2.25 should undergo USSNe, not ECSNe. 

 - COMPAS v02.27.03

reinhold-willcox:
In `HeHG::ResolveEnvelopeLoss()`, I readded back in check for IsSupernova() before the stellar type switch. The code compiles and the USSNe are reproduced. However, I haven't looked closely at other areas of impact, so I'm keeping this as a draft PR for now. Specifically, it would be good to understand why this was removed in the first place, and whether there were other functions that were modified in a similar way which also have to be corrected. 

I also think it would be good to add another SN type (type 32?), to distinguish ultra-stripped supernovae undergoing traditional core-collapse from ultra-stripped electron capture supernovae. The difference should still arise from a comparison of the CO core mass to MCBUR2, but the remnants and natal kicks may be qualitatively different. Perhaps that falls under the category of requested feature though, and for this PR, we should focus on restoring previous functionality.

### DIFF
--- a/src/GiantBranch.cpp
+++ b/src/GiantBranch.cpp
@@ -1638,16 +1638,17 @@ STELLAR_TYPE GiantBranch::ResolveElectronCaptureSN() {
 
 
 /*
- * Resolve Type Ia Supernova
- *
+ * Resolve Type IIa Supernova
+ * 
+ * This is a possibly made up SN type which would look like a Type Ia + H (see Hurley)
  * Zero attributes - leaves a Massless remnant
  *
  *
- * STELLAR_TYPE ResolveTypeIaSN()
+ * STELLAR_TYPE ResolveTypeIIaSN()
  *
  * @return                                      Stellar type of remnant (always STELLAR_TYPE::MASSLESS_REMNANT)
  */
-STELLAR_TYPE GiantBranch::ResolveTypeIaSN() {
+STELLAR_TYPE GiantBranch::ResolveTypeIIaSN() {
 
     m_Mass              = 0.0;
     m_Radius            = 0.0;
@@ -1840,8 +1841,8 @@ STELLAR_TYPE GiantBranch::ResolveSupernova() {
 
             stellarType = ResolvePairInstabilitySN();
         }
-        else if (utils::Compare(snMass, OPTIONS->MCBUR1()) < 0) {                                   // Type Ia Supernova
-            stellarType = ResolveTypeIaSN();
+        else if (utils::Compare(snMass, OPTIONS->MCBUR1()) < 0) {                                   // Type IIa Supernova - like a Type Ia + H (see Hurley)
+            stellarType = ResolveTypeIIaSN();
         }
         else if (utils::Compare(snMass, MCBUR2) < 0) {                                              // Electron Capture Supernova
             stellarType = ResolveElectronCaptureSN();
--- a/src/GiantBranch.cpp
+++ b/src/GiantBranch.cpp
@@ -1638,16 +1638,16 @@ STELLAR_TYPE GiantBranch::ResolveElectronCaptureSN() {
 
 
 /*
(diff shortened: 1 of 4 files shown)


ilyamandel:
I agree that's the correct fix and recommend merging it in.

ilyamandel (changes requested):
You have ResolveTypeIIaSN() in GiantBranch.cpp and ResolveTypeIaSN() in GiantBranch.h

ilyamandel (approved):
Thank you for the fix!

PR #763, opened 2022-02-15
Adding --hmxr-binaries option

IsobelMarguarethe:
Adding flag --store-hmxr-binaries, which adds any binary with 1 compact object and a donor that is > 80% Roche Lobe filling to the BSE_RLOF output. Also store fraction RL filling before and after the time step. Unsure about naming in file (currently Radius|RL>step) and about naming in code (''Roche Lobe Tracker'' is not intuitive). Also changed some minor typos in Options.cpp and the year for the most recent few updates in the changelog


IsobelMarguarethe:
Still to do:

- [x] Decide on naming of new variables in code
- [x] Decide on naming of new columns in BSE_RLOF
- [x] Add description of changes in changelog

@jeffriley @ilyamandel @ryosuke-hirai

reinhold-willcox:
This looks good to me (following the TODOs noted above). I would add that it might be good to put the 80% magic number as a value in constants.h for better clarity, and so that in the event that future users would like to use this feature, it can be implemented as a command line argument (though I don't think that's necessary at this point).

reinhold-willcox:
And thanks for catching the numerous typos!

IsobelMarguarethe:
Good idea! I've added that in my most recent commit.

jeffriley (changes requested):
Your function

    BaseBinaryStar::HasOnlyOneOf(STELLAR_TYPE_LIST p_List)

takes a list of stellar types as the parameter, and your description of the function is

    Determines if the binary contains only one star which is one of a list of stellar types passed

and your description of the return value is

    true if only one of the stars of the binary is in list, false if neither or both

I don't think that's actually what the code does.  For example, let's say the function is called with COMPACT_OBJECTS as the parameter, which is defined in constants.h as

    const std::initializer_list<STELLAR_TYPE> COMPACT_OBJECTS = {
        STELLAR_TYPE::HELIUM_WHITE_DWARF,
        STELLAR_TYPE::CARBON_OXYGEN_WHITE_DWARF,
        STELLAR_TYPE::OXYGEN_NEON_WHITE_DWARF,
        STELLAR_TYPE::NEUTRON_STAR,
        STELLAR_TYPE::BLACK_HOLE,
        STELLAR_TYPE::MASSLESS_REMNANT
    };

If the binary for which you call HasOnlyOneOf(COMPACT_OBJECTS) consists of a HeWD and a NS, I think your function will return TRUE after it counts only one HeWD.  Is that what you want?

Also, I think I would just use "hmxr-binaries" as the option name - I think that's more consistent with other options ("hmxr-binaries" would then mean turn that functionalit on/off, similar to the (commented) "BE-binaries" option).  The only other option that starts with "store-" is used to store the input files (grid file, logfile-definitions file) in the output container (directory), which is conceptually different functionality.  But it's up to you - we have some work to do one day to rationalise all the option names anyway.

IsobelMarguarethe:
You're right, that's not what I wanted - I have fixed this by moving the
check of the match counter outside of the loop over stellar types.
I've also replaced that flag with hmxr-binaries and the associated member
variable as m_HMXRBinaries.

jeffriley:
Ok, small nitpick.  I think

    return matchCount == 1

is more elegant than

    if (matchCount == 1) return true;
    return false;

Edit:

You could end the loop early if matchCount increments beyond 1, but the performance improvement would probably be marginal overall (adding a check that might be superfluous ~half the time might just create unnecessary overhead for that "~half of the time", so on balance not really improve anything...)

IsobelMarguarethe:
Oh, yes, that is better!

IsobelMarguarethe (commented):
I'm not sure which changes you requested that haven't already been made...

jeffriley (changes requested):
Changes look good.  Just a couple of things:

In BaseBinaryStar::BinaryPropertyValue() where you've changed the property names, can you check that the text ("value = ") is still aligned with the rest of the switch?  Not a big issue, but (I think) easier on the eyes (so easier to scan) if they align (it doesn't look like they do if I look at the file on github).  Same for BINARY_PROPERTY_LABEL in constants.h (there are probably a few other places, but they are the main ones - I don't want to be too picky... and maybe it's just me - I think if things get ragged in places like that the code is harder to read).

You have "BinaryConstituentStar::m_StarToRocheLobeRadiusRatio" as the variable name for STAR_TO_ROCHE_LOBE_RADIUS_RATIO_1 and STAR_TO_ROCHE_LOBE_RADIUS_RATIO_2 in standard-logfiles-record-specification-binary.rst, but I don't think that's right.  Isn't it m_RLOFDetails.props[Pre/PostMT]→starToRocheLobeRadiusRatio[1/2]?

And finally, you've tested it and it does what you expect?  And if you run without the new option specified, there is no change from a run with the current version?

IsobelMarguarethe:
Alignment: you're absolutely right (dangers of using sed without actually opening any files) - I've made the spacing consistent everywhere that I've found it misaligned. Totally agree that the code should be as readable as possible.

BinaryConstituentStar::m_StarToRocheLobeRadiusRatio: This was again achieved using sed to switch "m_RocheLobeTracker" to "m_StarToRocheLobeRadiusRatio". You can this at [this line of the diff](https://github.com/TeamCOMPAS/COMPAS/pull/763/commits/282bd612dd8b78a937419517359c25bdb5e5781d#diff-1dbaa5b1a570859a3b58e6c09077b26181b0d668debd621d45463ad965dee306L1207). I think this is because a BinaryConstituentStar just tracks its own ratio to its Roche Lobe radius with its companion.

I've tested that the output is as expected, and consistent with the master copy of the code. There is still a change in the output without the new flag as there are 4 new columns in BSE_RLOF, but other than that there is no change.

jeffriley:
Thanks for fixing the alignment.

Ok, it looks like @ilyamandel  removed the m_RocheLobeTracker variable in v02.13.11 and changed the code to calculate the value on-the-fly if it is requested to be included in any output file - but the documentation wasn't changed to reflect that code change.  So there was no BinaryConstituentStar::m_RocheLobeTracker variable in the code that you changed, and there is no BinaryConstituentStar::m_StarToRocheLobeRadiusRatio variable in your modified code.  I suggest changing the description of the COMPAS variable from "BinaryConstituentStar::m_StarToRocheLobeRadiusRatio" to "calculated using BinaryConstituentStar::StarToRocheLobeRadiusRatio()" (if that's where it's calculated - that's what I got at a brief look, but you'll know better).  (The COMPAS variable name is just there to help users - mainly developers - find the code related to the property being printed, so whatever description you think works).


Also, I just noticed that these are not in alphabetical order in BaseBinaryStar.cpp

        case BINARY_PROPERTY::STAR_TO_ROCHE_LOBE_RADIUS_RATIO_1:                    value = StarToRocheLobeRadiusRatio1();                                      break;
        case BINARY_PROPERTY::STAR_TO_ROCHE_LOBE_RADIUS_RATIO_2:                    value = StarToRocheLobeRadiusRatio2();                                      break;


And...  last one :-)  You might consider changing the name of the PR - it probably doesn't matter that much, but in case someone looks for PRs by name or something.

IsobelMarguarethe:
Thank you for such a thorough review @jeffriley! I've acted on all of the suggested changes.

jeffriley:
Changes look good! (I fixed a typo in changelog.h - I guess I can change your branch...)

I hope I haven't put you off ever opening another PR...

jeffriley (approved):

PR #764, opened 2022-02-16
fixed Heuval for Heuvel

reinhold-willcox:
typo in detailed plot


ilyamandel (approved):
Thanks for the quick fix :)

PR #765, opened 2022-02-16
Updates to runSubmit and compasConfig files

### LINKED ISSUE #740: Documentation -- runSubmit example
Generate an example workflow using runSubmit and update documentation accordingly.

### LINKED ISSUE #752: Documentation improvements
This is a running list of proposed improvements and corrections to the documentation. Includes fixes described in other issues, which are tagged appropriately. Since documentation is more/less tangential to producing good science results, these all fall to a lower priority than some of the other git issues. But good documentation also reduces the onboarding time for new users and enables contributions from a wider community. So if anyone has time/interest, please feel free to jump on any of the tasks below, grouped by relative priority. User tags indicate assignment.

High
- [ ] Add a "Changes in this release" page, which is an overview of key recent changes in the changelog, aimed at typical users. 
- [ ] Add stroopwafel interface file to compas, and a quick description to the documentation #741 @FloorBroekgaarden @stellarphysicsandpopulations 
- [ ] Add detailed evolution example using runSubmit #740 @reinhold-willcox @themikelau 
- [x] Add yaml install #750 @avigna 
- [ ] Ensure links to websites, images, and other documentation pages are up to date

Medium
- [ ] Clean up online docs file structure for clarity (c.f [ozstar docs](https://supercomputing.swin.edu.au/docs/index.html))
- [ ] Add in ComponentSpeed parameter @reinhold-willcox 
- [ ] Add in MassTransferTrackerHistory values 8 and 9 #733 @reinhold-willcox 
- [ ] Clarify code defaults #582 
- [ ] Automatically import input and output parameter details from the compas source code

Low
- [ ] Better documentation of unintended consequences of CHE #346
- [ ] Add API for COMPAS src code, so developers can see what functions take which input/output. This is a little more involved but it you automatically set up an API reference for every single function in COMPAS (built using the docstring comments). 
- [ ] Add API for python post-processing functions, so users can make better use of them. [This page](https://devblogs.microsoft.com/cppblog/clear-functional-c-documentation-with-sphinx-breathe-doxygen-cmake/) explains it quite nicely for C++ code. [Example from LEGWORK](https://legwork.readthedocs.io/en/latest/modules.html)
- [ ] Copy code buttons: You can add “copy to clipboard” buttons to all of your code blocks super easily just by adding the sphinx_copybutton extension
- [ ] Inheritance chain: You could explain the inheritance chain stuff with the COMPAS functions [using this](https://www.sphinx-doc.org/en/master/usage/extensions/inheritance.html)
- [ ] Icons: Go into your conf.py and set html_favicon and html_logo to the COMPAS icon and logo respectively and it’ll update the icon in the browser tab and the logo in the top left of the docs respectively

### LINKED ISSUE #759: Make sure online documentation is up-to-date after sorting runSubmit / yaml upgrade.
Make sure online documentation is up-to-date after sorting out the following issues:

- [x] #740 
- [x] #741 
- [x] #750 

This implies removing all the notes suggesting any of the previous issues is pending or that the runSubmit / yaml  is not ready.

reinhold-willcox:
- Updated compasConfigDefault.yaml to include all options for COMPAS that make sense for the yaml (there were a few key ones missing)
- Updated the runSubmit to include explicit FALSE boolean options. Previously, these were ignored, but this produces incorrect behavior if the compas default is TRUE but the user wants to turn the option off.
- Replaced demo pythonSubmit with runSubmit + compasConfig
- Confirmed that the detailed output image is correctly reproduced


ilyamandel:
@reinhold-willcox -- I see you are updating things in utils/example_plots/methods_paper_plots/detailed_evolution -- but it looks like the plotting script there is out of date relative to the one in utils/PythonScripts .  Should they not be the same (actually, can we force them to be the same -- does a git repo allow soft links?)

reinhold-willcox:
@ilyamandel You are correct that the two are divergent. This was done mostly because the one in utils/PythonScripts has been continually updated for a while, e.g to include the van den Heuvel diagrams, while the methods paper version is supposed to exactly reproduce what is seen in the methods paper. 

Some of the arguments used in the pythonSubmit (now runSubmit) used to generate the methods paper version are non-default, and the plotting script also explicitly sets the plot bounds and custom legend (so that it is neatly split in the mass axes). I can reformat these to match the general plotting script if you'd like, but there will always be differences between the two so its just a bit of extra overhead whenever we update the general script.

ilyamandel:
We can't reproduce exactly what was in the methods paper because the C++ code has changed, anyway.  I'd opt for keeping the plotting script the same -- less chance of someone using the wrong script...  But OK, that's a separate issue, so will approve this one for now.

ilyamandel (approved):

reinhold-willcox:
> We can't reproduce exactly what was in the methods paper because the C++ code has changed, anyway. I'd opt for keeping the plotting script the same -- less chance of someone using the wrong script... But OK, that's a separate issue, so will approve this one for now.

@ilyamandel can you clarify what you mean? I see the same plot when I run with this runSubmit, even though that means using non-default parameters.

ilyamandel:
What I mean is that the current version produces a binary with masses of 8.27 and 17.12 solar masses with a coalescence time of 220 Myr; cf. caption of Figure 15 in the methods paper.  That's fine, the code changed, but no need to try so hard to stick to the methods paper.

image


113.171330   GW merger in 104.9 Myr    8.272    14     17.122    14      7.080  0.004

ilyamandel:
@reinhold-willcox -- did you remember to check that your script ran? :)  It references utils/media/vanDenHeuvel_figures, but the figures are still in utils/media/vanDenHeuval_figures ;-)
[Working on a PR now.]

reinhold-willcox:
~~I did, and confirmed now that it still does. The plot looks nearly identical, but I see now that the final masses and merger time are indeed different. Although, weirdly, I merged in the latest version of dev and it still says Heuval with an 'a'. Is the name change part of a current PR? Edit: I see it, PR #767~~

~~I'm a bit confused though - if the goal is to produce a consistent plot for comparison with the method's paper and to check that the code still works the same, we should try to figure out which setting is causing this minor change and correct it. If the goal is to provide a tutorial for how to run the plotting script (and we don't care so much about consistent output) we should probably set this up differently, e.g in the tutorial directory and w/out reference to the methods paper.~~

Let's continue this in the active PR

PR #767, opened 2022-02-21
Detailed evolution improvements, default change

### LINKED ISSUE #766: van den Heuvel diagrams confusingly list donor masses after MT, rather than masses before MT, as one might expect
image

(Initially identified by @IsobelMarguarethe )

ilyamandel:
- Fixed incorrect location of figures for van den Heuvel diagrams
- Updated default example in example_plots/methods_paper_plots/detailed_evolution (a la figure 15 of methods paper) to not allow CE from radiative donors and have BH natal kicks
- Added change in mass to van den Heuvel diagram
- Replaced detailed_evol_plotter.py with soft link to plot_detailed_evolution.py from utils/PythonScripts/ to avoid keeping track of multiple scripts
- Updated documentation

Resolved #766 .


reinhold-willcox:
I can confirm that the changes reflect what is written above and the plotter works as advertised, with the exception that I don't see any documentation changes. Are you referring to the changelog?

As noted in other places, I'm not sure this is the best approach for the methods paper plot specifically, where users will expect to be able to get back what they see in the paper. 

Additionally, I see that the demo grid file was modified so that the kicks are now 0, meaning all the kick angles can be removed (were we actually kicking the black holes before??). But the masses and separation have also been adjusted, can you comment on why? The original settings were chosen to try to reproduce GW151226, so if we're no longer concerned about that, is the goal simply to produce a binary with a lot of interesting interactions?

ilyamandel:
@reinhold-willcox -- good catch on the documentation, I did that in the wrong file, now properly fixed. :)

The reason we can't reproduce figure 15 from the methods paper exactly is that we had a bug in the code related to the evolution of mass losing HG stars when the methods paper was written.  

As a temporary hack, I picked parameters that only reproduced a GW151226-like binary if radiative-envelope donors could go through CE (not our default!).  I now chose more sensible parameters.  I happen to think that moderately massive BHs probably get negligible natal kicks, hence the new choice of zero kick magnitudes.

If you look at files changed, you can compare the old and new files.

reinhold-willcox (approved):

PR #769, opened 2022-02-22
added fix for SN separation in van den heuvel plots

### LINKED ISSUE #768: Delayed update to binary properties after SN
When a SN happens, the detailed BSE output file only updates the orbital parameters (semi-major axis, eccentricity) on the next step.  This makes plotting look awkward and could confuse users.  See the following example, from example_plots/methods_paper_plots/detailed_evolution [the semi-major axis rightly grows to 2155 Rsun by the next time step]. 

image

reinhold-willcox:
SN events in van den heuvel plots now show correct separation

### DIFF
--- a/utils/PythonScripts/plot_detailed_evolution.py
+++ b/utils/PythonScripts/plot_detailed_evolution.py
@@ -324,8 +324,12 @@ class Event(object):
             self.aprev  = Data['SemiMajorAxis'][ii-1]
         # Cheap kludge for SN separations - later, should clean up when detailed printing is called
         if eventClass == 'SN':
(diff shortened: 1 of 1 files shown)


ilyamandel (changes requested):
Is it guaranteed that there will be an entry with index [ii+1] -- e.g., if the SN unbind the binary and the user does not evolve unbound binaries?  Might be good to include a check on the existence of that index before using it, to avoid a possible error.

reinhold-willcox:
Addressed by adding a try: except: clause. Generally bad python programming, but suitable for now. 

I verified that running a binary that disrupts on SN (without evolving disrupted systems) shows the correct (negative) separation post SN.

PR #776, opened 2022-04-04
Fix logo

QRK1TT13:
README had broken path for COMPAS logo.  Path has been updated.  Logo now shows on repository home


FloorBroekgaarden (approved):
Looks good. This was broken due to a recent move/cleaning of the directories. We should actually check if more things were broken due to this move.

FloorBroekgaarden:
thanks for fixing this and contributing to COMPAS!

Issue #780, opened 2022-04-05
runSubmit boolean parsing


ilyamandel:
runSubmit does not correctly set boolean options to False.  Copying @jeffriley 's suggestion verbatim:

I think this in the runSubmit file

for boolKey, boolVal in self.booleanChoices.items():
    if boolVal is True:
        self.command.update({boolKey: ''}) 
        
needs to change to this

for boolKey, boolVal in self.booleanChoices.items():
    if boolVal is True:
        self.command.update({boolKey: ''}) 
    else:
        self.command.update({boolKey: 'False'})
        
or this (matches my changes to pythonSubmit from a while ago when this problem surfaced (and may be better since Python is dynamically typed - but probably not really necessary...))

for boolKey, boolVal in self.booleanChoices.items():
    if boolVal is True:
        self.command.update({boolKey: ''}) 
    elif boolVal is False:
        self.command.update({boolKey: 'False'})
        
That would be affecting all boolean options (that the user wanted to set false).
I haven't tested it (but it should fix the problem).

PR #781, opened 2022-04-07
Fixed runSubmit.py parsing of booleans from yaml files

avigna:
I added the fix @jeffriley suggested for #780 and it is working as it should.

I include a BSE plot of the mass of the primary as a function of time setting `use-mass-loss: True` (blue), `use-mass-loss: False` (red), and `use-mass-loss: ` (yellow). It is working as expected. Did the same test of `use-mass-loss: True` and `use-mass-loss: False` for SSE (not attached).  

BSE_test
.


jeffriley (approved):
Looks good, thanks @avigna

Issue #774, opened 2022-03-25
Misleading detailed output: wrong (post-MT) stellar type stored during mass transfer episodes


IsobelMarguarethe:
**Describe the bug**
The mass transfer flag happens later than the mass transfer, leading to misleading detailed output. 

**Label the issue**
`urgency_moderate` - This is a moderately urgent issue

**To Reproduce**
1. Run `COMPAS --random-seed 23046 --number-of-systems 1 --detailed-output`
2. Compare the pre- and post-mass transfer stellar type of the primary stored in COMPAS_Output.h5 to BSE_Detailed_Output_0.h5:
```
import h5py

f1 = h5py.File("COMPAS_Output/Detailed_Output/BSE_Detailed_Output_0.h5", "r")
f2 = h5py.File("COMPAS_Output/COMPAS_Output.h5", "r")

times_1 = f1['Time'][()]
types_primary_1 = f1['Stellar_Type(1)'][()]
primary_rl_ratio_1 = f1['Radius(1)|RL'][()]
mt_tracking_1 = f1['MT_History'][()]

primary_rl_ratio_2 = f2['BSE_RLOF']['Radius(1)|RL<step'][()]
types_primary_2_pre_MT = f2['BSE_RLOF']['Stellar_Type(1)<MT'][()]
types_primary_2_post_MT = f2['BSE_RLOF']['Stellar_Type(1)>MT'][()]
times_2 = f2['BSE_RLOF']['Time<MT'][()]

stored_mass_transfer_times_map = (mt_tracking_1 == 1)

print("Times of RLOF: {}".format(times_2))
print("Times at which MT_tracking = 1: {}".format(times_1[stored_mass_transfer_times_map]))
print("Type of primary pre MT: {}".format(types_primary_2_pre_MT))
print("Type of primary post MT: {}".format(types_primary_2_post_MT))
print("Types of primary when MT_tracking = 1: {}".format(types_primary_1[stored_mass_transfer_times_map]))

# Times of RLOF: [4.35815622 6.07168313]
# Times at which MT_tracking = 1: [4.35815622 4.36934683]
# Type of primary pre MT: [ 2 14]
# Type of primary post MT: [ 7 14]
# Types of primary when MT_tracking = 1: [7 7]
```

**Versioning:**
 - MacOS Mojave 10.14.6
 - COMPAS=v02.27.05

reinhold-willcox:
Hey, sorry for the delay, I haven't kept up with git issues!

[...]

Issue #783, opened 2022-04-11
Mergers not always showing up in BSE_RLOF


reinhold-willcox:
**Describe the bug**
When two stars merge in COMPAS, the details of the RLOF show up in the `BSE_RLOF` output (although by default the 'Merger' flag itself is not included in `BSE_RLOF`). 

I'm seeing some cases where there is a merger between a BH and a high mass companion, but no associated output in `BSE_RLOF`. The systems are correctly flagged as a 'Merger' in `BSE_System_Parameters`, as well as the output on the command line. Furthermore, the `BSE_Supernovae` output only contains one entry, for the birth of the BH, even though the other star is very massive (between 40 and 110 Msun) and the orbit is intact following the SN (the kicks are always very weak due to significant fallback in these systems). In a (mostly default) run of 1e4 binaries, I see 89 systems that experience a merger without corresponding merger output.

**Label the issue**
Please label the 'severity' and 'urgency' of this issue. You can choose:

`urgency_moderate` - This is a moderately urgent issue
`severity_moderate` - This is a moderately severe bug

**To Reproduce**
```
COMPAS --initial-mass-1 47.052345928066174 --initial-mass-2 46.64348463422359 -a 1.109865820053667 -
 -random-seed 140 --wolf 0.1
```
Note that wolf-rayet winds do not need to be set to 0.1 to reproduce this, but I happened to have them on when I discovered the bug. 

**Expected behavior**
All mergers should show up in `BSE_RLOF`

**Versioning (please complete the following information):**
 - COMPAS  v02.27.07

reinhold-willcox:
Note: adding the `--hmxr-binaries` flag does not solve the problem.

PR #784, opened 2022-04-11
Missing mergers fix

reinhold-willcox:
Possible fix for issue #783 , in which some mergers between BHs and massive stars do not show up in `BSE_RLOF`. 

Based on the system I posted in that issue, I found that in the final evolution step (for BH+HG binary), within `EvaluateBinary`:
- CalculateMassTransfer determined there was no RLOF and nothing to do
- CalculateWindsMassLoss was removing a small amount of wind from the HG
- It was not a CE or stellar merger, and no SN was occuring, so the code entered the 'else' clause followed by `ResolveMassChanges`
- Here, the binary angular momentum changes s.t the components are now touching, hence m_Flags.stellarMerger is set to true. Then `EvaluateBinary` finishes and returns to `Evolve` where the next call is to `PrintRLOFParameters` which should be resulting in `BSE_RLOF` output. 

However, `PrintRLOFParameters` only logs RLOF Parameters if either star `IsRLOF()`, but by going through ResolveMassChanges, neither star has this set. To set it, the stars must have had `m_StarX->InitialiseMassTransfer()` called (which occurs when `CalculateMassTransfer()` leads to touching, but not when `ResolveMassChanges()` does). 

To fix this, I've added the lines:
```
m_Star1->InitialiseMassTransfer(m_CEDetails.CEEnow, m_SemiMajorAxis, m_Eccentricity);
m_Star2->InitialiseMassTransfer(m_CEDetails.CEEnow, m_SemiMajorAxis, m_Eccentricity); 
```
immediately below `m_Flags.stellarMerger = true;` when the stars are found to be touching following ResolveMassChanges. 

I can verify that this logs the merger in `BSE_RLOF` correctly and thus fixes the issue, but I am not sure that this is necessarily the best way to accomplish this and avoids introducing other complications, so I defer to @jeffriley 's judgement.

### DIFF
--- a/src/BaseBinaryStar.cpp
+++ b/src/BaseBinaryStar.cpp
@@ -2135,7 +2135,7 @@ double BaseBinaryStar::CalculateAngularMomentum(const double p_SemiMajorAxis,
 
 	double Is1  = ks1 * m1 * R1 * R1;
 	double Is2  = ks2 * m2 * R2 * R2;
-    double Jorb = ((m1 * m2) / (m1 + m2)) * std::sqrt(G1 * (m1 + m2) * p_SemiMajorAxis * (1.0 - (p_Eccentricity * p_Eccentricity)));
+    	double Jorb = ((m1 * m2) / (m1 + m2)) * std::sqrt(G1 * (m1 + m2) * p_SemiMajorAxis * (1.0 - (p_Eccentricity * p_Eccentricity)));
 
 	return (Is1 * w1) + (Is2 * w2) + Jorb;
 }
@@ -2250,9 +2250,10 @@ void BaseBinaryStar::EvaluateBinary(const double p_Dt) {
         ResolveMassChanges();                                                                                           // apply mass loss and mass transfer as necessary
         if (HasStarsTouching()) {                                                                                       // if stars emerged from mass transfer as touching, it's a merger
             m_Flags.stellarMerger = true;
-            // Initialise MT for both stars so that the show correct RLOF status
-            m_Star1->SetRocheLobeFlags(m_CEDetails.CEEnow, m_SemiMajorAxis, m_Eccentricity);                                       // initialise mass transfer for star1
-            m_Star2->SetRocheLobeFlags(m_CEDetails.CEEnow, m_SemiMajorAxis, m_Eccentricity);                                       // initialise mass transfer for star2
+		
+            // Set Roche lobe flags for both stars so that they show correct RLOF status
+            m_Star1->SetRocheLobeFlags(m_CEDetails.CEEnow, m_SemiMajorAxis, m_Eccentricity);                            // set Roche lobe flags for star1
+            m_Star2->SetRocheLobeFlags(m_CEDetails.CEEnow, m_SemiMajorAxis, m_Eccentricity);                            // set Roche lobe flags for star2
         }
     }
 
--- a/src/BaseBinaryStar.cpp
+++ b/src/BaseBinaryStar.cpp
@@ -1576,7 +1576,7 @@ void BaseBinaryStar::ResolveCommonEnvelopeEvent() {
         m_Flags.stellarMerger        = true;
     }
     else if ( (m_Star1->DetermineEnvelopeType()==ENVELOPE::RADIATIVE && !m_Star1->IsOneOf(ALL_MAIN_SEQUENCE)) ||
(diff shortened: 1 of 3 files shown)


jeffriley:
I haven't checked, but I would guess that calling 

    m_Star1->InitialiseMassTransfer(m_CEDetails.CEEnow, m_SemiMajorAxis, m_Eccentricity);
    m_Star2->InitialiseMassTransfer(m_CEDetails.CEEnow, m_SemiMajorAxis, m_Eccentricity);

causes the entries to be recorded in  BSE_RLOF because BinaryConstituentStar::InitialiseMassTransfer() calls BinaryConstituentStar::SetRocheLobeFlags(), so sets the RLOF flags correctly for the HG star after the winds mass loss.  The only other thing that BinaryConstituentStar::InitialiseMassTransfer() does is set m_MassTransferDiff = 0.0, which is probably superfluous, but also probably not harmful in this case.

So perhaps just calling 

    m_Star1->SetRocheLobeFlags(m_CEDetails.CEEnow, m_SemiMajorAxis, m_Eccentricity);
    m_Star2->SetRocheLobeFlags(m_CEDetails.CEEnow, m_SemiMajorAxis, m_Eccentricity);

is cleaner.

ilyamandel:
How does the binary go from no RLOF to touching if all that happened in between was a small amount of mass loss through winds?

jeffriley:
> How does the binary go from no RLOF to touching if all that happened in between was a small amount of mass loss through winds?

I don't know how much mass was lost - I assume enough to change the radius to turn the RLOF flag on... It must be on after the mass loss, othwerwise @reinhold-willcox's changes wouldn't have fixed the problem.  Maybe @reinhold-willcox can answer that.

Edit: if it is just mass loss and not mass transfer, shouldn't that widen the system?  So now I'm confused...  For @reinhold-willcox's fix to work (and apparently it does), one of the stars must be in RLOF, so either the star's radius increased or its RL radius decreased.  If the binary is widening wouldn't that mean the RL radius is decreasing?  But for the stars to be now touching one of the stars must have increased its radius...  @reinhold-willcox, maybe we need more details...

jeffriley (changes requested):
Needs changelog and version updates.

Also needs some more investigation before applying this fix - see questions/comments above.

reinhold-willcox:
@ilyamandel By peppering `EvaluateBinary` with std::cout lines, I think I've found the answer. The commented values below are the relevant attributes when printed at that location. The change occurs only after ResolveMassChanges(). What seems to be happening is that, although the timestep is fairly small and the mass loss is minimal, ResolveMassChanges triggers a change in stellar type of the secondary from HG to CHeB, which apparently leads to a huge jump in radius, from 150 to 2000 Rsun, leaping over the 1000 Rsun separation. 
 
@jeffriley You are correct though, the mass loss does in fact widen the binary by a bit, but not nearly enough to offset the radius jump.

 ```
 void BaseBinaryStar::EvaluateBinary(const double p_Dt) {
 
    //p_Dt = 5.74378e-05
 
    //PreMT
    //M1   = 12.8231
    //M2   = 57.1315
    //R1   = 5.43698e-05
    //R2   = 151.701
    //a_Ro = 949.682
    //St1  = Black_Hole
    //St2  = Hertzsprung_Gap
 
    CalculateMassTransfer(p_Dt);
 
    //Between MT and winds
    //M1   = 12.8231
    //M2   = 57.1315
    //R1   = 5.43698e-05
    //R2   = 151.701
    //a_Ro = 949.682
    //St1  = Black_Hole
    //St2  = Hertzsprung_Gap
 
    CalculateWindsMassLoss();
 
    //PostWinds
    //M1   = 12.8231
    //M2   = 57.1315
    //R1   = 5.43698e-05
    //R2   = 151.701
    //a_Ro = 949.682
    //St1  = Black_Hole
    //St2  = Hertzsprung_Gap
 
    if (...some stuff that doesn't matter here...)
    else {
        ResolveMassChanges();
 
        //PostResolveMassChanges
        //M1   = 12.8231
        //M2   = 56.5602
        //R1   = 5.43698e-05
        //R2   = 2176.41
        //a_Ro = 957.567
        //St1  = Black_Hole
        //St2  = Core_Helium_Burning
 
 
        if (HasStarsTouching()) {
            m_Flags.stellarMerger = true;
                                                                                                                                                                                                                                                                                
        }                                                                                                                                                                                                                                                                       
    }                                                                                                                                                                                                                                                                           
```

jeffriley (approved):
All good @reinhold-willcox - I changed some comments around the calls to SetRocheLobeFlags()

reinhold-willcox:
Meant to add previously, this partially resolves #774

Issue #782, opened 2022-04-08
MergerAtZams missing equilibrated binaries which still RLOF


reinhold-willcox:
**Describe the bug**
The MergerAtZams flag is great for determining if two stars were initially positioned too closely. However, some systems do not merge at ZAMS, but very shortly after (1 timestep, <.02 Myr).

Looking closely at these systems, they all undergo RLOF and equilibrate at birth. These equal mass binaries are then checked for whether their radii overlap (2R < a), but not whether they overflow the RL _after_ the equilibration. The ones which overlap are correctly flagged as a merger at zams, and the ones which do not overlap or RLOF (a second time) evolve as CHE binaries as expected. But the subset which do not overlap but _do_ RLOF should be considered as mergers at birth. 

**Label the issue**
`urgency_low`      - This issue is not urgent
`severity_minor`    - This is a minor bug with minimal impact

**To Reproduce**
Steps to reproduce are a bit involved. You need to add the initial radii and simulation time to the `BSE_System_Parameters` output, so create a logfile_defs.txt with the following:

logfile_defs.txt :
```
bse_sysparms_rec += {    
    STAR_1_PROPERTY::RZAMS,    
    STAR_2_PROPERTY::RZAMS,    
    BINARY_PROPERTY::TIME,    
}    
```

Add any of the following seeds to a gridfile:
 ```
1649389693 1649389695 1649389703 1649389722 1649389752 1649389771 1649389786 1649389843 1649389860 1649389868 164938 9875 1649389894 1649389901 1649389999 1649390043 1649390060 1649390063 1649390083 1649390194 1649390240 1649390323 16 49390335 1649390336 1649390342 1649390414 1649390423 1649390477 1649390506 1649390510 1649390512 1649390560 164939058 1 1649390584 1649390594 1649390601 1649390623 1649390624 1649390644  
```

And run with the following flags to allow for touching and rlow at ZAMS to avoid having these systems flagged as errors and thrown out.

`COMPAS --grid grid.txt --logfile-def logfile_defs.txt --semi-major-axis-max 10 --allow-touching --allow-rlof`

Then you can use the following script to see the error (or write your own, as you like).
```
SPs = h5.File('COMPAS_Output/COMPAS_Output.h5', 'r')['BSE_System_Parameters']

isEquil = SPs['Equilibrated_At_Birth'][()] == 1
RZams = SPs['Radius@ZAMS(1)'][()]
[...]

PR #785, opened 2022-04-12
Post-Zams overcontact binaries

reinhold-willcox:
Addresses issue #782, in which overcontact binaries at ZAMS (those which equilibrate at birth, but still have RLOF in both stars without them touching) merge immediately after ZAMS. I added a catch for two MS stars both in RLOF - this is an overcontact binary and so should avoid CEE/merger. Also, this means that two MS stars with nearly equal mass ratio which RLOF at the same time will also enter into an overcontact binary. In this latter case, the two stars do not equilibrate (though they are still nearly equal anyway). In practice, this just means that both stars are in a state of RLOF until they expand enough to touch. 

@ilyamandel We (Alejandro, Jeff, and I) discussed this at the code meeting, and think we covered all the corner cases, but can you confirm that this makes sense theoretically?

Note: leaving this as a draft because I still need to:
- Test the various use cases
- CHE stars which RLOF but do not touch - do they merge or evolve separately?

### DIFF
--- a/src/BaseBinaryStar.cpp
+++ b/src/BaseBinaryStar.cpp
@@ -2370,9 +2370,11 @@ EVOLUTION_STATUS BaseBinaryStar::Evolve() {
             }
             else if (StellarMerger() ) {                                                                                                    // have stars merged?
                 evolutionStatus = EVOLUTION_STATUS::STELLAR_MERGER;                                                                         // for now, stop evolution
+                m_Flags.stellarMerger = true;
             }
             else if (HasStarsTouching()) {                                                                                                  // binary components touching? (should usually be avoided as MT or CE or merger should happen prior to this)
                 evolutionStatus = EVOLUTION_STATUS::STARS_TOUCHING;                                                                         // yes - stop evolution
+                m_Flags.stellarMerger = true;
             }
             else if (IsUnbound() && !OPTIONS->EvolveUnboundSystems()) {                                                                     // binary is unbound and we don't want unbound systems?
                 m_Unbound       = true;                                                                                                     // yes - set the unbound flag (should already be set)
@@ -2391,9 +2393,11 @@ EVOLUTION_STATUS BaseBinaryStar::Evolve() {
                 // check for problems
                 if (StellarMerger()) {                                                                                                      // have stars merged?
                     evolutionStatus = EVOLUTION_STATUS::STELLAR_MERGER;                                                                     // for now, stop evolution
+                    m_Flags.stellarMerger = true;
                 }
                 else if (HasStarsTouching()) {                                                                                              // binary components touching? (should usually be avoided as MT or CE or merger should happen prior to this)
                     evolutionStatus = EVOLUTION_STATUS::STARS_TOUCHING;                                                                     // yes - stop evolution
+                    m_Flags.stellarMerger = true;
                 }
                 else if (IsUnbound() && !OPTIONS->EvolveUnboundSystems()) {                                                                 // binary is unbound and we don't want unbound systems?
                     evolutionStatus = EVOLUTION_STATUS::UNBOUND;                                                                            // stop evolution
@@ -2438,12 +2442,14 @@ EVOLUTION_STATUS BaseBinaryStar::Evolve() {
                 dt = std::min(m_Star1->CalculateTimestep(), m_Star2->CalculateTimestep()) * OPTIONS->TimestepMultiplier();                  // new timestep
                 if ((m_Star1->IsOneOf({ STELLAR_TYPE::MASSLESS_REMNANT }) || m_Star2->IsOneOf({ STELLAR_TYPE::MASSLESS_REMNANT })) || dt < NUCLEAR_MINIMUM_TIMESTEP) {
                     dt = NUCLEAR_MINIMUM_TIMESTEP;                                                                                          // but not less than minimum
-		}
+		        }
                 stepNum++;                                                                                                                  // increment stepNum
             }
         }
-        if (!StellarMerger())
-            (void)PrintDetailedOutput(m_Id);                                                                                                // print (log) detailed output for binary
+        if (StellarMerger())
+            (void)PrintRLOFParameters();                                                                                                    // print (log) RLOF parameters for mergers
+
+        (void)PrintDetailedOutput(m_Id);                                                                                                    // print (log) detailed output for binary
 
         if (evolutionStatus == EVOLUTION_STATUS::STEPS_UP) {                                                                                // stopped because max timesteps reached?
             SHOW_ERROR(ERROR::BINARY_EVOLUTION_STOPPED);                                                                                    // show error
--- a/src/BaseBinaryStar.cpp
+++ b/src/BaseBinaryStar.cpp
@@ -1831,8 +1831,13 @@ void BaseBinaryStar::CalculateMassTransfer(const double p_Dt) {
     }
 
     if (m_Star1->IsRLOF() && m_Star2->IsRLOF()) {                                                                               // both stars overflowing their Roche Lobe?
-        m_CEDetails.CEEnow = true;                                                                                              // yes - common envelope event - no mass transfer
-        return;                                                                                                                 // and return - nothing (else) to do
+        if (HasTwoOf({ STELLAR_TYPE::MS_LTE_07, STELLAR_TYPE::MS_GT_07 })) {                                                        // yes - but both are MS, so this is an over-contact binary
+            return;                                                                                                                 // so it stays in RLOF until they touch (no common envelope)
+        }
+        else {
+            m_CEDetails.CEEnow = true;                                                                                              // yes - and one is evolved so enter a common envelope event 
+            return;                                                                                                                 // - no mass transfer and return 
+        }
     }
 
     // one, and only one, star is overflowing its Roche Lobe - resolve mass transfer
--- a/src/BaseBinaryStar.cpp
+++ b/src/BaseBinaryStar.cpp
@@ -2370,7 +2370,6 @@ EVOLUTION_STATUS BaseBinaryStar::Evolve() {
             }
             else if (StellarMerger() ) {                                                                                                    // have stars merged?
                 evolutionStatus = EVOLUTION_STATUS::STELLAR_MERGER;                                                                         // for now, stop evolution
-                m_Flags.stellarMerger = true;
             }
             else if (HasStarsTouching()) {                                                                                                  // binary components touching? (should usually be avoided as MT or CE or merger should happen prior to this)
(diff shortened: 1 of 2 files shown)


jeffriley (commented):
@reinhold-willcox minor nit pick that is a matter of style and perhaps just personal preference, butI would write this

    if (m_Star1->IsRLOF() && m_Star2->IsRLOF()) {                                                                               // both stars overflowing their Roche Lobe?
        if (HasTwoOf({ STELLAR_TYPE::MS_LTE_07, STELLAR_TYPE::MS_GT_07 })) {                                                    // yes - but both are MS, so this is an over-contact binary
            return;                                                                                                             // so it stays in RLOF until they touch (no common envelope)
        }
        else {
            m_CEDetails.CEEnow = true;                                                                                          // yes - and one is evolved so enter a common envelope event 
            return;                                                                                                             // - no mass transfer and return 
        }
    }


as

    if (m_Star1->IsRLOF() && m_Star2->IsRLOF()) {                                                                               // both stars overflowing their Roche Lobe?
        if (!HasTwoOf({ STELLAR_TYPE::MS_LTE_07, STELLAR_TYPE::MS_GT_07 })) {                                                   // yes - but at least one is evolved so ... 
            m_CEDetails.CEEnow = true;                                                                                          // ...enter a common envelope event
        }                                                                                                                       // otherwise it's an over-contact binary and stays in RLOF until they touch (no common envelope)                                   
        return;                                                                                                                 // no mass transfer and return 
    }


I think it is cleaner and easier to read.

reinhold-willcox:
Good point. That's fixed now.

reinhold-willcox:
@jeffriley I found that many of these overcontact binaries were not showing up as mergers (in System Params or RLOF). I realized they were merging following the SSE module, not BSE, and we didn't have the appropriate catches in place. I've added them in, so now things print as expected, but let me know if there's a better way to do this.

jeffriley:
In BaseBinaryStar::Evolve() you have

    2371            else if (StellarMerger() ) {                                                                                                    // have stars merged?
    2372                evolutionStatus = EVOLUTION_STATUS::STELLAR_MERGER;                                                                         // for now, stop evolution
    2373                m_Flags.stellarMerger = true;

and

    2394                if (StellarMerger()) {                                                                                                      // have stars merged?
    2395                    evolutionStatus = EVOLUTION_STATUS::STELLAR_MERGER;                                                                     // for now, stop evolution
    2396                    m_Flags.stellarMerger = true;

The addition of lines 2373 and 2396 is not necessary - those lines are superfluous, because in BaseBinaryStar.h, StellarMerger() is 

    bool StellarMerger() const { return m_Flags.stellarMerger; }

so lines 2373 and 2396 will only be executed if m_Flags.stellarMerger is already true - no point setting it true again.

I'm not sure about the added lines to set m_Flags.stellarMerger = true in the "if (HasStarsTouching())" blocks - I haven't looked at the code in detail.  I'm wondering why we wouldn't have noticed before now that the stars had merged?  Ah, maybe because we now don't have a CE event for two MS stars?

I'm not sure what you mean by "I realized they were merging following the SSE module, not BSE".  They can only merge in the BSE code - what exactly do you mean?


Also, in BaseBinaryStar::Evolve() you have changed this

    2443        if (!StellarMerger())
    2444            (void)PrintDetailedOutput(m_Id);   

to this

    2449        if (StellarMerger())
    2450            (void)PrintRLOFParameters();                                                                                                    // print (log) RLOF parameters for mergers
    2451
    2452        (void)PrintDetailedOutput(m_Id);

Without your change, PrintDetailedOutput() is called only if there was no stellar merger.  With your change, PrintDetailedOutput() is always called - is that what you want?

I'm not sure about adding the call to PrintRLOFParameters() here - it will be called after the evolution of the binary completes if there was a stellar merger, so this is an added call for stellar mergers not involving over-contact binaries (it's an added call for all stellar mergers, but presumably mergers not involving over-contact binaries were printing ok before, so now there will be an added entry won't there?).

I'll take a better look later.

reinhold-willcox:
To your first point, yes I agree, I'll fix that now. 

Regarding the "HasStarsTouching" if they are touching, they are a merger, so it seems to me that these are just two different conditions for the same physics. There's probably a more elegant way to combine them, but functionally I just want to be able to identify anything that merges as a merger, and log the RLOF as the final step. That's why I added in the `if StellarMerger() {PrintRLOF;}` call at the end. I'm also not sure why we had the `PrintDetailedOutput` ignored for mergers, it seems like that would be a useful output to have for debugging.

jeffriley:
> Regarding the "HasStarsTouching" if they are touching, they are a merger, so it seems to me that these are just two different conditions for the same physics. 

Understand.  The check for stars touching is from the legacy code from memory, so I suspect that we were seeing binaries where the stars were touching but hadn't been flagged as a merger.  I suppose that might be possible on the very last timestep - maybe.  Or maybe it was put there to catch an anomalous situation.  I think rather than just flag the binary as a merger at that point we should try to find out if that code is ever executed, and if it is why the binary wasn't flagged as a merger before we got there.  If it's just that after the final timestep one or both stars have puffed up a bit and if there was another timestep we'd catch the merger, then flagging the merger there is probably ok.  But if something else is going on we should try to identify and fix it.  It might be legit, but it also might not be.

> I just want to be able to identify anything that merges as a merger, and log the RLOF as the final step.  

Yeah, I guess I'm wondering why we didn't do that earlier - same as above.  RLOF isn't merger, but it's a precursor, so if the stars are touching shouldn't there have been an RLOF event?  Why do we need to trigger it here?  Is it possible that this all happened in the very last timestep, and that's why we didn't trigger an RLOF event?  Triggering it here might just be putting a band-aid over a problem that we haven't yet identified.  Do we know?

This may expose my lack of understanding of the BSE_RLOF file.  When I refactored the code I just tidied up Coen's original RLOF implementation.  Then we removed it, then later Mike reinstated it, then you made some changes.  I don't really know in detail what it does - I thought an entry is put in the file whenever a star is undergoing RLOF, or at least that's what triggers the possibility of adding an entry to the file.  Is that so, or are there other triggers for adding an entry to the RLOF file?  The reason I ask is that you have said a number of times that you don't see the merger in the RLOF file - do you mean that there should be a merger event in the RLOF file, or are you just indicating that the RLOF event that led to the merger is not in the RLOF file?

> That's why I added in the if StellarMerger() {PrintRLOF;}

So the RLOF event is already in the stashed RLOF events, we just haven't written it to the file yet?  If that's the case, we probably should track what events have been written to the file, and if there are residual events that haven't yet been written to the file, we should write them - but the check should be for residual events not yet written to the file, not for a stellar merger.

> I'm also not sure why we had the PrintDetailedOutput ignored for mergers, it seems like that would be a useful output to have for debugging.

This is from BaseBinaryStar::EvaluateBinary():

        if (m_PrintExtraDetailedOutput == true && !StellarMerger()) { (void)PrintDetailedOutput(m_Id); }                    // print detailed output record if stellar type changed (except on merger, when detailed output is meaningless)

I think @ilyamandel  put that comment there, and I suspect he also added the check that you removed, so maybe talk to him.

reinhold-willcox:
The issues cropping up now are indeed when the radius changes rapidly enough that touching occurs without prior RLOF. That's not intrinsically a problem (in my opinion) as long as we log the RLOF and call it a merger. There are probably better, more robust fixes that we should implement in the long term (combining mergers and touching into one condition), but for now this band-aid fix solves the problem of missing mergers and is ready to go.

I generally expect BSE_RLOF output whenever the stars interact, since that makes intuitive sense. We do log RLOF whenever a star is RLOFing, but as noted earlier in this PR, and in PR #784 , we sometimes do not log RLOF correctly and thus miss some mergers. Meaning there is a merger event, but no corresponding output in BSE_RLOF. There are definitely other issues with the RLOF printing, e.g Isobel's other bug that a single RLOF event had two corresponding DetailedOutput columns flagged for it. But those will take a bit more time to sort out and fix cleanly. 

I disagree that printing detailed output at mergers is meaningless. Even if some of the parameters don't quite make sense, it's still useful to see what their values are.

jeffriley:
> The issues cropping up now are indeed when the radius changes rapidly enough that touching occurs without prior RLOF. That's not intrinsically a problem (in my opinion) as long as we log the RLOF and call it a merger. 

So the touching happens in the final timestep, and we check for RLOF before we modify the radius of the stars - that's why we don't trigger an RLOF event? 

> but for now this band-aid fix solves the problem of missing mergers and is ready to go.

Band-aid fixes have a habit of never being fixed properly, and we end up with poorly-written code peppered throughout the code base.  An issue should be opened to fix the problem properly, and the band-aid fix documented in that issue so we know what to remove when the issue is fixed properly.

This needs to be addressed (from above): "... but the check should be for residual events not yet written to the file, not for a stellar merger."  If you want to use the check for stellar merger as a band-aid fix, then (as above) an issue should be opened to fix the problem properly, and the band-aid fix documented in that issue so we know what to remove when the issue is fixed properly.

> I disagree that printing detailed output at mergers is meaningless. Even if some of the parameters don't quite make sense, it's still useful to see what their values are.

I have no preference, but if you're going to remove the check in one place you should remove it in both.

jeffriley:
> The issues cropping up now are indeed when the radius changes rapidly enough that touching occurs without prior RLOF. That's not intrinsically a problem (in my opinion) as long as we log the RLOF and call it a merger.

You added

    2449        if (StellarMerger())
    2450            (void)PrintRLOFParameters();   

PrintRLOFParameters() will only write a record to the BSE_RLOF file if the RLOF flag is set for one (or both) of the stars - so we *did* record an RLOF event prior to the stars touching.  The question is why didn't we see the merger if we saw the RLOF event?

I think we need to check for merger of over-contact binaries here

    if (m_Star1->IsRLOF() && m_Star2->IsRLOF()) {                                                                               // both stars overflowing their Roche Lobe?
        if (!HasTwoOf({ STELLAR_TYPE::MS_LTE_07, STELLAR_TYPE::MS_GT_07 })) {                                                   // yes - but at least one is evolved so ... 
            m_CEDetails.CEEnow = true;                                                                                          // ...enter a common envelope event
        }                                                                                                                       // otherwise it's an over-contact binary and stays in RLOF until they touch (no common envelope)                                   
        return;                                                                                                                 // no mass transfer and return 
    }

Prior to this change we would flag a common envelope event.  We changed that to only flag the CE for non-over-contact binaries, but we should, I think, add a check for over-contact binaries touching here, and if they are, call it a merger.

reinhold-willcox:
Based on discussions with the broader group, we've decided that what we currently do is acceptable for now. Specifically, when --allow-rlof is on, we implicitly "allow for over-contact binaries" which will merge early on the MS anyway. Currently, these will merge after the first evolution of a time dT, regardless of how big the timestep is. A better approach would be to ensure that these overcontact binaries follow a more careful prescription for overcontact binaries, but until some future user takes that on as a project, for now this will do. 

Closing this PR.

PR #787, opened 2022-04-19
improvements to rlof printing

reinhold-willcox:
Here's my first go at improving the BSE_RLOF output. I'll start with some comments on what's not working, then describe my changes. This is a draft PR while I get feedback on some of the points here, but it works reasonably well so we may choose to open it fully as is. 

What's not working:
- Some of the post MT parameters do not accurately reflect the post RLOF system. This is especially true when there is a SN in the same timestep, and the post RLOF values reflect those post SN, not just post RLOF (e.g negative separation). This is a result of the somewhat convoluted way we change parameters in a timestep and when during the timestep we log the ones for RLOF. 
- `BSE_RLOF` was missing some useful output parameters, and including some others which were less useful. I've added the useful ones, and commented the others out by default (they can be easily readded with a logfile). 

Changes to the `StashRLOFProperties(MASS_TRANSFER_TIMING::POST_MT)` calling locations:
- I played with this a lot, but it's a very non-trivial problem. The issue is that at different points throughout the EvaluateBinary code, we change the mass or the stellar type or the semimajoraxis (say), and at no point is everything aligned for what we would intuitively call "PostRLOF" for every timestep. This could be alleviated if we did not allow for RLOF and SNe in the same timestep, or if we tried some other major overhaul of when/how we call various parameters, with the goal of having everything consistently set at some interim point in the timestep. Both of these require signifiant work and thinking and are unlikely to be accomplished on a short timescale.
- I've removed the call contained in "PrintRLOFParameters". This call occurs after EvaluateBinary, and so after any SN, and is thus too late in the evolution. 
- I've added in several calls earlier in EvaluateBinary. These are designed to capture any changes from mass transfer that will affect the PostRLOF parameters, but not be affected by the SN (or a merger). 
- Note that ResolveMassChanges initiates the SN in the single star (if it's time to do so), so the mass, radius, etc. are all incorrect after this point, and the Prev values do not reflect the preSN values. So I did not add a `StashRLOF` after ResolveMassChanges, unless the system is also a merger.
- The best solution I could find to getting as many of the parameters as possible to be correct at the same time pre-SN was to set them all within EvaluateSupernova, using the m_Supernova values. This is a bit hacky and not a great long term solution, but as noted above, the long term solutions are all big overhauls. 

Added output parameters:
- The He and CO core masses for both stars (< and >MT) 
- Whether the RLOF resulted in a merger 
- We previously had `Time<MT` and `Time>MT`, but they were the same value. I've removed one of them, and replaced the other with just `Time`

Removed output parameters:
- Radius/RocheLobeRadius for both stars before and after. This was not the correct value, and can be easily calculated from other values in the post-processing.
- All the Zeta parameters.

Results:
- All the mass parameters (including the new He and CO core masses) are correct before and after, including for SNe. 
- These values are not correct if there is a merger, but the added merger flag allows the user to know when these can be ignored. 
- SemiMajorAxis and Eccentricity before and after are correct. This is especially crucial information to have for e.g USSNe, where the SN likely occurs after a RLOF episode and the separation post MT is of key interest. 
- The radius and stellar type are _incorrect_ post RLOF, and reflect the values of the post SN object. I could not find a way around this. At the very least, radius is not usually of huge interest, especially if the system immediately undergoes SN, and we can usually guess the post RLOF stellar type from the donor type.


reinhold-willcox:
Closing as outdated

Issue #741, opened 2022-01-18
Update stroopwafelInterface.py and documentation to use runSubmit.py


avigna:
Recently, the old classic `pythonSubmit.py` script was changed to a tandem of a `runSubmit.py` script plus a `compasConfigDefault.yaml` options file (solved in #720 and documented in #738 ).

However, we need to make sure that sampling done via Stroopwafel, e.g., in /utils/preProcessing/stroopwafelInterface.py, is

- [x] available for users, and 
- [x] properly documented.

 - OS: MacOS 12.0.1
 - COMPAS v02.27.00

@FloorBroekgaarden , I tagged you for this, but let me know if I should include someone else. Happy to give it a go it what would be probably small and simple hackaton.

avigna:
This is now complete following #791.

Issue #753, opened 2022-02-04
Generating binaries using stroopwafelinterface.py


syedalimohsinbukhari:
Hi,

When I generate binaries using `stroopwafelinterface.py` using the default values it gives me a 0 rate_of_hits. I even tried it for 10M binaries but still got the same result.

```python
### Include options from local pythonSubmit file      
usePythonSubmit = False #If false, use stroopwafel defaults

### Set default stroopwafel inputs - these are overwritten by any command-line arguments

compas_executable = os.path.join(os.environ.get('COMPAS_ROOT_DIR'), 'src/COMPAS')   # Location of the executable      # Note: overrides pythonSubmit value
num_systems = int(1e7)              # Number of binary systems to evolve                                              # Note: overrides pythonSubmit value
output_folder = 'output/'           # Location of output folder (relative to cwd)                                     # Note: overrides pythonSubmit value
random_seed_base = 0                # The initial random seed to increment from                                       # Note: overrides pythonSubmit value

num_cores = 10                      # Number of cores to parallelize over 
num_per_core = int(1e6)             # Number of binaries per batch
mc_only = True                      # Exclude adaptive importance sampling (currently not implemented, leave set to True)
run_on_hpc = False                  # Run on slurm based cluster HPC

output_filename = 'samples.csv'     # output filename for the stroopwafel samples
debug = False                       # show COMPAS output/errors
```
And the resulting log file is
```text
rate_of_hits = 0.000000
uncertainity = 0.000000
```
Also, I noticed that the `interesting_systems` function is reading the COMPAS CSV files. Is it because I'm not providing it with these files?
```python
def interesting_systems(batch):
    """
    This is a mandatory function, it tells stroopwafel what an interesting system is. User is free to define whatever looks interesting to them.
    IN:
        batch (dict): As input you will be given the current batch which just finished its execution. You can take in all the keys you defined in the configure_code_run method above
    OUT:
        Number of interesting systems
        In the below example, I define all the NSs as interesting, so I read the files, get the SEED from the system_params file and define the key is_hit in the end for all interesting systems 
    """
    try:
        folder = os.path.join(output_folder, batch['output_container'])
        shutil.move(batch['grid_filename'], folder + '/grid_' + str(batch['number']) + '.csv')
        system_parameters = pd.read_csv(folder + '/BSE_System_Parameters.csv', skiprows = 2)
        system_parameters.rename(columns = lambda x: x.strip(), inplace = True)
[...]

PR #789, opened 2022-04-21
minor changes in stroopwafelinterface to work with runSubmit

LiekeVanSon:
Resolving issues raised in #741 .

Stroopwafel now runs without an error. MC only runs properly.
Full AIS needs to be tested more thoroughly

### DIFF
--- a/utils/preProcessing/stroopwafelInterface.py
+++ b/utils/preProcessing/stroopwafelInterface.py
@@ -18,7 +18,7 @@ import argparse
 #######################################################
 
 
-### Include options from local runSubmit file      
+### Include options from local runSubmit + compasConfigDefault files      
 userunSubmit = False #If false, use stroopwafel defaults
 
 ### Set default stroopwafel inputs - these are overwritten by any command-line arguments
--- a/utils/preProcessing/stroopwafelInterface.py
+++ b/utils/preProcessing/stroopwafelInterface.py
@@ -234,10 +234,11 @@ if __name__ == '__main__':
     # Over-ride with pythonSubmit parameters, if desired
     if usePythonSubmit:
         try:
-            from pythonSubmit import pythonProgramOptions
-            programOptions = pythonProgramOptions()
-            pySubOptions = programOptions.generateCommandLineOptionsDict()
+            from runSubmit import pythonProgramOptions
+            programOptions = pythonProgramOptions()   # Call the programoption class from runSubmit
+            pySubOptions   = programOptions.command   # Get the dict from pythonProgramOptions
 
+            # Continue to work from the dict, by edditing SW related options
             # Remove extraneous options
             pySubOptions.pop('compas_executable', None)
             pySubOptions.pop('--grid', None)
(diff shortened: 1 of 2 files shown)


utils/preProcessing/stroopwafelInterface.py
```diff
@@ -234,10 +234,11 @@ def rejected_systems(locations, dimensions):
     # Over-ride with pythonSubmit parameters, if desired
```
avigna: Change this comment to:
Over-ride with runSubmit + compasConfigDefault parameters, if desired

Same in line 22 of this file, where the comment "Include options from local pythonSubmit file" should be "Include options from local runSubmit + compasConfigDefault files".

avigna (changes requested):
Address changes in comments.

avigna:
On the side: closing this issue *might* help with #753

avigna (approved):

stellarphysicsandpopulations:
(Stephen) 
Lieke's fix worked great for me, except that I needed to pass the full path to the YAML file to runSubmit, even though it's the same directory as my runSubmit. (Edit: Using "!" for comment lines below, since hashes give weird formatting.)

!! programOptions = pythonProgramOptions()
!! the above doesn't work, somehow... I needed to pass the explicit location of the yaml file, so:
yaml_file_location = os.path.join(path_to_runSubmit, 'compasConfigDefault.yaml')
programOptions = pythonProgramOptions(config_file=yaml_file_location)

I don't know why. I suspect this is *not* because of the changes in the Stroopwafel version I'm using. I guess it's because I'm using the Docker version of COMPAS?  (Or maybe my installation is also somehow weird in a different way...)

(So far I'm hard-coding an assumption about the name of the file, and that it is in the same location as runSubmit.py, but that can be changed.)

(Tagging @FloorBroekgaarden too, in case this saves time for others.)

stellarphysicsandpopulations:
(Stephen) 
Lieke's fix worked great for me, except that I needed to pass the full path to the YAML file to runSubmit, even though it's the same directory as my runSubmit:

##### programOptions = pythonProgramOptions()
##### the above doesn't work, somehow... I needed to pass the explicit location of the yaml file, so:
yaml_file_location = os.path.join(path_to_runSubmit, 'compasConfigDefault.yaml')
programOptions = pythonProgramOptions(config_file=yaml_file_location)

I don't know why. I suspect this is *not* because of the changes in the Stroopwafel version I'm using. I guess it's because I'm using the Docker version of COMPAS?  (Or maybe my installation is also somehow weird in a different way...)

(So far I'm hard-coding an assumption about the name of the file, and that it is in the same location as runSubmit.py, but that can be changed.)

(Tagging @FloorBroekgaarden too, in case this saves time for others.)

stellarphysicsandpopulations:
Thinking more clearly, it can't be anything to do with me using the Docker version, at least not directly... this setup is well before we get to the Docker call.  Interesting.  Oh well, it was fun to investigate, whatever the reason.

avigna:
What sort of error do you get, Stephen?

stellarphysicsandpopulations:
With the fix above, no error.  If I don't pass the full path to the YAML file then runSubmit somehow doesn't find the YAML file (even though it's in the same directory).   

Not an issue for me, now.  Whether it's important or not for other depends on how weird my case it.

PR #792, opened 2022-04-29
Fryer2022

LiekeVanSon:
New Fryer 2022 prescription for the remnant mass! :)
Based on paper that just appeared on the ArXiv (https://ui.adsabs.harvard.edu/abs/2022arXiv220413025F/abstract)

One thing that is not immediately clear from the paper is what is used to calculate "the fallback fraction" because there is no longer a need for Mproto. I've emailed Chris Fryer about this and he confirmed that the fallback fraction is still calculated assuming the Mproto masses from the 2012 paper. 

I have run and tested the outcome on v02.26, and mass distributions look largely the same whether you use Mproto from the Delayed or the Rapid prescription. See plots attached: This shows the mass distribution of more massive componentes from BBHs and BHNS at z < 0.5. These are runs of 10^6 binaries using fiducial COMPAS but with Fryer 2022, R for Mproto from RAPID, and  D for Mproto from DELAYED. I attached the extremes fmix = 0.5 (closest to old DELAYED) and fmix = 4.0 (closest to old RAPID). 

B1A3A45B-17B5-41EA-9C4A-BFF4C6BFD14D
51C361C3-ADD7-41F9-AD8A-2D29F41FE533
DB3AB369-6B61-4D51-BF92-99BBA48088CE
7CD395AE-6C40-4C37-9141-D72AE391A3BE


jeffriley (changes requested):
Hi Lieke, thanks for doing this.  I'll leave the code review to @ilyamandel , but I note you haven't updated the changelog, the version number, or the docs - you'll need to do that.  Since this isn't a defect repair, and it's more than a minor enhancement, I suggest you increment the version minor number (the middle digits), and reset the revision number (the rightmost digits) to 0.  (So the current version is (I think) 02.27.09, you should change that to 02.28.00).

~~Also, did you mean to leave this in:~~

      SAY(" baryonicRemnantMass = " << baryonicRemnantMass << " OPTIONS->Fryer22fmix() = " << OPTIONS->Fryer22fmix() << " p_COCoreMass = " << p_COCoreMass);

~~Do we want that printed every time GiantBranch::CalculateRemnantMassByFryer2022() is called?  If you really do want to see it for every binary it probably should go into one of the logfiles.~~

Ignore that - I must have been looking at one of your earlier commits...

ilyamandel (changes requested):
@LiekeVanSon -- Looks great, thank you!  But a few more requests in addition to @jeffriley's comments:

Line 1472 of GiantBranch.cpp -- "* @param [IN] OPTIONS->Fryer22fmix()" -- that's not an input parameter to the function, so please drop that line.

Line 1434 of Options.cpp -- "Critical CO core mass mass" -- repeated 'mass'

Line 732 of Options.h -- did the comment explaining "double m_NeutrinoMassLossValueBH;" get deleted?

Lines 735, 736 of Options.h -- looks like the comments got flipped for fmix and mcrit

LiekeVanSon:
Hi Ilya and Jeff, 

Thanks both for checking this out, and sorry for waiting so long with implementing your very minor comments!

I've adjusted the comments and updated the online docs/ program options list default. Please let me know if that is what you meant or if I missed some part of the docs.

ilyamandel (approved):
Looks great -- thank you, @LiekeVanSon !

ilyamandel (approved):
All good.

ilyamandel:
It seems that I can't approve changes that @jeffriley requested -- Jeff, can you check?

jeffriley (approved):
Looks good.  Thanks Lieke.

Issue #596, opened 2021-07-22
Implement ECSN fixes from my kicks paper


reinhold-willcox:
**Is your feature request related to a problem? Please describe.**
We should implement the ECSN fixes proposed in my NS kicks paper (arxiv:2107.04251). 

**Describe the solution you'd like**
In the paper, we propose that the ZAMS mass range for ECSN progenitors should be reduced to below 0.2Msol, but this is not that practical to implement directly in COMPAS because we don't trigger ECSNe based on the ZAMS mass. The simplest reasonable solution is to adjust the trigger for ECSN to account for previous MT events (or simply H env mass). 

For stripped progenitors which _should_ go ECSN, we can play with the values of e.g Mcbur1,2 and see how that affects predictions for DNS merger rates. That aspect might be worth a follow up (mini) paper, perhaps around the time of the O3b data release to better compare with any updated rates from the LVK gang. 

**Describe alternatives you've considered**
Leave things as is, I guess?

ilyamandel:
Suggestion:
add a new option (something like noECSNinSingleStars); if that option is set, check if a star that goes into the ECSN branch has a hydrogen envelope, and if it does, it stays a WD, avoiding an ECSN.

ilyamandel:
Addressed in #795 , closing.

PR #795, opened 2022-05-05
Ecsn implement willcox et al2021

reinhold-willcox:
Implemented new option "--allow-H-rich-ECSN" which defaults to _false_. 

If false, stars which would otherwise undergo ECSN are checked for whether they are H poor, as a proxy for whether they have interacted with a binary companion. If they are H rich (i.e non-interacting), they just switch over to ONeWD instead of undergoing ECSN, consistent with the findings of Willcox+ 2021. 

If true, ECSN candidates are not checked for H content (the current default). 

Addresses issues #596. I verified that when the flag is set to true, the behavior is identical to current dev, and that when false, there are no H rich ECSNe.


jeffriley (changes requested):
EDIT: Mixed up my reviews - had review for PR796 here...

Code changes look ok.

Also, because this is a non-trivial enhancement, I suggest you increment the version minor number (the middle digits), and reset the revision number (the rightmost digits) to 0 (the revision number is really for defect repairs and trivial enhancements).  So rather than go from 02.27.09 to 02.27.10, I think you should instead go to 02.28.00.

@ilyamandel can comment on the physics (and any other code issues)

reinhold-willcox:
@jeffriley I'm a bit lost, do you still want me to update the minor and revision digits, or was that only for the other PR?

jeffriley:
Both :-)  Any non-trivial enhancement should advance the version number - the minor number (middle digits) if it's just adding new/changed functionality that doesn't cause users too much disruption in how they use the code (i.e. essentially backwards compatible - maybe there's a new option), and the major number (leftmost digits) if the change is significant enough that users will be impacted a bit more significantly (maybe not so backwards compatible - maybe some of the fundamentals are different and users need to adapt their input files/post-processing code to the changes).  The revision number should just be for defect repairs and trivial enhancements.

ilyamandel (changes requested):
When making changes like this, I generally prefer leaving the default the same as the old behaviour initially and changing that later, after getting the group's consensus, so I'd set the option to be true by default for now.

In GiantBranch.cpp:
line 1644: m_Mass = m_COCoreMass; // defined in constant.h
I don't think you need this comment (also, there's no such thing as constant.h :) )

lines 1648 and 1649 -- these lines are commented out, please remove them

The new option should be incorporated into the default .yaml file.

reinhold-willcox:
I think I have addressed all the concerns, let me know if I missed any.

ilyamandel (approved):
Thanks, @reinhold-willcox , looks good to me.  
@jeffriley will need to approve as well, since he also requested changes.
Note that the changelog will need to be updated if Lieke's PR is merged first.

jeffriley (approved):
Approved - but as @ilyamandel pointed out, there a some merge conflicts now that Lieke's PR was approved and merged

reinhold-willcox:
done

PR #796, opened 2022-05-08
Added stable MT angular momentum loss gamma prescription from MacLeod & Loeb 2020

reinhold-willcox:
Added MacLeod_LINEAR prescription for angular momentum loss in stable mass transfer. 

This is a linear interpolater between gamma_ac and gamma_L2 (ang. mom. taken from the accretor and the L2 point respectively). The fraction between gamma_ac and gamma_L2 is also a new input option.


jeffriley (changes requested):
Prior to these changes each case block in the switch statement at line 1671 in BaseBinaryStar.cpp was just a single line of code (except for the default case), so each block could be written clearly on a single line, with the 'break' statement at the end of the line, and the existence of the 'break' statement was obvious. With the addition of cases that contain multiple lines of code, having the 'break' statement at the end of the last line is somewhat less obvious. I think it would be better to move the 'break' statement to a new line for case blocks that have multiple lines of code.

Also, because this is a non-trivial enhancement, I suggest you increment the version minor number (the middle digits), and reset the revision number (the rightmost digits) to 0 (the revision number is really for defect repairs and trivial enhancements). So rather than go from 02.27.09 to 02.27.10, I think you should instead go to 02.28.00.

@LiekeVanSon can comment on the physics (and any other code issues)

reinhold-willcox:
@jeffriley agreed on the clarity, both changes have been made.

ilyamandel (changes requested):
Don't forget to update the default .yaml file when adding a new option

ilyamandel (changes requested):
Why is the L2 point specific angular momentum the same as that of a circumbinary disk at a radius of 2 a?

I think it's discussed around Eq. (11) and (12) of https://articles.adsabs.harvard.edu/pdf/1998CoSka..28..101P, and is a function of the mass ratio...

reinhold-willcox:
Generally you are correct and they are not the same. But on the line above, we have:

```
        case MT_ANGULAR_MOMENTUM_LOSS_PRESCRIPTION::CIRCUMBINARY_RING    :                                                  // Based on the assumption that a_ring ~= 2*a*, Vinciguerra+, 2020 
            gamma = (M_SQRT2 * (p_DonorMass + p_AccretorMass) * (p_DonorMass + p_AccretorMass)) / (p_DonorMass * p_AccretorMass); break; // same as L2 point
```

I added the extra comment `// same as L2 point`, which I agree is not strictly correct. But, the prefactor here, sqrt(2), is within 2% of the value given by MacLeod & Loeb of (1.2)^2. So in practice, it does not matter, but if we are strict about the MacLeod prefactor, we may want to implement a more flexible value for the circumbinary ring prefactor.

ilyamandel:
If I understand correctly, the MacLeod prefactor is itself an approximation to the expression from the article I cited, https://articles.adsabs.harvard.edu/pdf/1998CoSka..28..101P.

jeffriley (approved):
I'm ok with the changes I requested

reinhold-willcox:
@ilyamandel I agree, the MacLeod & Loeb prefactor comes directly from the Pribulla paper you cite. They comment that it is the dimensionless change in separation of the L2 point (squared), and a slowly varying function of the mass ratio. I could implement this more carefully as the q-dependent fluctuation of L2, if you'd like.

ilyamandel (approved):
All good.

Issue #213, opened 2020-04-08
Initial conditions: Mind your p's and q's


ilyamandel:
Include the option of sampling from the correlated initial distribution in mass, mass ratio, separation and binarity proposed by Moe and Di Stefano rather than our current uncorrelated initial condition distribution.

SimonStevenson:
I agree. I think this will be much easier using the grid infrastructure. I think this would be a good thing to have in the python pre-processing scripts to generate initial conditions which we are hoping to put together (see issue #71).

ilyamandel:
Assigning to @Thillai96 -- but we agreed this is best handled through a grid file, outside the C++ code.

ilyamandel:
Implemented in #798 , to be closed once that's pulled.

PR #798, opened 2022-05-11
added moe di stefano sampler

reinhold-willcox:
This sampler follows the Moe & DiStefano 2017 distributions. It's adapted from Max Moe, so it should be correct. Only free parameters are m1Min, m1Max, and nSamples, defaulted to 5, 100, and 1e5 respectively. 

Output is a grid file that can be immediately submitted to COMPAS.


ilyamandel:
Do we have permission from Mads Sorensen (and Max Moe) to make this public as part of the COMPAS repo?

reinhold-willcox:
I didn't think to ask! I'll do that and report back when I get an answer.

reinhold-willcox:
Max said he's fine with it

ilyamandel:
Did you ask Mads, since this code is his?

reinhold-willcox:
I can't seem to find his contact. I'll keep digging.

ilyamandel:
Responded offline.

ilyamandel:
This would close #213 .

ilyamandel (changes requested):
@reinhold-willcox -- can you provide an example of how to run this to create a grid file, ideally both in the comments at the top of this file and in the documentation on the docs web page?  In particular

```
Due to the correlated nature of these parameters, it is not 
straightforward to supply alternative constraints, e.g
including only short period systems. Thus the only free
parameters are the maximum and minimum total mass, 
and the number of systems sampled.
```

that makes sense, but how exactly should a user specify these when calling this script?

reinhold-willcox:
Got the green light from Max and Mads, will update with the comments shortly.

Mads asked that we cite the paper where we first implemented the script: https://arxiv.org/abs/1808.06488

ilyamandel:
Let's add that citation to the relevant bit of code, and we will be sure to add that to the next release of the COMPAS methods paper.

reinhold-willcox:
@ilyamandel Is this PR good to go, or are there other changes you were hoping for?

ilyamandel (approved):

PR #799, opened 2022-05-14
Minimal core mass following Case A mass transfer

ilyamandel:
Added option retain-core-mass-during-caseA-mass-transfer to preserve a larger donor core mass following case A MT, set equal to the expected core mass of a newly formed HG star with mass equal to that of the donor, scaled by the fraction of its MS lifetime


jeffriley (changes requested):
I haven't finished my review, but it looks like you added a new option and didn't update the documentation...

Edit: Oh, just noticed it's still a draft :-)

ilyamandel:
OK, ready for review -- do your worst,  @jeffriley ! ;-)

jeffriley (changes requested):
Very nice - changes look good.

A couple of questions:

(1) Why did you remove the "--" from "mass-transfer-jloss-macleod-linear-fraction" in program-options-list-defaults.rst, line 999?

(2) I assume you've tested the changes and they produce the results you expect.  Have you checked backwards compatibility (i.e. with you new option set false, does everything work the way it used to)? What about the detailed plot?

ilyamandel:
Thanks for the review, @jeffriley !

(1) I hadn't removed the dash, I added the previously missing option (from a previous PR) but somehow dropped the dash. ;-) Anyway, fixed now -- thanks for the catch. 

(2) Yes.  Proof coming in a second.

ilyamandel:
Doing the check (all good) prompted me to fix a broken soft link from when example files got moved -- thanks!

image

jeffriley (approved):
Ok, all looks good to me.  I'll remove my block on merging. @IsobelMarguarethe to review?

ilyamandel:
For illustration purposes, here is what happens when I start a 60+40 Msol binary at different separations and evolve until the primary reaches TAMS (actually, first time step of HG), when it has developed a core.  

The new v02.31.00 results with retain-core-mass-during-caseA-mass-transfer set to False are identical to v02.28.00 results -- so it's all backward compatible.  

When retain-core-mass-during-caseA-mass-transfer is set to True, the total mass at TAMS is the same as before (no change to MT), but the core mass can be significantly (by up to 8 solar masses) larger for donors that experienced case A MT, as expected.  

image

ilyamandel:
@jeffriley , @IsobelMarguarethe -- given the checks above, I'll merge the PR; however, if any further issues are uncovered during more elaborate testing, please ping me!

PR #804, opened 2022-05-16
fixed help string for new ecsn in H rich stars option

reinhold-willcox:
The help string was copied from another option, now it is shows the correct string.


jeffriley (changes requested):
It's a change, so it needs to go in changelog and be reflected in the version number (revision number - it's a defect repair).

It's not cosmetic - it changes the output of the program.  If we don't increment the version number, we could have 2 binaries reporting the same version number yet producing different outputs.  Harmless as the difference in output might be in this case, it's still not an acceptable situation - a user could reasonably wonder what else might be different that they don't know about.

Or maybe you could save this for a future, larger change, or incorporate it into an existing PR?

reinhold-willcox:
done

jeffriley (approved):
Approved

PR #807, opened 2022-05-18
Minor defect repairs

jeffriley:
Defect repairs:
- Fixed STAR_PROPERTY_LABEL entries in constants.h for INITIAL_STELLAR_TYPE and INITIAL_STELLAR_TYPE_NAME - both labels missing the prefix "INITIAL_".
  Only caused a problem if a user wanted to add either of those to the logfile-definitions file - but since they are in the system parameters files (SSE and BSE)  by default encountering the problem would probably be unlikely.
- Fixed error identifier in Log::UpdateAllLogfileRecordSpecs() - was (incorrectly) ERROR::UNKNOWN_BINARY_PROPERTY, now (correctly) ERROR::UNKNOWN_STELLAR_PROPERTY


reinhold-willcox:
Confirmed that the bug fix works. Do we need to update the documentation?

Also, there is a deleted pythonsubmit in this PR, was that intended?

jeffriley:
No documentation changes required.

The pythonSubmit file appeared in the pushed files as a modified file.  I didn't change it - at least I don't remember changing it, and if I did it wasn't to deliberately fix or enhance anything - so I deleted it from my PR (on github, when I noticed it).  That won't delete the file from the repo will it?  I don't want to do that - I just didn't want to push a changed version of it...

reinhold-willcox:
I think it will delete it from the repo (it shows up as a deleted file in list of `files changed`). Could you copy it back in from dev and recommit it?

jeffriley:
File added @reinhold-willcox

reinhold-willcox (approved):
looks good, thanks Jeff

PR #810, opened 2022-05-21
Fixed missing CEE flag for unstable MS+MS MT

### LINKED ISSUE #809: MS+MS unstable mergers not flagged as CEE
**Describe the bug**
Mergers between two MS stars are ony flagged as unstable if both are undergoing RLOF. Of coursre, if only one is a donor, you can still have unstable MT, but the output flag `CEE>MT` is not raised in this case. 

**Label the issue**
`urgency_moderate`      - This issue is not urgent

**To Reproduce**

run: `COMPAS -n 1 --random-seed 1653046919 --initial-mass-1 8.0 --initial-mass-2 1.0 --semi-major-axis 0.034129411420349316 --logfile-definitions 'logdef.txt'`

with logdef.txt:
```
BSE_RLOF_REC += {
    BINARY_PROPERTY::STELLAR_MERGER,
}  
```

The output shows a merger at first encounter between an 8+1 MS+MS merger, which should be unstable from the mass ratio, but has `CEE>MT = 0`. See below.

 |  SEED                 |  (units)      |  1653046919  |
 |  --------             |  -----------  |  ----------  |
 |  CEE>MT               |  Bool         |  0.000000    |   
 |  Eccentricity<MT      |  -            |  0.000000    |   
 |  Eccentricity>MT      |  -            |  0.000000    |   
 |  MT_Event_Counter     |  Count        |  1.000000    |   
 |  Mass(1)<MT           |  Msol         |  7.985811    |   
 |  Mass(1)>MT           |  Msol         |  7.985811    |   
 |  Mass(2)<MT           |  Msol         |  1.000000    |   
 |  Mass(2)>MT           |  Msol         |  1.000000    |   
 |  Merger               |  Bool         |  1.000000    |   
 |  RLOF(1)<MT           |  Bool         |  0.000000    |   
 |  RLOF(1)>MT           |  Bool         |  1.000000    |   
 |  RLOF(2)<MT           |  Bool         |  0.000000    |   
 |  RLOF(2)>MT           |  Bool         |  0.000000    |   
 |  Radius(1)<MT         |  Rsol         |  4.131629    |   
 |  Radius(1)>MT         |  Rsol         |  4.131629    |   
 |  Radius(1)|RL<step    |  -            |  1.003590    |   
 |  Radius(1)|RL>step    |  -            |  1.003590    |   
 |  Radius(2)<MT         |  Rsol         |  0.894479    |   
 |  Radius(2)>MT         |  Rsol         |  0.894479    |   
 |  Radius(2)|RL<step    |  -            |  0.551509    |   
 |  Radius(2)|RL>step    |  -            |  0.551509    |   
 |  SemiMajorAxis<MT     |  Rsol         |  7.350505    |   
 |  SemiMajorAxis>MT     |  Rsol         |  7.350505    |   
 |  Stellar_Type(1)<MT   |  -            |  1.000000    |   
 |  Stellar_Type(1)>MT   |  -            |  1.000000    |   
 |  Stellar_Type(2)<MT   |  -            |  1.000000    |   
 |  Stellar_Type(2)>MT   |  -            |  1.000000    |   
 |  Time<MT              |  Myr          |  15.894403   |   
 |  Time>MT              |  Myr          |  15.894403   |   
 |  Zeta_Hurley(1)       |  -            |  0.300000    |   
 |  Zeta_Hurley(2)       |  -            |  0.300000    |   
 |  Zeta_Hurley_He(1)    |  -            |  0.300000    |   
 |  Zeta_Hurley_He(2)    |  -            |  0.300000    |   
 |  Zeta_Soberman(1)     |  -            |  -0.333333   |   
 |  Zeta_Soberman(2)     |  -            |  -0.333333   |   
 |  Zeta_Soberman_He(1)  |  -            |  -0.333333   |   
 |  Zeta_Soberman_He(2)  |  -            |  -0.333333   |   

 - COMPAS v02.31.01

reinhold-willcox:
Easy fix for #809


jeffriley (changes requested):
A suggestion and a question; first, the suggestion:

You have changed this

        else {                                                                                         // Unstable Mass Transfer
            if (m_Donor->IsOneOf( MAIN_SEQUENCE )) {
                m_Flags.stellarMerger = true;
                isCEE                 = true;
            }
            else {
                m_CEDetails.CEEnow = true;
                isCEE              = true;
            }
        }

to this

        else {                                                                                         // Unstable Mass Transfer
            if (m_Donor->IsOneOf( MAIN_SEQUENCE )) {
                m_Flags.stellarMerger = true;
                m_CEDetails.CEEnow = true;
                isCEE                 = true;
            }
            else {
                m_CEDetails.CEEnow = true;
                isCEE              = true;
            }
        }

It could be refactored as

        else {                                                                                         // Unstable Mass Transfer
            m_CEDetails.CEEnow = true;
            isCEE              = true;

            if (m_Donor->IsOneOf( MAIN_SEQUENCE )) {
                m_Flags.stellarMerger = true;
            }
        }

And the question (actually questions, but all related):

How does 'm_CEDetails.CEEnow' interact with the local variable 'isCEE'?  How are they different?  Why is 'isCEE' not set in this block of code 

    if (qCritFlag && m_Donor->IsMassRatioUnstable(m_Accretor->Mass(), m_Accretor->IsDegenerate()) ) {
        m_CEDetails.CEEnow = true;
    }

We should reflect on why the local variable 'isCEE' exists and why it was implemented to be slightly different from 'm_CEDetails.CEEnow' - was the difference deliberate, or was it a mistake in the original implementation?  If it was deliberate, what was the reason?  Is that reason still valid?  Do we still need the local variable 'isCEE', or can it be removed and we just test 'm_CEDetails.CEEnow'?

reinhold-willcox:
@jeffriley I'm happy to refactor the first suggestion, that makes sense and improves readability. 
Edit: done

To your second point: I don't know what the difference is. It looks like `isCEE` is pretty limited in scope, but there are a few places in the `CalculateMassTransfer` function where `m_CEDetails.CEEnow` gets set but not `isCEE`, so I'm hesitant to mere the two until we better understand the effect of that. I'm happy to leave them separate for now, but perhaps this could be a future fix for someone who has time to look into it?

jeffriley:
I think we should understand the code before we change it, otherwise we mightn't understand the consequences of the change.  And if we can remove a variable and tidy up the code we should - otherwise we'll be refactoring again in a couple of years. It should only take a few minutes to understand what the code is doing.

In BaseBinaryStar::CalculateMassTransfer(), the local variable 'isCEE' is apparently used only to check for recycled pulsars (at the end of the function).  In that function, 'isCEE' is set true everywhere 'm_CEDetails.CEEnow' is also set true, except for two places: the first is when both stars are undergoing RLOF, and in that instance we immediately exit the function, so setting a local variable before we leave is pointless.  The other places is the block of code I mentioned above:

    if (qCritFlag && m_Donor->IsMassRatioUnstable(m_Accretor->Mass(), m_Accretor->IsDegenerate()) ) {
        m_CEDetails.CEEnow = true;
    }

Why isn't 'isCEE' also set true there?  The check for recycled pulsars will still be executed after this block of code is executed.  Presumably 'isCEE' indicates whether there is a CEE happening - isn't that also what 'm_CEDetails.CEEnow' indicates, or are they subtlely different?  Does the block of code above NOT indicate the possibility of recycled pulsars?  But even if so, not setting 'isCEE' is probably not the way to handle that - either a CEE is happening or it is not.

The obvious difference between 'isCEE' and 'm_CEDetails.CEEnow' is that 'isCEE' has no meaning outside the scope of BaseBinaryStar::CalculateMassTransfer(). That tells us two things: (a) 'isCEE' is used only for the recycled pulsars check, and (b) 'm_CEDetails.CEEnow' is used elsewhere in the code.  

For (a), we just need to understand where (in BaseBinaryStar::CalculateMassTransfer()) we might want to flag that a CEE is occurring so that we can check for recycled pulsars (hence the question about the block of code above). 

For (b), where else in the code is 'm_CEDetails.CEEnow' used, and what will setting it true in BaseBinaryStar::CalculateMassTransfer() do (other than what you expect it to do)?  Are there any possible unintended consequences of setting 'm_CEDetails.CEEnow' true?  I don't think so, but have you checked?  Have you run a test with the change in, and checked the results against a run with it not in (i.e. the current version of COMPAS) to see that the only change in output is what you are expecting?

jeffriley:
Another difference between 'isCEE' and 'm_CEDetails.CEEnow' just occurred to me - 'isCEE' is set false at the start of BaseBinaryStar::CalculateMassTransfer(), whereas 'm_CEDetails.CEEnow' could enter the function already true.  Is the difference between them that 'isCEE' means 'we just noticed a CE event', and 'm_CEDetails.CEEnow' means 'a CEE is occurring'?  Is that subtle difference meaningful in BaseBinaryStar::CalculateMassTransfer()?  But either way, the question is still why is 'isCEE' not set true in the block of code described above - is that a defect that we haven't yet noticed, or should it truly not be set there?

reinhold-willcox:
I agree this is an area we could probably clean up the code, but for this particular PR I think trying to dissect and integrate isCEE vs m_CEDetails.CEEnow might be a bit out of scope. Would it be alright with you if I added your comments above to a new issue and restricted this PR to just fix the small MS+MS merger bug?

jeffriley:
Have you tested your change against the current version of the code?  What about the detailed plot - that's one of our standard tests - have you run that?

I'd prefer to resolve the 'isCEE' issue now, but if you don't want to spend an hour or so looking at the code open a new issue - but we both know that opening an issue won't get it looked at for months, if at all.  We keep doing this and the code bloats and we have redundant, if not just plain wrong, code.

I don't see how it can take very long at all to discern the meaning of 'isCEE' - it's a local variable and only used in a single check in the function.  And is it really out of scope for this PR?  You're changing code that is directly related.  No change to the code should be performed in isolation - you need to take a holistic look at the area of code you are changing and determine whether your change is (a) correct in the context of the code around it, and (b) all that needs to be changed (especially if your change might have side-effects - and a possible side-effect here is that you're making 'isCEE' redundant - or exposing a defect).    You said when you opened the underlying issue here that the problem is minor and not urgent - so why not take a little bit of time and improve the code?

Edit: NOT CEE is a necessary condition for recycled pulsars isn't it?  So 'isCEE' must be a flag for a CEE.  But isn't that what 'm_CEDetails.CEEnow' is?  The only place 'isCEE' is not set true at the same time 'm_CEDetails.CEEnow' is in the block of code indicated - can you speculate as to why that might be?  Might it be an oversight, or should we not flag a CEE in that circumstance?  It is possible that 'm_CEDetails.CEEnow' enters BaseBinaryStar::CalculateMassTransfer() already true - does that matter (i.e. could/should 'isCEE' stay false in that case)?

reinhold-willcox:
Looking at the logic and comments in this function, it is clear that whoever added `isCEE` intended it to capture all occurences of common envelope getting flagged (see the comment `// Check for recycled pulsars. Not considering CEE as a way of recycling NSs.`). 

Assuming `m_CEDetails.CEEnow` is false at the beginning of this function, the only time it is set to true without `isCEE` also being set to true (or the function returning) is in the if clause for the qCritical stability flag. I believe `isCEE` was erroneously not added here because the qCritical option is even now not fully configured. 

Additionally, a quick search shows that `m_CEDetails.CEEnow` is only ever set to true in this function. So bearing all that in mind, I think it is clear that the two parameters have the same intended functionality and thus are redundant. I'll make the fix and do a quick 100k comparison run between this branch and `dev` to ensure that the only difference is in the CEE flag of MS+MS mergers.

reinhold-willcox:
Confirmed, the two parameters were redundant and combing them did not produce any changes between runs of 100k binaries. Also, the detailed plotter is identical. Hopefully this satisfies all the requested changes.

jeffriley (approved):
Great!  Thanks for taking the time to improve the code @reinhold-willcox

PR #811, opened 2022-05-21
plotting script fixes

### LINKED ISSUE #808: Add HR diagram to detailed plotter
Currently the detailed plotter features 4 plots, which show the time evolution of (1) mass parameters, (2) length parameters, (3) stellar types, and (4) eccentricity. It would be really useful to debugging and general investigation to include an HR diagram. Since we like having 4 plots in a 2x2 grid, I propose we put the eccentricity and stellar type plots together (this was done previously) since they are not very busy plots anyway, and put the HR diagram in the bottom right quadrant.

reinhold-willcox:
- added periapsis to length plot, 
- added HR diagram with lines of const radii and event markers. 
- made the 'merging in X Myr' text sci notation

addresses issue #808


ilyamandel (changes requested):
eventString = r'Double compact object ({}+{}) merging in {:.2e} Myr'.format(self.stypeName1, self.stypeName2, Tdelay)

We should not be doing this if the binary is unbound, i.e., if a<0 or e>1.  In fact, the moment either of those happens, we should instead print out that the binary is unbound.

reinhold-willcox:
True, that should be fixed now.

ilyamandel (approved):
Thanks, @reinhold-willcox !

PR #812, opened 2022-05-31
Split major events and timesteps

### LINKED ISSUE #805: Improve internal consistency for reporting (previously: Implement separating major events into separate timesteps)
As discussed on the compas dev call

reinhold-willcox:
Major code update:
- split key events into separate timesteps to ensure output is always consistent (including stellar type changes from SSE, mass transfer events, and supernovae)
- add adaptive shortening of timesteps to ensure that major events _should_ not happen in the same timestep

This is a draft so we can collaborate in parallel and check that things work as expected.


jeffriley:
@reinhold-willcox What's the rationale for the changes to BaseBinaryStar::EvaluateSupernovae() ?

reinhold-willcox:
The code currently allows for two SNe in one timestep, which leads to unphysical output such as the stellar type preSN set to a BH. Since we're trying to avoid multiple events in a timestep, there should never be a SN_STATE::BOTH.

jeffriley:
Right - so don't forget to change the comments describing the function (and where it is called), and to remove the SN_STATE::BOTH state from the available states.  Also the documentation will need to be changed.

It might actually be better to remove the function entirely and just handle the different cases where it is currently called - if we make these changes the function doesn't really do anything that can't be handled in the calling code.

reinhold-willcox:
Agreed. This isn't meant to be a polished PR just yet, I'm just trying to get the code to do the right thing for now.

jeffriley:
Understood - I just put the comment here so we'd remember to do it.  Also, in answer to your question in the code, I don't think you need to catch the "else" case.  In such cases we tend to be passive - if neither condition is true, just do nothing - even if we shouldn't be there if neither condition is true.  If you think it could happen, and it would be catastrophic if it does, then sure, put a catch in, but I think passive is ok here.

reinhold-willcox:
Closing as outdated

PR #819, opened 2022-06-15
Moe Di Stefano sampler bug fix

reinhold-willcox:
Fixed a bug that tried to call an array as a function. Applied only at very low masses (m1<0.8).


ilyamandel (approved):
Looks good!

Issue #817, opened 2022-06-15
FastCosmicIntegration maximum chirp mass error


reinhold-willcox:
**Describe the bug**
This bug was discovered by @adam-boesky who is a summer student working with Floor. 

When running the FastCosmicIntegration pipeline on a typical data file, the following error was thrown. 

```
UserWarning: Maximum chirp mass used for detectability calculation is below maximum binary chirp mass * (1+maximum redshift for detectability calculation)
```

Looking at the source code, the error comes from the following:
```
if max(chirp_masses)*(1+max_redshift_detection) < Mc_max:
        warnings.warn("Maximum chirp mass used for detectability calculation is below maximum binary chirp mass * (1+maximum redshift for detectability calculation)", stacklevel=2)
```
where Mc_max = 300 and max_redshift_distance = 1 by default. 

By inspection, the inequality looks like it should go the other way. Can someone who is more familiar with the FastCosmicIntegration confirm this? If so, the fix is easy.

**Label the issue**

`urgency_high`     - This is a very urgent issue and should be resolved as soon as possible
`severity_minor`    - This is a minor bug with minimal impact

**Versioning (please complete the following information):**
 - COMPAS v02.31.03

jeffriley:
This was already raised in issue #591 and was fixed in PR #628 

So when did we lose the fix?  Was it backed out unintentionally?  That would raise questions about other changes and fixes that we might also have lost...

reinhold-willcox:
This is a weird one...

You can see the history of commits here: https://github.com/TeamCOMPAS/COMPAS/commits/6b13662f893cfcf10eaffd7462021001d65e8c73?after=6b13662f893cfcf10eaffd7462021001d65e8c73+69&branch=6b13662f893cfcf10eaffd7462021001d65e8c73&qualified_name=6b13662f893cfcf10eaffd7462021001d65e8c73

If you run 
```
$ git checkout 5eb0bc5 
$ cd '...'/Folders/CosmicIntegration/PythonScripts/ 
$ git diff bb83bdf f707f87 FastCosmicIntegration.py | grep Mc_max
-    if max(chirp_masses)*(1+max_redshift_detection) < Mc_max:
+    if max(chirp_masses)*(1+max_redshift_detection) > Mc_max:
```
you'll see that this is where the incorrect < was swapped in back for the >, seemingly in `bb83bdf`. But that PR doesn't show any changes to the FastCosmicIntegration file.

[...]

PR #636, opened 2021-10-04
Fix MassTransferTrackerHistory

ilyamandel:
Remove unnecessary IsPrimary() / BecomePrimary() functionality

PR #820, opened 2022-06-20
supposedly merged back in the missing commits - NEED TO VERIFY

reinhold-willcox:
Fix for issue #817, and the related discussion lower down in that issue. 

The underlying problem was that a few commits from PR #628 were quietly reverted in #636, and we did not notice until now. As best I can tell, the undoing commit was this one: [Cleaning Commit](https://github.com/TeamCOMPAS/COMPAS/pull/636/commits/4010d38c5cb2544ea8510b6bbe94c476fb8be4d0). I think what happened is that Ilya (apologies for calling you out) accidentally recommited older versions of these files that he had updated in previous commits. 

The files in question include an .eps produced in the detailed output was modified, the `getting_started.md` documentation was cleaned up for whitespace, the newly created `generate_frame_file.py` file was removed, and some `FastCosmicIntegration.py` fixes, including the Mc test flagged in issue #817, were removed. I don't believe any other changes were made, so I think this was an erroneous commit, which makes the fix here much easier.

For this PR, I've done a `git revert` on this specific commit. The eps is recreated often, and getting_started.md is now merged into the online_docs (and the whitespace issues addressed), so I've ignored both of these files - they are not longer relevant. There were a few more involved changes in the `FastCosmicIntegration.py` file that required manual input, i.e I had to assume which parts of the older and newer versions we wanted to keep, but I think I did this right. I've also readded the `generate_frame_file.py`, since this looks like it could be useful. That said, apparently no one has looked for or needed it, because it was never actually part of the repo. So @ilyamandel could you let me know if this script is worth cleaning up and keeping around?

### DIFF
--- a/utils/CosmicIntegration/FastCosmicIntegration.py
+++ b/utils/CosmicIntegration/FastCosmicIntegration.py
@@ -109,10 +109,10 @@ def find_metallicity_distribution(redshifts, min_logZ_COMPAS, max_logZ_COMPAS,
     sigma = sigma_0* 10**(sigma_z*redshifts)
     
     ##################################
-    # Follow Langer & Norman 2007? in assuming that mean metallicities evolve in z as:
+    # Follow Langer & Norman 2006 in assuming that mean metallicities evolve in z as:
     mean_metallicities = mu0 * 10**(muz * redshifts) 
         
-    # Now we re-write the expected value of ou log-skew-normal to retrieve mu
+    # Now we re-write the expected value of the log-skew-normal to retrieve mu
     beta = alpha/(np.sqrt(1 + (alpha)**2))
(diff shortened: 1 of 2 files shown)


ilyamandel:
Thank you for figuring this out, @reinhold-willcox !  You are absolutely right to call me out.  I am still puzzled that the reversions do not show up in the list of overall changes for that PR, https://github.com/TeamCOMPAS/COMPAS/pull/636/files .  It would be  good to know why that is, because that's what I usually check when testing / approving PRs, rather than going through a multi-commit PR one commit at a time...

ilyamandel:
P.S. generate_frame_file.py is a work in progress, but yes, we should keep it (in fact, I'll be making changes to it shortly).

ilyamandel (approved):

PR #828, opened 2022-07-25
Non stripped ecsn fix

### LINKED ISSUE #827: --allow-H-rich-ECSN flag does not work
I found ECSNe labeled as H-poor (I ran with --allow-H-rich-ECSN: False ) in binaries with huge separations — sometimes >1000 AU.

This is odd, because H-poor SNe should only happen in stripped stars, and stripping can only happen in interacting binaries.

On investigation, it turns out that we assume that H-poor stars are those that formed from objects of type HeMS and later.  Of course, ECSNe are formed from ONe WDs, so always satisfy this — i.e., all ECSNe are always H-poor, and the flag is meaningless.

What the flag was supposed to represent is whether the star had been stripped of its H envelope earlier, but it does not detect this as currently implemented.

P.S. A probably unrelated issue — but still one to fix -- is that our ONeWDs can have total masses larger than He core masses.

reinhold-willcox:
Addresses issue #827, regarding the allow-H-rich-ecsn option. Effecitvely single ECSN progenitors will usually be ONeWDs, and so H-poor anyway, meaning the option did not capture the intended functionality of removing effectively single progenitors.

I replaced the condition for H-richness with a check on whether the MT donor history is empty, i.e if the progenitor has ever been a donor. I also fixed the option name where appropriate. 

I checked that the only ECSNe which occur now are ones with a non-empty MT history.

### DIFF
--- a/online-docs/pages/User guide/Program options/program-options-list-defaults.rst
+++ b/online-docs/pages/User guide/Program options/program-options-list-defaults.rst
@@ -62,8 +62,8 @@ Default = GRID
    * - NEVER
      - indicates that the program options should `not` be added to the sysparms file
 
-**--allow-H-rich-ECSN** |br|
-Allow ECSNe in effectively single progenitors (those which retain a H envelope). |br|
+**--allow-non-stripped-ECSN** |br|
+Allow ECSNe in effectively single progenitors. |br|
 Default = FALSE
 
 **--allow-rlof-at-birth** |br|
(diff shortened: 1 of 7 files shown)


ilyamandel (approved):
Thank you for the quick turnaround!

ilyamandel:
Actually, this is getting more and more confusing.  
/BSE_Supernovae/Experienced_RLOF(SN) can be 1 while /BSE_Supernovae/MT_Donor_Hist(SN) is NA.  
How is that possible?

reinhold-willcox:
That's tricky, I'm not sure. Do you have an h5?

Issue #825, opened 2022-07-08
Odd system: merger of HeMS with massless remnant


nrsegovia:
**Describe the bug**
As I do not know if the following is a bug, I have just labeled it as a question.

I have been running tests with systems where at least one component goes through the HeMS phase, and I'm using the RLOF log file to check for changes in separation. Amongst these, some have `CEE>MT == 1` but `SemiMajorAxis>MT` does not show up in my plots as, for some reason, the separation is `nan`. Running one of these systems under the detailed mode shows no problems with the separation, but the simulation stops after the MT episode, and the message in the terminal says that the stars merged. However, in the last time step saved in the detailed output, the radii are ~150 and ~0.1 R_Sun while the separation is ~260 R_Sun (so no `HasStarsTouching()`). On top of that, the bigger star ends as an EAGB star according to the detailed output, but the final message in the terminal says it is a massless remnant. I checked SN flags just in case (`Supernova_State`), but nothing pops up.

So, to summarize:
* Why am I suddenly getting a `nan` value for the separation? My guess is that the "merger" caused it
* But then, why am I getting a merger when the stars do not seem to be close enough? Is a CEE being resolved but not logged?
* And finally, any clues about what leads to the printing of `massless_remnant` as final phase of the EAGB star? The `--rlof-printing` indicates that there is a CE episode before the end of the simulation, which seems to be linked to the evolution into a massless remnant.

**To Reproduce**
```
--random-seed 1952  --maximum-evolution-time 15000 --initial-mass-1 2.906743424402851 --initial-mass-2 2.183152794529362 -a 0.14299254627993116 -e 0.20219885610022265 -n 1 --common-envelope-alpha 0.25 --common-envelope-lambda-prescription LAMBDA_LOVERIDGE -z 0.005 --mass-transfer-accretion-efficiency-prescription THERMAL --allow-rlof-at-birth TRUE --detailed-output TRUE
```
And a similar run turning `--detailed-output` off, while turning `--rlof-printing` on (to check for changes before/after MT episodes).

**Versioning:**
COMPAS v02.31.04

ilyamandel:
@solfreludio -- I think what happens to these stars is fine, but the DetailedOutput is not. :(

[...]

PR #830, opened 2022-08-01
Print to DetailedOutput after merger, addresses issue 825

ilyamandel:
- Print to DetailedOutput after merger, addresses issue #825 
- Ensures no ONeWDs are formed with masses above Chandrasekhar mass
- Minor comment tweaks and a bit of defensive programming


src/GiantBranch.cpp
```diff
@@ -1703,7 +1706,11 @@ STELLAR_TYPE GiantBranch::ResolveElectronCaptureSN() {
     }
     else {                                                                                  // -no, treat as ONeWD 
         
-        m_Mass       = m_COCoreMass;                                                        
+        m_Mass       = std::min(m_COCoreMass,MCH);                                          // no WD masses above Chandrasekhar mass
```
reinhold-willcox, line 1712: This change seems to me like a bandaid fix. If we're getting core masses above MCH, they should be undergoing SN, so they should not end up here anyway. I would suggest throwing an error here if the mass is too high, instead of resetting it to the lower MCH value.

reinhold-willcox (changes requested):

src/GiantBranch.cpp
jeffriley: We don't throw errors.  We could display a warning or error message that the coremass was reset, but we can't just then return from the function - the show must go on, so we have to return with values that are sensible enough to continue the evolution.

One day we'll figure out how we want to respond to errors/events/situations like this (e.g. we might set an error and bubble it all the way back to the Evolve() function and terminate the evolution there with the appropriate error), but right now our strategy is to just plough on regardless...

reinhold-willcox: Fair point. I'd be happy if we kept the new change but added an Error message.

ilyamandel: That was an attempt at defensive programming -- but I accept the criticism, will fix. :)

ilyamandel: See fix.

ilyamandel:
@reinhold-willcox , @jeffriley , ready for your review.

jeffriley (commented):
Changes look good.  Just a couple of (related) questions:

You have this comment:

//Should we use the prescription in the last paragraph of section 6.1 of Hurley+ 2000, instead? return (utils::Compare(m_COCoreMass, CalculateCoreMassAtSupernova_Static(m_Mass0)) >= 0);

Did you mean to leave that in?  HeHG::CalculateGBParams_Static() sets gbParams(McBAGB) = p_Mass0 - have you checked whether gbParams(McBAGB) is actually equal to m_Mass0 at this point? Is it worth checking?

jeffriley (changes requested):
Requesting changes just to answer my questions.

ilyamandel:
@jeffriley -- you are right, thank you, fixed!  
Also took the opportunity to add the "--muller-mandel-sigma-kick" option to the default .yaml file (the option has been there, we just didn't have it in the default file)

ilyamandel:
@reinhold-willcox -- are you happy with your change request being addressed?

jeffriley (approved):
All good

reinhold-willcox (approved):

PR #832, opened 2022-08-03
Accretion-Induced Collapse

### LINKED ISSUE #827: --allow-H-rich-ECSN flag does not work

reinhold-willcox:
- Added Accretion-Induced Collapse as an alternative SN, only ever applied to ONeWDs which accrete above the Chandrasekhar mass. 
- Added relevant functions for IsAIC() and ExperiencedAIC()
- Updated documentation for relevant sections
- Updated changelog, including a typo in the previous entry


jeffriley (commented):
Changes look ok to me, with one small nit-pick:

This comment

 * @return        Stellar type of remnant (always STELLAR_TYPE::NEUTRON_STAR)
 
in ONeWD::ResolveAIC() is not strictly true.  It is true if the sanity check passes (i.e. if IsSupernova() is true), but it won't (or shouldn't) be true if not.

Otherwise, certainly from a code perspective, I approve.  The physics also looks ok, but I'm not familiar enough with AIC to comment sensibly...

Edit:

Just a thought.  Instead of this construct:

if (IsSupernova()) {
    ...
    ...
    return STELLAR_TYPE::NEUTRON_STAR;
}
else {
    return m_StellarType;
}

you could instead do this:

STELLAR_TYPE stellarType = m_StellarType;

if (IsSupernova()) {
    ...
    ...
    stellarType = STELLAR_TYPE::NEUTRON_STAR;
}

return stellarType ;

It's marginally cleaner, and more in keeping with the aim to have a single entry and single exit in any function.
If you don't like that, this is more in keeping with the way we do sanity checks at the start of a function:

if (!IsSupernova()) return m_StellarType;

...
...

return STELLAR_TYPE::NEUTRON_STAR;

so we exit immediately if we shouldn't be there.  Just a bit easier to see at a glance what's happening.

Up to you.

ilyamandel (changes requested):
I second @jeffriley 's suggestions above.  Also, keep in mind that, strictly speaking, WDs can supernovae -- type IAs.  So please clarify that comment.  

Also, given other recent commits, please merge dev into this one and update changelog.h, including version number.

Otherwise, looks awesome, @reinhold-willcox !

reinhold-willcox:
Issues should now be addressed, let me know if I missed anything.

jeffriley (changes requested):
Thanks @reinhold-willcox.  You don't actually need the "else" for the sanity check.

If you just have:

if (!IsSupernova()) return m_StellarType;

as the first line of executable code, that gets you out of there if you shouldn't be there - the "else" is superfluous (because of the return).

ilyamandel (approved):

jeffriley (approved):
All good.

PR #833, opened 2022-08-09
Finely sampled detailed outputs

reinhold-willcox:
- Added a detailed output printer after all major points where the mass might change (all in EvaluateBinary). 
- Added a flag to specify whether the printing happened at the end of a timestep, as a warning to users that mid-timestep detailed output may be self-inconsistent.
- Moved m_Unbound and related lines to the appropriate spot alphabetically (unrelated bit of tidying)

Note: the detailed plotter will probably not work following these changes, I'll look at that at a future date.

### DIFF
--- a/src/BaseBinaryStar.cpp
+++ b/src/BaseBinaryStar.cpp
@@ -372,6 +372,8 @@ void BaseBinaryStar::SetRemainingValues() {
 	m_CEDetails.preCEE.rocheLobe2to1             = DEFAULT_INITIAL_DOUBLE_VALUE;
 	m_CEDetails.preCEE.semiMajorAxis             = DEFAULT_INITIAL_DOUBLE_VALUE;
 
+    m_DetailedOutputEndOfTimestep                = false; // set false by default since there are many extra detailed outputs in the code and only one at the end of the timestep
+
     m_Flags.stellarMerger                        = false;
     m_Flags.stellarMergerAtBirth                 = false;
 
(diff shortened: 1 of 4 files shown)


jeffriley (changes requested):
@reinhold-willcox Overall looks good, but I'll suggest some changes later.

ilyamandel:
Nice!  Two queries:

Do we need m_DetailedOutputEndOfTimestep ?  Would it be cleaner to just pass an additional boolean parameter to the PrintDetailedOutput() function [and let that function handle that extra column, but most of us never look at the logging code]?

Do we definitely not want to print on merger?  Yes, the output is messy -- but we do want to know what happened...

reinhold-willcox:
This way seemed the cleanest - I wasn't sure how to include the flag as an output parameter if it was just used as an input to the PrintDetailedOutput script. 

We should still print on merger, there is a detailed printing call after all the iterations are done. We just don't need to do that twice.

jeffriley:
The suggestions for change I mentioned above were (what I think is) a better way to implement the end-of-timestep flag.  I'll get to those probably tomorrow.  I'm happy to make the changes since I'm more familiar with the logging code than anyone else, or help you make them if you want to take a leap into the logging code.  They won't actually go deep into the logging code really - it's more about making that output column more general than just a boolean flag.  More later.

reinhold-willcox:
I didn't want to make any changes that were too ingrained since we don't want this feature to be longlasting - but I'm happy for you to implement something that reads simpler. 

I'm also curious what other information you are thinking of including in the flag.

jeffriley:
I don't think it should be a temporary change.  Instead of recording the column value as a boolean, I suggest we use an integer.  That changes  the number of different values we can record from 2 to lots.  We can then define the integer value to be anything we want.  Right now we want to indicate whether a record written to the file was written part-way through a timestep or after the completion of the timestep - maybe that can be indicated by writing a 0 or 1 to the column value.  Maybe we also want to indicate that the record was written immediately prior to a particular event/calculations, or immediately following a particular event/calculation - maybe those can be indicated by writing a 2, 3, 4, etc. to the column value.  Or maybe the column value could be a bitmap that lets us combine record types.  It just allows us to record more detail than just a boolean flag.  We may or may not use many different values, but at least we can if necessary.  And that functionality doesn't have to be limited to the detailed output file.  We could, if we thought it would be useful, have the same functionality in, say, the SN file - and we could record two (or more) different types of record  in the SN file - pre- and post-SN for example (maybe - that's just an example).  Maybe the functionality would be useful in the RLOF file.  We don't have to implement them all now, but if we implement the column up-front as an integer, it lets us expand the functionality later.

jeffriley:
Actually, if we implement this and we then have the ability to effectively label the detailed output records, does that obviate the need to ensure we only have one event per timestep?  Can we not achieve the output you want in the detailed output file by having multiple calls to PrintDetailedOutput() sprinkled throughout the code, and ensure the record type is set so that anyone looking at the output file knows what each record indicates (e.g. start of timestep, after first SN, after MT, after second SN, after timestep, etc.).  Would that achieve what we need with much less modification of the code (and the associated risks)?  I wonder if we could make the binary/stellar attributes consistent at way-points in the timestep that would allow us to print self-consistent, suitably labelled, detailed output records.

jeffriley:
@reinhold-willcox @ilyamandel Just pushed new branch 'log-record-type' - needs documentation before I open a PR for it.  Feel free to have a look and/or try it out.

jeffriley:
@reinhold-willcox can we close this since PR #843 was merged?

reinhold-willcox:
Yes, thanks Jeff for your help

PR #834, opened 2022-08-09
max time and timesteps

### LINKED ISSUE #818: maximum-evolution-time per binary
When running with a modified maximum time in a gridfile, I get the error:

```
WARNING: Option not supported in grid file: 'maximum-evolution-time': ignored
```

This should be allowed, since we may want to evolve with different maximum times on a per binary basis. For example, I am looking at some very low mass stars, where the MS age is very long. If I evolve every binary at these longer total times, it slows down the whole run by a lot because the more massive systems are now evolving for much longer than needed.

reinhold-willcox:
Maximimum evolution time and maximum number of timesteps now enterable on a per-binary basis in the gridfile.

Edit: I would change the documentation, but I don't think there's anything to update for it. We do not specify in the docs whether a given option is not gridfile-able.

### DIFF
--- a/src/Options.h
+++ b/src/Options.h
@@ -217,8 +217,6 @@ private:
         "logfile-system-parameters",
         "logfile-type",
 
-        "maximum-evolution-time",
-        "maximum-number-timestep-iterations",
         "mode",
 
         "notes-hdrs",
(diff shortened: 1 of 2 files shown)


ilyamandel (commented):
Looks good to me, but I will wait for @jeffriley to approve. :)

jeffriley (approved):

PR #839, opened 2022-08-11
Critical mass ratios

### LINKED ISSUE #838: Bug in BaseStar::CalculateZadiabaticHurley2002
**Describe the bug**
The function captures equation (56) from Hurley+ 2002. In the paper, x is the positive part of the M-R exponent for giant stars, `R \propto M^{-x}`, with x specified as approx 0.3. 

The equation for zeta adiabatic is then given as 

Zad \approx Zeq \approx -x + 2(Mc/M)^5

This should be negative for small core mass (and large convective envelope mass), which makes sense because convective envelopes expand due to mass loss, thus zeta should be negative. 

In the code (in BaseStar::CalculateZadiabaticHurley2002), we set x = -0.3, which is incorrect and leads to positive zeta, such that convective giants might be able to avoid unstable MT. Presumably this is because this captures the "entire exponent" including the negative, but I believe this is incorrect. 

Can someone confirm that this is a bug? If so I'm happy to fix it. 

**Label the issue**

 - COMPAS v02.31.08

reinhold-willcox:
Implemented critical mass ratios from Claeys+ 2014.

Credit split with Alejandro, who did most of the work setting this up a while ago. I just uncommented the sections and made sure the options ran correctly. 

I thought this might be a moderate version change, so I incremented the middle set of digits. @jeffriley let me know if this is not the right use case.


jeffriley:
Yes, correct change to version number - the middle digits are for minor changes to functionality that are not defect repairs (so mostly enhancements).  Really the right-most digits should only increment for defect repairs, but they are sometimes used (by COMPAS developers) for very minor enhancements - not strictly correct, but hey...

avigna:
Fantastic Reinhold, this will be a very nice enhancement to have available. If not urgent, I can give it a go next week. The first check I'd suggest is a plot with mass ratio distributions per stellar type in the CEE file for our default model, this new enhancement with both Claeys+2014 default choices and some ad-hoc mass ratio variations (i.e., not default values from Claeys+2014). This will convince us it is working, and then the review will be mostly about code and formating.

reinhold-willcox:
@avigna For sure. Do you have any suggestions for the ad hoc variations, or is it just any alternatives to what's currently the default? There were some values listed previously that were not in Claeys, which I replaced with the Claeys values, but which might have been motivated for other reasons. Do you know if those came from some other paper?

avigna:
Just alternatives to default to see things work if you modify them... can be all the same mass ratio, e.g., q=0.343, or something quite arbitrary.

> There were some values listed previously that were not in Claeys, which I replaced with the Claeys values, but which might have been motivated for other reasons. Do you know if those came from some other paper?

I can't remember from the top of my head.

ilyamandel (changes requested):
Happy to leave the main check to @avigna , but I do have a couple of requests:

1. In the documentation, I think we should be more clear on things like "Specify both HG flags to use" -- just say which options/flags need to be specified (ditto elsewhere).  "Requires --use-critical-mass-ratio-HG" -- maybe say that this threshold is used only if that option is set to true (can use language that @jeffriley suggested in the temperature threshold PR).  Ditto for Options.cpp.  Clarify in the documentation which way the mass ratio is defined (accretor over donor mass or donor over accretor mass).

2. In BaseBinaryStar.cpp, do we really need to set the merger if the donor is a MS star here?  Feels like this should happen inside the CE code, rather than being a check every time we have MT:

        m_CEDetails.CEEnow = true;
        if (m_Donor->IsOneOf( MAIN_SEQUENCE )) {
            m_Flags.stellarMerger = true;
        }

avigna:
(I was thinking to wait with the review until the requested plot was uploaded, but let me know if I should get to it asap)

reinhold-willcox:
I'll get to the plot now.

Separately, I'm wondering if we need to have a flag to turn on qCrit for each stellar type. Is that a realistic use case, or would it be cleaner to have one flag "--use-critical-mass-ratios-claeys"?

reinhold-willcox:
@avigna Here's the plot you requested. Stype is a shorthand for Stellar Type. The random model has all qCrit values set to 0.5. I also offset the histogram baselines a bit to improve readability, so e.g the histogram for stype=6 shows that there is no difference between the various models, which I interpret to mean that this is going to be unstable MT regardless. 

You can see where the qCrits make a noticeable difference because of a sharp spike at the qCrit value (esp in Stype=1 and 2). The dearth of CEEs from the `random` model compared to the othes in Stype=3 and Stype=5 is a bit suspicious, because the Claeys qCrit is 0 for non-degenerate accretors, so I would expect fewer CEEs from Red the Green in the [0, 0.5] regime. I would guess that what's happened here is that so many more `random` systems went CEE during the HG, especially in this q regime, that they were removed and not available to undergo CEE as FGBs. Curious to get your interpretation though. 

image

avigna:
The Claeys implementation looks good to me, and very similar to our default, which I didn't expect. I calculated the Q_crit for GB and AGB stars according to Claeys+2014 (see attached plots). It looks like it is doing what it is supposed to do, particularly because only stars with very large core fractions (Mcore/M>0.8 or so) will result in dynamically stable mass transfer (i.e., their envelopes are mostly/fully convective according to this implementation).
Screenshot 2022-08-18 at 14 26 47
Screenshot 2022-08-18 at 14 23 45

avigna:
What I don't get is why the random model is not sharp in all of the plots; if you are "forcing" it to be a fixed value, it shouldn't matter if it a degenerate or non-degenerate accretor, right? I'd suggest to make one plot with degenerate accretors and another with non-degenerates (to understand the problem). In any case, the Claeys implementation seems fine, it is the fixed/ad-hoc one the one that I'm not convinced about...

reinhold-willcox:
The reason is that even when critical mass ratios are used, we still implement the zeta criteria regardless. So really, anything that is unstable in `default` will be unstable in `random`, but there will be (possibly) some additional systems which are newly unstable in `random` because of the mass ratio criteria. I had assumed this was the desired approach, but maybe we need to rethink this...

reinhold-willcox:
@avigna What you said made me realize the condition for stability with qCrits was not correct. I've tried to clean this up (see the recent changes to BaseBinaryStar), but there are still some bugs to work out. I'll make this a draft PR for now.

reinhold-willcox:
So here are the new plots, with the corrected stability prescriptions using either qCrits or zetas (but not both). 

This looks much closer to what you described, with sharp vertical lines at the cutoff mass ratios for the models that use them. I've renamed `random` to `fixed0.5` for clarity. I was a bit surprised by the region in the Stype=1 plot between q=[0.2, 0.25] where the `default` curve (black line) rises above the `fixed0.5` curve (green line), because I expected that the `fixed0.5` curve would not miss any of the systems with that mass ratio, while the `default` may miss some. I looked into this more closely, and found that all of these extra systems in `default` are coming from the second CE in systems where the first CE was more equal mass. So `fixed0.5` does not see any CEEs, while `default` sees a first one that leads to mass ratio reversal, and then a second unequal mass ratio CEE from a MS onto a He star. 

image

reinhold-willcox:
@ilyamandel @jeffriley I've made some modifications to how we specify the stability criteria for MT. There was nothing wrong with the previous formulation, but it was not straightforward to extend with the qCrits, hence the modifications. Hopefully the new version is clear to read and accurately captures the physics, but let me know if something looks off.

jeffriley:
It's hard to tell without actually pulling the branch down - which I'll do later - but for now just a quick comment/question:

These are commented:
```
        /*
        "critical-mass-ratio-giant-degenerate-accretor",
        "critical-mass-ratio-giant-non-degenerate-accretor",
        "critical-mass-ratio-helium-giant-degenerate-accretor",
        "critical-mass-ratio-helium-giant-non-degenerate-accretor",
        "critical-mass-ratio-helium-hg-degenerate-accretor",
        "critical-mass-ratio-helium-hg-non-degenerate-accretor",
        "critical-mass-ratio-helium-ms-degenerate-accretor",
        "critical-mass-ratio-helium-ms-non-degenerate-accretor",
        "critical-mass-ratio-hg-degenerate-accretor",
        "critical-mass-ratio-hg-non-degenerate-accretor",
        "critical-mass-ratio-ms-high-mass-degenerate-accretor",
        "critical-mass-ratio-ms-high-mass-non-degenerate-accretor",
        "critical-mass-ratio-ms-low-mass-degenerate-accretor",
        "critical-mass-ratio-ms-low-mass-non-degenerate-accretor",
        "critical-mass-ratio-white-dwarf-degenerate-accretor",
        "critical-mass-ratio-white-dwarf-non-degenerate-accretor",
        */
```
but I think still have getters and are still in the docs.  Are they going to be removed?  I thought that's what you meant by your comment:

"Also, the option to use critical mass ratios for each stellar type has been superseceded by a single option to set the critical mass ratio prescription (defaulting to NONE)"

I'll have a better look later.

jeffriley (changes requested):
Preliminary review comments above - more later.

avigna:
So now the fixed criteria looks good, but the Claeys for stellar types above HG (type 2) doesn't look good to me. The condition for dynamically unstable mass transfer is Q=Ma/Md<Qcrit, so most systems in your plot should end up experiencing a CEE and therefore in the histograms. I.e., what you had before for Claeys made sense to me.

reinhold-willcox:
@avigna That's true, the Claeys' value for qCrit for Giant donors with non-degen accretors is a function of the donor core/total mass ratio, not a constant. I was incorrectly setting this to 0 for simplicity in the code, and not updating it later to be more accurate. I'll fix that. 

@jeffriley the part I believe you're referring to is in the list of options that are BSE_only. The qCrit values do not have any meaning for SSE, but they are still intrinsically single star attributes, so I opted to remove them completely (so they can be implemented as SSE if so desired, for whatever reason).

reinhold-willcox:
@avigna Here are the plots with the corrected qCrit function implemented for Giants.

@ilyamandel Incidentally, this qCrit function also uses the same Hurley x \approx 3 that we discussed for the Hurley zetas in the other PR #842 . In that PR, I left x as a constant for simplicity, but since this fix also uses it, I looked around and found that Jeff already implemented the function call (of course!). So I've fixed the x call in both places in this PR. 

image

jeffriley:
> @jeffriley the part I believe you're referring to

No, I just misread the changes...  ignore what I said :-)

Pending @avigna and @ilyamandel I'm good with this.

avigna:
Fantastic, plot looks good now. Will get to the actual review in the next couple/few days.

reinhold-willcox:
@avigna and @jeffriley I made another edit, though this one in principle doesn't affect binary evolution. I removed IsMassRatioUnstable and replaced it with CalculateCriticalMassRatio. This is a bit more conceptually correct for a method in BaseStar, since a single star should know it's own critical mass ratio, but shouldn't really "know" if the mass ratio is unstable. It also makes the code a bit easier to read through, especially with the addition of alternative critical mass ratio formulations (i.e Ge et al., which I will push once this PR goes through). 

I've verified that the plot above which compares the various stability prescriptions is identical to the one produced after this change, so it's unlikely that I did anything spuriuos / unintended.

Separately, @ilyamandel I looked into your comment above re:  setting MS CEEs to mergers within the CalculateMassTransfer function. We set this again in ResolveCommonEnvelope, and somewhat more robustly (it allows for double RLOF MS CEEs), so I've removed this from the CalculateMassTransfer function.

src/BaseBinaryStar.cpp
```diff
@@ -1883,12 +1883,10 @@ void BaseBinaryStar::CalculateMassTransfer(const double p_Dt) {
 
 
 
+    // RTW
```
avigna: Are these your initials? If so, I suggest you remove them (no major changes and therefore no need for them, IMO).

online-docs/pages/User guide/Program options/program-options-list-defaults.rst
```diff
@@ -201,6 +201,93 @@ VINK_MASS_LOSS_MINIMUM_TEMP (default 12500K). |br|
...
+Critical mass ratio for MT from a HG star to a degenerate accretor.
+0 is always stable, <0 is disabled. Only used for "--critical-mass-ratio-prescription CLAEYS", ignored otherwise. |br|
+Default = 0.210000
+
+**--critical-mass-ratio-HG-non-degenerate-accretor** |br|
+Critical mass ratio for MT from a HG star to a non-degenerate accretor.
+0 is always stable, <0 is disabled. Only used for "--critical-mass-ratio-prescription CLAEYS", ignored otherwise. |br|
+Default = 0.250000
+
+**--critical-mass-ratio-MS-high-mass-degenerate-accretor** |br|
```
avigna, line 214: In Options.cpp you say:
> // A value of 0.0 means the mass ratio will never be unstable - this does not guaruntee stability of the MT, just that instability is not based on the mass ratio

This is not consistent with this comment. Which one is right? Fix accordingly.

online-docs/pages/User guide/Program options/program-options-list-defaults.rst
```diff
@@ -201,6 +201,93 @@ VINK_MASS_LOSS_MINIMUM_TEMP (default 12500K). |br|
...
+
+**--critical-mass-ratio-MS-high-mass-non-degenerate-accretor** |br|
+Critical mass ratio for MT from a MS star to a non-degenerate accretor.
+0 is always stable, <0 is disabled. Only used for "--critical-mass-ratio-prescription CLAEYS", ignored otherwise. |br|
+Default = 0.625000
+
+**--critical-mass-ratio-MS-low-mass-degenerate-accretor** |br|
+Critical mass ratio for MT from a MS star to a degenerate accretor.
+0 is always stable, <0 is disabled. Only used for "--critical-mass-ratio-prescription CLAEYS", ignored otherwise. |br|
+Default = 0.000000
```
avigna: Shouldn't this be 1, following Claeys?

online-docs/pages/User guide/Program options/program-options-list-defaults.rst
```diff
@@ -201,6 +201,93 @@ VINK_MASS_LOSS_MINIMUM_TEMP (default 12500K). |br|
...
+``NONE`` defaults to the zeta prescription for stability, 
+``CLAEYS`` uses qCrit values from Claeys et al. 2014. |br|
+Default = NONE
+
+**--critical-mass-ratio-white-dwarf-degenerate-accretor** |br|
+Critical mass ratio for MT from a white dwarf to a degenerate accretor.
+0 is always stable, <0 is disabled. Only used for "--critical-mass-ratio-prescription CLAEYS", ignored otherwise. |br|
+Default = 1.600000
+
+**--critical-mass-ratio-white-dwarf-non-degenerate-accretor** |br|
```
avigna, line 286: Same comment as before.
In Options.cpp you say:

    // A value of 0.0 means the mass ratio will never be unstable - this does not guaruntee stability of the MT, just that instability is not based on the mass ratio

This is not consistent with this comment. Which one is right? Fix accordingly.

src/BaseStar.h
```diff
@@ -148,7 +147,9 @@ class BaseStar {
 
             void            CalculateBindingEnergies(const double p_CoreMass, const double p_EnvMass, const double p_Radius);
 
-            double          CalculateDynamicalTimescale() const                                                 { return CalculateDynamicalTimescale_Static(m_Mass, m_Radius); }         // Use class member variables
+    virtual double          CalculateCriticalMassRatio(const bool p_AccretorIsDegenerate) const                 { return 0.0; }                                                     // Default is 0.0
```
avigna, line 150: Why does it needs to be a virtual function? Not saying it is wrong, just a legit question.

src/Remnants.cpp
```diff
@@ -1,86 +1,86 @@
...
- * @param   [IN]    p_Time                      Current age of star in Myr
- * @return                                      Suggested timestep (dt)
- */
-double Remnants::ChooseTimestep(const double p_Time) const {
-
-    double dtk = std::min(std::max(0.1, 10.0 * std::max(0.1, 10.0 * p_Time)), 5.0E2);
-    double dte = dtk;
-
-    return std::max(dte, NUCLEAR_MINIMUM_TIMESTEP);
+#include "Remnants.h"
```
avigna, line 1: I don't see any changes in this file. Am I missing something? Of not, maybe worth unstaging.

src/Remnants.h
```diff
@@ -34,7 +34,9 @@ class Remnants: virtual public BaseStar, public HeGB {
 
     double          CalculateCoreMassOnPhase() const                                                            { return m_Mass; }                                                      // Return m_Mass
 
-    double          CalculateEddingtonCriticalRate() const                                                      { return 2.08E-3 / 1.7 * m_Radius * MYR_TO_YEAR; }       // Hurley+, 2002, Eq. (67)
+    double          CalculateCriticalMassRatio(const bool p_AccretorIsDegenerate) const                         { return 0.0; }                                                         // Should never be called...
```
avigna, line 37: Okay, here you do make changes... is this somewhat correlatd to changes made in Remnats.cpp? (See my previous comment).

avigna: Why should it never be called?

src/constants.h
```diff
@@ -785,6 +787,15 @@ const COMPASUnorderedMap<ZETA_PRESCRIPTION, std::string> ZETA_PRESCRIPTION_LABEL
     { ZETA_PRESCRIPTION::ARBITRARY, "ARBITRARY" }
 };
 
+// Critical Mass Ratio prescription
+enum class QCRIT_PRESCRIPTION: int { NONE, CLAEYS}; //, GE15, GE15_IC};
+const COMPASUnorderedMap<QCRIT_PRESCRIPTION, std::string> QCRIT_PRESCRIPTION_LABEL = {
+    { QCRIT_PRESCRIPTION::NONE,    "NONE" },
+    { QCRIT_PRESCRIPTION::CLAEYS,  "CLAEYS" },
+    //{ QCRIT_PRESCRIPTION::GE15,    "GE15" },
+    //{ QCRIT_PRESCRIPTION::GE15_IC, "GE15_IC" },
```
avigna, line 858: What is GE15_IC? For my own reference.

src/constants.h
```diff
@@ -1284,6 +1295,16 @@ const std::initializer_list<STELLAR_TYPE> ALL_HERTZSPRUNG_GAP = {
 };
 
 
+// (convenience) initializer list for GIANTS (excluding HG)
+const std::initializer_list<STELLAR_TYPE> GIANTS = {
+    STELLAR_TYPE::FIRST_GIANT_BRANCH,
+    STELLAR_TYPE::CORE_HELIUM_BURNING,
```
avigna: Hmn. CORE_HELIUM_BURNING does not necessarily constitute giants...

jeffriley (changes requested):
Generally looks good. Couple of changes:

1. Add new options to category listing in the docs - alphabetically in category if possible.
2. Add brief description of changes to 'What's new' section of docs

Other than that, I approve, pending @avigna and @ilyamandel 

Oh, and I think our testing has slipped recently (me too...).  Have you run tests to compare pre-changes and post-changes?  Do you get what you expect?  The detailed plot?

reinhold-willcox:
@jeffriley Added the sections you requested. 

Regarding testing: I verified that for default options and identical input, this branch and dev produce identical results using `h5diff`, except for the parameters describing the version, start/end/clock times, and the values for Zeta_Hurley (which is expected because that parameter was updated, but the updates were not used).

jeffriley:
Thanks @reinhold-willcox - works for me.  Just awaiting @avigna and @ilyamandel

src/Options.cpp
```diff
@@ -391,37 +391,33 @@ void Options::OptionValues::Initialise() {
...
 
-    m_MassTransferCriticalMassRatioHG                               = false;
-    m_MassTransferCriticalMassRatioHGNonDegenerateAccretor          = 0.40;                                                 // Claeys+ 2014 = 0.25
+    m_MassTransferCriticalMassRatioHGNonDegenerateAccretor          = 0.25;                                                 // Claeys+ 2014 = 0.25
     m_MassTransferCriticalMassRatioHGDegenerateAccretor             = 0.21;                                                 // Claeys+ 2014 = 0.21
 
-    m_MassTransferCriticalMassRatioGiant                            = false;
-    m_MassTransferCriticalMassRatioGiantNonDegenerateAccretor       = 0.0;
-    m_MassTransferCriticalMassRatioGiantDegenerateAccretor          = 0.87;                                                 // Claeys+ 2014 = 0.81
+    m_MassTransferCriticalMassRatioGiantNonDegenerateAccretor       = -1.0;                                                 // Value not used! Claeys+ 2014 uses an equation in mass-radius exponent and core mass (equivalent to Hurley zeta adiabatic definition), so if -1, this value is overwritten later
```
avigna, line 408: What's the difference between 0 and -1?

src/Options.cpp
```diff
@@ -4276,6 +4270,8 @@ COMPAS_VARIABLE Options::OptionValue(const T_ANY_PROPERTY p_Property) const {
         case PROGRAM_OPTION::PPI_LOWER_LIMIT                                : value = PulsationalPairInstabilityLowerLimit();                               break;
         case PROGRAM_OPTION::PPI_UPPER_LIMIT                                : value = PulsationalPairInstabilityUpperLimit();                               break;
 
+        case PROGRAM_OPTION::QCRIT_PRESCRIPTION                             : value = static_cast<int>(QCritPrescription());                                break;
```
avigna, line 4273: Why is this static? Again, probably fine, just question.

avigna (commented):
Done with this check, overall looks good @reinhold-willcox . There is something I think should be further clarified before I approve this PR. There are some numerical choices and conditions that I find somethwat unclear (see, e.g., Options.cpp). I copy some fo the statements found on Options.cpp:

> // A value of 0.0 means the mass ratio will never be unstable - this does not guaruntee stability of the MT, just that instability is not based on the mass ratio

> m_MassTransferCriticalMassRatioGiantNonDegenerateAccretor = -1.0; // Value not used! Claeys+ 2014 uses an equation in mass-radius exponent and core mass (equivalent to Hurley zeta adiabatic definition), so if -1, this value is overwritten later

> ("Critical mass ratio (mA/mD) for MT from a giant star to a degenerate accretor (default = " + std::to_string(p_Options->m_MassTransferCriticalMassRatioGiantDegenerateAccretor) + ")\n 0 is always stable, <0 is disabled.\n Only used for --critical-mass-ratio-prescription CLAEYS, ignored otherwise.").c_str()

This makes me uncertain about how will COMPAS proceed with numerical choices q>0, q=0, q<0. Originally, my mindset was: q>0 leads to use that particular value, q=0 leads to always stable mass transfer (as the condition for instability is never met), and q<1 disables the option and leads to something else (e.g., Soberman or default). This approach still makes sense to me. However, it doesn't need to be the one remaining. In any case, I do think we should have a clear idea of what is going on.

Therefore, what I request is that we are clear on how these options work, and we adapt fix accordingly (if necessary). I suggest we can do this by making similar plots as before, but for three different values: q<0, q=0.0, and q=0.5 (or something). Once we are sure how this options work, we agree that is the sensible approach, and they are correctly described in the code, I will be happy to approve.

P.S.: also several minor questions, which should not be difficult to address/reply.

reinhold-willcox:
@avigna Thanks for the feedback. I'll go through comment by comment, and add something here for non-trivial fixes. 

Edit: All points addressed. If you want me to remove the GIANTS list, let me know, otherwise I think this should be all set.

src/BaseBinaryStar.cpp
reinhold-willcox: Yes, those are my initials, and put there to remind myself to double check that part of the code. But this file is marked as outdated, and I don't see the same comment in my local version, so I think this one is already addressed.

online-docs/pages/User guide/Program options/program-options-list-defaults.rst
reinhold-willcox: That's a mistake, thanks for catching it. I wrote that early on, when the critical mass ratios were checked first and if they were not flagged, the other instability checks were performed (i.e the zetas). I rewrote that part of the code to not do that, but forgot to update this comment. I'll push the change shortly.

online-docs/pages/User guide/Program options/program-options-list-defaults.rst
reinhold-willcox: Yes, I'll fix that.

online-docs/pages/User guide/Program options/program-options-list-defaults.rst
reinhold-willcox: Fixed.

src/BaseStar.h
reinhold-willcox: This function calculates the critical mass ratio on a per stellar type basis (i.e, each type calls the Option for its type to get the constants laid out in Options.cpp). So this one is virtual, and defaults to 0.0 just to give it something to default to, but the BaseStar version should never get called.

src/Remnants.cpp
reinhold-willcox: This was a weird one, the file was full of the single character `^M` or something like that, at the end of each line. I'm not sure why they were there, but I removed them and the code ran fine. My guess is that github recognizes a diff but doesn't know how to show the character...

src/Remnants.h
reinhold-willcox: This is unrelated to the Remnants.cpp weird issue. I copied this from a similar function in Remnants.cpp (maybe something to do with zeta's, I can't recall). The comment was apt, so I kept it. It should only be called for a star that is a donor, which should never happen for NSs/BHs.

src/constants.h
reinhold-willcox: The Ge+15 paper (and others in the series) calculate zetas and critical mass ratios for two cases: adiabatic expansion, and keeping the envelope isentropic. They are similar, but distinct, and it was straightforward to include both in COMPAS. In their downloadable datafile, they use the IC suffix to identify the isentropic model (it stands for something, but I forget what). I'll talk a lot about this tomorrow on the call.

src/constants.h
reinhold-willcox: That's true, I was maybe a bit sloppy with this one. This is a relic from an earlier attempt to setup the critical mass ratios. I didn't end up using this, but figured it might be useful to others in the future. If you think it is more misleading than helpful, I can remove it.

src/Options.cpp
reinhold-willcox: I comment on this a bit to the right. The qCrit for Giants is not a constant but a function, but users may still want to define it as a constant. Elsewhere in the code, it checks if this value is still -1 (and hasn't been updated by a user option), and if so, calculates the value for the function at runtime.

src/Options.cpp
reinhold-willcox: I'll defer to the wisdom of @jeffriley to give you a more enlightened answer. I just copied the formatting from one of the other prescriptions.

src/Remnants.cpp
jeffriley: @reinhold-willcox @avigna 
^M is a carriage return (CR - see the ASCII table).  I'd say someone has edited the file in DOS/Windows.  Unix systems end lines with a line feed (LF) character (see ASCII table...), or '\n'.  DOS/Windows systems end lines with a CRLF sequence, or '\r\n'.  The ^M is the CR, or '\r'.  Unix systems cope with it, but github thought every line had changed...  I thought you had accidentally removed the file, so you were adding it back - I guess not :-)

src/constants.h
jeffriley: @reinhold-willcox are you going to leave the commented parts of this in?  For a future update?

src/Options.cpp
jeffriley: @avigna @reinhold-willcox 

@avigna I will assume you want to know about 'static_cast' (vs any other type of cast) rather than assume you mistake this for a static function call...

This function returns the values of program options so they can be printed/logged to the log files.  In this case, the program option is an 'enum class':

    enum class QCRIT_PRESCRIPTION: int { NONE, CLAEYS };

Let's say we remove the comments, and make it

    enum class QCRIT_PRESCRIPTION: int { NONE, CLAEYS, GE15, GE15_IC};

We use enum classes to help prevent developers accidentally using numbers outside the range of values permitted for the variable.  For example, in this case, the valid qcrit prescriptions are NONE and CLAEYS, GE15, and GE15_IC.  The way it is declared, the integer values are NONE = 0, CLAEYS = 1, GE15 = 2, and GE15_IC = 3.  Our code (presumably) does different things for those values - but only those values.  We don't want someone inadvertently assigning, say, 15 to the qcrit prescription - we don't know what that is.  So that's why we use enum classes - it restricts the values we can assign to a variable of that type and hopefully helps prevent inadvertent errors.

So that's the background.  

When we want to print the value of the option to a log file, we want to print the integer value - so we need to cast the variable to an int.  We could just use

    value = (int)QCritPrescription();

instead of

    value = static_cast<int>(QCritPrescription());

and get the same result.  In this case...  The classic C cast '(int)QCritPrescription()' does no checking - it just converts the value to an int.  'static_cast' on the other hand does some checking to ensure that the variable being cast to an int can actually be cast to an int with no errors.  In this case we know it can, but it is good practice to use 'static_int' over '(int)' - not in all cases, but in most (as always, there are exceptions, and pragmatism should prevail over dogma).  'static_cast' assumes that there is a valid conversion in the language, or an appropriate constructor that makes the cast possible (assuming static typing).

There are other types of cast in C++: 'dynamic_cast', 'const_cast', and 'reinterpret_cast'. Some of those are dangerous (in that if you don't know for sure what you're doing you'll end up causing problems), and they don't apply here anyway (they could, but static_cast is the best choice).  Google them if you really want to know about them - the discussion here would be way too long if I described them...

jeffriley:
@reinhold-willcox  are you going to leave the commented text in here?  Perhaps for a future update?  If so, maybe add a comment to say why it is there
 
    // Critical Mass Ratio prescription
    enum class QCRIT_PRESCRIPTION: int { NONE, CLAEYS}; //, GE15, GE15_IC};
    const COMPASUnorderedMap<QCRIT_PRESCRIPTION, std::string> QCRIT_PRESCRIPTION_LABEL = {
        { QCRIT_PRESCRIPTION::NONE,    "NONE" },
        { QCRIT_PRESCRIPTION::CLAEYS,  "CLAEYS" },
        //{ QCRIT_PRESCRIPTION::GE15,    "GE15" },
        //{ QCRIT_PRESCRIPTION::GE15_IC, "GE15_IC" },
    };

reinhold-willcox:
@jeffriley The next update is ready for a PR, but I am waiting to push it until this one goes through. The commented text will be part of that, so it will only be commented out very briefly.

jeffriley (approved):
Approved

ilyamandel (approved):
Happy with the response to my queries, but have not reviewed the physics carefully -- so please do not pull until @avigna approves.

reinhold-willcox:
@ilyamandel and @jeffriley , I cleaned up the yaml here, and made the PPI/PISN options lowercase. Both are very minor fixes.

jeffriley (changes requested):
> ... and made the PPI/PISN options lowercase

You've only changed the strings in Options.cpp - the same strings exist in Options.h, so you should also change them there.

EDIT: They are also in mixed case in the online docs, so you'll need to change them there too. 

But...

What's the motivation for the change to those options in particular?  If you want option strings to be lowercase (why?  Is there a compelling reason not to have mixed case?), why those ones only and not all the others that are mixed case:

    allow-non-stripped-ECSN
    case-BB-stability-prescription
    critical-mass-ratio-HG-degenerate-accretor
    critical-mass-ratio-HG-non-degenerate-accretor
    critical-mass-ratio-MS-high-mass-degenerate-accretor
    critical-mass-ratio-MS-high-mass-non-degenerate-accretor
    critical-mass-ratio-MS-low-mass-degenerate-accretor
    critical-mass-ratio-MS-low-mass-non-degenerate-accretor
    critical-mass-ratio-helium-HG-degenerate-accretor
    critical-mass-ratio-helium-HG-non-degenerate-accretor
    critical-mass-ratio-helium-MS-degenerate-accretor
    critical-mass-ratio-helium-MS-non-degenerate-accretor
    kick-magnitude-sigma-CCSN-NS
    kick-magnitude-sigma-ECSN
    kick-magnitude-sigma-USSN 
    mass-transfer-thermal-limit-C
    muller-mandel-kick-multiplier-BH 
    muller-mandel-kick-multiplier-NS 
    neutrino-mass-loss-BH-formation
    neutrino-mass-loss-BH-formation-value
    
I say either leave all the mixed case options the way they are, or change them all to lowercase.  But the question remains - is there some reason you particularly want to do this?  I don't really have a preference - other than to do the same thing to all of them - just curious as to your reason.

src/constants.h
avigna: Yes, I think it is more misleading than helpful, particularly if you don't use it. Fully remove to avoid any potential problems.

src/Options.cpp
avigna: Oh wow Jeff, this was a great explanation! Thanks :)

avigna (commented):
(Ignore this)

avigna (changes requested):
Just remove the
```
// (convenience) initializer list for GIANTS (excluding HG)
const std::initializer_list<STELLAR_TYPE> GIANTS = {
    STELLAR_TYPE::FIRST_GIANT_BRANCH,
    STELLAR_TYPE::CORE_HELIUM_BURNING,
```
part and good to go.

reinhold-willcox:
@jeffriley You are right, I was not being consistent. I was sorting the options and the PPI's were first because they were capitalized, so I was trying to fix that. But it's not hugely important, so I've reverted it. 

@avigna Too easy. Let me know if anything else is needed.

avigna:
> Done with this check, overall looks good @reinhold-willcox . There is something I think should be further clarified before I approve this PR. There are some numerical choices and conditions that I find somethwat unclear (see, e.g., Options.cpp). I copy some fo the statements found on Options.cpp:
> 
> > // A value of 0.0 means the mass ratio will never be unstable - this does not guaruntee stability of the MT, just that instability is not based on the mass ratio
> 
> > m_MassTransferCriticalMassRatioGiantNonDegenerateAccretor = -1.0; // Value not used! Claeys+ 2014 uses an equation in mass-radius exponent and core mass (equivalent to Hurley zeta adiabatic definition), so if -1, this value is overwritten later
> 
> > ("Critical mass ratio (mA/mD) for MT from a giant star to a degenerate accretor (default = " + std::to_string(p_Options->m_MassTransferCriticalMassRatioGiantDegenerateAccretor) + ")\n 0 is always stable, <0 is disabled.\n Only used for --critical-mass-ratio-prescription CLAEYS, ignored otherwise.").c_str()
> 
> This makes me uncertain about how will COMPAS proceed with numerical choices q>0, q=0, q<0. Originally, my mindset was: q>0 leads to use that particular value, q=0 leads to always stable mass transfer (as the condition for instability is never met), and q<1 disables the option and leads to something else (e.g., Soberman or default). This approach still makes sense to me. However, it doesn't need to be the one remaining. In any case, I do think we should have a clear idea of what is going on.
> 
> Therefore, what I request is that we are clear on how these options work, and we adapt fix accordingly (if necessary). I suggest we can do this by making similar plots as before, but for three different values: q<0, q=0.0, and q=0.5 (or something). Once we are sure how this options work, we agree that is the sensible approach, and they are correctly described in the code, I will be happy to approve.
> 
> P.S.: also several minor questions, which should not be difficult to address/reply.

@reinhold-willcox actually, I realized I'm still missing the answer to this, as well as the pertinent plot.

jeffriley:
> I was sorting the options and the PPI's were first because they were capitalized, so I was trying to fix that.

If you're sorting in python and you want to ignore case, look at str.casefold.  Or downshift all your strings first.

jeffriley (approved):
> I was sorting the options and the PPI's were first because they were capitalized, so I was trying to fix that.

If you're sorting in python and you want to ignore case, look at str.casefold. Or downshift all your strings first (and maybe use numpy.argsort if you want to keep the mixed case for printing)

reinhold-willcox:
@avigna good point. I haven't implemented that the way you described. I haven't added any checks for values outside of the [0, 1] range, so as you can see for the `fixed_neg0.5` model, the code runs fine, but everything is stable. I would argue that we should not allow a value outside of this range, and that the -1 for Giant donors should be recognized as a placeholder (which cannot actually be implemented, i.e I don't want users to manually set -1). But I take your point that we could make negative values default to another prescription. It depends a bit on how much flexibility and variation you think users will want. Will some people want to use Claeys critical mass ratios for some types, and Ge for others? Or qCrits for some types, and zetas for others? 

If we think these options need a lot of flexibility, it might be good to chat ~in person~ over zoom about them. Let me know what you think. 

image

avigna:
@reinhold-willcox I think that plot shows that the code is doing what it is supposed to do and my comments/thoughts are beyond the scope of this PR (I will raise them in the coding channel and raise an issue if needed, though). I am happy to approve this PR.

avigna (commented):
It all seems good, happy to approve.

avigna (approved):

avigna:
Apologies, I accidentally closed and then re-opened the PR. In any case, it has been approved by the three reviewers and all checks have passes! So @reinhold-willcox , feel free to merge when you consider pertinent, and then we can close this PR.

PR #840, opened 2022-08-12
Add option flexible T convective

reinhold-willcox:
If we set `--envelope-state-prescription FIXED_TEMPERATURE`, we can now specify an arbitrary cutoff temperature with `--convective-envelope-temperature-threshold`

### DIFF
--- a/online-docs/pages/User guide/Program options/program-options-list-defaults.rst
+++ b/online-docs/pages/User guide/Program options/program-options-list-defaults.rst
@@ -190,6 +190,11 @@ Default = :math:`1.5 \times 10^{13}`
 Slope for the Kruckow lambda (see Kruckow et al. 2016 as implemented by Vigna-Gomez et al. 2018). |br|
 Default = −0.833333
 
+**convective-envelope-temperature-threshold** |br|
+Temperature [K] threshold, below which the envelope's of giants are convective. 
+Requires --envelope-state-prescription is FIXED_TEMPERATURE. |br|
+Default = 5370
+
 **--cool-wind-mass-loss-multiplier** |br|
 Multiplicative constant for wind mass loss of cool stars, i.e. those with temperatures below the
 VINK_MASS_LOSS_MINIMUM_TEMP (default 12500K). |br|
--- a/online-docs/pages/User guide/Program options/program-options-list-defaults.rst
+++ b/online-docs/pages/User guide/Program options/program-options-list-defaults.rst
@@ -1001,7 +1001,7 @@ Go to :ref:`the top of this page <options-props-top>` for the full alphabetical 
 
 **Mass transfer physics**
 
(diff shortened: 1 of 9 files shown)


jeffriley (changes requested):
In program-options-list-defaults.rst

    **convective-envelope-temperature-threshold** |br|
    Temperature [K] threshold, below which the envelope's of giants are convective. 
    Requires --envelope-state-prescription is FIXED_TEMPERATURE. |br|
    Default = 5370

Remove the apostrophe

"Requires --envelope-state-prescription is FIXED_TEMPERATURE." - should that not be something like

"Only used for --envelope-state-prescription is FIXED_TEMPERATURE, ignored otherwise"

"Requires" implies the user specifying the option without also specifying "--envelope-state-prescription is FIXED_TEMPERATURE" will result in an error

What about the options category listing in the docs?



"Default = 5370"

In constants.h the value is 5.3703E3


Is the "// RTW" at line 917 of Options.cpp there for a reason?

@ilyamandel should review for physics

reinhold-willcox:
Agreed with most of your points, those are now fixed

> What about the options category listing in the docs?
> 
> "Default = 5370"
> 
> In constants.h the value is 5.3703E3

I skipped the .3 because we don't know temperatures nearly this well, and it comes from 10^3.73, so we're approximating anyway.
 
> Is the "// RTW" at line 917 of Options.cpp there for a reason?

Nah, thanks for catching, that's how I know where I've partially edited some code to come back to it later. It is removed now.

jeffriley:
Thanks @reinhold-willcox 

> I skipped the .3 because we don't know temperatures nearly this well, and it comes from 10^3.73, so we're approximating anyway.

Sure, I get we're approximating, but my point was the documentation should match the code - so one of them should change to match the other.  We shouldn't say we do something but actually do something slightly different.  If we don't know temperature that well, why have that precision in constants.h?

Edit: Just occurred to me that the help text and RunDetails file will show the value from constants.h, so more reason to make the the documentation show what's in constants.h (whether you change what's in constants.h is up to you, but the documentation should match the help text, RunDetails content, and the code)

jeffriley:
Oh - what about adding the new option to the category listing in the docs?

jeffriley (approved):

Physics looks ok to me too, but @ilyamandel might want to review

ilyamandel (changes requested):
Very minor change request: @reinhold-willcox , can you please apply the same changes to Options.cpp that @jeffriley requested for the documentation?

        (
            "convective-envelope-temperature-threshold",                               
            po::value<double>(&p_Options->m_ConvectiveEnvelopeTemperatureThreshold)->default_value(p_Options->m_ConvectiveEnvelopeTemperatureThreshold),                                                                  
            ("Temperature [K] threshold, below which the envelope's of giants are convective. Requires --envelope-state-prescription is FIXED_TEMPERATURE (default = " + std::to_string(p_Options->m_ConvectiveEnvelopeTemperatureThreshold) + ")").c_str()
        )

ilyamandel (approved):

PR #842, opened 2022-08-16
fixed Hurley Z adiabat -x for +x

### LINKED ISSUE #838: Bug in BaseStar::CalculateZadiabaticHurley2002

reinhold-willcox:


ilyamandel (changes requested):
@reinhold-willcox -- This is correct, but if you are using this, why not use the full version of x from Eq. (47) of Hurley+ (2000)?  Otherwise, maybe replace the comment with
"Approximate version of  Eq. (47) of Hurley+ (2000)"

reinhold-willcox:
@ilyamandel That's a good point. I'll add the comment for now, since I don't think anyone is using this model in practice, but if we decide to it would be good to implement it in greater detail.

reinhold-willcox:
This PR is now superceded by #839. Converting this one to draft until the other is accepted, then this one can be closed.

PR #843, opened 2022-08-16
Adds record type property (column) to all log files (except the switch logs)

jeffriley:
Draft until I write the documentation - but documentation is at the top of Log.h

### DIFF
--- a/online-docs/index.rst
+++ b/online-docs/index.rst
@@ -37,6 +37,7 @@ Contents
    
    ./pages/references
    ./pages/contact-us
+   ./pages/whats-new
 
 
 If you use COMPAS in the preparation of a publication, please :doc:`cite COMPAS <./pages/how-to-cite>`.
--- a/online-docs/notebooks/CosmicIntegration.ipynb
+++ b/online-docs/notebooks/CosmicIntegration.ipynb
@@ -175,12 +175,12 @@
     "    \n",
     "Mlower:\n",
     "\n",
-    "    lower limit used for M1 in the user-defined options of the simulation. \n",
+    "    lower limit used for M1 in the pythonSubmit of the simulation. \n",
     "    Needed to recover `true` amount of Msun evolved (see step 4)\n",
     "    \n",
     "Mupper:\n",
     "\n",
-    "    upper limit used for M1 in the user-defined options of the simulation. \n",
+    "    upper limit used for M1 in the pythonSubmit of the simulation. \n",
     "    Needed to recover `true` amount of Msun evolved (see step 4)\n",
     "\n",
     "binaryFraction:\n",
@@ -3339,7 +3339,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.7.6"
+   "version": "3.6.5"
   }
  },
  "nbformat": 4,
--- a/online-docs/pages/User guide/COMPAS output/standard-logfiles-record-specifiers.rst
+++ b/online-docs/pages/User guide/COMPAS output/standard-logfiles-record-specifiers.rst
@@ -25,6 +25,6 @@ All standard log files, except the switch log files, have a record type property
 to idenitify and filter records within a standard log file. The record type property was introduced primarily to support different types of records
 in the detailed output files (BSE and SSE), and is currently unly used in those files, but could be useful for other log files. All record types for
 all standard log files are printed to the log files by default, but records of specific record types can be enabled and disabled using the
-appropriate program option (see e.g. see option ``--logfile-detailed-output-record-types`` in :doc:`program-options-list-defaults` for the detailed
+appropriate program option (see e.g. ``--logfile-detailed-output-record-types`` in :doc:`program-options-list-defaults` for the detailed
 output files).
 
--- a/online-docs/pages/User guide/COMPAS output/standard-logfiles-record-specifiers.rst
+++ b/online-docs/pages/User guide/COMPAS output/standard-logfiles-record-specifiers.rst
@@ -25,5 +25,6 @@ All standard log files, except the switch log files, have a record type property
 to idenitify and filter records within a standard log file. The record type property was introduced primarily to support different types of records
 in the detailed output files (BSE and SSE), and is currently unly used in those files, but could be useful for other log files. All record types for
 all standard log files are printed to the log files by default, but records of specific record types can be enabled and disabled using the
-appropriate program option (see e.g. :doc: PUT SOMETHING HERE for the detailed output files).
+appropriate program option (see e.g. see option ``--logfile-detailed-output-record-types`` in :doc:`program-options-list-defaults` for the detailed
+output files).
 
--- a/online-docs/pages/User guide/COMPAS output/standard-logfiles-record-specifiers.rst
+++ b/online-docs/pages/User guide/COMPAS output/standard-logfiles-record-specifiers.rst
@@ -20,3 +20,10 @@ available for use is:
 The stellar property types (all types except BINARY_PROPERTY AND PROGRAM_OPTION) must be paired with properties from the stellar property list, 
 the binary property type BINARY_PROPERTY with properties from the binary property list, and the program option type PROGRAM_OPTION with properties 
 from the program option property list.
+
+All standard log files, except the switch log files, have a record type property (column). The record type property is an integer that can be used
+to idenitify and filter records within a standard log file. The record type property was introduced primarily to support different types of records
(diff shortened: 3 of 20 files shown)


jeffriley:
@reinhold-willcox I think I may have missed specifying some of the intra-timestep detailed output prints as intra-timestep - can you check?

ilyamandel (commented):
@jeffriley :

In BaseBinary::EvaluateBinary(), should we not print on stellar merger, too, given that we are now providing the inter-step flag?  Otherwise, we might still miss key info from DetailedOutput...

    if (m_PrintExtraDetailedOutput && !StellarMerger()) {
        (void)PrintDetailedOutput(m_Id, BSE_DETAILED_RECORD_TYPE::INTRA_TIMESTEP);                                  // print detailed output record if stellar type changed (except on merger, when detailed output is meaningless)
    }

ilyamandel (commented):
In Log.cpp, I see that you renamed the fields of the fields of the Switch log by changing the capitalisation:

fileDetails.hdrStrings.push_back("Star_Switching"); // append header string for field

...

fileDetails.hdrStrings.push_back("Switching_From");                                                                 // append header string for field
                            fileDetails.hdrStrings.push_back("Switching_To");                                                                   // append header string for field
                            
While I realise this may look prettier, this will make my post-processing code break, and possibly that of other users.  So, unless it's critical for some reason, I'd prefer not to change output parameter names for "beautification" purposes.

Aside: I was going to say that if output record names are changed, they should also be changed in the documentation -- but I don't think these are in the documentation at all; if so, they should probably be added. :)

jeffriley:
> In BaseBinary::EvaluateBinary(), should we not print on stellar merger, too, ...

This implementation was intended to facilitate adding extra calls to things like PrintDetailedOutput(), not actually add them :-)  But no worries, I'll change that - and add the extra calls that Reinhold put in his PR.

> In Log.cpp, I see that you renamed the fields of the fields of the Switch log by changing the capitalisation

Yes, I vacillated between doing that and not doing it.  They were uppercase because when I built those files they really weren't mainstream code, but now they are, and all other header strings are mixed case.  I vacillated for the reason you mention - I might break someone's post-processing code.  In the end I figured you were probably the only person using the switch logs, so the impact wouldn't be great - you can change your post-processing code :-)  But if you really think I should change them back I will... (Edit: and I was going to talk to you about it before moving this from draft...)

> ... but I don't think these are in the documentation

Are the switch log header strings not in the documentation?  If not I'll put them there.

jeffriley:
Added/rationalised calls to PrintDetailedOutput() for binary systems
Removed m_PrintExtraDetailedOutput variable (and associated code) from BaseBinaryStar class
Some minor cleanups:
-    minor formatting and typo fixes
-    removed IncrementOmega() function from the BaseStar and Star classes (anti-patterm and no longer used - if it ever was)
-    tidied up description of MainSequence::UpdateMinimumCoreMass()


PrintDetailedOutput() for binary systems can now be called with the following record types (and the record type is now a required parameter - no default):

```
enum class BSE_DETAILED_RECORD_TYPE: unsigned int {     // BSE_DETAILED_OUTPUT file record type
    INITIAL_STATE,                                      //  0 - record describes the initial state of the binary
    POST_STELLAR_TIMESTEP,                              //  1 - record was logged immediately following stellar timestep (i.e. the evolution of the constituent stars for a single timestep)
    POST_BINARY_TIMESTEP,                               //  2 - record was logged immediately following binary timestep (i.e. the evolution of the binary system for a single timestep)
    TIMESTEP_COMPLETED,                                 //  3 - record was logged immediately following the completion of the timestep (after all changes to the binary and components)
    FINAL_STATE,                                        //  4 - record describes the final state of the binary
    STELLAR_TYPE_CHANGE_DURING_CEE,                     //  5 - record was logged immediately following a stellar type change during a common envelope event
    STELLAR_TYPE_CHANGE_DURING_MT,                      //  6 - record was logged immediately following a stellar type change during a mass transfer event
    STELLAR_TYPE_CHANGE_DURING_MASS_RESOLUTION,         //  7 - record was logged immediately following a stellar type change during mass resolution
    STELLAR_TYPE_CHANGE_DURING_CHE_EQUILIBRATION,       //  8 - record was logged immediately following a stellar type change during mass equilibration for CHE
    POST_MT,                                            //  9 - record was logged immediately following a mass transfer event
    POST_WINDS,                                         // 10 - record was logged immediately following winds mass loss
    POST_CEE,                                           // 11 - record was logged immediately following a common envelope event
    POST_SN,                                            // 12 - record was logged immediately following a supernova event
    POST_MASS_RESOLUTION,                               // 13 - record was logged immediately following mass resolution (ie. after winds mass loss & mass transfer complete)
    POST_MASS_RESOLUTION_MERGER                         // 14 - record was logged immediately following a merger after mass resolution
};
```

I have run two rudimentary tests:

1000 binary systems with detailed-output and switch-log enabled
1000 single stars with detailed-output and switch-log enabled

Both tests ran to completion and produced the expected outputs.

I have not updated the documentation (I'll get to it). The code shows how to use the new 'record-type' functionality - we can add/delete calls to PrintDetailedOutput() as necessary.

This code does cause the detailed output file to have more records than the current COMPAS version. Users don't need to consume all the records in the file - they can filter by record type. The detailed plotter could just plot record types 0, 3, & 4 for example (or whatever else is required).

At the moment all record types are logged to the file. We could implement an option to limit the record types logged to specific files. I'm not sure that's really necessary because users can just ignore the records they don't want to consume, but if we don't do that this change is not backward-compatible - post-processing code that processes the BSE detailed output file will be affected (it won't necessarily fail, but it will process more records than maybe the user wants...).

@reinhold-willcox you referred to the very first call to PrintDetailedOutput() in BaseBinaryStar::Evolve() as being "post the initial timestep". That's only true if you consider the creation and initialisation of the binary as the first timestep - there is no timestep prior to the call to BaseBinaryStar::Evolve(). I have used "INITIAL_STATE" as the record type for that call - we can change it to "TIMESTEP_COMPLETED" if that's preferable (same for the final call to PrintDetailedOutput() - I've used "FINAL_STATE", but we can change that to "TIMESTEP_COMPLETED" if that's preferable).

@ilyamandel I have not changed the switchlog header strings back to uppercase yet...

Looking for feedback.

jeffriley:
In response to @ilyamandel 's comment that changing the case of the switchlog header strings may break post-processing code, and the fact implementing the record type code will affect post-processing code, I suggest we add a new section to the readthedocs documentation pages: a "what's new in this release" section/page that allows us to draw the attention of users to new/changed features that might affect how they use COMPAS.  The new section/page could be scrollable so users reading it could scroll back through changes implemented in earlier versions. We could just dump the new changelog file/entry there, or actually write a more conversational version of the latest change to the changelog file.

At some stage we will be in a position where we have  to make changes that will affect post-processing code - otherwise COMPAS will not improve (very much).  In the  case of the header strings, what happens if we find one of them is wrong - either spelled incorrectly, or maybe the name is misleading or the wrong name was used etc?  We have to fix it, and it might affect post-processing code.  We should try to minimise the need for post-processing code to change in response to new COMPAS versions, but sometimes change will be inevitable (I'm not suggesting the switchlog header strings is one of those occasions...).  I think highlighting changes, certainly significant changes or changes that might affect post-processing code, by describing the  changes on the documentation pages is the best way to minimise disruption.

ilyamandel (commented):
This looks great.  Documentation and changes to detailed output plotter to do [the latter, presumably, in a separate but coordinated PR?], so not clicking approve yet.  One query: do we still need the following comment in Star.cpp?

// log record JR: this should probably be before the star switches type, but this way matches the original code

reinhold-willcox:
@jeffriley This looks good to me. I made some changes to the detailed plotter, but it otherwise works fine with the updates here (based on some brief checks, it's hard to do a comprehensive check that all possible evolution channels look good in the plotter). 

Happy to approve once you've added the documentation fixes mentioned above.

jeffriley:
@ilyamandel @reinhold-willcox 

New changes made:

- Add new option for each standard log file to allow specification of which record types to print. See e.g. '--logfile-detailed-output-record-types'
- Online documentation updated for record types and new options
- Detailed plotter changed to work with record type column (thanks RTW)
- Added new section to online documentation: 'What's new'. Documented record types changes in this new section

@reinhold-willcox I changed the record types from 0-based to 1-based - I looked at your changes to the detailed plotter - I don't think that change will affect the plotter (it looks like you don't actually check the record type?)

I've moved the PR out of draft - I think it's ready to go.

reinhold-willcox (approved):

PR #844, opened 2022-08-16
Adding Alvaro's magnitude calculation module

ilyamandel:
Adding Alvaro's magnitude calculation module as an untracked python utility


jeffriley (approved):
Looks ok to me.

PR #849, opened 2022-09-01
Critical mass ratios ~and Zeta's~ from Ge et al. 2020 

### LINKED ISSUE #837: Add zeta prescription options to help output
The help output states that for `--stellar-zeta-prescription`, the default is SOBERMAN, but doesn't list the other options. Very easy fix, good first issue.

reinhold-willcox:
Implementation of the Critical mass ratios and Donor radial response Zetas as defined in Ge et al. 2020 (and the related papers in this series, Ge et al. 2015 and 2010). 

The datafile from the Ge et al 2020 paper (which specifically discusses only Giant stars) includes the results from the other two papers (for MS and HG stars) and this is what has been copied over into constants.h. 

The prescription provides qCrit or zeta values for solar metallicity stars as a function of mass and radius, presented as a grid. I interpolate on this grid, which is defined between 0.1 and 100 solar masses (and uses end points for masses outside this range). This approach assumes that the radius of a star is a good proxy for its evolutionary state, so probably is not so valid for stripped stars. It also assumes that the radius at a given evolutionary state is consistent with what COMPAS says - in my new paper, I'll show a plot that suggests this is a reasonable assumption, at least at the boundaries between stellar types. However, this assumption likely breaks down for metallicities that deviate significantly from solar. 

One last point to bring up is that in this current implementation, the critical mass ratios (or zetas) are either on or off. You can't set them on a per-stellar-type basis. I'll leave that to a future PR, if there is a consensus that this is desirable.


jeffriley (changes requested):
Looks good @reinhold-willcox .  A few comments after a quick look:

In BaseBinaryStar.cpp:

- BaseBinaryStar::CalculateMassTransfer(): I would remove line 1891, and change line 1878 to "double qCrite = 0.0;" - that way the default is set at declaration, and will be used if it is not set explicity in the code path (It's abour defensive programming - in this case it makes no difference, but if someone later adds some code and doesn't set qCrit, at least it will be covered by the default set at declaration).

- Lines 1904,5, & 6 - why 3 blank lines?  2 is probably enough...

In BaseStar.cpp:

- BaseStar::CalculateZetaAdiabatic(): what is zetaStar for the default case?  Best to initialise zetaStar when it is declared at the start of the function, then that is returned as the default (defensive programming again).

- Lines 1299, 1300, 1301, & 1302 - why 4 blank lines?
- Lines 1320, 1321, 1322, 1323, & 1324 - why 5 blank lines?
- Lines 1346, 1337, 1338, 1339, & 1340 - why 5 blank lines?
- Lines 1464, 1465, 1466, & 1467 - why 4 blank lines?

Why have you renamed CalculateZeta() to CalculateZetaByStellarType()?  It seems to me that the function differentiates between different envelope types.

Documentation needs to be updated for the addition of GE20, GE20_IC options to --critical-mass-ratio-prescription.

changelog.h needs to be updated

reinhold-willcox:
@jeffriley All points addressed. The reason for all the whitespace is it helps me to see where I've been working when I quickly scroll back and forth through a file, but I usually remember to delete it when I'm done. 

I played around a bit with renaming the zeta functions. It was a bit tricky to find a way to name them that would be straightforward to understand for future devs. For the Ge20 prescriptions, the zeta is calculated independently of a stellar type, but for the other ones, zeta depends on either the stellar type or the envelope prescription (or potentially both). By default we have one value for MS, one value for HG, and then a function for other giants, but that's only because we assume HG are all radiative and all other giants are convective. If we mess around with that, e.g with the Teff threshold for convective envelopes, then things change. But I have the `CalculateZetaByStellarType()` function in both `MainSequence.cpp` and `GiantBranch.cpp`, and not in `BaseStar`, so I tried to name it in a way that reflected that distinction.

jeffriley:
CalculateZetaByStellarType()...  My first thought when I saw the name was "Of course it's by stellar type - it's object-oriented".

I get what you want to convey, but I don't think using that name does it.  I think 'CalculateZetaByEnvelopeType()' is better - that's what it actually does, and the "by stellar type" part is implied by OO.

Also, for BaseStar::CalculateInterpolatedQCritOrZetaGe2020()

    /* 
     * Interpolate Ge+20 Critical Mass Ratios and Zetas
     * 
     * double BaseStar::CalculateInterpolatedQCritOrZetaGe2020()
     * 
     * @return                                      Critical mass ratio or zeta for given stellar mass / radius
     */ 

Can you add a more detailed description?  What's the OR in the name for?  By the name I assume one of two possibilities is returned, but what is returned under what circumstances?  And how does the caller know what's returned - or doesn't it matter to them?  And if  it doesn't matter then isn't it really just one thing?  Even though it might be calculated as one thing or another, if the caller doesn't know which is returned and it doesn't matter to them which it is, isn't there a single name for whatever it is that's returned?

reinhold-willcox:
> CalculateZetaByStellarType()... My first thought when I saw the name was "Of course it's by stellar type - it's object-oriented".
>
> I get what you want to convey, but I don't think using that name does it. I think 'CalculateZetaByEnvelopeType()' is better - that's what it actually does, and the "by stellar type" part is implied by OO.

But it doesn't just go by envelope type. It's sort of a mix of envelope and stellar type. But the point is, the functions for it are in different stellar class types, and not just in BaseStar. Admittedly, the naming scheme is not perfect, and I don't know what the best route forward is. But it's an improvement on the current naming scheme, where we sometimes have Z and sometimes Zeta, and also "CalculateZadiabatic" vs "CalculateZeta" even though both are calculating zeta in the adiabatic case, and "CalculateZadiabatic" is only called in the convective case. 



> Can you add a more detailed description? What's the OR in the name for? By the name I assume one of two possibilities is returned, but what is returned under what circumstances? And how does the caller know what's returned - or doesn't it matter to them? And if it doesn't matter then isn't it really just one thing? Even though it might be calculated as one thing or another, if the caller doesn't know which is returned and it doesn't matter to them which it is, isn't there a single name for whatever it is that's returned?

The function returns either a qCrit value or a zeta, but the function mechanics are the same for both so I only wanted to write the function once. Its job is to get a value by interpolating from a few points in the Ge20 data object, and the difference between whether we're getting a qCrit or a zeta, or from the adiabatic or isentropic models, is only down to which index to use. Since the switch to pick the index uses on a mix of different object types, either a `QCRIT_PRESCRIPTION` or `ZETA_PRESCRIPTION` depending on which option is set, I wasn't sure how to cleanly make it a function input argument. I suppose I could make a new enum class in constants.h, for just these 4 prescription types, and then set each one as appropriate in the BaseBinaryStar.cpp call, but that seems like overkill. 

If I made this explicitly clear in the function description, would that be satisfactory?

jeffriley:
>  But the point is, the functions for it are in different stellar class types, and not just in BaseStar. 

That's one of the key concepts of object-oriented programming - runtime polymorphism.  That's what virtual functions are for.  It is assumed that the function acts per stellar type, so adding "ByStellarType" to the name is superfluous and likely to be confusing - runtime polymorphism guarantees it will be "by stellar type".  In this case many of the stellar type classes don't implement the function, so they will execute the function they inherit from either the MainSequence class or the GiantBranch class.

EDIT: For example, the BaseStar class has a virtual function named "CalculateRadiusOnPhase()", and each of the stellar classes has its own function named "CalculateRadiusOnPhase()", where the radius of the star is calculated by stellar type. We didn't name that function "CalculateRadiusOnPhaseByStellarType()" - it is implicit in the nature of OOP that the radius is calculated by stellar type.

> If I made this explicitly clear in the function description, would that be satisfactory?

That's what I was asking for :-)

reinhold-willcox:
@jeffriley I've addressed your requested changes in a way that I hope is satisfactory. I've made the functions "CalculateCriticalMassRatio" and "CalculateZetaAdiabatic" to be very general and to live in BaseStar, where they choose which more specifc funtion to call based on the various options and inputs. Only these functions are ever called in BaseBinaryStar, so users should know right away what they are generally retrieving. 

`CalculateCriticalMassRatio` either calls `CalculateCriticalMassRatioGe2020` or `CalculateCriticalMassRatioClaeys14` depending on which option is set.

`CalculateZetaAdiabatic` either calls `CalculateZetaGe2020` or `CalculateZetaConstantsByEnvelope` depending on which option is set. 

Both  `CalculateCriticalMassRatioGe2020` and `CalculateZetaGe2020` are really just wrappers around `InterpolateGe2020DataObjectForEitherQCritOrZeta`, which I know still has the OR in the name that you don't like, but between the new name and the function description, I think it's clear what is going on. Also, this function is never called directly but only indirectly via the functions above. 

The `CalculateCriticalMassRatioClaeys14` and `CalculateZetaConstantsByEnvelope` replace the functions in all the individual stellar types to call the constants set in the Options (or default) or to calculate some function. But either way, I like that the name clearly shows that these are models for specific use cases of the more general functions above (which have shorter names). 

Let me know if this suits.

reinhold-willcox:
@jeffriley I've also cleaned up how we deal with donor mass loss in CalculateMassTransfer. Previously, our if-then calculated the donor mass loss (either full envelope or root finder), and then called a lot of the same functions with whatever the donor mass loss variable was called. I've tightened this up to remove redundancies and improve readability.

reinhold-willcox:
There was a bug in the previous way I cleaned the mass loss functions, now fixed. I confirmed that dev and this branch produce identical output for default settings.

avigna:
(In the interest of time) I'm happy to review this once @reinhold-willcox have addressed all of @jeffriley's requests.

jeffriley (approved):
Approved.

@avigna's turn...

reinhold-willcox:
@jeffriley and @avigna I was previously misunderstanding the Zeta parameter from Ge et al. (2020). It is not, as I thought, a value from the beginning of the MT sequence, but a value based on the critical mass ratio. In fact, it is functionally equivalent, and thus redundant, with the qCrits from that paper. I've removed references to it in this PR (except I've left the rows in the data object in constants.h, but they are never used). 

Let me know if that was unclear. The critical mass ratio part of the code should still work fine.

ilyamandel (changes requested):
@reinhold-willcox : I finally found a bit of time to spend on COMPAS work today, so here's my review (feel free to use it a second review in lieu of @avigna if he doesn't get a chance to review it).  I split it into requested items and suggestions.

Requested changes:

Make sure latest dev is merged in (changelog, etc.)

BaseStar::InterpolateGe20QCrit() has a fair bit of commented-out code; we should just remove that, since it bloats code.

Suggestions:

I find it rather counter-intuitive that massLossDonor has a negative value (unlike it's previous name dM, which I expect to be negative when mass is lost); I suggest changing the name or changing the sign.

Since zetas are now recomputed only when needed, will they be logged correctly if they weren't needed (i.e., if running with qcrit)?

When running with current default options, has it been checked that the outputs are identical before & after this change?

avigna:
Thanks for jumping in on this @ilyamandel! I will not have time to review this before the end of the year, but high on my priority afterwards. So @reinhold-willcox, depending on your urgency or availability, feel free to use Ilya's review instead. If you want a triple check, give me until the end of January. :)

reinhold-willcox:
Updated, cleaned up the comments, changed the name to massDiffDonor, and verified that the default works the same :)

image

I expect the reported zeta's will not be correct if the option is not used. In my mind, that is not problematic, though we may want to say something in the docs about it. Alternatively, we'd have to only include the output if the option is turned on, which is a bit clunkier and not consistent with what is already done in other parts of the code.

SimonStevenson:
I spent a bit of time looking at this at the code meeting today. I ran a few example binaries using --critical-mass-ratio-prescription GE20 and checked that changing the initial masses below the critical mass ratio leads to a merger. Given that this already has a couple of reviews, I think this is probably ready to go.

ilyamandel:
I expect the reported zeta's will not be correct if the option is not used. In my mind, that is not problematic, though we may want to say something in the docs about it.

I suggest we indeed include this in the docs.

Otherwise, please merge in dev to avoid conflicts and pull into dev. :)  

I'm removing my "Request changes".

ilyamandel (approved):

ilyamandel:
@reinhold-willcox , please merge dev into this branch before resubmitting PR.  Sorry for my mistake.

ilyamandel:
Wait, dev was merged in.  Something must have just gone very wrong on my end.  I am terribly sorry.  

The compilation warnings I mentioned above may still be relevant, though.

PR #854, opened 2022-09-14
fixed yaml log file names

reinhold-willcox:
Related to issue JD mentioned in the devel_coding slack channel.


jeffriley (approved):
Approved

PR #856, opened 2022-09-26
fixed macleod gamma interpolation

reinhold-willcox:
The previous interpolation for the Macleod gamma (for specific angular momentum) was incorrect. Previously, it took the gamma value at the accretor and the gamma value at the L2 point, and linearly interpolated between those two values. This is not necessarily incorrect, but also not really the most intuitive approach. 

Now, it calculates the position of the accretor and the position of the L2 point in the orbit, and linearly interpolates on the position, before calculating the appropriate gamma value at that position. Because gamma is a quadratic function of this orbital position, the difference is negligible at the boundaries but strongest in between.

### DIFF
--- a/src/BaseBinaryStar.cpp
+++ b/src/BaseBinaryStar.cpp
@@ -1654,6 +1654,7 @@ double BaseBinaryStar::CalculateRocheLobeRadius_Static(const double p_MassPrimar
  *
  * This is gamma (as in Pols's notes) or jloss (as in Belczynski et al. 2008
  * which is the fraction of specific angular momentum with which the non-accreted mass leaves the system.
+ * Macleod_linear comes from Willcox et al. (2022)
  *
  * Updates class member variable m_Error      JR: todo: revisit error handling (this could be a const function)
  * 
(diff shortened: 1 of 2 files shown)


jeffriley (commented):
Looks good.  Just one comment.

I'm not a fan of using snake case for local variable names.  Our style guide indicates camel case for local variables.  The only time we come close to snake case is when we use the "p_" and "m_" prefixes for function parameters and class member variables - but the rest of the name is camel case.  I don't really care which one we settle on (I prefer camel case, but in the end if most people prefer snake case then we'll use snake case), but 95% of the code uses camel case (that's why the style guide suggests camel case for local variables).  Using a different style pushes the code base to inconsistency and messiness that we're trying to avoid.  So unless there's a strong reason for not sticking to the same style as the rest of the code, I think you should use camel case instead of snake case.

reinhold-willcox:
Nah, you're absolutely right, apologies for the oversight.

jeffriley (changes requested):
Thanks @reinhold-willcox 

I think you missed one:

            double a_gamma = aAcc + (aL2 - aAcc)*OPTIONS->MassTransferJlossMacLeodLinearFraction();
            gamma = aGamma*aGamma*(1+q)*(1+q)/q;

Happy to approve once that's fixed

jeffriley (approved):
Approved

PR #863, opened 2022-11-04
begin adding support for COMPAS python interface

avivajpeyi:
There is a bunch of awesome COMPAS python code in the `utils` folder. For users to use the code, they need to do something like: 
```python
# Import COMPAS root directory and set data
compasRootDir   = os.environ['COMPAS_ROOT_DIR']
# Import COMPAS specific scripts
sys.path.append(compasRootDir + '/postProcessing/PythonScripts/CosmicIntegration/')
```

This can be a bit clunky. Instead, maybe the useful python scripts can be made `pip` installable so that users can just do something like 
```python
from compas.post_processing import CosmitIntegrator
```

This PR is a start on this. 
At the moment, I have only added `h5view` to the `compas` pip install. 
I have added a rudimentary test to check that this is executing (albeit this test isn't too useful atm) 

The interface looks something like the following:
Screen Shot 2022-11-04 at 12 35 16 pm

### DIFF
--- a/.github/workflows/lint-and-pytest.yml
+++ b/.github/workflows/lint-and-pytest.yml
@@ -1,0 +1,36 @@
+# Tests that py code is formatted correctly, can install, and runs tests
+name: lint-and-pytest
+
+on:
+  push:
+    paths: ['compas/**', 'tests/**']
+  pull_request:
+    paths: ['compas/**', 'tests/**']
+
+jobs:
+#  pre-commit: # COMMENTING OUT FOR NOW AS THERE WILL BE ALOT OF CHANGES
+#    runs-on: ubuntu-latest
+#    steps:
+#    - uses: actions/checkout@v2
+#    - uses: actions/setup-python@v2
+#    - uses: pre-commit/action@v2.0.3
+
+  build-n-test:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.8]
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install -e .[dev]
+    - name: Test with pytest
+      run: |
+        pytest tests/
--- a/.gitignore
+++ b/.gitignore
@@ -42,3 +42,7 @@ COMPAS
 
 Makefile-reinhold
 .gitignore
+
+# python setup files
+compas.egg-info
+!compas/
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,19 @@
+# file: .pre-commit-config.yaml
+# lints the .py files to use flake8 compliant 'black' formatting (and sorts imports)
+
+repos:
+-   repo: https://github.com/ambv/black
+    rev: 22.3.0
+    hooks:
+    - id: black
+-   repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v2.0.0
+    hooks:
+    - id: flake8
+    - id: requirements-txt-fixer
+    - id: trailing-whitespace
+-   repo: https://github.com/pre-commit/mirrors-isort
+    rev: v5.10.1
+    hooks:
+    - id: isort
+
--- a/compas/__init__.py
+++ b/compas/__init__.py
@@ -1,0 +1,11 @@
+# -*- coding: utf-8 -*-
+
+__all__ = []
+
+__author__ = "Team COMPAS"
+__email__ = "compas email"
+__uri__ = "https://github.com/TeamCOMPAS/COMPAS"
+__license__ = "MIT"
+__description__ = "COMPAS"
+__copyright__ = "Copyright 2022 COMPAS developers"
+__contributors__ = "https://github.com/TeamCOMPAS/COMPAS/graphs/contributors"
--- a/compas/h5view.py
+++ b/compas/h5view.py
@@ -15,8 +15,6 @@ h5view.py [-h] [-f FILENAME_FILTER] [-r [RECURSION_DEPTH]] [-S] [-H]
                  [-V SEED_LIST [SEED_LIST ...]]
                  input [input ...]
 
-
-
 HDF5 file content viewer.
 
 positional arguments:
@@ -775,4 +773,4 @@ def main():
 
 
 if __name__ == "__main__":
-    main()
+  main()
--- a/compas/h5view.py
+++ b/compas/h5view.py
@@ -15,6 +15,8 @@ h5view.py [-h] [-f FILENAME_FILTER] [-r [RECURSION_DEPTH]] [-S] [-H]
                  [-V SEED_LIST [SEED_LIST ...]]
                  input [input ...]
 
+
+
 HDF5 file content viewer.
 
 positional arguments:
--- a/compas/h5view.py
+++ b/compas/h5view.py
@@ -773,4 +773,4 @@ def main():
 
 
 if __name__ == "__main__":
-  main()
+    main()
--- a/online-docs/pages/Getting started/building-COMPAS.rst
+++ b/online-docs/pages/Getting started/building-COMPAS.rst
@@ -72,3 +72,17 @@ If you are using MacOS and running into linking issues with the boost libraries,
 
 In some Mac installations, the GNU C++ compiler is not installed how we might expect, so trying to compile and link with ``clang++`` might help.
 
+Installing python interface
+---------------------------
+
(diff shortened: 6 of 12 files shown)


jeffriley (changes requested):
@avivajpeyi Is this ready for review?

A couple of quick comments:

(a) I think you have included files you didn't mean to - e.g. RunDetails, and h5View.py from test (or is that test data?)?
(b) Does the pip install work?  If it doesn't work yet, you shouldn't include it in the public documentation.

avivajpeyi:
> @avivajpeyi Is this ready for review?
> 
> A couple of quick comments:
> 
> (a) I think you have included files you didn't mean to - e.g. RunDetails, and h5View.py from test (or is that test data?)? 
> (b) Does the pip install work? If it doesn't work yet, you shouldn't include it in the public documentation.

Hey @jeffriley 

(a) I think `RunDetails` is generated by running Compas? I thought it was being used in `h5view`?  I had to move `h5view` from `utils`--> `compas/h5view.py` 

(b) Ah yes, i haven't set up the pypi stuff yet, so this is meant to be `pip install .`

jeffriley:
@avivajpeyi 

 > I think RunDetails is generated by running Compas? 

Yes it is, but we don't push it to the repo - the only person that needs it is the person  that ran COMPAS.  Or is this just test data for something?

> I thought it was being used in h5view?

h5View.py will read a COMPAS hdf5 output file and print details in human-readable form - but we don't need a copy of RunDetails on the repo to do that.  Oh... so it's test data for h5view for you notebook?  Why do you need a notebook for h5view - there is already documentation for h5view...  And h5view.py doesn't actually read a RunDetails file - it reads the RunDetails group from an hdf5 file - they are different (same contents, but different objects).

> I had to move h5view from utils--> compas/h5view.py

Why?  For your pip install?

Incidentally, you need to make it clear in the documentation what this is actually installing - if I don't have COMPAS installed and I do this, can I run COMPAS?

Btw, h5copy.py (in utils) is the tool most, if not all, users of COMPAS would use.

jeffriley (commented):
One more comment: I don't know that the toctree for the "COMPAS Output" documentation page (online-docs/pages/User guide/COMPAS output/output.rst) is the right place for the link to the documentation for the python  interface - might have to give that some more thought.

avivajpeyi:
**Comment 1:**
> Or is this just test data for something?

Yes, the files in `test/test_data/COMPAS_Output` is just test data. Ideally, I would have liked to generate this by running `COMPAS --number-of-systems 5` before testing begins (this can be done in the `setUp()` function of the test.

To start, i thought it would be ok to just include the output from `COMPAS --number-of-systems 5`


**Comment 2**
> utils/h5View.py --> compas/h5View.py 


Yeah, I think the goal will be to start moving files to the `compas` dir -- this is the dir that will be installed when a user runs `pip install .`, and eventually `pip install compas` (without having to clone the repo) 

**Comment 3**
>  you need to make it clear in the documentation what this is actually installing


Yeah! ok ill update the docs 

**Comment 4**
>  if I don't have COMPAS installed and I do this, can I run COMPAS?

I think eventually itd be great if we can make COMPAS cpp code also get installed via the pip install (so both the CPP and the python code are installed simultaneously :)  There may be a way to avoid making users edit the make file/deal with boost stuff. But this is probably a future PR 


**Comment 5**
> Btw, h5copy.py (in utils) is the tool most, if not all, users of COMPAS would use.

Cool! will add that in the next PR

jeffriley:
@avivajpeyi 

> Yeah, I think the goal will be to start moving files to the compas dir -- this is the dir that will be installed when a user runs pip > install ., and eventually pip install compas (without having to clone the repo)

If you are going to move utilities out of the utils directory then you need to make sure the documentation reflects where the utilities are - and that needs to take into account that people may not use pip install - they may just clone the repo and get the utilities that way.

I'm not sure it makes complete sense to move the utilities just for the pip install - don't forget, not everyone will use pip install.  If I clone the repo I get the utilities without the need to run pip install - and I can run them perfectly well without having to do the pip install - so let's think about making changes that affect everyone.  Why can't we have the utilities in compas/utils - is pip not capable of creating and installing into subdirectories?  Let's talk about this before we just start moving things around.

Also, if I clone the repo and get the utilities installed that way, I can run them without the need to do this:

    # Import COMPAS root directory and set data
    compasRootDir   = os.environ['COMPAS_ROOT_DIR']
    # Import COMPAS specific scripts
    sys.path.append(compasRootDir + '/postProcessing/PythonScripts/CosmicIntegration/')

h5copy.py and h5view.py can be run just by executing something like:

    python3 path-to-h5view/h5view.py -H path-to-hdf5-file/filename

There is no need to set COMPAS_ROOT_DIR or anything else.  That needs to remain - users should be able to clone the repo and run things like h5view.py and h5copy.py from the commandline just as they can now.

> I think eventually itd be great if we can make COMPAS cpp code also get installed via the pip install 

Sure - but if you compile and link the code and install an executable, the executable needs to be for generic architectures, and will presumably just be the latest version of COMPAS.  Users sometimes have a need to run older versions of COMPAS - for that they need the docker images, or to recompile with an older version of the code.  What versions of the dependencies (boost, gsl, hdf5 etc.) would we be baking into the executable file we install for the user?

The dependencies - gsl, boost, hdf5 libraries etc. - need to be installed with the pip install.

If we do this (install the entire COMPAS suite with pip) we need, again, to make sure the documentation clearly reflects what is being installed.


> There may be a way to avoid making users edit the make file/deal with boost stuff.

Nobody needs to edit the Makefile, for boost or anything else.  Fully-qualified paths to all the dependencies can be passed to make from the commandline - it is not necessary (and not recommended) to modify the Makefile.  See the example in the online documentation for how to specify the c++ compiler without the need to modify the Makefile - specifying paths to include files and libraries is the same:

https://compas.readthedocs.io/en/latest/pages/Getting%20started/building-COMPAS.html

Also note this comment from above:

> One more comment: I don't know that the toctree for the "COMPAS Output" documentation page (online-docs/pages/User 
> guide/COMPAS output/output.rst) is the right place for the link to the documentation for the python interface - might have 
> to give that some more thought.

The page you have added the link to the python interface documentation is the index page for the section of the documentation that describes the COMPAS output files.  The utilities (h5copy, h5view) etc. manipulate hdf5 files.  Some users will not use hdf5 files - they might have COMPAS output csv files instead - in which case h5copy, h5view etc. are irrelevant.

I think the link to the python interface should be in the section of the documentation that describes utilities (i.e. the post-processing tools section), not the output files.  So somewhere here:

https://compas.readthedocs.io/en/latest/pages/User%20guide/Post-processing/post-processing.html

avivajpeyi:
Hey Jeff, sorry I probably missed docs for the utils folder.


Sure, happy to stop on this move. It seems like if folks are fine with this i shouldn't butt in :) 

Feel free to close this and the other PR

jeffriley:
@avivajpeyi 

You're not butting in...  I'm very happy for you to do this, as are other people, but we need to make sure that everyone is catered for, so we need to do this in a way that works for everyone.  It's doable - it just needs some thought and care.

So no, don't stop - maybe take a bit of time and run COMPAS a few times to get yourself familiar with how it works, what its output files are, how it is installed, how it is used - then you can work out how to make it better. I'm sure lots of people would appreciate a pip install, but it needs to work for everyone, that's all.

avivajpeyi:
ah ok -- maybe in that case, can I get on a call with you to sort out the best way going forward? We can maybe schedule a time on slack

I noticed that COSMIC a popsynth code has packaged their python + fortran code in quite a clean way.  
https://github.com/COSMIC-PopSynth/COSMIC

jeffriley:
@avivajpeyi

No idea about COSMIC, but it sounds like you should be able to make the same thing work for COMPAS.

The bottom line for me is that cloning the repo and installing via pip should co-exist - a user should be able to choose to do either.  What happens if I clone the repo, and then do a pip install?  Or vice versa?  I have no idea...  But as long as I can choose to do either, I'm happy.  Right now if I clone the repo I have to also find and install the dependencies - boost, hdf5 libraries etc.  That can be a pain, so if pip can do that for me, even better. We shouldn't lose any flexibility - users should always have the choice. And the documentation has to be clear. Do that and we're good :-)

I don't have time this week, but anytime after Fri 11 is good.

avivajpeyi:
Ah ok this might just be some miscommunication on my part. 


Installing from source with pip: 
```
git clone compos
cd compas
pip install . -e 
```
This allows the users to install from source -- and use symlinks. Ie they can make changes to the python files without having to re-install  (however, if one eventually adds the COMPAS cpp src to get built with the pip-install, they would need to run pip install again to build the cpp code after changes to the cpp code) 


If the user doesn't need to make changes to compas, once we add COMPAS to pypi (this can be automated with a CI action), users will be able to do something like: 
```
pip install compas
```
to get a working version of compas. 

Happy to wait till you're free to chat more on this

PR #864, opened 2022-11-04
Add h5copy to pip install

avivajpeyi:
Adds in h5copy to the pip install, adds some starter tests and docs. 

Requires #863 to be completed


ilyamandel:
Converted to Draft since it's waiting on #863 .

PR #867, opened 2022-11-28
Setting up the two-stage common envelope formalism

ilyamandel:
This sets up the framework for the two-stage common envelope formalism of Hirai & Mandel (2022).  The default behaviour (ENERGY formalism) is unchanged.  In fact, for the time being, the TWO_STAGE formalism also produces identical results to the ENERGY formalism -- but this does all of the background set-up work so that Lewis can implement the 2-stage formalism by only modifying two functions, `GiantBranch::CalculateConvectiveEnvelopeMass()` and `BaseStar::CalculateConvectiveEnvelopeBindingEnergy`.


jeffriley (changes requested):
Just a quick code inspection (I didn't try to compile or run - I assume you have run tests and are satisfied your new code does what's expected):

(1) program-options-list-defaults.rst, line 138 typo misspells 'formalism':

    ``ENERGY`` is the standard alpha-lambda formlism; ``TWO_STAGE`` is the formalism of Hirai & Mandel (2022) |br|


(2) BaseBinaryStar.cpp, around line 1556.  The conditionals 

    if( OPTIONS->CommonEnvelopeFormalism() == CE_FORMALISM::ENERGY ) { ... }
    if( OPTIONS->CommonEnvelopeFormalism() == CE_FORMALISM::TWO_STAGE ) { ... }

are not sufficient if someone adds a third formalism and forgets, or doesn't know, to modify this section of code.  Defensive programming would ensure a default here.

It looks like lines 1592 and 1593 are not aligned (indent-wise) with the following code.


(3) BinaryConstituentStar.cpp, line 277.  You removed my comment, but didn't answer the question...  Why is this statement required?  Can we add a comment that explains why it is there?  Does the value "0.00001" need a name (in constants.h)?

    if (m_CEDetails.lambda < 0.00001) m_CEDetails.lambda = 0.0;

ilyamandel:
Thanks, @jeffriley !

All issues addressed.

Yes, it passes all of my checks.  Basically, it does nothing for now (as in, the default "ENERGY" version is identical to what we had before, and the new TWO_STAGE version is for now producing the same results, too) -- this just sets up the framework to make it easier for a COMPAS newbie to implement the "astrophysics" bits of TWO_STAGE in the future.  So your review is exactly what's needed at this stage. :-)

jeffriley (approved):
All good - thanks @ilyamandel

PR #869, opened 2022-11-30
Fix Time pre-MT

### LINKED ISSUE #868: Time pre and post MT
**Describe the bug**
Pointed out to me by Andreas Konstantinou. The parameters `Time<MT` and `Time>MT` in BSE_RLOF are always identical, since in both cases we're just storing the current time. This should be changed, so that `Time<MT` reflects the time in the previous timestep. 

**Label the issue**

 - COMPAS v02.33.02

reinhold-willcox:
Resolves issues #868. Previously, `Time<MT` and `Time>MT` were always identical. Now `Time<MT` shows the previous timestep's time.


jeffriley (changes requested):
Changes generally look good, but a couple of changes to the documentation are required:

First, the definition of the default RLOF record:

https://compas.readthedocs.io/en/latest/pages/User%20guide/COMPAS%20output/standard-logfiles-default-record-specifications-BSE-rlof.html#bse-rlof-parameters


And also the description of the binary proerties:

https://compas.readthedocs.io/en/latest/pages/User%20guide/COMPAS%20output/standard-logfiles-record-specification-binary.html#binary-properties


And since you've changed the column header (and added a new one), I think you should add an entry to "What's new" so that people are aware (or can be if they look at "What's new":

https://compas.readthedocs.io/en/latest/pages/whats-new.html

reinhold-willcox:
done

jeffriley (approved):
All good

PR #872, opened 2022-12-14
Fixes problem with no content in switchlog files (issue #870)

### LINKED ISSUE #870: Switch log is empty
**Describe the bug**
First identified by @solfreludio. When the switch log is enabled (in SSE or BSE), it creates an empty directory. None of the stellar type switches are correctly printed. 

This issue was introduced in v2.32.00, when the logfile record types were added, so I'm assigning @jeffriley. 

**Label the issue**

**To Reproduce**
```COMPAS --switch-log ```

**Expected behavior**
Stellar type switches should be printed.

**Version**
 - COMPAS v02.34.00

jeffriley:
Fixes problem with no content in switchlog files (issue #870 - introduced in v2.32.00)

Close #870 when merged.


jeffriley:
Hmm, just noticed a couple of extraneous changes crept in - they were in my branch from when we were checking the p_NoCheck parameter...  They'll make no substantial difference to the code (and actually I think they make the code closer to correct and certainly more consistent).  I can remove them from here or leave them - whatever works.  When we rationalise the whole p_NoCheck thing these changes will be made anyway.

nrsegovia:
I think you left a couple of testing-related files in there (`src/ssegrid.txt` and `src/output`)

jeffriley:
I did.  That's what I get for hurrying. Ok, I'll clean it up later today.

jeffriley:
Ubuntu 18.04 brownout - will need to wait a day to test the build.

@reinhold-willcox I think we should move to Ubuntu 20.04 or 22.04.  I run 20.04 at home with no problems, but I think 22.04 is the latest LTS.

ilyamandel (approved):
Looks good, thank you for the fixes!

PR #875, opened 2022-12-17
updated ubuntu version in compile workflow from 18.04 to 22.04

reinhold-willcox:


jeffriley (approved):

Issue #876, opened 2023-01-05
Units on post-CEE semi-major axis


nrsegovia:
**Describe the bug**
A comparison between the separation given by the RLOF logs against the CEE ones showed me that separation post CEE was incredibly small, implying that most stars were touching post-CEE. 

Investigating this issue, I found that the ratio of `semiMajorAxisPostRLOF/semiMajorAxisPostCEE` is equal to `46238.77350477` in all cases, which is consistent with ~ `10000000/215` where `215` is just ~1AU in Rsol units. So I believe this is just a misplaced/missing unit conversion.

Note that, as far as I have been able to test, this issue only affects the log files. The flag for stellar merger is not triggered by this unusual post-event separation value.

**Label the issue**

`urgency_low`      - This issue is not urgent

`severity_minor`    - This is a minor bug with minimal impact

**To Reproduce**
- Run ~100,000 binaries, create RLOF and CEE logs.
- Select stars evolving into HeWDs (previous stellar type !10), although I believe this would happen no matter the stellar type (as long as it is the same episode).
- Compare the separation post-event. Other properties such as pre/post-event mass, pre/post-event radius, pre-event separation match perfectly.

**Expected behavior**
The separation post event shown in the CEE and RLOF logs should be similar, as both should be expressed in Rsol units (according to the documentation).

**Versioning (please complete the following information):**
 - COMPAS v02.34.02

nrsegovia:
[Line 1659 in BaseBinaryStar.cpp](https://github.com/TeamCOMPAS/COMPAS/blob/950b01aaacbb7ffc1743edbfb81a224d851a741d/src/BaseBinaryStar.cpp#L1659) seems to be the issue. The / operator should be *, I will submit a PR ~tomorrow.

nrsegovia:
Fixed in #877

PR #877, opened 2023-01-06
Fixed semi-major axis units on post-CEE logs

nrsegovia:
Small fix, the conversion from AU to Rsol was wrong. See #876.


jeffriley (changes requested):
Good catch.  The value is only used to print to the logfiles at the moment, so the defect should have had no other effects.

Need to update changelog.

nrsegovia:
> Good catch. The value is only used to print to the logfiles at the moment, so the defect should have had no other effects.
> 
> Need to update changelog.

I just followed the style from previous updates in `changelog.h`, please let me know if further changes are needed.

jeffriley (approved):
Looks good - thanks!

jeffriley:
@reinhold-willcox Looks like we're failing on one of t he dependencies - can you take a look?

PR #887, opened 2023-01-19
Add che demo

FloorBroekgaarden:


FloorBroekgaarden:
ready to be reviewed now :)

reinhold-willcox (changes requested):
Looks really good so far! Here are some (mostly cosmetic) fixes that I think would improve the PR. 

One small change in the CHE_evolution_demo would be below Question 7, I would change the hard-coded path to use the variable set at the top of the notebook. Same with the BH spin demo (you can simply comment out the `path = ` lines in later cells). 

In `calculate_black_hole_spin_pop_synth.ipynb`, you have `from spin_class import *`. As I understand, this is generally bad practice, but for this particular case it might not matter so much. But maybe since this is a tutorial, we should try to stick to cleaner coding. If possible, I would prefer to have this cleaned up, but I'm happy for others to disagree. 

In the same file, you also have an anomalous line of html that slipped in somehow `<div class="alert ...">`. There are some commented lines here too in the plotting scripts, which should probably be removed. Lastly, I think the `plt.show()` lines should be removed. When they are present, each plot is reprinted every time a new `plt.show()` is called, so that by Example 5 I have to scroll through a bunch of plots I've already seen previously, and it's a bit confusing. 

In both these tutorials, the download time for the datasets is quite large. I would suggest putting all the relevant links at the top of the files so students can be working through the files while the downloads are going (instead of getting halfway through and having to start a new download). I might also request that different test files be used, i.e variations that hold only the DCO systems you want, though that might be an annoying bit of extra work for you. 

Also, you have a few files in your demo which differ from the existing one in the repo by a few parameter names (since the datafile is from an older version of Compas) and a bit of whitespace. This includes ClassCOMPAS.py, selection_effects.py, totalMassEvolvedPerZ. Would it be possible to rename the files e.g to have `-old.py` or something like that appended, to highlight this difference? Or would that cause other issues?

Similarly, there are differences between the FastCosmicIntegration files, but these seem to be more dramatic, i.e the code routines seem to have been updated. Is it worth updating the version in your tutorial, or does it not matter? I would also like to see this one renamed with a `-old` suffix or similar. 

I quickly glanced at the spin_class.py file, and it looks really nice! Though you've unfortunately had to create 2 versions of the class, one to work with older Compas datasets. This highlights a concern going forward: do we need 2 versions of all post-processing routines to handle older data? Not something to fix here, but should be discussed at some point.

jeffriley:
> Though you've unfortunately had to create 2 versions of the class, one to work with older Compas datasets. 
> This highlights a concern going forward: do we need 2 versions of all post-processing routines to handle older data?
> Not something to fix here, but should be discussed at some point.

I don't think it would be too difficult to write a tool to convert older COMPAS output files to the new format.  We can then convert the data for tutorials etc., and also make the tool available on the repo so other users who have older data can convert to the new format if necessary.

reinhold-willcox:
> > Though you've unfortunately had to create 2 versions of the class, one to work with older Compas datasets.
> > This highlights a concern going forward: do we need 2 versions of all post-processing routines to handle older data?
> > Not something to fix here, but should be discussed at some point.
> 
> I don't think it would be too difficult to write a tool to convert older COMPAS output files to the new format. We can then convert the data for tutorials etc., and also make the tool available on the repo so other users who have older data can convert to the new format if necessary.

This would likely work for most use cases, but not if a parameter is used that didn't exist in older compas versions. This isn't a reason not to do it, just something to be mindful of.

jeffriley:
> This would likely work for most use cases, but not if a parameter is used that didn't exist in older compas versions.
> This isn't a reason not to do it, just something to be mindful of.

That's true, but well-written post-processing tools would either handle missing data, or fail gracefully...

reinhold-willcox:
> > This would likely work for most use cases, but not if a parameter is used that didn't exist in older compas versions.
> > This isn't a reason not to do it, just something to be mindful of.
> 
> That's true, but well-written post-processing tools would either handle missing data, or fail gracefully...

I'm all about falling gracefully!

FloorBroekgaarden:
thank you both so much for the feedback. Here are the changes that I made: 
TDLR Summary: I implemented all the suggestions, all done & ready to be merged 

Detailed reply:
 - In calculate_black_hole_spin_pop_synth.ipynb, you have from spin_class import *. As I understand, this is generally bad practice, but for this particular case it might not matter so much. But maybe since this is a tutorial, we should try to stick to cleaner coding. If possible, I would prefer to have this cleaned up, but I'm happy for others to disagree. 
--> done : changed it to ```from spin_class import calculate_spin''' 

 - In the same file, you also have an anomalous line of html that slipped in somehow <div class="alert ...">. 
 --> done

 - There are some commented lines here too in the plotting scripts, which should probably be removed.
 --> done 

-  Lastly, I think the plt.show() lines should be removed. When they are present, each plot is reprinted every time a new plt.show() is called, so that by Example 5 I have to scroll through a bunch of plots I've already seen previously, and it's a bit confusing.
--> done 

- In both these tutorials, the download time for the datasets is quite large. I would suggest putting all the relevant links at the top of the files so students can be working through the files while the downloads are going (instead of getting halfway through and having to start a new download). I might also request that different test files be used, i.e variations that hold only the DCO systems you want, though that might be an annoying bit of extra work for you 
- -> done (added a link to the downloads in the beginning). Many of the datasets already do use a subset of the data (only DCOs). 


 - Also, you have a few files in your demo which differ from the existing one in the repo by a few parameter names (since the datafile is from an older version of Compas) and a bit of whitespace. This includes ClassCOMPAS.py, selection_effects.py, totalMassEvolvedPerZ. Would it be possible to rename the files e.g to have -old.py or something like that appended, to highlight this difference? Or would that cause other issues?
 - -> done 

- Similarly, there are differences between the FastCosmicIntegration files, but these seem to be more dramatic, i.e the code routines seem to have been updated. Is it worth updating the version in your tutorial, or does it not matter? I would also like to see this one renamed with a -old suffix or similar.
- -> done!

FloorBroekgaarden:
thank you both so much for the feedback. Here are the changes that I made:
TDLR Summary: I implemented all the suggestions, all done & ready to be merged

Detailed reply:

In calculate_black_hole_spin_pop_synth.ipynb, you have from spin_class import *. As I understand, this is generally bad practice, but for this particular case it might not matter so much. But maybe since this is a tutorial, we should try to stick to cleaner coding. If possible, I would prefer to have this cleaned up, but I'm happy for others to disagree.
--> done : changed it to ```from spin_class import calculate_spin'''

In the same file, you also have an anomalous line of html that slipped in somehow

.
--> done
There are some commented lines here too in the plotting scripts, which should probably be removed.
--> done

Lastly, I think the plt.show() lines should be removed. When they are present, each plot is reprinted every time a new plt.show() is called, so that by Example 5 I have to scroll through a bunch of plots I've already seen previously, and it's a bit confusing.
--> done

In both these tutorials, the download time for the datasets is quite large. I would suggest putting all the relevant links at the top of the files so students can be working through the files while the downloads are going (instead of getting halfway through and having to start a new download). I might also request that different test files be used, i.e variations that hold only the DCO systems you want, though that might be an annoying bit of extra work for you

-> done (added a link to the downloads in the beginning). Many of the datasets already do use a subset of the data (only DCOs).

Also, you have a few files in your demo which differ from the existing one in the repo by a few parameter names (since the datafile is from an older version of Compas) and a bit of whitespace. This includes ClassCOMPAS.py, selection_effects.py, totalMassEvolvedPerZ. Would it be possible to rename the files e.g to have -old.py or something like that appended, to highlight this difference? Or would that cause other issues?

-> done

Similarly, there are differences between the FastCosmicIntegration files, but these seem to be more dramatic, i.e the code routines seem to have been updated. Is it worth updating the version in your tutorial, or does it not matter? I would also like to see this one renamed with a -old suffix or similar.

-> done!


> Looks really good so far! Here are some (mostly cosmetic) fixes that I think would improve the PR.
> 
> One small change in the CHE_evolution_demo would be below Question 7, I would change the hard-coded path to use the variable set at the top of the notebook. Same with the BH spin demo (you can simply comment out the `path = ` lines in later cells).
> 
> In `calculate_black_hole_spin_pop_synth.ipynb`, you have `from spin_class import *`. As I understand, this is generally bad practice, but for this particular case it might not matter so much. But maybe since this is a tutorial, we should try to stick to cleaner coding. If possible, I would prefer to have this cleaned up, but I'm happy for others to disagree.
> 
> In the same file, you also have an anomalous line of html that slipped in somehow `<div class="alert ...">`. There are some commented lines here too in the plotting scripts, which should probably be removed. Lastly, I think the `plt.show()` lines should be removed. When they are present, each plot is reprinted every time a new `plt.show()` is called, so that by Example 5 I have to scroll through a bunch of plots I've already seen previously, and it's a bit confusing.
> 
> In both these tutorials, the download time for the datasets is quite large. I would suggest putting all the relevant links at the top of the files so students can be working through the files while the downloads are going (instead of getting halfway through and having to start a new download). I might also request that different test files be used, i.e variations that hold only the DCO systems you want, though that might be an annoying bit of extra work for you.
> 
> Also, you have a few files in your demo which differ from the existing one in the repo by a few parameter names (since the datafile is from an older version of Compas) and a bit of whitespace. This includes ClassCOMPAS.py, selection_effects.py, totalMassEvolvedPerZ. Would it be possible to rename the files e.g to have `-old.py` or something like that appended, to highlight this difference? Or would that cause other issues?
> 
> Similarly, there are differences between the FastCosmicIntegration files, but these seem to be more dramatic, i.e the code routines seem to have been updated. Is it worth updating the version in your tutorial, or does it not matter? I would also like to see this one renamed with a `-old` suffix or similar.
> 
> I quickly glanced at the spin_class.py file, and it looks really nice! Though you've unfortunately had to create 2 versions of the class, one to work with older Compas datasets. This highlights a concern going forward: do we need 2 versions of all post-processing routines to handle older data? Not something to fix here, but should be discussed at some point.

reinhold-willcox (approved):

reinhold-willcox:
Looks great, thanks Floor!

PR #892, opened 2023-01-30
added orbit inclination angle to SN output

reinhold-willcox:


jeffriley (changes requested):
Documentation?

jeffriley:
Oh, and changelog...

jeffriley:
And just a question - does it need to be in the default record?  Can you just make it available, and users that need it can add it via the logfile definitions file at run time?  Just so we don't continue to bloat the default output files.  No problem if you think it would be useful as a default - just asking the question.

reinhold-willcox:
It's a pretty useful output option, I would propose advocate including it by default. 

Docs and changelog updated.

jeffriley (approved):
All good

PR #900, opened 2023-02-01
Fixing "black hole mass loss" issue

### LINKED ISSUE #884: BHs are losing mass
**Describe the bug**
BHs appear to be losing mass following the Vink(?) prescription. This was discovered by Andreas Konstantinou. The m_Mdot parameter appears to have related, anomalous behavior. 

**Label the issue**

**To Reproduce**

```
COMPAS --random-seed 1 --number-of-systems 10000 --metallicity 0.0142 --kick-magnitude-distribution MULLERMANDEL --remnant-mass-prescription  MULLERMANDEL --muller-mandel-kick-multiplier-NS 520 --logfile-definitions COMPAS_Output_Definitions.txt
```

Together with the following `COMPAS_Output_Definitions.txt`
```
BSE_RLOF_Rec += { BINARY_PROPERTY::SYSTEMIC_SPEED, 
    STAR_1_PROPERTY::LUMINOSITY,
    STAR_2_PROPERTY::LUMINOSITY,
    STAR_1_PROPERTY::TEMPERATURE,
    STAR_2_PROPERTY::TEMPERATURE,
    STAR_1_PROPERTY::ANGULAR_MOMENTUM,
    STAR_2_PROPERTY::ANGULAR_MOMENTUM,
    STAR_1_PROPERTY::DYNAMICAL_TIMESCALE,
    STAR_2_PROPERTY::DYNAMICAL_TIMESCALE,
    STAR_1_PROPERTY::THERMAL_TIMESCALE,
    STAR_2_PROPERTY::THERMAL_TIMESCALE,
    STAR_1_PROPERTY::NUCLEAR_TIMESCALE,
    STAR_2_PROPERTY::NUCLEAR_TIMESCALE,
    STAR_1_PROPERTY::MDOT,
    STAR_2_PROPERTY::MDOT,
    STAR_1_PROPERTY::MASS_TRANSFER_DIFF,
    STAR_2_PROPERTY::MASS_TRANSFER_DIFF}
```

Then mask for accreting BHs. 

**Expected behavior**
BHs should not lose mass, and m_Mdot should accurately reflect the mass loss in a given timestep. 

**Screenshots**
image

 - COMPAS v02.34.03

ilyamandel:
This fixes #884 .

Black holes are not losing mass and their Mdot is correctly reported as zero.

Fixes:

We did incorrectly report Mdot for single stars, because of the line void ResolveMassLoss() { } // NO-OP in Remnants.h, which meant we never called ResolveMassLoss for these, so Mdot was not being reset correctly, and just held onto its last value.  

For binaries, meanwhile, the problem was with BaseBinaryStar::CalculateWindsMassLoss(), where we stopped winds for binaries in mass transfer — but never updated Mdot.

Along with fixing these issues, I fixed a few other things, like the incorrect application of Dutch winds to remnants, a typo in the calculation of NS luminosities, and a bit of light cleaning.

### DIFF
--- a/online-docs/pages/User guide/Program options/program-options-list-defaults.rst
+++ b/online-docs/pages/User guide/Program options/program-options-list-defaults.rst
@@ -1126,7 +1126,13 @@ Default = 1.0
 :ref:`Back to Top <options-props-top>`
 
 **--use-mass-loss** |br|
-Enable mass loss. |br|
+Enable mass loss through winds. |br|
+Note that setting this option to false can have unexpected consequences, e.g., TPAGB stars that are prevented from losing mass 
+cannot become white dwarfs, so will end up as massless remnants.  This is a useful option for testing, but this setting is not recommended
+for production. It is better to use specific wind prescription controls, such as --cool-wind-mass-loss-multiplier, 
+--luminous-blue-variable-prescription, --luminous-blue-variable-multiplier, 
+--mass-loss-prescription, --overall-wind-mass-loss-multiplier, --wolf-rayet-multiplier
+--cool-wind-mass-loss-multiplier. |br|
 Default = TRUE
 
 .. _options-props-V:
--- a/src/BH.h
+++ b/src/BH.h
@@ -54,6 +54,8 @@ protected:
 
     double  CalculateLuminosityOnPhase() const                                      { return CalculateLuminosityOnPhase_Static(); }
 
+    double          CalculateMassLossRate()                                         { return 0.0; } // Ensure that BHs don't lose mass in winds
+    
     double  CalculateMomentOfInertia(const double p_RemnantRadius = 0.0) const      { return (2.0 / 5.0) * m_Mass * m_Radius * m_Radius; }
     double  CalculateMomentOfInertiaAU(const double p_RemnantRadius = 0.0) const    { return CalculateMomentOfInertia(p_RemnantRadius * RSOL_TO_AU) * RSOL_TO_AU * RSOL_TO_AU; }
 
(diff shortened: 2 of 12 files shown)


reinhold-willcox:
@jeffriley This looks good to me, but the new function `SetMdot()` looks like it might not align with your preferred OOP style, so I'm deferring to you.

jeffriley:
You're right @reinhold-willcox, I don't like it (as evidenced by the comments @ilyamandel removed for the calls to SetMassLossDiff().  I meant to find a better way  than expose setters, but I haven't got to it yet (I think we should have something somewhere that reminds us to revisit that).

@ilyamandel Rather than doing the following in BaseBinaryStar::CalculateWindsMassLoss():

    m_Star1->SetMassLossDiff(0.0);
    m_Star2->SetMassLossDiff(0.0);
    m_Star1->SetMdot(0.0);
    m_Star2->SetMdot(0.0);

what about a new virtual function in BaseStar something like:

    void BaseStar::HaltWinds() {
        m_MassLossDiff = 0.0;
        m_Mdot = 0.0;
    }

That would still leave a couple of calls to SetMassLossDiff(), but maybe we can resolve those later.

(But if you must leave SetMdot() in, can you change it to 'void SetMdot(const double p_Mdot)'?

ilyamandel:
@jeffriley -- OK, happy to hide the setter this way. :)

Partial change made:  m_MassLossDiff is a variable of BinaryConstituentStar, while m_Mdot is a variable of BaseStar, so I kept the former to be set via  SetMassLossDiff() but replaced SetMdot() with HaltWinds() for the latter.  I also 
added a comment “// JR would prefer to avoid a Setter for aesthetic reasons” :wink:

Finally, replaced LOG10_ZSOL with pre-computed value to avoid compilation issues.  

Tested, ready for your repeat review.

reinhold-willcox (approved):
Looks good, thanks Ilya

PR #908, opened 2023-02-02
Removed some setters in BinaryConstituentStar

jeffriley:
Removed some setters in BinaryConstituentStar.  Will address others later.

### DIFF
--- a/src/BaseBinaryStar.cpp
+++ b/src/BaseBinaryStar.cpp
@@ -1841,7 +1841,7 @@ void BaseBinaryStar::CalculateWindsMassLoss() {
         if (OPTIONS->UseMassLoss()) {                                                                                           // mass loss enabled?
 
             double mWinds1 = m_Star1->InitialiseWindsMassLoss();                                                                // initialise winds mass loss for star1
-            double mWinds2 = m_Star1->InitialiseWindsMassLoss();                                                                // initialise winds mass loss for star2
+            double mWinds2 = m_Star2->InitialiseWindsMassLoss();                                                                // initialise winds mass loss for star2
 
             double aWinds  = m_SemiMajorAxisPrev / (2.0 - ((m_Star1->MassPrev() + m_Star2->MassPrev()) / (mWinds1 + mWinds2))); // new semi-major axis for circularlised orbit
 
--- a/src/BaseBinaryStar.cpp
+++ b/src/BaseBinaryStar.cpp
@@ -1832,25 +1832,20 @@ void BaseBinaryStar::CalculateWindsMassLoss() {
 
     m_aMassLossDiff = 0.0;                                                                                                      // initially - no change to orbit (semi-major axis) due to winds mass loss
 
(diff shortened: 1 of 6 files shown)


jeffriley:
@ilyamandel I tweaked some of your code, and removed a couple of setters.

I am not able to produce the detailed output plot for comparison - can you check that you can?  If you can then I'll need to look at my configuration, but if you can't then we might need to look at the code.

ilyamandel (changes requested):
Do we really need to split off InitialiseWindsMassLoss from CalculateMassLossValues?  It seems like we are adding an extra layer of functions which will just be more confusing to future coders -- can we avoid this?

BaseStar.cpp, line 1939:
//JR: todo: pick a better name for this...

Maybe now is a good time?  How about CalculateMassAfterWindMassLoss() [to make it clear this is all about winds?]

1942: 
should be update*d*

1932: calculated *stellar* mass *after mass loss* 

Note that because of another pull, changelog.h is now outdates, so dev needs to be merged in.

ilyamandel:
P.S. 

> I am not able to produce the detailed output plot for comparison - can you check that you can? If you can then I'll need to look at my configuration, but if you can't then we might need to look at the code.

The following works for me on v. 2.34.06 and produces the same detailed output as previously:
make clean
make -j 4
$COMPAS_ROOT_DIR/src/COMPAS -v
cd $COMPAS_ROOT_DIR/utils/examples/methods_paper_plots/detailed_evolution/
rm -r COMPAS_Output
python3 runSubmitDemo.py
python3 plot_detailed_evolution.py

jeffriley:
> Do we really need to split off InitialiseWindsMassLoss from CalculateMassLossValues? It seems like we are adding an extra
> layer of functions which will just be more confusing to future coders -- can we avoid this?

Unless I'm misremembering, they do different things - we don't always want to do the total of both wherever they are called individually, do we?

ilyamandel:
InitialiseWindMassLoss does two things: it calls CalculateMassLossValues and sets m_MassLossDiff.  Is the goal to hide a setter?  I think it just makes for harder to read, messier code.  Anyway, I won't argue with you on that one, but see my other requests for changes (all minor). :)   Also, dev needs to be merged in (e.g., changelog.h is out of date).

jeffriley:
@ilyamandel I'll get to this in a bit, but the quick answer to this:

Is the goal to hide a setter?

It is, yes (well, to remove, rather than hide, a setter) - I removed SetMassLossDiff() which just set a class member variable.  Is the new code messier?  Well, I suppose it's more indirect than using a setter, so could be considered messier, but that's not a good case for a setter in OO code...  I'll have another look and see if I can make it less messy.

ilyamandel:
Just a gentle reminder on this one too, @jeffriley ; I suspect that with the major changes to the code between now and then, it might be easier to start whatever you wanted to accomplish from scratch than to resuscitate these old PRs, but obviously it's your call. :)

PR #910, opened 2023-02-04
Revert "Critical mass ratios ~and Zeta's~ from Ge et al. 2020 "

ilyamandel:
Reverts TeamCOMPAS/COMPAS#849

@reinhold-willcox , I am getting a lot of warnings like this after trying to compile your merged PR #849 .   I wonder why this wasn't picked up by CI (@avivajpeyi ?).  In any case, please fix them or revert #849 .

In file included from ./WhiteDwarfs.h:9:
./Remnants.h:38:21: warning: 'Remnants::CalculateCriticalMassRatio' hides overloaded virtual functions [-Woverloaded-virtual]
    double          CalculateCriticalMassRatio(const bool p_AccretorIsDegenerate) const                         { return 0.0; ...
                    ^
./BaseStar.h:158:29: note: hidden overloaded virtual function 'BaseStar::CalculateCriticalMassRatio' declared here: different qualifiers
      (none vs const)
    virtual double          CalculateCriticalMassRatio(const bool p_AccretorIsDegenerate); 
                            ^
./BaseStar.h:158:29: note: hidden overloaded virtual function 'BaseStar::CalculateCriticalMassRatio' declared here: different qualifiers


jeffriley (approved):

ilyamandel:
Also, this takes us back several versions (I should have noticed that github shows no conflicts relative to where this was branched off, not the current head of dev).

themikelau:
@ilyamandel @avivajpeyi  CI would not fail on warnings, only errors. It looks like COMPAS successfully compiled and I wonder whether those warnings cause any actual problems—they look like they might be specific to your compiler settings. In the output of the CI, I do not see such warnings (https://github.com/TeamCOMPAS/COMPAS/actions/runs/4090934683/jobs/7054689459).

jeffriley:
> I wonder whether those warnings cause any actual problems

They would if they were real.  The warning indicates the BaseStar virtual function "CalculateCriticalMassRatio()" is being hidden by the "CalculateCriticalMassRatio()" declared in Remnants.h because they have a different signature (one of the missing a "const" qualifier - or at least there is a difference in the "const" qualifiers in the declarations).  If that was real then the function declared in Remnants.h is not overloading the base class function (because they have different signatures).

However, as far as I can tell after inspecting the two .h files, there doesn't seem to be a difference in the signatures, so I think @themikelau is right - the problem seems to be on @ilyamandel's system (it would seem @ilyamandel you have out-of-date/out-of-sync source files).

Edit: did you do a 'make clean'?  I would have thought this is a compile error rather than a link error, but it's worth a try I suppose.

ilyamandel:
I did make clean, but it could indeed be a temporary issue.  I again apologise for messing this up.  Hope @reinhold-willcox can make another PR from the same branch -- sorry!

PR #912, opened 2023-02-06
removed anomalous const in Remnants.h

reinhold-willcox:


ilyamandel:
Are you sure this will be enough?   There seems to be a bit of a range of when const is / is not used...

$ grep CalculateCriticalMassRatio *.h
BaseStar.h:    virtual double          CalculateCriticalMassRatio(const bool p_AccretorIsDegenerate); 
BaseStar.h:    virtual double          CalculateCriticalMassRatioClaeys14(const bool p_AccretorIsDegenerate) const         { return 0.0; }                                                     // Default is 0.0
BaseStar.h:            double          CalculateCriticalMassRatioGe20(const QCRIT_PRESCRIPTION p_qCritPrescription)        { return InterpolateGe20QCrit(p_qCritPrescription); }
FGB.h:    double          CalculateCriticalMassRatioClaeys14(const bool p_AccretorIsDegenerate) const             { return GiantBranch::CalculateCriticalMassRatioClaeys14(p_AccretorIsDegenerate); }                     // Skip HG 
GiantBranch.h:            double          CalculateCriticalMassRatioClaeys14(const bool p_AccretorIsDegenerate) const; 
HG.h:    double          CalculateCriticalMassRatioClaeys14(const bool p_AccretorIsDegenerate) const;
HeGB.h:    double      CalculateCriticalMassRatioClaeys14(const bool p_AccretorIsDegenerate) const ;
HeHG.h:            double          CalculateCriticalMassRatioClaeys14(const bool p_AccretorIsDegenerate) const;
HeMS.h:            double          CalculateCriticalMassRatioClaeys14(const bool p_AccretorIsDegenerate) const;
MS_gt_07.h:    double      CalculateCriticalMassRatioClaeys14(const bool p_AccretorIsDegenerate) const ;
MS_lte_07.h:    double      CalculateCriticalMassRatioClaeys14(const bool p_AccretorIsDegenerate) const ;
Remnants.h:    double          CalculateCriticalMassRatio(const bool p_AccretorIsDegenerate) const                         { return 0.0; }                                                         // Should never be called...
Star.h:    double              CalculateCriticalMassRatio(const bool p_AccretorIsDegenerate) const                         { return m_Star->CalculateCriticalMassRatio(p_AccretorIsDegenerate); }
Star.h:    double              CalculateCriticalMassRatioClaeys14(const bool p_AccretorIsDegenerate) const                 { return m_Star->CalculateCriticalMassRatioClaeys14(p_AccretorIsDegenerate); }

jeffriley:
It should be consistent throughout...

I compiled COMPAS a few minutes ago with '-Woverloaded-virtual' and got lots (and lots) of warnings (which may be why -Woverloaded-virtual isn't enabled with -Wall).  I haven't trawled through them yet, and they may not all be problems, but we should probably check one day...

jeffriley:
@reinhold-willcox I think what happened here is that all declarations of CalculateCriticalMassRatio() ere const - but then Nicolas? changed the function in BaseStar so that it was no longer const - so all declarations in all classes now need to have 'const' removed.

jeffriley:
@reinhold-willcox I think the need for this PR was obviated by PR https://github.com/TeamCOMPAS/COMPAS/pull/974

I think we can close this PR without merging.

jeffriley (commented):
I think we should close this without merging - see previous comment

ilyamandel:
@jeffriley -- I still get loads of warnings when I compile COMPAS, so at some point we should fix them, but no rush... [And no objection to closing this, of course]

jeffriley:
Yes, I do too when I compile on my Mac - but not on my Linux system. I will
try to fix that soon.

PR #914, opened 2023-02-10
Super eddington CE

### LINKED ISSUE #859: Super-Eddington accretion during a common-envelope event
It would be a useful enhancement to have the possibility of super-Eddington accretion onto a neutron star or black hole during a common-envelope event. This is in a similar vein to what was done and considered in [van Son et al. (2020)](https://ui.adsabs.harvard.edu/abs/2020ApJ...897..100V/abstract), but that version never made it to the main COMPAS repo. One could think of having a multiplicative factor to the Eddington limit, as well as alternative criteria.

See, e.g., a brief discussion here: https://compashq.slack.com/archives/CMCEQ0RS8/p1666102881607499

LiekeVanSon:
Adding an extra option `CHEVALIER' to Common Envelope accretion, following model 2 from [van Son et al. 2020. ](https://iopscience.iop.org/article/10.3847/1538-4357/ab9809/pdf)

This also allows BHs to accrete during a common envelope event if CommonEnvelopeMassAccretionPrescription is set to anythning but ZERO. I.e., CommonEnvelopeMassAccretionPrescription now sets the accretion of either a neutron star or a black hole in a common envelope (previously, only accretion onto NSs was implemented). 

This resolves #859


jeffriley (changes requested):
Thanks Lieke.  A few things:

(a) changelog.h needs to be updated.

(b) BinaryConstituentStar::ResolveCommonEnvelopeAccretion() does not need the stellar type parameter you added - it should be removed.  When BinaryConstituentStar::ResolveCommonEnvelopeAccretion() is called, it is called on an object that already knows its stellar type.

(c) Do we need the switch() in BinaryConstituentStar::ResolveCommonEnvelopeAccretion()?  Neutron stars are treated differently in  the existing version, and you've added black holes, but they (NS and BH) are treated identically (I think), so why not just change the existing code from:

if (IsOneOf({ STELLAR_TYPE::NEUTRON_STAR})) {

to

if (IsOneOf({ STELLAR_TYPE::NEUTRON_STAR, STELLAR_TYPE::BLACK_HOLE})) {

Are you expecting we might treat them differently in the future?

Edit: per my comment in BinaryConstituentStar::ResolveCommonEnvelopeAccretion(), we probably should use the class hierarchy here to calculate the mass accreted.  While it was only NS that was treated differently I was a little lax, but now we have two stellar types treated differently (albeit the same as each other), I think we should consider making CalculateMassAccretedForCO() (in your code) a class function for each of the stellar types (named something like CalculateMassAccretedDuringCE()).  It's not an easy change (because it kind-of mixes stellar and binary attributes - what does an individual star know about common envelopes...?), so maybe if black holes are the last addition here we could relent...  I'm in two minds - this code (the original, not your changes) is not really in keeping with the OO philosophy, but to make it so is not a simple change.  If we decide not to, a comment along these lines should be maintained in the code so if a developer decides to add another stellar type, the comment should trigger a rewrite to more OO code...

ilyamandel (changes requested):
A few more things on top of what @jeffriley mentioned:

On line 151 of BinaryConstituentStar.cpp, there's a spurious print statement (left behind from testing?)

Should double CalculateMassAccretedForCO(const double p_Mass, const double p_CompanionMass, const double p_CompanionRadius)
have a static classifier since it does not change any internal variables?

I am a bit concerned about the apparent lack of a check for mass conservation in the new prescription: it seems that the BH can grow in mass by more than the mass loss from the companion.  Would it make sense to clamp the mass accretion to be no more than the mass lost (the pre-CE envelope mass of the companion)?

On the special treatment for NSs and BHs -- the issue is that we are assuming that only NSs and BHs can accrete during a CE, not other stars.  I think that's quite weird, because it's harder for NSs and BHs to accrete than, say, for MS stars (far more feedback that can suppress accretion).  E.g., if BHs can accrete super-Eddington in the new mode, why not MS stars?  I think we should just be more consistent in our treatment of CE accretion across stellar types, as we are for normal accretion (with the usual class inheritance for CE accretion options, just as for normal MT options).

LiekeVanSon:
HI Ilya and Jeff, thanks for reviewing :

(a) updated

(b) thanks, removed p_StellarType from the function again

> On line 151 of BinaryConstituentStar.cpp, there's a spurious print statement (left behind from testing?)

Yes sorry, removed.  this was left from running a couple of test binaries to see if the CE accretion behaved

> Should double CalculateMassAccretedForCO(const double p_Mass, const double p_CompanionMass, const double p_CompanionRadius) have a static classifier since it does not change any internal variables?

I don't know enough of the OO programming structure to know the pro's and con's of this?
To do this, would I just change the definition in BinaryConstituentStar.h to the following?
` static double              CalculateMassAccretedForCO(const double p_Mass, ...)`


> I am a bit concerned about the apparent lack of a check for mass conservation in the new prescription: it seems that the BH can grow in mass by more than the mass loss from the companion. Would it make sense to clamp the mass accretion to be no more than the mass lost (the pre-CE envelope mass of the companion)?

Good point! I added a check right before returning 'deltaMass' to make sure that it never exceeds the preCE envelope mass. This should now be a safety net for all cases


(c) 
> On the special treatment for NSs and BHs -- the issue is that we are assuming that only NSs and BHs can accrete during a CE, not other stars. I think that's quite weird, because it's harder for NSs and BHs to accrete than, say, for MS stars (far more feedback that can suppress accretion). E.g., if BHs can accrete super-Eddington in the new mode, why not MS stars? I think we should just be more consistent in our treatment of CE accretion across stellar types, as we are for normal accretion (with the usual class inheritance for CE accretion options, just as for normal MT options).

Yes I initially intended to add two separate options for the CommonEnvelopeMassAccretionPrescription, one for NS and one for BH behaviour, because this was how we had it in the old COMPAS code, but then I had the same doubts as Ilya: physically it doesn't really make sense to have super Eddington accretion for one, but not the other?
 I agree that extending CE accretion to more stellar types is more consistent (at least indeed MS, He stars, CHE stars and white dwarfs) but I am a little bit reluctant to a general extension to all stellar types. 
In particular I'm a bit afraid for unexpected behavior.  For example: what would happen for double core common envelopes? Will both cores accrete even though they might be 2 HG stars? In this case most of the  CalculateMassAccretedForCO() function as it is probably doesn't make sense. 

1) Are there other exceptions that you can think of that could cause problems if we extend the CE accretion to all stellar types? 
2) If we do extend it to everything, what do you mean exactly with "with the usual class inheritance for CE accretion options, just as for normal MT options" Do you mean we should make CalculateMassAccretedForCO() a class function for each of the stellar types as Jeff suggests? (this question is just my lacking programming knowledge again :)

ilyamandel (approved):
Sorry, 
Should double CalculateMassAccretedForCO(const double p_Mass, const double p_CompanionMass, const double p_CompanionRadius) have a static classifier since it does not change any internal variables?
that was a typo -- I meant to say that it probably should have a const classifier, but it's not a big deal either way.

Happy with the resolution of other issues I noted, leaving it to @jeffriley to check the ones he noted.

jeffriley (approved):
I made a few minor formatting changes, and I added the const qualifier to BinaryConstituentStar::CalculateMassAccretedForCO().

I'm not sure why you left the switch in BinaryConstituentStar::ResolveCommonEnvelopeAccretion() instead of using

    if (IsOneOf({ STELLAR_TYPE::NEUTRON_STAR, STELLAR_TYPE::BLACK_HOLE})) { ... }

but that's up to you.

Otherwise, good to go - except you're now a few commits behind dev, so you're going to have to merge dev - I'll leave that up to you.

LiekeVanSon:
Thanks!

I had left the switch in case we wanted to change to something more consistent but complex for other stellar types. But as we're sticking with just the NS and BHs for now I've changed it back to an if statement .

jeffriley:
Thanks Lieke!  Yeah, I think if we include more than just NS and BH we might have to think about having separate functions for each of the stellar types, rather than a SWITCH or IF.

PR #915, opened 2023-02-11
Minor fixes

ilyamandel:
Minor fixes to documentation and python comments, not touching C++ files.

### DIFF
--- a/online-docs/pages/User guide/COMPAS output/standard-logfiles-record-specification-binary.rst
+++ b/online-docs/pages/User guide/COMPAS output/standard-logfiles-record-specification-binary.rst
@@ -1660,7 +1660,7 @@ both. If both are printed then the file will contain two columns with the same h
    * - Description:
      - Post-supernova systemic (centre-of-mass) velocity (\ :math:`km s^{-1}`).
    * - Header String:
-     - Systemic_Velocity
+     - SystemicSpeed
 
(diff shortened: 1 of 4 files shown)


jeffriley (changes requested):
Looks good save for the following:

In online-docs/pages/User guide/Program options/program-options-list-defaults.rst:

Remove the apostrophe in "zeta's" on line 287:
    Warning: if running with ``--critical-mass-ratio-prescription``, zeta's will not be computed, 

Fix terminating quotes around "MULLERMANDEL" on lines 499 and 1055:
    however, if using ``MULLERMANDEL'', it is recommended to keep them consistent by doing so for both. |br|
    however, if using ``MULLERMANDEL'', it is recommended to keep them consistent by doing so for both. |br|


Edit: @ilyamandel if you're ok with the suggested changes I can make them...

ilyamandel:
Thanks for spotting these, @jeffriley !  Fixed along with a couple of other quotation marks while I was at it. :)

jeffriley (approved):
Thanks @ilyamandel .  All good from me - I'll leave the final decision to @reinhold-willcox and @IsobelMarguarethe

ilyamandel:
I'll just pull it in -- we have too many PRs hanging around, and this is documentation only, so trivial to add more things if others find more issues. :)

PR #916, opened 2023-02-12
Kick output vectors

### LINKED ISSUE #821: Binary properties missing from documentation
The following binary properties are exposed as properties able to be added to binary log files, but are not described in the online documentation.  They should be added to the documentation:

RLOF_POST_STEP_STAR_TO_ROCHE_LOBE_RADIUS_RATIO_1
RLOF_POST_STEP_STAR_TO_ROCHE_LOBE_RADIUS_RATIO_2
RLOF_PRE_STEP_STAR_TO_ROCHE_LOBE_RADIUS_RATIO_1
RLOF_PRE_STEP_STAR_TO_ROCHE_LOBE_RADIUS_RATIO_2 

My best guesses at the respective descriptions are:

Ratio of the primary star's radius to the radius of its Roche lobe immediately following the time step
Ratio of the secondary star's radius to the radius of its Roche lobe immediately following the time step
Ratio of the primary star's radius to the radius of its Roche lobe immediately prior to the time step
Ratio of the secondary star's radius to the radius of its Roche lobe immediately prior to the time step

but I think these were added by @reinhold-willcox when he was updating the BSE_RLOF-related code (?), so he might change/elaborate as necessary.

 - OS: [e.g. Ubuntu 20.04]
 - COMPAS [e.g. v02.31.03]

reinhold-willcox:
Added post-SN orbital AM vector (in pre-SN frame of reference) to the possible BINARY_PROPERTIES that can be included in the output. The parameters are 

`SUPERNOVA_ORBIT_INCLINATION_VECTOR_X`, 
`SUPERNOVA_ORBIT_INCLINATION_VECTOR_Y`, and 
`SUPERNOVA_ORBIT_INCLINATION_VECTOR_Z`. 

They are only useful in some rare use cases, so they are not included in the default `BSE_SUPERNOVAE` logfile, but can be added with a custom `logfile-definitions` file.

Separately, added documentation for the following properties, addressing issue #821: 
```
RLOF_POST_STEP_STAR_TO_ROCHE_LOBE_RADIUS_RATIO_1
RLOF_POST_STEP_STAR_TO_ROCHE_LOBE_RADIUS_RATIO_2
RLOF_PRE_STEP_STAR_TO_ROCHE_LOBE_RADIUS_RATIO_1
RLOF_PRE_STEP_STAR_TO_ROCHE_LOBE_RADIUS_RATIO_2
```


ilyamandel (changes requested):
Thanks for the changes, @reinhold-willcox .  A few queries/thoughts:

- In the documentation, it would be great to have more information about the angular momentum vector.  E.g., how is the frame defined (is \hat{z} along the pre-SN orbital along momentum vector?  what about \hat{x}?)?  What are the units (I see in the code that you've normalised the vector, but I wouldn't know that without looking at the code... and if it is normalized, why not just output two angles, theta and phi?)?

- Should the label be something like Orbit_Inclination_Vector_Z>SN rather than SN_Orbit_Inclination_Vector_Z?

- Vector3d m_OrbitalAngularMomentumVector; // Systemic velocity vector, relative to ZAMS Center of Mass
I think this comment in BaseBinaryStar.h is incorrect.

reinhold-willcox:
@ilyamandel I've now addressed your comments. Regarding the unit vector components vs two new angles, I chose the former simply because it's more directly useful for my project with Ryo, and the code already has the vector component information readily stored (so I don't have to define two new angles and then write up documentation about them), though I agree that the latter is slightly more storage efficient. If we decide these parameters are likely to be useful in a broader context, it may be worth discussing which is the better approach.

ilyamandel (approved):
All good.

PR #922, opened 2023-02-15
Issue #797 - auto yaml file creation

### LINKED ISSUE #797: Auto-generated .yaml

jeffriley:
- Addressed issue https://github.com/TeamCOMPAS/COMPAS/issues/797 - implemented functionality to create YAML file. Two new options (--create-YAML-file and --YAML-template). See documentation for details.
- Modifed runSubmit.py to work with new yaml file format (i.e. all options could be commented...)
- Minor defect repairs in options code
- Minor fixes to online documentation; also clarified make arguments

Some highlights:

Added two new options: --create-YAML-file and --YAML-template
Added two new source files: yaml.cpp and yaml.h

--create-YAML-file, as you might expect, creates a YAML file.
The YAML file created is formatted according to a template.

Users can specify a YAML template using the --YAML-template option. If no template is specified, or the file does not exists or is unreadable, the COMPAS default template will be used - this is declared in yaml.h. Allowing users to specify a template file gives them the flexibility to order and group the options the way they want (most people will accept the default, but some may want to have their own groupings).

The YAML file is formatted as you requested above - mostly :-) I added a "default" field.

The option values written to the YAML file are COMPAS default values, and the line in the YAML file will be commented, UNLESS options were specified on the command line when the YAML file is created, in which case the values will be as specified on the command line, and the line in the YAML file will not be commented. This lets users create project-specific YAML file if the wish (again, most people won't, but some might).

e.g.

    ./compas --create-yaml-file

will create a YAML file using the default template in yaml.h, with all values COMPAS defaults, and all (option definition) lines in the YAML file commented (choices headers are not commented).

    ./compas --create-yaml-file --logfile-type csv

will create a YAML file using the default template in yaml.h, with all values COMPAS defaults, EXCEPT for the value for the --logfile-type option, and all (option definition) lines in the YAML file commented EXCEPT for the definition line for --logfile-type.

I had to modify runSubmit.py to work with a YAML file that had all options in a choice group (i.e. stringChoices, booleanChoices, etc.) commented. I tested the new runSubmit.py against the old YAML file format and it seemed to work ok.  (I also updated runSubmitDemo.py)

I tried various ways to break the functionality - I'm sure there is a way. It parses a template file, and that template file could have just about anything in it... I've tried to make the code robust - if the template file is poorly formatted and has unexpected characters in it, then you mightn't see the results you expect - but I'm oping it won't actually fail and stop COMPAS. @reinhold-willcox, feel free to beat it up as much as you can.

I've added documentation - I added a "PreProcessing" section - and moved "Sampling" under that. The documentation is a bit sparse - feel free to add to it...

Also, in testing I tried to follow the instructions on the "Running COMPAS via Python" page of the online documentation and it didn't work for me (and it seems to be very Mac-centric). Can you check the instructions there for correctness?


reinhold-willcox:
This might take me some time to review fully, but at first glance this looks great! I love the idea that the yaml can be generated simultaneously with other inputs, so that users can get a project-custom yaml for free. 

I notice a lot of whitespace changes (or capitalization fixes / float precision truncation), which are really just fixes to formatting inconsistencies - not a huge concern, but it's always nice to make the code look better. There are also a lot of places where the git diff highlights lines or parts of lines that are the same before and after. I assume this is because of the previous PR which included extra files by accident, and github doesn't really know how to recognize this and ignore the diffs. Again, not a concern, it just makes the review look bigger than it is. 

I'll go through and add my review on a change by change basis, so that I can come back and add more later as need be.

src/utils.cpp
```diff
@@ -1,6 +1,5 @@
 #include <iostream>
 #include <stdarg.h>
-#include <fstream>
```
reinhold-willcox, line 3: why is this removed?

src/utils.h
```diff
@@ -1,11 +1,12 @@
 #ifndef __utils_h__
 #define __utils_h__
 
+#include <algorithm> 
+#include <vector>
```
reinhold-willcox: why are these added?

utils/examples/methods_paper_plots/detailed_evolution/runSubmitDemo.py
```diff
@@ -39,12 +39,18 @@ def __init__(self, config_file='compasConfigDefault.yaml', grid_filename=None, r
...
+        if self.stringChoices    is None: self.stringChoices    = {}
+        if self.listChoices      is None: self.listChoices      = {}
+
         compas_executable_override = os.environ.get('COMPAS_EXECUTABLE_PATH')
         print('compas_executable_override', compas_executable_override)        
 
         if (compas_executable_override is None):
             # Make sure you have set and exported the COMPAS_ROOT_DIR environment variable
             git_directory = os.environ.get('COMPAS_ROOT_DIR')
+            print('<', git_directory, '>')
```
reinhold-willcox: Is this supposed to be here, or was this accidentally included while testing?

reinhold-willcox (changes requested):
Overall looks good, and a useful functionality. 

Some requests:
- There are some conflicts that need to be addressed
- For boolean and numerical options, when the new yaml file is created, the updated values are correct, but the listed defaults also show the updated value. This does not occur for string options.
- When `--create-yaml-file` is included, COMPAS does not run - no binaries are produced (thought the yaml file is successfully created). This is true if `--yaml-template` is present or not.
- From the documentation, it looks like `--yaml-template` requires `--create-yaml-file`, but COMPAS runs fine with the former without the latter. A warning would be useful here. 
- If the new yaml file has the same name as an existing file, you are prompted to overwrite it. I like this, I think it's a great idea. But the end of the prompt is "overwrite (Y/N)". With apologies for the extreme nit-pickiness, could you make the non-default letter lowercase?

Otherwise looks great, thanks Jeff!

Edit: the inline replies below were from when I previously started this review - I forgot they were there, so some might be a bit out of date.

jeffriley:
Thanks @reinhold-willcox - I can fix these (I'll try to get it done tomorrow...).

Just one question:

> When --create-yaml-file is included, COMPAS does not run - no binaries are produced (thought the yaml file is successfully
> created). This is true if --yaml-template is present or not.

That was deliberate. --help, --version, also don't run -  they just do their job (display help, display version) and stop.  I did the same for --create-yaml-file.  I think that's reasonable - is there a good reason for having it behave differently?  It's trivial to do, so it's really just a question of how we want that option to behave.

Note, --yaml-template is always optional - if it is not supplied, the default template in yaml.h is used.

jeffriley:
@reinhold-willcox 

> There are some conflicts that need to be addressed

I think I fixed them.

> For boolean and numerical options, when the new yaml file is created, the updated values are correct, but the listed defaults
> also show the updated value. This does not occur for string options.

Now fixed (I believe...)

> When --create-yaml-file is included, COMPAS does not run - no binaries are produced (thought the yaml file is successfully
> created). This is true if --yaml-template is present or not.

Discussed above - I'm ok to leave it as is, but am open to changing it (as I said, the change is trivial).

> From the documentation, it looks like --yaml-template requires --create-yaml-file, but COMPAS runs fine with the former
> without the latter. A warning would be useful here.

**--create-yaml-file** is stand-alone - it will run just fine without **--yaml-template**. As noted above, **--yaml-template** is always optional - if it is not supplied, the default template in yaml.h is used.  We don't issue warnings for specified options that are not relevant to the run.  For example, if the user specifies:

    ./compas -n 1 --mode sse --initial-mass-1 19.2 --initial-mass 23.8

we just ignore the extraneous **--initial-mass-1 19.2**.  The documentation should not imply that **--create-yaml-file** is required if **--yaml-template** is specified, just that one applies to the other.

> If the new yaml file has the same name as an existing file, you are prompted to overwrite it. I like this, I think it's a great idea.
> But the end of the prompt is "overwrite (Y/N)". With apologies for the extreme nit-pickiness, could you make the non-default
> letter lowercase?

There is no default - I just keep asking until I get a "Y" (or "y") or "N" (or "n").

I could change "Y/N" to "y/n" - hopefully indicating that there is no default, or I could add a default.  If we add a default I would suggest "N" (or "n") be the default.


Ah... I forgot the inline replies.  I don't recall changing runSubmitDemp.py, but if I did we can undo the change - it doesn't look like it's absolutely necessary.  I'll check the others - I may have added some includes when I was testing (I originally put the yaml code in utils.cpp, then decided it should have its own file) - they may be able to be removed.  The removal of "#include \<fstream\>" is probably because I noticed we don't actually need it.

jeffriley:
Ok, I removed

    #include <algorithm> 
    #include <vector>

from utils.h - 'algorithm' is not required (may have been at some stage, but isn't now), and 'vector' is included in constants.h, so we don't need to include it again.

'fstream' isn't required, so removing that is ok.

src/constants.h
```diff
@@ -209,10 +210,10 @@ constexpr double GAUSS_TO_TESLA                         = 1.0 / TESLA_TO_GAUSS;
 // constants
 
 constexpr double _2_PI                                  = M_PI * 2;                                                 // 2PI
-constexpr double SQRT_M_2_PI                            = 0.7978845608028653558798921198687637369517;               // sqrt(2/PI)
+constexpr double SQRT_M_2_PI                            = 0.79788456080286536;                                      // sqrt(2/PI)
 constexpr double DEGREE                                 = M_PI / 180.0;                                             // 1 degree in radians
 
-constexpr double GAMMA_E                                = 0.57721566490153286060651209008240243104215933593992;     // Euler's Constant (probably don't need so many digits after the decimal point...)
+constexpr double GAMMA_E                                = 0.57721566490153286;                                      // Euler's Constant
```
reinhold-willcox, line 216: Mostly unrelated to this PR, but why is this Euler's constant? It's not the usual Euler constant, e...

utils/examples/methods_paper_plots/detailed_evolution/runSubmitDemo.py
reinhold-willcox: Bump on this, since I don't think it has been addressed yet.

reinhold-willcox:
Hey @jeffriley , this all looks good. I'm happy to approve once you remove the print() in runSubmit, and respond to the Euler constant question.

jeffriley:
Thanks @reinhold-willcox

I left a print in?  Oops - I'll fix that.

> Mostly unrelated to this PR, but why is this Euler's constant? It's not the usual Euler constant

Euler's constant, not Euler's number...  https://en.wikipedia.org/wiki/Euler%27s_constant

Actually, looking at the code it is never used - I wonder why it is there?  All I did was reduce the number of decimal digits - beyond 17 or 18 they make no difference at all...

jeffriley:
Oh, the print in runSubmitDemo...  I really don't recall doing that, but I guess I did.  Now removed.

SimonStevenson:
Re Euler's constant, I think it is probably there because it was used in a _very_ early version of COMPAS (before COMPAS was called COMPAS) in this paper https://arxiv.org/abs/1703.06873 , following the equations in this paper (https://arxiv.org/abs/1302.4442), where we were doing post-Newtonian evolution of black hole spins. That functionality was largely removed from COMPAS, but I guess the number stuck around. It could be removed.

reinhold-willcox:
Lol, that makes sense. Looks good, thanks Jeff & Simon

reinhold-willcox (approved):

PR #925, opened 2023-02-19
Change Roche Lobe radius output in BSE_DETAILED_OUTPUT_REC for GWLandscape

jeffriley:
Minor change and defect repair:

- Changed units of ROCHE_LOBE_RADIUS_1 and ROCHE_LOBE_RADIUS_2 from orbital separation to RSOL
- Changed header string for ROCHE_LOBE_RADIUS_1 from "RocheLobe(1)|a" to "RocheLobe(1)" - ditto for ROCHE_LOBE_RADIUS_2
- removed STAR_TO_ROCHE_LOBE_RADIUS_RATIO_1 ("Radius(1)|RL")and STAR_TO_ROCHE_LOBE_RADIUS_RATIO_2 ("Radius(2)|RL") from the default output for BSE_DETAILED_OUTPUT_REC (can be calulated from other values in the default output)
- changed plot_detailed_evolution.py to accomodate the removal of STAR_TO_ROCHE_LOBE_RADIUS_RATIO_1 and STAR_TO_ROCHE_LOBE_RADIUS_RATIO_2 from the default output
- changed online documentation to reflect:
(a) removal of STAR_TO_ROCHE_LOBE_RADIUS_RATIO_1 and STAR_TO_ROCHE_LOBE_RADIUS_RATIO_2 from the default output
(b) change of header strings for ROCHE_LOBE_RADIUS_1 and ROCHE_LOBE_RADIUS_2 (units already (accidentally...) correct)
- fixed minor defect in call to m_Accretor->CalculateMassAcceptanceRate() in BaseBinaryStar::CalculateMassTransfer() (only affected runs with mass-transfer-thermal-limit-accretor = RADIUS_TO_ROCHELOBE)


jeffriley:
Detailed output plot looks same before/after these changes (at least as far as I could see...)

@reinhold-willcox I updated plot_detailed_evolution.py for these changes - is the that the only post-processing script that is affected?

Also @reinhold-willcox I noticed the lines (handles) in the legend for the right-most plots is missing in the detailed output plot - I don't think that's something I did...

ilyamandel (changes requested):
Looks great, @jeffriley , except one thing, in the plotter:

starToRLradius1 = Data['Radius(1)'][()][mask] / (Data['RocheLobe(1)'][()][mask] * (1.0 - Data['Eccentricity'][()][mask]))

Why do you need to multiply by (1-e) again, since the Roche Lobe radius already includes this term?  I think this may be a mistake.

jeffriley:
Are you sure @ilyamandel?

'Radius(1)|RL' is calculated in the c++ code in BinaryConstituentStar::StarToRocheLobeRadiusRatio() as:

	double  BinaryConstituentStar::StarToRocheLobeRadiusRatio(const double p_SemiMajorAxis, const double p_Eccentricity) {
            if ((utils::Compare(p_SemiMajorAxis, 0.0) <= 0) || (utils::Compare(p_Eccentricity, 1.0) > 0))
                return 0.0;         // binary is unbound, so not in RLOF
    
            double rocheLobeRadius = BaseBinaryStar::CalculateRocheLobeRadius_Static(Mass(), m_Companion->Mass());
            return (Radius() * RSOL_TO_AU) / (rocheLobeRadius * p_SemiMajorAxis * (1.0 - p_Eccentricity));
	}

'RocheLobe(1)' is calculated in the c++ code in BaseBinaryStar.h as:
	
	double RocheLobeRadius1() const { return CalculateRocheLobeRadius_Static(m_Star1->Mass(), m_Star2->Mass()) * SemiMajorAxisRsol(); }

(was 'RocheLobe(1)|a' but is now in RSOL - 'RocheLobe(1)|a' is no longer exposed as a print variable)

All I did here was replace 'Radius(1)|RL' in the plotting code with a calculated value - which I think is the same.  If I missed something when I changed the code, then I'm still missing it...

ilyamandel:
Ahh, you are right, I missed that -- two changes are necessary, not just one. ;-) 

I think we should just agree for consistency that the Roche lobe for an eccentric binary is computed at periapsis (technically, there's no such thing as a Roche potential for an eccentric binary, because there's no co-rotating reference frame in which the system looks fixed, but we have to make a consistent assumption.)

That means 
double RocheLobeRadius1() const { return CalculateRocheLobeRadius_Static(m_Star1->Mass(), m_Star2->Mass()) * SemiMajorAxisRsol(); }
should instead read
double RocheLobeRadius1() const { return CalculateRocheLobeRadius_Static(m_Star1->Mass(), m_Star2->Mass()) * SemiMajorAxisRsol() * (1-Eccentricity()); }
and similarly for  RocheLobeRadius2().

And then the change in post-processing that I mentioned. :)

jeffriley:
@ilyamandel  Scope creep addressed ;-)  Also added entry to "What's New" so users are aware of header string changes and changes to BSE detailed output default record.

ilyamandel (approved):
Thank you for indulging the scope creep, @jeffriley . :-)

PR #927, opened 2023-02-22
added fryer young 2007 bh kicks

reinhold-willcox:


jeffriley (changes requested):
Looks good @reinhold-willcox.  Just a few things:

"break" missing from case BLACK_HOLE_KICKS::FRYERYOUNG2007 in switch statement in BaseStar::ApplyBlackHoleKicks().  Without the "break", when OPTIONS->BlackHoleKicks() == BLACK_HOLE_KICKS::FRYERYOUNG2007 the code will fall through the BLACK_HOLE_KICKS::FRYERYOUNG2007 case to the default case - I'm surprised you got the results you were expecting.
Can you put a comment on the BLACK_HOLE_KICKS::FRYERYOUNG2007 case (for consistency with the other cases)?

Can you put a comment on or about line 2913 in BaseStar::DrawSNKickMagnitude() to indicate where the constant 3.5 comes from - so later readers of the code know.  I imagine it comes from FryerYoung2007, or maybe Banerjee et al. 2020?, but it would be good to explicitly say so.
Same statement - why the two IF statements and not just a single IF with && ?  Are you expecting possible additions/changes to the code inside the outer IF?  Doesn't really matter - just a matter of style I guess.

Lines 257 & 258 in Options.cpp seem to be mis-aligned (just formatting).


Make a note (either to self or to me :-)) to change line 1476 in Options.cpp after the YAML PR is merged (the default string can be automatically formatted after that PR is merged).

Lines 708 - 711 in Options.h seem to be misaligned (just formatting).  I think this might happen because we have different tab values et for our editors/IDEs.  It's not a huge problem, but every now and then we should tidy up the code I think - ragged/messy code is too hard to read (imo).

Need to add FRYERYOUNG2007 to the documentation.

jeffriley:
Oops - I didn't notice it's a draft :-)  Sorry...

reinhold-willcox:
All good, thanks @jeffriley ! This is meant to be a template, Oliver Anagnostou from Melbourne is thinking of adding a few different BH kick prescriptions. I hadn't actually run the code yet, just made sure that it compiled. But your advice is helpful, I'll make sure things are looking good. 

Edit: made all your requested changes, except for documentation.

reinhold-willcox:
Closing for now, we can re-open if needed.

PR #930, opened 2023-03-06
Fix for issue #855

### LINKED ISSUE #855: RLOF resulting in merger instead of mass transfer episode
I used v02.33.00 to reproduce a single system that lead to BBH merger in v02.21.01. However, during the first mass transfer episode, the primary in v02.33.00 significantly expands above its Roche lobe, meeting the contact condition and resulting in a stellar merger.

You can reproduce the system by executing the following command `./COMPAS -n 1 --mode BSE --metallicity 0.0001 --initial-mass-1 1.191637951222871e+02 --initial-mass-2 79.681699985633131 --detailed-output --semi-major-axis 2.1602553656995424625` using v02.33.00.

The system should not overfill its Roche lobe so much and should experience case B mass transfer.

I attach a plot where you can see the comparison between the old and new version.
bug_hunt

We should figure out in which version does the deffect was introduced and take it from there.

For a short Slack discussion on the issue, see https://compashq.slack.com/archives/CMCUJ859T/p1663784521632899 .

jeffriley:
Fix for issue #855 - changed condition in HG::UpdateInitialMass() from "<" to "<=" - refer to issue text for details.

Close issue #855 when merged


ilyamandel (changes requested):
I believe this fix may cause other problems, reducing the remnant mass of stars that lose parts of their envelope during the HG.  I urge further tests.  E.g., with f_WR=0, to kill mass loss from stripped He stars, would this change impact the final pre-SN mass of a star that loses mass in winds on the HG?

jeffriley:
Agreed @ilyamandel . I knew worst case this would just back-out your changes, but I wanted to get it in as a placeholder.

src/CHeB.cpp
```diff
@@ -1490,8 +1490,8 @@ STELLAR_TYPE CHeB::ResolveEnvelopeLoss(bool p_NoCheck) {
         m_Tau = ((m_Age - tHeIPrime) / tHePrime);                                   // Hurley et al. 2000, just after eq 81
         m_Age = m_Tau * HeMS::CalculateLifetimeOnPhase_Static(m_Mass0);             // Hurley et al. 2000, eq 76 and following discussion
 
-        CalculateTimescales(m_Mass0, m_Timescales);
         CalculateGBParams(m_Mass0, m_GBParams);
+        CalculateTimescales(m_Mass0, m_Timescales);
```
reinhold-willcox, line 1494: This seems like a very dramatic change, what is the reasoning here?

jeffriley: Timescales are calculated using GBParams.  All other calls to CalculateTimescales() follow a call to CalculateGBParams(). I think this one was mistakenly reversed.

reinhold-willcox: Has it been reversed this whole time? I would expect that to change functionality for CHeB stars. 

I originally thought this was related to the HeHG p and q constants (PR #744), which was complicated to resolve because of the order of the function calls, or something like that.

jeffriley:
> Has it been reversed this whole time? I would expect that to change functionality for CHeB stars.

It has - I just happened to notice it while researching this problem.  I suspect it was reversed because the functionality was reversed in the legacy code, and when I refactored I tried to not change too much wrt physics algorithms - but it may not have been - it might have just been my mistake.  Either way, it has been reversed the whole time (I think - I think I checked back a bit and didn't see a change).  Should it be?  Well, as I said, timescales are calculate using gbparams.  We could not change it here, but do some tests with it both ways to see what the effect actually is.  There will be an effect, but I'm not sure if it will be positive or negative.

ilyamandel (changes requested):
@jeffriley -- I don't think the part that fixes #855 should be included (the changes you made here for that fix just return us back to the initial problem; the fix is in #956 ) -- but the other changes you made here look good, so would you be able to remake this PR to just keep the other changes?

ilyamandel:
@jeffriley -- see my previous comment: I don't think the part that fixes https://github.com/TeamCOMPAS/COMPAS/issues/855 should be included ( part of the fix is in https://github.com/TeamCOMPAS/COMPAS/pull/956 and more work is needed,  but as we discussed, the changes you made here for that fix just return us back to the initial problem).  However, the other changes you made here look good, so would you be able to remake this PR to just keep the other changes?

jeffriley:
@ilyamandel Yes, will do.  I'll try to get to it in the coming week.

ilyamandel:
Hi @jeffriley,
The issue this PR was intended to fix, #855 , has been closed due to other fixes, so I suspect this PR is by now unnecessary . & outdated.  Can we close it, or is there code from here that you still want to apply?

PR #931, opened 2023-03-13
Non zams initialization

reinhold-willcox:


reinhold-willcox:
This PR encompasses and supercedes PR #732, which is now closed.

SimonStevenson:
It's great that you're working on this, I think it would be a very useful feature. I'd be happy to help review this.

reinhold-willcox:
@SimonStevenson  It should be ready to review now, if you have time.

jeffriley (changes requested):
I'll review this wrt terminology and code consistency

reinhold-willcox:
@jeffriley It says you requested changes but I don't see any requests. Is that a placeholder for future requests, or am I just not seeing existing ones?

jeffriley:
It's a placeholder - I did it so the PR wouldn't merge until we've made some changes.  I will have time to look at this after the end of this week - but I will be travelling for most of May, so may not get much done until I get back at the start of June.

ilyamandel:
@jeffriley -- I notice that there is a note from you from April that says you are about to look at this PR. ;-)  I am not sure if it's still relevant (@reinhold-willcox ?) but I think we should probably clean up very old PRs one way or another.

reinhold-willcox:
@ilyamandel This functionality would be very useful in COMPAS, in my opinion, so I would advocate addressing the remaining issues and getting the code committed. 

As it stands, the code here (at some point) worked for the initialization of MS, HeMS, WD, NS, and BH. From memory, Jeff and I were debating whether to allow for the initialization of other stellar types, and stars with tau != 0 (e.g stars partway along the HG), but this could be an extension in a later PR. There was also a question about the naming convention of the variables, but this is cosmetic so I am happy to adopt whatever convention works for the majority. If now is the right time to address these issues, I'll go ahead and update this branch to the latest version of `dev`.

ilyamandel:
I am not aware of any other major changes in progress, @reinhold-willcox , just bug fixes -- so I agree now would be a good time to do this, if you have spare cycles.  Thank you!

reinhold-willcox:
Closing: outdated wrt PR #1036

PR #934, opened 2023-03-14
Pulsar evo 2

yuzhesong:
In this PR, some of the pulsar evolution issues from before were resolved, and described as below:

1. Previously, for whatever reason pulsars weren't evolving correctly with NS.cpp code. I have re-wrote parts of NS.cpp and now pulsars are evolving according to descriptions in https://arxiv.org/pdf/1912.02415.pdf

2. The majority of the function: 
`void NS::UpdateMagneticFieldAndSpin()` 
is now written as a new function:
`void NS::spinDownIsolatePulsar(const double p_Stepsize)`
and the spinDownIsolatePulsar() function describes how isolated pulsars are spinning down due to magnetic braking.

3. Added specific pulsar time step description. 

4. In function `void NS::UpdateMagneticFieldAndSpin()` I added another else if condition to let the pulsar continue to spin down after the recycling is finished.


ilyamandel (changes requested):
Thanks, Robert!  A few questions/suggestions:

- Do we need song_oct_18.txt in the git repo?

- What is m_PulsarDetails.birthSpindown?  Do you mean m_PulsarDetails.birthSpinDownRate?  Similarly for other functions/outputs related to spin-down rates [e.g., Pulsar_BirthSpindown() -> Pulsar_BirthSpinDownRate() by analogy with Pulsar_SpinDownRate()]

- spinDownIsolatePulsar() -> SpinDownIsolatedPulsar()

- Why define bool PrintPulsarEvolutionParameters() as returning a boolean if you then ignore the output and typecast it as void?  Also, what's the intention of returning true when no printing is required?

- in typedefs.h, "double birthSpindown; // Pulsar birth period (s s-1)" -- that's not the right comment

- some strange comments in NS.cpp, e.g.:
" * This is changed to the form "
"* @param [IN] p_Omega Mass in Msol"
"* @param [IN] p_Alpha Mass in Msol"
"Spin down rate of an isolated Neutron Star in SI" -- isn't the spin-down rate dimensionless?
Suggest carefully going through and checking all comments.

- CalculateSpinDownRate() seems to return something in units of 1/s^2; why is it not just Pdot?

- CalculateSpinDownRate_Static() seems to have almost the same code and functionality as CalculateSpinDownRate(); one should just call the other

- " * JR: todo: description?
 * Robert: I think this function calculates the pulsar parameters at birth. "
 Is that true?  If so, we should just say this, without names and prior discussion :-)

jeffriley (changes requested):
Thanks Robert.  A few suggestions:

1. You have uploaded your .gitignore - do we need your makefiles in the project .gitignore?
2. You have uploaded song_oct_18.txt - can you remove it please?
3. You have added two calls to PrintPulsarEvolutionParameters() - prior to this PR we only printed one record to the Pulsars file in each timestep, and only then if the binary included at least one NS.  Your change possibly adds two records per timestep, and further more will add those records even if the binary does not include a NS.  Is that what you want?  If so, then you need to document that change in behaviour in the documentation (at least in the "What's New" section). Current COMPAS users that look at pulsars won't be expecting multiple records in the Pulsars file per timestep - they will need to know (a) that they may be there, and (b) how to interpret them.  You should probably also use record types to distinguish the different types of Pulsars records - see the discussion of record type at the top of Log.h. 
4. You have "Given in the form discussed by Simon & Robert" in the description of NS::ChooseTimestep(). That's not a reasonable description - nobody but Simon and Robert will understand that (and in five years' time we may not even know who Simon and Robert are...).
5. Please add ".0" to the constants in the conditionals in lines NS.cpp 44, 47-48, 50-54. c++ is a strongly-typed language: "1" is an integer, whereas "1.0" is a double.  It may make no difference now, but it might if someone changes things later - it's good practice to use the correct data type even for constants.
6. Please change "pow" to "PPOW" in NS.cpp line 54 (we use PPOW for profiling).
7. Why so many blank lines at/about 62, 355, 415 in NS.cpp?  We try to stick to 2 blank lines between functions throughout the code - consistency of formatting makes the code easier to read.
8. You need to update changelog.h
9. Line 2726 in constants.h:
{ ANY_STAR_PROPERTY::PULSAR_BIRTH_SPINDOWN, { TYPENAME::DOUBLE, "Pulsar_Birth_Spindown", "s s-1",  <- what units does "s s-1" represent?
10. You need to update the documentation for:
(a) changed default record specifier for BSE_DETAILED_OUTPUT_REC (actually, since we're discussing BSE_DETAILED_OUTPUT_REC, do we really need all those pulsar columns all the time - can we remove them from the default and let users who are studying pulsars add them for their runs via the logfile-definitions file?  Just a question - our output files are bloated, so anything we can do to reduce the bloat is useful).
(b) changed default record specifier for BSE_PULSAR_EVOLUTION_REC
(c) exposed new variables to be available for printing

jeffriley:
@ilyamandel 

> Why define bool PrintPulsarEvolutionParameters() as returning a boolean if you then ignore the output and typecast it as
> void? Also, what's the intention of returning true when no printing is required?

That's from the logging implementation - the logging functions return a boolean to indicate success (or not), but until we start bubbling all errors up via try/catch blocks we just throw the result away (head in the sand... see no evil... etc.).

yuzhesong:
Thanks for the feedbacks @ilyamandel @jeffriley I'll go through them in details later. 

@jeffriley per your number 3 - I forgot to document that in the pull request as it was done a bit while ago. The duplicated call of PrintPulsarEvolutionParameters() was because the first timestep of pulsar evolution wasn't actually printing the first timestep. It was instead printing the second timestep. Reinhold and I worked on this fix during the last powow, where we call the printing twice in the first time step only so we can have the pulsar information in the actual first timestep.

ilyamandel (changes requested):
A few more things:

---
    m_PulsarDetails.spinDownRate  = CalculateSpinDownRate_Static(m_PulsarDetails.spinFrequency, m_MomentOfInertia, m_PulsarDetails.magneticField,  NEUTRON_STAR_RADIUS * RSOL_TO_KM, 1.0);   // was using m_radius, alpha = 1.0

Why should it not be m_radius?  Regardless, I'd just say alpha = 1.0 in the comments, no reason to tell future developers what this used to look like.

Ditto here: "m_MomentOfInertia = CalculateMomentOfInertia_Static(m_Mass, NEUTRON_STAR_RADIUS * RSOL_TO_KM) ; //m_Radius * RSOL_TO_KM) ;"

---

In NS.cpp, line 374, another misplaced comment: "* @return Tuple containing the Maximum Mass Acceptance Rate and the Accretion Efficiency Parameter"

---

    double magFieldLowerLimit = PPOW(10.0, OPTIONS->PulsarLog10MinimumMagneticField()) * GAUSS_TO_TESLA;    
    double magFieldLowerLimit_G = magFieldLowerLimit * TESLA_TO_GAUSS; 
    
    Why convert back and forth? :)
    
    ---
    
    double momentOfInertia = m_MomentOfInertia; //* unitsMoI;
    
    Don't leave behind commented out code; and since you are keeping the original units, I don't think you really need to define this new variable.
   
   Remove other examples of commented out code, e.g., there are lots in NS::UpdateMagneticFieldAndSpin() such as 
   //constexpr double NSRadius_IN_M = NEUTRON_STAR_RADIUS * RSOL_TO_KM * KM_TO_M ;
 
    ----
    
    NS.cpp, line 383 -- this is never used, I think:
    "constexpr double unitsMoI = G_TO_KG * CM_TO_M * CM_TO_M;"
    
    ----
    
        // calculate the spin down rate for isolated neutron stars, see Equation 4 in arXiv:0903.3538v2 (Our version is in SI)      
    double pDotTop               = constant_2 * TESLA_TO_GAUSS * TESLA_TO_GAUSS * m_PulsarDetails.magneticField * m_PulsarDetails.magneticField;
    
    I am pretty sure your version is in CGS... ;-)
    
---

Commented out code and outdated comments are a confusing combination:
double radius = NEUTRON_STAR_RADIUS * RSOL_TO_CM;// m_Radius * RSOL * 100; // in meters

See earlier question on using NEUTRON_STAR_RADIUS here.

-----

I don't understand the logic in NS::UpdateMagneticFieldAndSpin().  There are three different conditions under which spinDownIsolatePulsar() is called.  Yet I am not sure whether the various if/else clauses actually catch all possibilities.  Can you simplify/clarify?

ilyamandel (commented):
//This is partially recycling the pulsars in a CE system
//This is the ''classical'' isolated pulsars

Should refer to a single pulsar; also, the first comment above is in an if clause that can apply even if p_CommonEnvelope is false (see line 475)

jeffriley:
@yuzhesong 

> The duplicated call of PrintPulsarEvolutionParameters() was because the first timestep of pulsar evolution wasn't actually
> printing the first timestep. It was instead printing the second timestep.

That must be because neither of the stars has become a NS at that point. If that's true, then do you really want to print the details of the stars/binary in the timestep that one of the stars becomes a NS?  In most cases we only switch stellar type at the end of a timestep, though there are times that we do the switch mid-timestep - is the problem you were seeing (the record not printing) due to that?

With the changes you have made (the addition of the calls to PrintPulsarEvolutionParameters()), because you are not checking the current stellar type of the stars, or the stellar type they are about to switch to if that's the problem, you will get records printed to the Pulsars file regardless of whether either of the stars is an NS now or is about to become one.

I would encourage you to work out why you were not getting the record printed in the timestep you want, and if it is because one of the stars is about to become a NS and hasn't yet switch, but you want the details in that timestep printed, then use record  types, similar to the way we use record types for the detailed output file.

yuzhesong:
@ilyamandel 

> I don't understand the logic in NS::UpdateMagneticFieldAndSpin(). There are three different conditions under which spinDownIsolatePulsar() is called. Yet I am not sure whether the various if/else clauses actually catch all possibilities. Can you simplify/clarify?

The main frame of NS.cpp is still intact from Debatri's version of the scripts. So from what it looks like, originally there are only two conditions where pulsars are evolved here:
1. pulsar is not recycled, and is not interacting with the companion (the first IF condition)
2. pulsar is recycled, there is a common envelope in the system, and accretion is happening. 

The last else if stated is what I added, so that after the recycling described above happened and finished, the pulsar will keep evolving/being spun down by magnetic braking.

And yes, this is absolutely not all of the conditions there could be. Effectively, this means COMPAS can't produce MSPs right now. Simon and I will add in the formation channels for MSPs and LMXBs later on.

yuzhesong:
@ilyamandel 

> CalculateSpinDownRate() seems to return something in units of 1/s^2; why is it not just Pdot?

The output file records the spin-down rate in the form of the derivative of spin-frequency, which is what CalculateSpinDownRate() returns.

jeffriley:
@yuzhesong @ilyamandel 

Sorry to step in the middle, but

>> Yet I am not sure whether the various if/else clauses actually catch all possibilities.
>
> And yes, this is absolutely not all of the conditions there could be. Effectively, this means COMPAS can't produce MSPs 
> right now. Simon and I will add in the formation channels for MSPs and LMXBs later on.

if the conditions (if/else if/else if) don't capture all possibilities, what happens when one of those uncaptured possibilities occurs?  The way the code is written, the function just does nothing - it returns without changing anything.  Is that what you intend?  If so, you should document that in the description of the function. Or should there be a catch-all "else {}" at the end where you explicitly handle "other" conditions (do you need to emit a warning that an unexpected condition has occurred?).

yuzhesong:
@jeffriley 

> Sorry to step in the middle, but
> 
> Yet I am not sure whether the various if/else clauses actually catch all possibilities.
> 
> And yes, this is absolutely not all of the conditions there could be. Effectively, this means COMPAS can't produce MSPs
> right now. Simon and I will add in the formation channels for MSPs and LMXBs later on.
> 
> if the conditions (if/else if/else if) don't capture all possibilities, what happens when one of those uncaptured possibilities occurs? The way the code is written, the function just does nothing - it returns without changing anything. Is that what you intend? If so, you should document that in the description of the function. Or should there be a catch-all "else {}" at the end where you explicitly handle "other" conditions (do you need to emit a warning that an unexpected condition has occurred?).

This is not when none of the cases is satisfied but, one of the cases that I found was not behaving properly was a binary system with a pulsar accreting from a companion low mass MS star throughout the lifetime of the system. It satisfies the second else if statement in this function, but the evolutionary path of the pulsar (fig below) is obviously not properly calculated for that scenario. 

image

But to answer your question, it should be that when the criteria aren't met, indeed nothing should happen. For now I guess I can just add an else {} statement and print out a warning?

yuzhesong:
@ilyamandel 

> Why should it not be m_radius? Regardless, I'd just say alpha = 1.0 in the comments, no reason to tell future developers what this used to look like.

I don't know why there was a separate NEUTRON_STAR_RADIUS instead of m_Radius, I also tried to just print them out, and they are slightly different from each other (while the same for all neutron stars)

NEUTRON_STAR_RADIUS = 1.42857e-05, m_Radius 1.4374e-05 (both in R_sol)

jeffriley:
>> Why should it not be m_radius? Regardless, I'd just say alpha = 1.0 in the comments, no reason to tell future developers
>> what this used to look like.
>
> I don't know why there was a separate NEUTRON_STAR_RADIUS instead of m_Radius, I also tried to just print them out, and
> they are slightly different from each other (while the same for all neutron stars)
>
> NEUTRON_STAR_RADIUS = 1.42857e-05, m_Radius 1.4374e-05 (both in R_sol)

NEUTRON_STAR_RADIUS is defined in constants.h as (1.0 / 7.0) * 1.0E-4 (= 1.42857e-05)
According to the comment in constants.h this comes from Hurley et al. 2000, just after eq 93, and is supposed to be 10km in R_sol.  However, Hurley lists it as 1.4E-5, so I imagine (1.0/7.0)E-4 is someone's better approximation of 10km in R_sol.

m_Radius = 1.4374e-05 comes from setting the radius to 10km in NS.cpp, then applying the conversion constant in constants.h (KM_TO_RSOL = 1 / RSOL_TO_KM = 1.0 / 6.957E5) - so 10.0 / 6.957E5 = 1.4374e-05.

Maybe we should change the value of NEUTRON_STAR_RADIUS in constants.h to be 10.0 * KM_TO_RSOL
And maybe we should replace the instances of 10.0 in NS::CalculateRadiusOnPhaseInKM_Static() with a constant defined in constants.h?

EDIT: I think you should use m_Radius - it is set to a different value in NS.cpp if **--neutron-star-equation-of-state ARP3** is used instead of the default (**--neutron-star-equation-of-state SSE**)

yuzhesong:
@ilyamandel @jeffriley 
The latest commit should covered all the suggestions but these two:
1). logging the first time step of pulsar evolution
2). recycling channels of pulsars. 
Both of these are stated in changelog.h

jeffriley:
Thanks @yuzhesong . I will look at the changes in detail in the next few days.  The following jumped out at me:

1. You need to change the version string in changelog.h

2. If the units are second per second, shouldn't the units string be either "s/s", or "s s^-1" (not just "s s-1")?

3. Can you actually remove ".gitignore" and "song_oct_18.txt" (I don't think ".gitignore" needs to be pushed if it is not to be changed, but if you can't remove it from the PR then I guess pushing an unchanged version also works (I have trouble removing accidentally pushed files - maybe ask @reinhold-willcox - he knows more about github than I do))

yuzhesong:
@jeffriley 

1. version number updated;
2. unit updated to s/s
3. "song_oct_18.txt"  should already be removed from the last push from what I can tell, and I also made sure to reverse .gitignore to be identical to the project .gitignore, but do let me know if you still see them in the PR somehow.

jeffriley:
@yuzhesong You're right - .gitignore and song_oct_18.txt are gone - I must have been hallucinating...

jeffriley:
A couple of comments for the moment - I'll look more later.

In the description of NS::ChooseTimeStep():

     * Given in the form as follow, with T being age of pulsar
     * 0 < T < 0.01 Myr, time stpe is 1e-3 Myr;
     * 0.01 < T < 0.1 Myr, time stpe is 1e-2 Myr;
     * 0.1 < T < 1 Myr, time stpe is 0.1 Myr;
     * 1 < T < 10 Myr, time stpe is 1 Myr;
     * 1 < T < 500 Myr, time stpe is linear in log space;
     * T > 500 Myr, time step is always 500 Myr
     *
     * ChooseTimestep(const double p_Time)

the description doesn't actually match the code (specifically, the description doesn't allow for the "=" cases).

And

     * 1 < T < 10 Myr, time stpe is 1 Myr;
     * 1 < T < 500 Myr, time stpe is linear in log space;
 
I think the lower bound on the second of the statements above should be 10?
    
Also, just repeating the code in the description isn't all that useful - more useful would be a description of the rationale you used for deciding on the stepped nature of the timestep, and why you chose the step boundaries, and the value of the timestep.

There are some typos in the description ("stpe" in several places).

And finally, this:

     * ChooseTimestep(const double p_Time)

should be

     * double ChooseTimestep(const double p_Time)
  

@yuzhesong and I will talk about the additional calls to **PrintPulsarEvolutionParameters()** next week sometime with a view to resolving that issue.

jeffriley:
I forgot to mention:

If **p_Time** == 500.0, will the result be 500.0 if the value is calculated by interpolation?  If so, you should change the conditional:

    else if (p_Time <= 500) {

to

    else if (p_Time < 500.0) {

No point doing the interpolation if we already know what the result is.

And...  Some of your conditionals compare to integer constants.  e.g.:

    else if (p_Time <= 500) {

Please change them to floating-point values.  The compiler will do the promotion implicitly in those circumstances, but it's good to specify the data type when we know it.  Here's the reason it's important to get into the habit of specifying the correct data type, even with constants.  The following c++ code, when executed:

    std::cout << "M_PI           = " << M_PI << "\n";
    std::cout << "2.0/3.0 * M_PI = " << 2.0/3.0 * M_PI << "\n";
    std::cout << "2/3 * M_PI     = " << 2/3 * M_PI << "\n";

    M_PI           = 3.14159
    2.0/3.0 * M_PI = 2.0944
    2/3 * M_PI     = 0

The compiler didn't know that we wanted the "2/3" evaluated as a floating-point operation, so return the integer value.  If we get into the habit of specifying the correct data type even for constants, we avoid these sorts of problems - we know the context, the compiler has to guess, and doesn't always guess right.

jeffriley:
In NS::UpdateMagneticFieldAndSpin():

    std::cout<<"Pulsar Evolution Not Invoked."<<std::endl;

That's not how we emit warnings and errors in COMPAS - look for instances of "SHOW_WARN(" in the code to see examples.  Think about putting the warning string in ERROR_CATALOG in constants.h rather than just having plain text in the code - we don't have all our error/warning message in the ERROR_CATALOG, but we should be trying to get them all there.

jeffriley:
> @jeffriley per your number 3 - I forgot to document that in the pull request as it was done a bit while ago. The duplicated
> call of PrintPulsarEvolutionParameters() was because the first timestep of pulsar evolution wasn't actually printing the first
> timestep. It was instead printing the second timestep. Reinhold and I worked on this fix during the last powow, where we call
> the printing twice in the first time step only so we can have the pulsar information in the actual first timestep.

@yuzhesong and I discussed this today in the COMPAS dev meeting.  I think the problem here is not that the first timestep of pulsar evolution wasn't being printed, but rather that the attributes of the pulsar are being changed in the first timestep by the following code in BaseBinaryStar::EvaluateBinary():

    m_Star1->UpdateMagneticFieldAndSpin(m_CEDetails.CEEnow, m_Dt * MYR_TO_YEAR * SECONDS_IN_YEAR, EPSILON_PULSAR);      // update pulsar parameters for star1
    m_Star2->UpdateMagneticFieldAndSpin(m_CEDetails.CEEnow, m_Dt * MYR_TO_YEAR * SECONDS_IN_YEAR, EPSILON_PULSAR);      // update pulsar parameters for star2

@yuzhesong should confirm this hypothesis and, if it is correct, we need to determine how best to resolve the issue - do we execute those function calls **before** evolving the stars instead of after?  That would have the effect of updating magnetic field and spin at the start of the timestep rather than the end, so it would not affect the birth attributes of the pulsar - but we need to think about any side-effects.

yuzhesong:
I can confirm with @jeffriley about the fixes we need to make with the printing of neutron star properties. 

It also turns out that defining NEUTRON_STAR_RADIUS is necessary @ilyamandel because the birth properties of a NS is set  in the time step before it turns into a NS, which means that if we don't define NEUTRON_STAR_RADIUS,  it would use the radius of the star before going supernova to set the moment of intertia and P dot of a NS and give really weird values. So I will be reverting to this in a later change.

jeffriley:
@yuzhesong 

> ... because the birth properties of a NS is set in the time step before it turns into a NS ...

That's not quite right - let's talk about this before you make changes.

yuzhesong:
@jeffriley I printed out all the variables used in NS::CalculateSpinDownRate_Static() for one system, and it prints the radius of the star prior to supernova explosion (at 6e13 cm)

jeffriley:
@yuzhesong I'd have to look at the source for NS::CalculateSpinDownRate_Static() and when it is called.  A static function does not require an object, so it can be called anytime - it does not require that the star is a neutron star.  The NS attributes are set in NS::Initialise() when the switch is done.

EDIT: bear in mind that a static function cannot access class member variables - so you pass things in, you get results out, but no class member variables are interrogated or changed.

yuzhesong:
@jeffriley I'll have to look at it later, but for some reason the neutron star birth Pdot can be set with the wrong radius, which would end up with a very unphysical Pdot, e.g. 1e20 s/s instead of 1e-10 s/s

jeffriley:
See my added note above re static functions. NS::CalculateSpinDownRate_Static() is not changing any attributes of a neutron star - or any other object for that matter.

yuzhesong:
@jeffriley Yeah, the function itself is not changing any attributes; it gets called in NS::CalculateAndSetPulsarParameters(), which is used to set the pulsar parameters at birth; and if m_Radius is passed as an argument to NS::CalculateSpinDownRate_Static() within NS::CalculateAndSetPulsarParameters(), it can use the weird radius that's not 10km, and return an even weirder Pdot at birth for the pulsar. I'm assuming this all happens within the timestep where all the things happens before the star is labeled as a NS, but I'll have to confirm this tomorrow

jeffriley:
Ok, but NS::CalculateAndSetPulsarParameters() is not static and can't be called without an object of class NS existing.  So, back to this:

> ... because the birth properties of a NS is set in the time step before it turns into a NS ...

That's not possible.  The birth properties of a neutron star are set in NS::Initialise() and, as you point out, possibly changed in NS::CalculateAndSetPulsarParameters(), and neither of those functions can be called without an object of class NS existing.

And this:

> I printed out all the variables used in NS::CalculateSpinDownRate_Static() for one system, and it prints the radius of the star
> prior to supernova explosion (at 6e13 cm)

That may be so, but if NS::CalculateSpinDownRate_Static() was being called from NS::CalculateAndSetPulsarParameters(), or any other non-static NS class member function, then the NS object exists. An object of class NS is only instantiated as a result of a supernova event, so if you are printing the attributes of the star prior to the SN event, then you are printing the attributes of the progenitor, not the neutron star.

Let's talk about this - maybe in the next dev meeting.

EDIT:

> ... before the star is labeled as a NS

A star is not labelled NS.  A star is an object of NS class, and that object will have m_StellarType = STELLAR_TYPE::NS.  If an object of NS class is not instantiated, there is no NS star, and there is no object that has m_StellarType = STELLAR_TYPE::NS.

jeffriley (changes requested):
Thanks @yuzhesong.  I think we're nearly done - just one thing:

The "f" at the end of line 23 of standard-logfiles-example-definitions-file.rst - typo?
    # start with the default SSE Parameters specification and add ENV_MASSf

Ok, two things... you'll need to fix the merge conflicts.

yuzhesong:
@jeffriley should be fixed

jeffriley (approved):
Thanks Robert - all good.

I think we're waiting on @ilyamandel now.

jeffriley:
@ilyamandel ping

ilyamandel (changes requested):
Sorry, I had missed that this was in my court.  Only a few very small things, mostly checking over my comments from last time:

spinDownIsolatePulsar() -> SpinDownIsolatedPulsar()
[you fixes the capitalisation, but note the *d* in Isolated]

Units of spin-down rate: OK, but can you specify them in the output format documentation (as you do for the birth period)?

enum class PULSAR_RECORD_TYPE: unsigned int {                                                                       // BSE_PULSAR_EVOLUTION file record type
    DEFAULT = 1,                                                                                                    //  1 - record describes the initial state of the binary
    POST_BINARY_TIMESTEP,                                                                                           //  3 - record was logged immediately following binary timestep (i.e. the evolution of the binary system for a single timestep)
};

Why do you need to list specific values like 1 and 3?  (Also, would it even be 3, since it's the second option?).  [I am surprised that the compiler doesn't complain about the apparently spurious comma after POST_BINARY_TIMESTEP]

I think @jeffriley requested changing s-1 to s^-1 (see, e.g., "double birthSpinDownRate; // Pulsar birth down rate(s s-1)" in typedefs.h)

    else if (p_Time < 500.0) {
        double slope      = (log10(500.0) - log10(1.0)) / (log10(500.0) - log10(10.0));
        double intersect  = log10(1.0) - slope * log10(10.0);
        double log10_step = intersect + slope * log10(p_Time);
        result = PPOW(10.0, log10_step);
    }
    
    You are making this far too complicated. :) Also, there's no need to calculate log10(1) [it's zero] or log10(10) [it's 1]. ;-) 

Why does double CalculateSpinDownRate_Static(const double p_Omega, const double p_MomentOfInteria, const double p_MagField, const double p_Radius, double const p_Alpha)
take a parameter p_Alpha which is never used?  It gives the false impression that the misalignment is taken into account during spin evolution, when in fact it's not.

    else if ((m_PulsarDetails.spinFrequency < 2.0 * M_PI * 1000.0) && (p_RecycledNS || p_CommonEnvelope) && utils::Compare(p_MassGainPerTimeStep, 0.0) > 0) {
        //This is partially recycling the pulsars in a CE system 
  Contrary to the comment, the code condition seems to sometimes evaluate to true for non-CE systems (i.e., when p_CommonEnvelope is false)
  
  I still don't understand the three conditions in that function, BTW.  If the last if is supposed to catch everything else, why does it have an else statement?  If it's not meant to catch everything else, what happens when none of the three conditions are met?
  
Don't forget to merge in dev and update changelog once ready. :)

ilyamandel:
Thanks, @yuzhesong !  I fixes a few typos in comments, happy to approve. :)

ilyamandel (approved):

PR #902, opened 2023-02-01
Refactoring python scripts into a module "compas_python_utils"

themikelau:
- Refactors various COMPAS python utility scripts into an installable python module
- Adds CLI tools to make access to some scripts easier (eg `compas_h5view`)

Closes #863, #864

PR #935, opened 2023-03-14
Fix the workflow to work successfully on MacOS 12. This required chan…

emanehab99:
- Fix the workflow to work successfully on MacOS 12. This required chanding the dependencies and commenting plt.show() in the plotting script
- Use setup-python action to cache pip dependencies
- Use cache action to cache macos dependencies, which saved ~7mins of building time
- Restrict the paths that triggers the workflow to avoid unncessary runs not related to the workflow


ilyamandel (changes requested):
@emanehab99 , I understand that commenting out plt.show() from plot_detailed_evolution.py may be good for automatic testing, but won't this make the script less useful for a typical user who actually wants to see the plots?

emanehab99:
@ilyamandel I agree. Hence I suggested in my email to either use a separate copy of the script for CI/CD purposes (this needs to be maintained whenever the plotting script changes) or add comments in the script to clarify why this line is commented and it's safe to be uncommented

ilyamandel:
@emanehab99 : Following discussion with @avivajpeyi , @jeffriley et al., we'll address this by having an optional noshow argument in the plotting script; CI will use that argument to avoid creating plots.

ilyamandel:
@avivajpeyi -- now that PR #902 is in, can you please update this PR with the do not plot flag?

avivajpeyi:
Hey Ilya -- will update this before the dev meeting tomorrow!

avivajpeyi:
i _think_ the mac-os-12 container is known to be slow: https://github.com/actions/runner-images/issues/6147

This means while the CI will take ~ 1.5 min to install ubuntu CI dependencies, the CI can take > 15 min to install the mac dependencies (much better than the 1hr we noticed here https://github.com/TeamCOMPAS/COMPAS/pull/938, due to @emanehab99 's cool optimization hacks :D ) 

Maybe instead of running the mac-os checks as frequently as the ubuntu checks, we can instead do a weekly run or smth?

avivajpeyi:
Manually canceled the macOs workflow after 1.5Hrs (offft)

ilyamandel:
@avivajpeyi -- fine on turning off MacOS compilation for now.  But what about the detailed evolution plotter with the no plot option -- is that good?

avivajpeyi:
yep the CI running on ubuntu runs the detailed plotter and saves the fig (not yet being deployed to the TeamCOMPAS.github.io, but downloadable) 

Turning off MacOS compilation --  cool -- in that case shall we just close this PR? I think the main thing this PR is adding is MacOS build.

ilyamandel:
Well, let's ask  @emanehab99: is there a way to get around this taking more than 1.5 hours for MacOS?

avivajpeyi:
@ilyamandel macos took ~35min (after caching) while ubuntu took ~8min

ilyamandel:
I think this has been overtaken by other developments / implemented by other PRs, so closing this one.

PR #938, opened 2023-03-20
add macos to CI-compile test

avivajpeyi:


ilyamandel (approved):
Looks good to me! :)

avivajpeyi:
Killing this, as its taking way too long. Maybe we can come back to this in the future when the macos builds are more stable.

PR #944, opened 2023-03-21
add gitignore

avivajpeyi:


jeffriley (approved):
Thanks @avivajpeyi !

reinhold-willcox:
We had a .gitignore, what happened to it?

jeffriley:
Apparently I deleted it accidentally.  I'll figure git out one day...

reinhold-willcox:
classic

ilyamandel:
I asked Jeff about that when reviewing the PR that deleted .gitignore and
he convinced me that it was only on his own machine -- which doesn't make
sense in retrospect, of course.  So it's my fault too for being too easily
convinced. :)

PR #949, opened 2023-03-26
Merge conflict repairs

jeffriley:
Repairs after bad merge conflicts fix

### DIFF
--- a/online-docs/pages/whats-new.rst
+++ b/online-docs/pages/whats-new.rst
@@ -6,17 +6,20 @@ Following is a brief list of important updates to the COMPAS code.  A complete r
 
 **LATEST RELEASE** |br|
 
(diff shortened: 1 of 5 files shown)


jeffriley:
@reinhold-willcox @ilyamandel I didn't add an entry to changelog.h, or increment the version number, for this change.  I vacillated over doing that - technically I probably should since it changes the code (changelog.h and a few code tidy-ups), but it is really just putting things back the way they should have been.

Now that I've typed that I think I should add an entry to changelog.h and chanve the version number to 02.37.01 - it is a defect repair and it does change the code, and we should document why we did that.  I'll put this in draft until I make that change.

jeffriley:
Ok, changelog updated - ready for review

ilyamandel (approved):

PR #953, opened 2023-03-27
add docs for compas-py utilities 

avivajpeyi:
Requires #954


avivajpeyi:
Closing for now as there is not much need for this

PR #955, opened 2023-04-05
Added missing const in WD files and new option for qCrit prescription

nrsegovia:
- Fixed the missing `const` in WD-related functions
- Added the Hurley+ 2002 prescription with the Hjellming & Webbink 1987 condition for mass transfer from a Giant, as required for the basic binary runs in the code comparison project.

Caveats: I'm not sure about my modification to `Remnants::CalculateCriticalMassRatio`. Also, the remaining warnings are all related to `-Woverloaded-virtual`, however I do not know how to deal with them. If it is straightforward, I can add the fix to this PR.

~~Pending: documentation updates, `changelog.h` update. I will do this if there are no major issues related to the changes in this PR.~~


jeffriley (changes requested):
I can look at fixing the overloaded-virtual warnings after this is merged.

Two things at first glance (I haven't reviewed the code, these just stood out):

(i) changelog.h and the version number haven't been updated
(ii) I don't see any changed documentation files - there should be if you added a new prescription

EDIT: sorry - I just saw your "Pending" comment...  Maybe change this to a draft?

nrsegovia:
@jeffriley I have included those changes in the latest commit 😀

jeffriley (changes requested):
Thanks @solfreludio .  A few comments/requested changes:

In CHeB.h you define

    double CalculateCriticalMassRatioHurleyHjellmingWebbink() const { return 0.33; }

and it is similarly defined elsewhere with a different constant returned.  Can you add a comment to all instances, similar to the definition of e.g. double CalculateGyrationRadius() in CHeB.h, that explains the choice of constant - otherwise it is just a magic number that other developers might not recognise or know the source.

The description (in the comments immediately prior to the implementation) of GiantBranch::CalculateCriticalMassRatioHurleyHjellmingWebbink()
don't match the function.  Also, in the comment "Defined as mDonor/mAccretor in paper", make sure it is clear which paper you're referring to. 
Also in that function you use some integer constants. c++ is a strongly-typed language: "1" is an integer, whereas "1.0" is a double. The compiler will do the promotion implicitly in those circumstances, but it's good to specify the data type when we know it - so please change them to floating-point values. 

Here's the reason it's important to get into the habit of specifying the correct data type, even with constants. The following c++ code, when executed:

    std::cout << "M_PI           = " << M_PI << "\n";
    std::cout << "2.0/3.0 * M_PI = " << 2.0/3.0 * M_PI << "\n";
    std::cout << "2/3 * M_PI     = " << 2/3 * M_PI << "\n";


    M_PI           = 3.14159
    2.0/3.0 * M_PI = 2.0944
    2/3 * M_PI     = 0

The compiler didn't know that we wanted the "2/3" evaluated as a floating-point operation, so returns the integer value. If we get into the habit of specifying the correct data type even for constants, we avoid these sorts of problems - we know the context, the compiler has to guess, and doesn't always guess right. It may make no difference to your code right now, but it might if someone changes things later.

Can you explain the functionality of Remnants::CalculateCriticalMassRatio() a little more (in the comments immediately prior to the implementation)?  Why do we do nothing for most prescriptions?  Are we ever likely to implement another prescription for remnants?  If not, then we probably don't really need the switch statement - maybe a plain if statement will suffice?  Should we emit an error or warning if this function is called when the prescription is not QCRIT_PRESCRIPTION::HURLEY_HJELLMING_WEBBINK?  If not, why not? Also, please put a comment explaining the source of the constant (1.6).

nrsegovia:
Thanks for the comments @jeffriley , I think I have gone through all of them. 

> In CHeB.h you define
> 
> ```
> double CalculateCriticalMassRatioHurleyHjellmingWebbink() const { return 0.33; }
> ```
> 
> and it is similarly defined elsewhere with a different constant returned. Can you add a comment to all instances, similar to the definition of e.g. double CalculateGyrationRadius() in CHeB.h, that explains the choice of constant - otherwise it is just a magic number that other developers might not recognise or know the source.
> 

Added comments specifying where the numbers come from.

> The description (in the comments immediately prior to the implementation) of GiantBranch::CalculateCriticalMassRatioHurleyHjellmingWebbink() don't match the function. Also, in the comment "Defined as mDonor/mAccretor in paper", make sure it is clear which paper you're referring to. Also in that function you use some integer constants. c++ is a strongly-typed language: "1" is an integer, whereas "1.0" is a double. The compiler will do the promotion implicitly in those circumstances, but it's good to specify the data type when we know it - so please change them to floating-point values.
> 

I believe you meant that the description pointed to a different function related to Claeys? If so, I have changed that. The paper has been specified and the integers have all been removed.

> Can you explain the functionality of Remnants::CalculateCriticalMassRatio() a little more (in the comments immediately prior to the implementation)? Why do we do nothing for most prescriptions? Are we ever likely to implement another prescription for remnants? If not, then we probably don't really need the switch statement - maybe a plain if statement will suffice? Should we emit an error or warning if this function is called when the prescription is not QCRIT_PRESCRIPTION::HURLEY_HJELLMING_WEBBINK? If not, why not? Also, please put a comment explaining the source of the constant (1.6).

I had the same question so there is a discussion in the disc_mass_transfer channel on slack. Basically remnants are not considered since evolution stops at double white dwarf formation. I have reverted the changes as @ilyamandel pointed out that we would have to implement  GW evolution, too.

jeffriley (approved):
Thanks Nicolas.  I'm good with this - @ilyamandel ?

ilyamandel (changes requested):
Hi @solfreludio,

Thank you for this PR, and sorry for the belated comments.

1.
double CHeB/HeMS/MS_gt_07::CalculateCriticalMassRatioHurleyHjellmingWebbink() const { return 0.33; }
double HG::CalculateCriticalMassRatioHurleyHjellmingWebbink() const { return 0.25; }
This refers to Hurley+ 2002 section 6.2.1, but I cannot find these prescriptions there.  What am I missing?  If you are assuming something else, please provide the correct reference or explain your choice.

2. double HeGB/HeHG::CalculateCriticalMassRatioHurleyHjellmingWebbink() const { return 1.28; }
double MS_lte_07::CalculateCriticalMassRatioHurleyHjellmingWebbink() const { return 1.44; }
I presume these are inverse of the listed values of 0.784 and 0.695; please explain that in the comments so future developers can track this down.

3. double TPAGB::CalculateCriticalMassRatioHurleyHjellmingWebbink() const { return GiantBranch::CalculateCriticalMassRatioHurleyHjellmingWebbink(); }
You don't actually need this because TPAGB will automatically inherit from EAGB. :)

4. https://github.com/TeamCOMPAS/COMPAS/pull/955/files shows that a bunch of files have lost their final newline characters?

5. The constants.h file is a bit out of date (dev needs to be merged back in and any merge conflicts manually resolved).

nrsegovia:
> double CHeB/HeMS/MS_gt_07::CalculateCriticalMassRatioHurleyHjellmingWebbink() const { return 0.33; } double HG::CalculateCriticalMassRatioHurleyHjellmingWebbink() const { return 0.25; } This refers to Hurley+ 2002 section 6.2.1, but I cannot find these prescriptions there. What am I missing? If you are assuming something else, please provide the correct reference or explain your choice.
> 
I had the same issue as you did, since the values are not mentioned in the paper. Looking at Hurley's BSE code you can find that 3 (1/3) and 4 (1/4) are used. I guess I will make it explicit, i.e. "As coded in BSE".

>     2. double HeGB/HeHG::CalculateCriticalMassRatioHurleyHjellmingWebbink() const { return 1.28; }
>        double MS_lte_07::CalculateCriticalMassRatioHurleyHjellmingWebbink() const { return 1.44; }
>        I presume these are inverse of the listed values of 0.784 and 0.695; please explain that in the comments so future developers can track this down.
> 
Okay, I think I only commented something along those lines in Giants.cpp

>     3. double TPAGB::CalculateCriticalMassRatioHurleyHjellmingWebbink() const { return GiantBranch::CalculateCriticalMassRatioHurleyHjellmingWebbink(); }
>        You don't actually need this because TPAGB will automatically inherit from EAGB. :)
>
Thanks, I'm still not used to working with inheritance.
 
>     4. https://github.com/TeamCOMPAS/COMPAS/pull/955/files shows that a bunch of files have lost their final newline characters?
> 
I can see that as well, but it is not something done on purpose (I have no idea how it happened). I will try to fix it.

>     5. The constants.h file is a bit out of date (dev needs to be merged back in and any merge conflicts manually resolved).
Sure, but I'm pretty sure these are changes introduced recently (I checked for changes in dev before my latest commits). Thanks for the heads-up. **EDIT:** are you sure it was constants.h? I did not get any conflicts nor merges when syncing.

ilyamandel:
> I guess I will make it explicit, i.e. "As coded in BSE"

Sounds good!

> Sure, but I'm pretty sure these are changes introduced recently

Yes, they were, so this is definitely not your fault.  Actually, it's mine, for rolling in another PR while this one was under review.  Sorry!

ilyamandel:
@solfreludio -- have you made all the changes you wanted?  If yes, I can approve before we get out of sync again (I have another PR on the way :) ).

nrsegovia:
@ilyamandel I have just finished. Please check constants.h, though... As I did not get any merge conflicts (actually no merge at all for that file), so I'm wondering if I missed something. Besides that, I have restored the new lines and improved the description for the values being used.

ilyamandel (approved):
Looks good, @solfreludio !  Thank you for the fixes!

PR #956, opened 2023-04-07
Bug Fix for issue #855

ilyamandel:
Resolves issue #855 by using Mass0 rather than Mass to determine ages and timescales


jeffriley (changes requested):
Thanks @ilyamandel.  The only addition I'd like to see is the rationale for HG::CalculateCoreMassOnPhaseIgnoringPreviousCoreMass - maybe in the preamble comments to the function.  Just so people looking at the code in the future have some idea why it is there and the context in which it is (and should be?) used.

jeffriley (approved):
Thanks for the explanation. Approved.

PR #960, opened 2023-04-20
Fix CI errors

avivajpeyi:
@ilyamandel found that the `plot_detailed_evolution` is failing in the CI due to `latex` not being found: 
https://github.com/TeamCOMPAS/COMPAS/actions/runs/4749210567/jobs/8436237386
```
=========================== short test summary info ============================
FAILED py_tests/test_plot_detailed_evolution.py::test_plotter - RuntimeError: Failed to process string with tex because latex could not be found
```

This PR will fix this error

### DIFF
--- a/.github/workflows/compas-compile-ci.yml
+++ b/.github/workflows/compas-compile-ci.yml
@@ -32,7 +32,9 @@ jobs:
           cd misc/cicd-scripts
           chmod 755 linux-dependencies
           ./linux-dependencies
-          echo "checking latex:"
+          echo "Running which python"
+          which python
+          echo "Running which latex"
           which latex
 
       - name: Build Compas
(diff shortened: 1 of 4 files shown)


avivajpeyi:
Ok, things seem to be working with e6ee4855b9af5fc49645641eb230aaca1f629649 😄 

Its not the best solution -- as now we're spending more than 3 min installing Latex stuff (i couldn't figure out which latex package was leading to the issue so I'm just installing a bunch of standard latex packages, although we probably don't need most) 

In the future if this comes up again, it may be worthwhile looking into souring TinyTex:
```
      - uses: r-lib/actions/setup-tinytex@v2
      - name: add TeX deps
        run: |
          tlmgr install cm-super
          tlmgr install type1cm
          tlmgr install underscore
          tlmgr install dvipng
          tlmgr install psfrag
```
https://www.rdocumentation.org/packages/tinytex/versions/0.44

This is a lightweight version of Latex that has some helpful features for installing missing packages

avivajpeyi:
Oh weird, the dea6538dfe0224a01f42f6adc898e04eb9ca20b8 CI failed due to an error with sampleh5 -- I think this is giving different results maybe because of an issue with setting the random seed?

ilyamandel (approved):
Looks good -- thank you, @avivajpeyi !

PR #962, opened 2023-04-20
2 stage CE upgrade

ilyamandel:
Included Picker et al. (2023, in prep.) fits for the convective envelope mass in the TWO_STAGE common envelope treatment.

Default behaviour is unchanged.


ilyamandel:
Note: because this branch initially started off as being branched off from another branch (the one changing the default options), https://github.com/TeamCOMPAS/COMPAS/pull/962/files still shows changes relative to the original branching off point rather than relative to current dev.  @reinhold-willcox , @jeffriley -- I think you looked into this, is there a way to fix it?  (Not a big deal, but would make it easier for reviewers...)

ryosuke-hirai (approved):
Looks good to me.

LewisPicker (approved):
The functions appear to be applied correctly. 

P.s you were correct about applying Tnorm sorry for the confusion.

PR #963, opened 2023-05-02
add compas_h5copy as a CLI program

avivajpeyi:
This will allow users to run 
```
compas_h5copy [options]
```

Users can still run
```
python /path/to/compas_h5copy.py [options]
```


jeffriley (approved):
Thanks @avivajpeyi

PR #971, opened 2023-05-30
Analytic expression for star forming mass per evolved binary

### LINKED ISSUE #378: Analytic solution for rate normalisation from COMPAS mass cuts
No, I just think there is a slightly simpler way of doing this.

In [totalMassPerZ.py](https://github.com/TeamCOMPAS/COMPAS/blob/production/postProcessing/Folders/CosmicIntegration/PythonScripts/totalMassEvolvedPerZ.py), COMPAS finds the fraction of mass in COMPAS vs mass in total population in order to normalise the rates. It does this through inverse CDF sampling, but I believe it could be found analytically instead. This is both faster and also makes it so the fraction doesn't fluctuate each time.

I think you can implement this by calculating the fraction like this:
image
where you can change the binary fraction, primary mass limits and secondary mass lower limit. I will post code that implements in Slack that you can use if you think this is a useful change :)

**Additional context**
A comparison with the analytic value vs. 200 runs of the COMPAS code that calculates the same fraction with fbin=0.7, m1_min=5, m1_max=150, m2_min=0.1. I think this seems to be in agreement.
image

### LINKED ISSUE #970: Add analytical expression for average star-forming mass evolved per binary
We [sample from an inverse IMF CDF](https://github.com/TeamCOMPAS/COMPAS/blob/bc97c0c683977d9b29a4bc3df834b8871f9c30d6/compas_python_utils/cosmic_integration/ClassCOMPAS.py#L226) to compute the average star-forming mass evolved per binary. This can be done analytically and much faster (requiring less code!) 

@ilyamandel shared the following 
```
Mtot=(1/0.08)*(1/0.5)*(0.08^1.7-0.01^1.7)/1.7 + (1/0.5)*(0.5^0.7-0.08^0.7)/(0.7) + (-0.5^(-0.3))/(-0.3);
Ntot=(-(0.5)^(-1.3)+Mmax^(-1.3))/(-1.3)+(0.5^(-0.3)-0.08^(-0.3))/(-0.3)*(1/0.5)+(0.08^0.7-0.01^0.7)/0.7*(1/0.5)*(1/0.08);
Nint=(Mmax^(-1.3)-Mmin^(-1.3))/(-1.3);
Mtot*1.5/Nint
```
(where Mmax, Mmin are the COMPAS run's primary mass limits)

This does not account for the `compas_m2_min = 0.1 Msun`, and needs to be worked in to get the analytical function that we want.

TomWagg:
Here's is my code for calculating the expression using numerical integration rather than inverse CDF sampling. I didn't quite manage it in 4 lines like @ilyamandel in #970...but I'd argue that mine is tad easier to understand 😉

There are the usual parameters of the lower/upper limits on M1, lower limit on M2, binary fraction and all of the IMF parameters. I've additionally added the option to use a custom PDF for the mass ratio if you don't use one that is uniform which hopefully allays some of the worries from the original issue.

Fair warning - I've tested that this gives me the right fraction but I haven't actually run it on a COMPAS file as I didn't have one to hand and don't have COMPAS installed on this laptop (it is curretly 9:40pm so I'm not going to attempt that this evening haha!) - so please check I didn't do something dumb.

This will fix #378 and will close #970

p.s. It has been 979 days since I first brought this up in #378 - I sort of wish I'd waited 3 weeks more and made it beyond triple digits haha.


compas_python_utils/cosmic_integration/totalMassEvolvedPerZ.py
```diff
@@ -1,198 +1,137 @@
...
+    b1 = 1 / ( 
+                (m2**(1 - a12) - m1**(1 - a12)) / (1 - a12) \
+                + m2**(-(a12 - a23)) * (m3**(1 - a23) - m2**(1 - a23)) / (1 - a23) \
+                + m2**(-(a12 - a23)) * m3**(-(a23 - a34)) * (m4**(1 - a34) - m3**(1 - a34)) / (1 - a34)
+             )
+    b2 = b1 * m2**(-(a12 - a23))
+    b3 = b2 * m3**(-(a23 - a34))
+    
+    # evaluate IMF either at a point or for a list of points
+    if isinstance(m, float):
```
avivajpeyi: If we put a @np.vectorize decorator around this function I think you can skip the float VS array logic. 

I'd be happy to make this change if you're not sure what I'm talking about!

avivajpeyi:
There is also some duplicated/similar code at the bottom of the ClassCompas.py file -- is this doing the same thing? Maybe we should remove the duplicated code from there? 

But that can be a separate PR !

compas_python_utils/cosmic_integration/totalMassEvolvedPerZ.py
avivajpeyi: I made the suggested change

avivajpeyi:
I also added some tests for the IMF function -- does this look right? 
IMF


Additionally, @ilyamandel 's function is computing the star-forming mass per evolved binary without reading in COMPAS data -- not as a function of redshift.

compas_python_utils/cosmic_integration/totalMassEvolvedPerZ.py
TomWagg: This is magic, thanks for demonstrating this!

TomWagg:
> There is also some duplicated/similar code at the bottom of the ClassCompas.py file -- is this doing the same thing? Maybe we should remove the duplicated code from there?
> 
> But that can be a separate PR !

Ah right yes good point, this function could basically be a drop-in replacement for `find_star_forming_mass_per_binary_sampling` in the ClassCOMPAS file

TomWagg:
I think the IMF looks pretty good, though I'm used to seeing it in log-log

Yeah this code assumes that you have metallicity bins rather than smoothly sampled distributions, but the fraction function can be applied in both cases I think

avivajpeyi:
Added @ilyamandel's analytical function here as well (can be used for the cases when we don't need to load in the COMPAS data, and just have the m1, m2 limits) 

Using @ilyamandel's function in place of the IMF_sampling code 

@ilyamandel will you be able to review this PR?

ilyamandel (changes requested):
Thank you, @TomWagg and @avivajpeyi  .

My analytical function implicitly assumes that m1_min \geq 0.5 Msun (m3 in your notation); could you please add that to the comments?

Did you check that the two calculations agree?

Note that there is a conflict preventing a pull.

avivajpeyi:
Yes, I checked that the calculations agreed with the previous version in `ClassCompas` (which your version now replaces) [See here](https://github.com/TeamCOMPAS/COMPAS/issues/970#issuecomment-1569766888) 

I am not sure how to use @TomWagg's code to compute a  `star_forming_mass_per_binary(Mmax, Mmin, M2min...)` so I haven't added any tests comparing Ilya's analytical code with Toms' new code. @TomWagg could you add a function in `totalMassEvolvedPerZ.py` that does this, and maybe a test in `py_tests/test_total_mass_evolved_per_z.py` to ensure that the values are the same?

TomWagg:
I'm not _sure_ but I think this would just be done by taking the total of the initial masses in a COMPAS simulation, dividing by the fraction that I calculate with my function, and dividing by the number of binaries?

Does that sound right?

ilyamandel:
@TomWagg -- could you please walk me through the typical use case for your code?  How would I be expected to run it?

TomWagg:
Ah perhaps it is now obsolete @ilyamandel? What I would imagine is this (writing down a lot of things you already know but just to be thorough): You run some large COMPAS simulation/grid to work out the rate of occurrence of something. Since we don't usually sample the full parameter space (e.g. m1>5Msol), we need to account for the fact that you need more stellar mass to actually create each binary. So my function gives the ratio between the amount of mass in a COMPAS simulation, and the amount of mass you'd _actually_ need to create the binaries from a true IMF etc.

So the typical use case would be transforming the total mass of an initial COMPAS population and transforming it into a total star forming mass that you'd need to produce that population. You'd use the function to get the ratio between these two masses.

ilyamandel:
Sorry, @TomWagg , I should have asked for an example of how to call the function.   As you say, what we typically want is the amount of star forming mass that each simulated binary represents.  Can you walk me through how I should compute that (i.e., given an .h5 file from COMPAS, what calls should I make)?  I presume this is what @avivajpeyi was looking for in order to build a test, and what you were referring to here:

> this would just be done by taking the total of the initial masses in a COMPAS simulation, dividing by the fraction that I calculate with my function, and dividing by the number of binaries?

Aside: note that the test script is currently failing for this PR; can you check what the issue is?

avivajpeyi:
Hey Ilya -- the test that is failing is the placeholder test I've put in for toms' function (failing on purpose ATM)

ilyamandel:
Got it.  I believe @jeffriley is including this in a PR he is preparing for a binned (memory-efficient) version of FastCosmicIntegrator.

avivajpeyi:
I've removed the `numerical == analytical` check for now. 

I've added some placeholder code to create a mock COMPAS file with the datasets: 
```
    - Metallicity@ZAMS(1)
    - Mass@ZAMS(1)
    - Mass@ZAMS(2)
```


<details><summary>See here:</summary>
<p>

https://github.com/TomWagg/COMPAS/blob/2fe1a04255fe259d0144fd9bda5a35495fa633f0/py_tests/test_total_mass_evolved_per_z.py#L87

```python
def fake_compas_output(tmpdir)->str:
    """
    Create a fake COMPAS output file with a group 'BSE_System_Parameters' containing
    the following 1D datasets:
    - Metallicity@ZAMS(1)
    - Mass@ZAMS(1)
    - Mass@ZAMS(2)

    The values of these datasets should be from the IMF function.
    """
    compas_path = f"{tmpdir}/COMPAS.h5"
    n_systems = int(1e4)
    with h5.File(compas_path, "w") as f:
        f.create_group("BSE_System_Parameters")
        # TODO: Make the values of these datasets be from the IMF function
        f["BSE_System_Parameters"].create_dataset(
            "Metallicity@ZAMS(1)", data=np.linspace(0, 1, n_systems)
        )
        f["BSE_System_Parameters"].create_dataset(
            "Mass@ZAMS(1)", data=np.random.uniform(3, 300, n_systems)
        )
        f["BSE_System_Parameters"].create_dataset(
            "Mass@ZAMS(2)", data=np.random.uniform(3, 300, n_systems)
        )
    return compas_path
```


</p>
</details> 

If we can generate realstic mock `mass_1, mass_2, metallicity_at_zams` data (currently, I've made these some bogus values), then we can compare @TomWagg 's numerical version to the analytical version.

TomWagg:
Sorry for the delay @ilyamandel & @avivajpeyi (NASA ATP proposal is stealing away my time), I've added the example code to the PR now (just summing up Mass@ZAMS for 1 and 2 and then dividing by the number of binaries and the fraction I calculate) - there was a factor of `n_binaries` missing in the last commit I think.

The test still doesn't pass but I think this is just because the fake_compas_output function isn't sampling realistically yet. I changed some of the limits in the test and the values got closer so I'm hopeful that with some proper COMPAS output we'll match 🤞🏻

avivajpeyi:
Hey @ilya, @TomWagg, I added in the last bit we needed to compare Tom's numerical code against Ilya's analytical version. The test comparing the two versions is [here](https://github.com/TomWagg/COMPAS/blob/c95ef32c5e030906bee745036e24881687a6139e/py_tests/test_total_mass_evolved_per_z.py#L70). 



I've set the `nsamples=1e4` and the error tolerance ~ `+/- 1Msun`. 
We can crank up `nsamples` but the test will run slower. 

Plot comparing the numerical vs. analytical 'star_forming_mass_per_binary'  using different numbers of samples 
analytical_vs_numerical


**This is ready to be merged in!**

ilyamandel (approved):
Looks great, @avivajpeyi and @TomWagg !  Happy to approve, will leave it to you to decide whether to press the "Pull" button. :)

PR #974, opened 2023-06-04
Defect repair: Overloaded virtual warnings

jeffriley:
- Fixed "hides overloaded virtual function" warnings.
- Added "-Woverloaded-virtual" to compiler flags to enable warnings for g++ on linux systems.
- A little code cleanup

This should fix the current "hides overloaded virtual function" warnings.  None of these changes should affect COMPAS output.


ilyamandel (commented):
There are some questions for @solfreludio and @reinhold-willcox in COWD.cpp and HeWD.cpp ("// JR: Question: should this be set false if the condition is not satisfied?") -- it would be good to get answers from them.

ilyamandel (approved):

Issue #958, opened 2023-04-07
AllowedTimeExceeded?


ilyamandel:
When running COMPAS with populationDataPrinting on, I get messages like this:

Generating a new binary - 2745
Binary has masses 14.0044 & 8.80516 Msol
Binary has initial semiMajorAxis 0.124755 AU
RandomSeed 1680911027
2745: **Allowed time exceeded**: (Main_Sequence_>_0.7 -> Neutron_Star) + (Main_Sequence_>_0.7 -> Black_Hole)
Using supplied random seed 1680911028 for Binary Star id = 209722

Since we choose not to evolve double CO binaries, why is this yielding "Allowed time exceeded"?

jeffriley:
@ilyamandel  We don't evolve **_bound_** DCOs (well, the check is for DCO and BOUND).  By default we evolve unbound systems.  If you run this system with --evolve-unbound-systems FALSE, evolution stops when the system becomes unbound at 16.96 Myr, with a neutron star and a main sequence star.  If you add BINARY_PROPERTY::UNBOUND to the detailed output record and run with --evolve-unbound-systems TRUE (the default), you'll see that it evolves unbound from t = ~16.95 Myr through to ~14 Gyr.

Maybe we should indicate in the (console) output that the system evolved unbound, and possibly add the UNBOUND flag by default to the detailed output file?

EDIT: I'm still jet-lagged and tired... should we not evolve unbound systems that are comprised of two compact objects?  Even if that is the proper fix here, we should still indicate in the (console) output that the system evolved unbound (I think that should have been done when the code was added to evolve unbound systems, but it seems to have been missed).

ilyamandel:
@jeffriley -- I agree with your "EDIT".  The idea for evolving unbound systems was that we may want to follow through an unbound binary that's composed of, say, and NS and a MS star to see if the MS star will make an NS.  However, we should still stop evolution under conditions for which we stop evolving bound systems -- in particular, when both stars become compact objects.

ilyamandel:
Fixed by @jeffriley in #976 , closing.

PR #976, opened 2023-06-05
Fix for Issue 958

jeffriley:
Defect repair:
- Fix for issue #958 - evolving unbound systems that contain two compact objects.  Also added BINARY_PROPERTY::UNBOUND to BSE Detailed Output file default record.
- Changed makefile to be POSIX compliant for .o suffix rule.  No need to change docker Makefile - it is already POSIX compliant.
   - since GNU Make 4.3 a warning is issued for suffix rules that include prequisites - in our case the .o rule on line 125:
     "Makefile:125: warning: ignoring prerequisites on suffix rule definition" - and in future releases it will become an error.


ilyamandel (approved):
Thank you, @jeffriley !

Issue #972, opened 2023-06-01
Create a 'Binned-cosmic integrator' (avoid FastCosmicIntegrator's OOM error)


avivajpeyi:
As discussed in #968, the FastCosmicIntegrator's large vectors (used for faster-numpy calculations) lead to out-of-memory issues when dealing with more than ~50K DCOs. 

We need to 
- put a warning in fast cosmic integration saying not good for large datasets
- make a new version of FastComsicIntegration (binnedCosmicIntegrator) that returns the binned detection matrix -- doesn't bother with generating the formation rate array (avoid having the [Nbinary, Nredshifts] arrays that can be ~several gigs)

```
inputs
---------------

- COMPAS path
- Star Formation parameters {aSF, dSF, mu0, sigma0} 
- redshift-bins 
- chirp-mass bins

outputs
-----
detection-rate matrix: a 2d matrix with one value for each [redshifts bin, chirp mass bin]

```

avivajpeyi:
Completed with #979

PR #968, opened 2023-05-29
Fast cosmic integrator memory blow-up fix

avivajpeyi:
## Summary 
The FastCosmicIntegrator blows up in memory usage in three locations: 
- [x] the IMF sampling method in [`find_star_forming_mass_per_binary_sampling`](https://github.com/TeamCOMPAS/COMPAS/blob/bc97c0c683977d9b29a4bc3df834b8871f9c30d6/compas_python_utils/cosmic_integration/FastCosmicIntegration.py#L426)
- [ ] dPdlogz calculation in[ `find_metallicity_distribution`](https://github.com/TeamCOMPAS/COMPAS/blob/bc97c0c683977d9b29a4bc3df834b8871f9c30d6/compas_python_utils/cosmic_integration/FastCosmicIntegration.py#L128)
- [ ] In the [formation rate calculation](https://github.com/TeamCOMPAS/COMPAS/blob/bc97c0c683977d9b29a4bc3df834b8871f9c30d6/compas_python_utils/cosmic_integration/FastCosmicIntegration.py#L182)

I've attached a screenshot of some profiling info. 
Screenshot from 2023-05-29 18-36-07

This isn't an issue when dealing with COMPAS files with n-systems < 10M, but it requests large amounts of RAM (>50Gigs) when processing larger COMPAS files (causing the program to crash). 

 This PR will try to fix the memory issues in those three functions, re-working them in Cpp  (I suspect the issue involves multiple copies of the data). 
 

@jeffriley  -- how did you run FastCosmicIntegration with your 32M and 512M datasets?

 
----
## Memory profiles 

Using Jeff's 5M systems dataset

<details><summary>Mem profile of find_detection_rate</summary>
<p>

```
Line #    Mem usage    Increment  Occurrences   Line Contents
=============================================================
   314    172.8 MiB    172.8 MiB           1   @mem_profile
   315                                         def find_detection_rate(path, dco_type="BBH", merger_output_filename=None, 
   420    238.7 MiB     65.9 MiB           2       compas_data = COMPASData.from_hdf5(
   421    172.8 MiB      0.0 MiB           1           path=path, dco_type=dco_type,  weight_column=weight_column,
   422    172.8 MiB      0.0 MiB           1           merges_hubble_time=merges_hubble_time, pessimistic_CEE=pessimistic_CEE, no_RLOF_after_CEE=no_RLOF_after_CEE,
[...]

PR #979, opened 2023-06-16
Add binned integrator (with GPU support)

avivajpeyi:
As discussed in #968, #972, this adds a modified version of the cosmic-integrator (that can run on GPU/CPU), which bins the chirp-mass and redshift data while computing the detection-rate matrix (this will thereby avoid the multiple large arrays of [n_binary, redshift].

This also adds a test of the binned-cosmic integrator to ensure that it produces the same output if data is binned before/after running the cosmic integrator.

Finally, this adds some additional plotting code that can be used as diagnostics (the detection-rate plot, plot of the SNR sensitivity grid, and a plot of the SFR/metallicity distributions).  

## Visual test
Below I visually compare the binned integration outputs 
1. Binning after fast cosmic integration
binned_detection_matrix_plot
2. 'binned' cosmic integrator
binned_detection_matrix_plot_v2

The pytest also checks the values in the two grids are close w/ tol 1e-5

## Requirements
- Requires #971 to be merged in
- Requires @jeffriley 's binned-integrator to be merged in

Apologies for the miscommunication, @jeffriley, and @ilyamandel! I was unaware you were still working on adding your version of the cpp `FastCosmicIntegrator` in!


jeffriley:
@avivajpeyi @ilyamandel I just pushed my latest code for FastCosmicIntegration.py (I hope that's all I pushed...).

It is a monolithic file - it contains several classes.  We can separate them if required. I don't think this will merge with @avivajpeyi 's code easily.  I'm happy for you to use any, all, or none of my code.  It is documented - there is an explanation at the top of the file.

I have done some rudimentary tests and it seems to work as expected, but one of the many things I am not very good at is testing my own code, so it is very likely I have tested only the happy path...  Some further testing would be useful should you decide to use the code.

Let me know if you have any questions.

EDIT: @ilyamandel This code will produce the per-binary files according to your "Binary mergers: a common format" document, but it doesn't produce the yaml file.  I don't think this is the right vehicle for that.  I suspect it may just be better to let the user have COMPAS create a project-specific yaml file, and  then modify that to suit, but we can talk about that.

avivajpeyi:
This is awesome! I haven't been through the entire file -- but making my way through it. I think it would be great if we can add some tests for your code (I've managed to use set up some basic unit tests for my version of the comic integrator), and I think we can merge these (will require some work). Maybe we can meet up on a call to discuss how to proceed. Ill be back in the office next week.

avivajpeyi:
This is awesome! I haven't been through the entire file -- but making my way through it. I think it would be great if we can add some tests for your code (I've managed to use set up some basic unit tests for my version of the comic integrator), and I think we can merge these (will require some work). Maybe we can meet up on a call to discuss how to proceed. Ill be back in the office next week.

ilyamandel (approved):
For what it's worth, I am happy with the approach of having *both* a modified fast_cosmic_integrator as @TomWagg and @avivajpeyi have done (with a revised normalisation calculation, etc.) and @jeffriley's new binned_cosmic_integrator.  They provide a natural check on each other, have slightly different use cases, and appeal to users with different preferences.  Ultimately, we may deprecate one of them, but there's no obvious reason to merge them now.  So my approval refers to the entire PR.  

In fact, I'd be happy to pull this in now so people can start using these (which is the best way of testing them), and add further modifications/tests later, as separate PRs -- but will leave it to @jeffriley and @avivajpeyi to decide whether to press the "Pull" button. :)

jeffriley:
I'm ok to have two versions.  I'm not sure of all the modifications @avivajpeyi and @TomWagg have made.  Users can turn binning on or off with my version, and turn the per-binary .csv file on or off, so as far as the 'standard' behaviour goes they should be the same.  @avivajpeyi added support for gpus which I think is a good thing to have - my version doesn't have that.

So, I'll leave it to @avivajpeyi whether and when to pull this in.

avivajpeyi:
IMO i think we should return this to draft state and make some changes before merging this in.

#### 1. I think that having two versions of the same code will lead to confusion 😅 
The module I created `binned_cosmic_integrator` and Jeff's `FastCosmicIntegration.py` are doing mostly the same things.  Both have the ability to compute the detection matrix (with and without binning).  Both can store the rates to file. If we decide to merge this PR we should really try to do something about fixing this. 

#### 2. Deleting zombie-code

@jeffriley -- I noticed that both of us have some similar classes -- it seems like neither of us use `ClassCOMPAS` `ClassMSSFR`, etc. If that is the case, we should really delete these in this PR.

jeffriley:
> I think that having two versions of the same code will lead to confusion 😅

I think that's a fair comment - and it means doubling-up on maintenance.  Maybe @ilyamandel can expand further on his reasons for suggesting we keep both versions (maybe my version imposes OO too much...?)?

> Both can store the rates to file. 

But differently.  My version will only store the per-binary merger rates, and yields to (separate) csv files, and in the format specified in @ilyamandel 's document.  I think your version will store the rates arrays as they are produced (i.e. binned or per-binary), but not in Ilya's format.  That's easily fixed though - as I mention in my documentation, providing a small script that uses my version to create the arrays and then write them to a file, to the HDF5 file (per Lieke's code), plotting functionality, etc., is not hard at all.

> Deleting zombie-code ... 
> neither of us use ClassCOMPAS ClassMSSFR, etc. If that is the case, we should really delete these in this PR.

I have no problem deleting those with one caveat: I only provide the functionality in my version that I need for FastCosmicIntegration.  I think Coen provided some other functionality in those classes, and I don't know if people use that functionality.  If they do, we shouldn't delete those files unless/until we determine if the functionality is really needed, and if it is we provide it elsewhere.

ilyamandel:
Are ClassCOMPAS, etc.,  used in some of the Jupyter notebooks that are
provided as examples? (you can probably just do a search? not sure, since I
never use the notebooks)

avivajpeyi:
Some more things that we should consider

#### 3. Moving Jeff's `FastCosmicIntegrator.py` into the `compas_python_utils`

If we merge this in as is, this will result in there existing 3 `cosmic-integration` codes:
- Jeff's new `FastCosmicIntegrator.py` located in the repo root, 
- my `binned_cosmic_integrator` module, and 
- the old `FastCosmicIntegrator.py`

Leaving Jeff's new `FastCosmicIntegrator.py` file outside the `compas-python-utils`  module will prevent users from importing it without hackily adding it to their path... 

If we move it in, we can also run the basic tests we already have for the previous `FastCosmicIntegrator`. 
`pytest  py_tests/test_fast_cosmic_integration.py`




#### 4. Updating Examples and documentation
If we are going to get rid of the old `FastCosmicIntegrator` code with Jeff's version, we need to update the documentation and examples accordingly.

jeffriley:
> Are ClassCOMPAS, etc.,  used in some of the Jupyter notebooks that are provided as examples?
> (you can probably just do a search? not sure, since I never use the notebooks)

I think they are - I also never use notebooks.

avivajpeyi:
The CosmicIntegration notebook doesnt really work atm 😅 I remember trying them out when I started working on this stuff in ~Oct. 

IDK about the others.

jeffriley:
> Some more things that we should consider...

I'd be in favour of replacing the existing FastCosmicIntegration code with mine, except that as mine is, it doesn't provide all the functionality of the current FastCosmicIntegration code:

1. it can't be run from the command line
2. it doesn't provide plotting functionality
3. it doesn't provide Lieke's writing rates to HDF5 files
4. it doesn't write the rates arrays to file

All that is easily added, or provided as a wrapper script - but I won't have time to do it before I leave for Aspen

ilyamandel:
So that seems like a good reason to keep both for now, and then perhaps
streamline things in the future?

ilyamandel:
Consider moving the unnecessary classes into unsupported examples rather
than deleting them outright?

jeffriley:
> So that seems like a good reason to keep both for now, and then perhaps streamline things in the future?

Well, yeah, but really, adding a script to do that would take maybe an hour...  If I get a spare hour between now and Wednesday I'll do it.

jeffriley:
> Consider moving the unnecessary classes into unsupported examples rather than deleting them outright?

That's a good idea.  Just make sure we document what we do here so people know - maybe put an entry in "What's New"

avivajpeyi:
> Consider moving the unnecessary classes into unsupported examples rather than deleting them outright?

I personally think keeping unsupported zombie code is confusing... 

Instead, it sounds like the examples just need to be updated and run as a part of the CI. Then whoever makes some changes in the code will be responsible for fixing the examples etc to ensure that their changes do not result in the CI failing. 

I can set this up for the cosmic-integration code in this PR.

ilyamandel:
That sounds great!

avivajpeyi:
Cool! I've just updated the CosmicIntegrator example. 

I'll add the CI test for this shortly.

avivajpeyi:
Ok, I've updated the `CosmicIntegrator` example, and have set the CI to run the files in the `utils/examples` folder. 

I noticed that this file is not included in the online docs, but that maybe can wait for a future PR.

avivajpeyi:
Hey @ilyamandel  and @jeffriley,  it sounds like we'll include both  `binned_cosmic_integrator` and replace the current `FastCosmicIntegrator` with Jeff's `FastCosmicIntegrator` (with optimistic intentions of `merging` these in the near future)

As we want to have Jeff's `FastCosmicIntegrator` replace the current `FastCosmicIntegrator`, it sounds like there is some work that still needs to be done with Jeff's code (involving deleting the current versions of the files that Jeff has re-written, and I think Jeff wants to add some extra stuff as mentioned above).  

To prevent further bloating in this PR, would it be ok if we handle adding Jeff's file in a new PR?

jeffriley:
> To prevent further bloating in this PR, would it be ok if we handle adding Jeff's file in a new PR?

Ok by me.

avivajpeyi:
@jeffriley I also updated the tutorial as discussed [(see this)](https://compas--979.org.readthedocs.build/en/979/notebooks/CosmicIntegration.html)

The `py_tests/test_examples.py` allows us to run the example (and other examples in `utils/examples/` as a unit test to ensure it works. 

This is ready to be merged in, depending on your review :)

jeffriley (changes requested):
Ok, I'll get to this tomorrow probably.  This is just a placeholder.

jeffriley (approved):
I'm ok with this. One day we should probably rationalise the directory structure - e.g. the sampling (bootstrapping) module use cases aren't restricted to cosmic integration are they?  Yet the source file lives in a directory subordinate to the cosmic integration directory.  Perhaps it should be on the same level - CI & sampling are both tools and one doesn't depend on the other (does it? If it does it probably doesn't need to).

Thanks @avivajpeyi - good to go

PR #981, opened 2023-06-21
added che to the list of stellar types in the plotter

### LINKED ISSUE #980: Two massless remnants leading to errors in plotting script
The plotting script chokes on this example.  It shouldn’t — COMPAS runs just fine, though it does claim to produce two massless remnants.  (Whether that behaviour by COMPAS is problematic is, perhaps, a separate question? -- _Edit: in retrospect, just two PISNe following CHE, all good._)

COMPAS --initial-mass-1 80 --initial-mass-2 75 -a 0.1 -n 1 --detailed-output --random-seed 0 --metallicity 0.001

python3 plot_detailed_evolution.py COMPAS_Output/Detailed_Output/BSE_Detailed_Output_0.h5

reinhold-willcox:


ilyamandel (approved):
Looks good -- thank you!

PR #961, opened 2023-04-20
Updating defaults

ilyamandel:
Changing default settings as follows (see PR #957 ):

--allow-non-stripped-ECSN: FALSE (was true -- see Willcox+, 2021)
--evolve-unbound-systems: TRUE (was false -- this takes very little extra time and is often useful)
--wolf-rayet-multiplier: 0.1 (was 1.0 -- see, e.g., Stevenson & Clarke, 2022, showing that this is closest to more modern prescriptions; also Neijssel+, 2021, etc.)
--muller-mandel-kick-multiplier-NS: 520.00 (was 400 -- calibrated by Kapil+, 2023)
--remnant-mass-prescription: MULLERMANDEL (was FRYER2012 -- this is now well tested, see above)
--kick-magnitude-distribution: MULLERMANDEL (was MAXWELLIAN -- this is now well tested, see above)
--retain-core-mass-during-caseA-mass-transfer: TRUE (was false -- see Romero-Shaw+, 2023)

PR #982, opened 2023-06-22
Enable DWD evolution and change Evolve Unbound default

nrsegovia:
I have added the option to evolve DWDs by setting `--evolve-double-white-dwarfs TRUE`. The default is `FALSE`, so it should not affect runs where DWDs are not relevant. The documentation has been updated accordingly.

I also spotted a difference between the documentation and the code:

> --evolve-unbound-systems
> Continue evolving stars even if the binary is disrupted.
> Default = FALSE

But the code had `m_EvolveUnboundSystems = true` as default, so I have changed it to `false` to match the documentation.


jeffriley (changes requested):
I haven't actually reviewed the code changes yet, but I have a few quick questions/comments:

1. I don't see any changes to the documentation - am I missing something?
2. changelog needs to be updated, as does the version number
3. if we're changing defaults in the code, don't we also need to generate a new yaml file?

nrsegovia:
> I haven't actually reviewed the code changes yet, but I have a few quick questions/comments:
> 
>     1. I don't see any changes to the documentation - am I missing something?
Sorry, I had reverted some changes and forgot to add the documentation after.
>     2. changelog needs to be updated, as does the version number
Done.
>     3. if we're changing defaults in the code, don't we also need to generate a new yaml file?
I have not edited the yaml file before... are we talking about the yaml files in `src`?

jeffriley:
> I have not edited the yaml file before... are we talking about the yaml files in src?

The default yaml file is in compas_python_utils/preprocessing/compasConficDefault.yaml.  You can either edit that, or have COMPAS create a new yaml file using the '--create-YAML-file' option.  See documentation here:
https://compas.readthedocs.io/en/latest/pages/User%20guide/Pre-processing/pre-processing-yaml.html.

Actually, since you are changing the default behaviour of COMPAS, you should add an entry to the 'What's New' documentation page to document the change in behaviour there - at least then we've done our best to let users know something has changed.

nrsegovia:
> The default yaml file is in compas_python_utils/preprocessing/compasConficDefault.yaml. You can either edit that, or have COMPAS create a new yaml file using the '--create-YAML-file' option. See documentation here: https://compas.readthedocs.io/en/latest/pages/User%20guide/Pre-processing/pre-processing-yaml.html.
> 
Thanks !

> Actually, since you are changing the default behaviour of COMPAS, you should add an entry to the 'What's New' documentation page to document the change in behaviour there - at least then we've done our best to let users know something has changed.

I think it would be better to revert the change in the code and instead change the documentation, so people already working with the defaults in the yaml file won't be affected. Usually I do not use `runSubmit.py` so I was not expecting unbound systems in my simulations, according to what the current documentation shows.

ilyamandel:
Just a quick comment on this:

> I also spotted a difference between the documentation and the code:
> 
> --evolve-unbound-systems
> Continue evolving stars even if the binary is disrupted.
> Default = FALSE
> 
> But the code had m_EvolveUnboundSystems = true as default, so I have changed it to false to match the documentation.
> 

The intention was to have the new default be true: https://github.com/TeamCOMPAS/COMPAS/pull/961 .  It appears that I hadn't updated the documentation in that PR, so the fault is mine -- I apologise.  But the right behaviour is to keep that true and update documentation accordingly, as you've now done, thank you!

(In fact, the documentation should eventually be updated for all changes from that PR, but I'll add that as a separate issue.)

ilyamandel (approved):
Looks great -- thank you, @solfreludio !

ilyamandel (changes requested):
But please update the default .yaml file to include the new option, following Jeff's recommendation:

The default yaml file is in compas_python_utils/preprocessing/compasConficDefault.yaml. You can either edit that, or have COMPAS create a new yaml file using the '--create-YAML-file' option. See documentation here:
https://compas.readthedocs.io/en/latest/pages/User%20guide/Pre-processing/pre-processing-yaml.html.

jeffriley:
@solfreludio Since you are adding a new option, make sure you have followed the instructions at the start of the options.cpp file for adding new options (in particular, I don't think you have followed step 13 - but I haven't checked all steps)

nrsegovia:
I have gone through the instructions in `Options.cpp` and skipped steps 8, 9, 11, and 12 as I think they do not apply for this option. Let me know if I'm mistaken. The yaml file has been updated as well.

ilyamandel (approved):
Thank you, @solfreludio !  The changes look good to me. @jeffriley?

jeffriley (approved):
All good.  Thanks @solfreludio

PR #986, opened 2023-07-04
Add 'Evolution_Status' column to system parameters file (SSE and BSE)

jeffriley:
Enhancement, a little code cleanup:
 - Record and expose m_EvolutionStatus for both BaseStar and BaseBinaryStar as a variable available for selection for printing.  m_EvolutionStatus records the final evolution status - the reason evolution was stopped.  This was already printed to the console for each star or binary, and is now available to be recorded in the log files.
 - Add 'Evolution_Status' column to both SSE and BSE default system parameters records, and record m_EvolutionStatus there. 
 - Fixed a few typos, a little code cleanup.


jeffriley:
@ilyamandel when you're back from vacation

ilyamandel (commented):
It seems odd that a binary can have an error in either SSE or BSE, but a single star does not have an error as a possible status.  If binaries can have SSE errors as a status, does that not mean that single stars may report an error as the reason for termination -- and if so, should that not be captured in the documentation?

           * - SSE error for one of the constituent stars 
             - = 5
           * - Error evolving binary 
             - = 6

ilyamandel (commented):
Options.cpp, line 4161:  I see that you changed -1 to -2.  If the value matters, should this constant be defined somewhere?  

status = -2; // set error status

ilyamandel (approved):
@jeffriley -- looks great, modulo the two queries above.  Since both are minor, I'll approve, but will leave the merging to you in case you decide to tweak anything in response.

jeffriley:
@ilyamandel 

> It seems odd that a binary can have an error in either SSE or BSE, but a single star does not have an error as a possible status.
> If binaries can have SSE errors as a status, does that not mean that single stars may report an error as the reason for
> termination -- and if so, should that not be captured in the documentation?

I'm not sure if I understand.  There is a table for evolution status in 'standard-logfiles-record-specification-stellar.rst' (one of the files modified by this PR).  

In stellar evolution (SSE mode), if the star encounters an error during evolution that causes evolution to stop, that error is reported by the SSE code, hence the table in 'standard-logfiles-record-specification-stellar.rst'.

In binary evolution, if one of the constituent stars encounters an error during evolution that causes evolution to stop, that error is reported to the BSE code, and the BSE code reports the error, hence the table in 'standard-logfiles-record-specification-binary.rst'.

Am I misunderstanding your question?


> Options.cpp, line 4161: I see that you changed -1 to -2. If the value matters, should this constant be defined somewhere?

Oops.  Yes.  I'll fix that.

ilyamandel:
@jeffriley : in the ***-stellar.rst file, there is no general error report option, that's exactly my point:

           * - Simulation completed 
             - = 1
           * - Evolution stopped because an error occurred 
             - = 2
           * - Allowed time exceeded 
             - = 3
           * - Allowed timesteps exceeded 
             - = 4

jeffriley:
> ... in the ***-stellar.rst file, there is no general error report option, that's exactly my point:

Ohhh - you want the actual SSE error to be reported by the BSE code?  I don't think that's relevant - at least not in the context I think you mean.  The errors reported in the table in 'standard-logfiles-record-specification-stellar.rst' refer to SSE - that is, the evolution of a single star.  Those errors aren't relevant to BSE, because BSE doesn't actually evolve a single star to completion - it evolves a single star for a single timestep - so the time/timesteps exceeded errors aren't relevant, and neither is the 'simulation completed' status.  The only SSE error reported that is valid is the general error: 'an error occurred'.

One day (one day...) I plan to fix the error handling - right now it's not great.  We should really bubble-up any errors/warnings and let the calling code decide what to do with it - that way we could have much more granular (and informative) errors and warnings, and a sensible recovery process.  That's on my list, but now now...

Does that answer the question?

ilyamandel:
> The only SSE error reported that is valid is the general error: 'an error occurred'.

I was just wondering why BSE can report this general error, as well as a general BSE error, but SSE does not.  But never mind, if you are happy, I am happy. :-)

jeffriley:
Yeah, I think I'm missing what you mean.  Both have a general error:

"Evolution stopped because an error occurred " for SSE, and

"Error evolving binary" for BSE.

You might have to spell out what you mean - what other general error should SSE have? Maybe I'm still jet-lagged (or I'm just thick...) :-)

ilyamandel:
In the ...binary.rst file, I see 

           * - SSE error for one of the constituent stars 
             - = 5
           * - Error evolving binary 
             - = 6
             
In the ...stellar.rst file, I don't see any generic errors. 
         
I was querying the asymmetry.

jeffriley:
Oh.  The ...stellar.rst file has:

       * - Simulation completed 
         - = 1
       * - Evolution stopped because an error occurred 
         - = 2
       * - Allowed time exceeded 
         - = 3
       * - Allowed timesteps exceeded 
         - = 4

I see "Evolution stopped because an error occurred" as a generic error.  Maybe I should change the string to "Error evolving star".

ilyamandel:
Sorry, @jeffriley, I am just blind, all is well! :)

PR #993, opened 2023-08-27
readthedocs config changes

jeffriley:
Made a few small config changes - hoping that removing the pdf build might obviate the need for latex to get involved in the docs build...


ilyamandel (approved):
Looks good, @jeffriley !  Let's give it a try! :)

PR #996, opened 2023-09-27
Updated wind perscriptions 

jmerritt1:
Added a new `UPDATED` option to the mass-loss prescriptions. This is a wrapper applying new specific prescriptions on a case-by-case basis. These are selected via new program options for `OB`, `VMS`, `RSG`, and `WR-mass-loss`, which each have their own wrapper function selecting from a total of 15 new perscriptions. Previous default behavior is recoverable by specifying `VINK`. 

Commit history is in the updated_winds branch, which became relatively unmergable. The changes have been collected and added to an up-to-date branch here as Jeff suggested.


SimonStevenson:
Two things that will need updating:

- [x] Update `changelog.h` with description of changes, and bump version
- [x] Update documentation for new options

jmerritt1:
note in Options.cpp, that we didn't use, `(" + AllowedOptionValuesFormatted("` and instead list the options as before. This convention must have been changed in the last few months.  See https://github.com/TeamCOMPAS/COMPAS/blob/fe48a6b91f400ad0fb1803b37f348b98defe603a/src/Options.cpp#L1736 for example.

jeffriley:
> note in Options.cpp, that we didn't use, `(" + AllowedOptionValuesFormatted("` and instead list the options as before. This convention must have been changed in the last few months. See
> 
> https://github.com/TeamCOMPAS/COMPAS/blob/fe48a6b91f400ad0fb1803b37f348b98defe603a/src/Options.cpp#L1736
> 
> for example.

Yeah, that changed a while ago - we should use the new method in this PR (not difficult)

jeffriley:
> missed one new RSG option on ln 1736, but should change to AllowedOptionValuesFormatted() anyway?

Yes - that way the help text is automatically updated if any of the allowed values are changed - and the help text format is kept uniform with other options.

jmerritt1:
> > missed one new RSG option on ln 1736, but should change to AllowedOptionValuesFormatted() anyway?
> 
> Yes - that way the help text is automatically updated if any of the allowed values are changed - and the help text format is kept uniform with other options.

Thanks, one less spot to change options in the future. I've updated these.

ilyamandel:
Sorry, I am offline until October 8, so won't be able to review this -- defer to @jeffriley and @SimonStevenson !

jmerritt1:
> Two things that will need updating:
> 
> * [x]  Update `changelog.h` with description of changes, and bump version
> * [x]  Update documentation for new options

in `program-options-list-defaults.rst`, it's formatting some of the options lists with the bracket on the next line, but I'm not seeing why?

jeffriley:
> bumped version to 02.39.02

Let's make it 02.40.00 - this is new functionality, not a defect repair

xx.yy.zz -> xx is (supposed to be) major new release; yy minor new release/new functionality; zz defect repair

jmerritt1:
> > bumped version to 02.39.02
> 
> Let's make it 02.40.00 - this is new functionality, not a defect repair
> 
> xx.yy.zz -> xx is (supposed to be) major new release; yy minor new release/new functionality; zz defect repair

Should I list all the specific perscriptions in the changelog or are just the new program options fine?

jeffriley (changes requested):
Thanks @jmerritt1.  I have a few comments and suggested changes.  I'll number them so we can refer to them by number if necessary:

1. `prescription` is mis-typed as `perscription` in a few places:

    `program-options-list-defaults.rst`
    `BaseStar.cpp`
    `changelog.h`
    `constants.h`
    `Options.h`

2. Should we implement an option to set the `VERY_MASSIVE_MINIMUM_MASS` threshold, or is 100Msol either the universally accepted threshold or required for our implementation?

3. If you run your version of the executable with `./compas --help`, you will see:

    Options:
    ...
    --OB-mass-loss
       OB mass loss prescription (Options: [], default = 'NONE)
    ...
 
    Note no options listed in `[]` - that's because you missed step 9 in the instructions at the top of Options.cpp

    Also note that there is a missing single quote after `'NONE` - that needs to be added in Options::AddOptions():

    Change 

            `("OB mass loss prescription (" + AllowedOptionValuesFormatted("OB-mass-loss") + ", default = '" + p_Options->m_OBMassLoss.typeString + ")").c_str()`

    to

           `("OB mass loss prescription (" + AllowedOptionValuesFormatted("OB-mass-loss") + ", default = '" + p_Options->m_OBMassLoss.typeString + "')").c_str()`

    and similarly for `RSG-mass-loss`, `VMS-mass-loss`, and `WR-mass-loss`

4. If you run your version of the executable with `./compas --create-yaml-file myyaml`, and look at the resultant yaml file (`myyaml`).  At the bottom of the file you'll see:

        ### Additional COMPAS options not found in YAML template ###
        #    --RSG-mass-loss: 'NJ90'                                               # Default: 'NJ90'
        #    --VMS-mass-loss: 'NONE'                                               # Default: 'NONE'
        #    --WR-mass-loss: 'BELCZYNSKI2010'                                      # Default: 'BELCZYNSKI2010'
        #    --metallicity-distribution: 'ZSOLAR'                                  # Default: 'ZSOLAR'          Options: ['LOGUNIFORM','ZSOLAR']

    That's because your changes to `yaml.h`:

            "    --OB-mass-loss: 'NONE'                                                # Default: 'NONE'"
            "    --RSG-mass-loss: 'NJ90'                                               # Default: 'NJ90'"
            "    --VMS-mass-loss: 'NONE'                                               # Default: 'NONE'"
            "    --WR-mass-loss: 'BELCZYNSKI2010'                                      # Default: 'BELCZYNSKI2010'"

    are malformed. You need to remove the default values in both places.  e.g. change

            "    --OB-mass-loss: 'NONE'                                                # Default: 'NONE'"

    to

            "    --OB-mass-loss"

    and similarly for `RSG-mass-loss`, `VMS-mass-loss`, and `WR-mass-loss` (see examples above and below your additions in `yaml.h`).

5. `BaseStar::CalculateMassLossRateOBBjorklund2022()`:
    - function name in description does not match actual function name
    - is it worth noting the caveats documented in the description (i.e. "This prescription is calibrated to the following ranges:") in the online documentation for the relevant option?  Is it important a user knows that (particularly a user who may just use COMPAS without looking at the code)?

6. `BaseStar::CalculateMassLossRateOBVinkSander2021()`:
    - function name in description does not match actual function name
    - function has commented code that should be removed
    - equations at lines 2015-2020, 2030-2035, and 2045-2051 have ragged right edges with math operators.  Either have a single space between the end of the text and the operator, or line all operators up (like the equation at 1967-1973) - the way it is makes it too hard to read.
    - it's easier to read if the comments at the end of code lines are aligned with each other too (for humans maintaining the code - the compiler doesn't care of course...)

7. `BaseStar::CalculateMassLossRateOBKrticka2018()`:
    - function name in description does not match actual function name
    - `log10(m_Metallicity / ZSOL)` is `BaseStar::LogMetallicityXi()` - `log10()` is expensive, no need to call it if we don't have to (and twice...).

8. `BaseStar::CalculateMassLossRateRSGBeasor2020()` has commented code that should be removed.

9. `BaseStar::CalculateMassLossRateRSGDecin2023()`:
    - I would probably just rewrite this as:

           return PPOW(10.0, -20.63 - 0.16 * m_MZAMS + 3.47 * log10(m_Luminosity));

      but, can't we rewrite that as:

           return PPOW(m_Luminosity, -20.63 - 0.16 * m_MZAMS + 3.47);

      I think the latter would use fewer compute cycles - the former seems to do an unnecessary `log10()` operation, which isn't cheap.

10. `BaseStar::CalculateMassLossRateRSGKee2021()`:
    - has commented code that should be removed.
    - also, change  `2.` -> `2.0`,  `17000. -> 17000.0`, and `60000. -> 60000.0` (for the same reason as above - easier for humans to read (esp. at a glance))

11. `BaseStar::CalculateMassLossRateRSGVinkSabhahit2023()`:
    - change `-8.` -> `-8.0`, and `-24.` -> `-24.0` (same as above)

12. `BaseStar::CalculateMassLossRateVMSBestenlehner2020()`:
    - change `1.` -> `1.0` (same as above)
    - change constants `1` and `2` to `1.0` and `2.0`. c++ is a strongly-typed language: "1" is an integer, whereas "1.0" is a double. It may make no difference now, but it might if someone changes things later - it's good practice to use the correct data type even for constants.

13. `BaseStar::CalculateMassLossRateVMSVink2011()`:
    - reference/link to paper in the description (is 'Vink 2011' sufficient)?
    - function should be `const`
    - if the `else` clause is executed, `rate` is calculated, and is recalculated after the `if/else` using `logMdotdiff` which is undefined (only defined in the `if` clause is executed).  `teff` and `rate2001` only need to be calculated in the `else` clause, so those calculations should be moved into the `else` clause.  Bring the final calculation of `rate` up into the `if` clause (it requires `logMdotdiff`).

14. `BaseStar::CalculateMassLossRateVMSSabhahit2023()`:
    - reference/link to paper in the description  (is 'Sabhahit 2023' sufficient)?
    - function should be `const`
    - resolve and remove question on line 2260

15. `BaseStar::CalculateMassLossRateWolfRayetSanderVink2020()`:
    - change the conditional on line 2425 to use utils::Compare()
    - Mdot has already been initialised to `0.0` (at line 2412), so no need to set it `0.0` at line 2426. Instead, change the conditional to
        `if (utils::Compare(logL0, logL) < 0)` and delete lines 2426 and 2427.
    - move line 2434 into the `if` - no need to multiply by the option value if Mdot is `0.0` - just wastes compute cycles.
    - change `- 1` to `- 1.0` on line 2430

16. `BaseStar::CalculateMassLossRateVMS()` has commented code that should be removed

17. `BaseStar::CalculateMassLossRateWolfRayetTemperatureCorrectionSander2023()`:
    - change `141E3` to `141.0E3` on line 2457 (same as above re data type)
    - change `100E3` to `100.0E3` on line 2458 (same as above re data type) 
    - change the conditional on line 2465 to use Utils::Compare()

18. `BaseStar::CalculateMassLossRateHeliumStarVink2017()`:  I think I'd just use a single return line, but up to you.

19. `BaseStar::CalculateMassLossRateHeliumStarShenar2019()` has commented code that should be removed

20. `BaseStar::CalculateMassLossRateOB()`: parameter should be `const`

21. `BaseStar::CalculateMassLossRateRSG()`: parameter should be `const`; function should be `const`

22. `BaseStar::CalculateMassLossRateVMS()`: function should be `const`

23. `BaseStar::CalculateMassLossRateUpdatedPrescription()`:
    - "The structure is similar to above" - do you mean similar to `BaseStar::CalculateMassLossRateVink()`?  If so, say so (it may not be above if the code gets changed/refactored).

24. Add an entry to the `What's New` page of the online documentation to describe any changes to default behaviour (e.g. which mass loss prescription is now default?  Any other changes to default behavour?)

25.  Your question: "Should I list all the specific perscriptions in the changelog or are just the new program options fine?"
    - Just the new options are fine for changelog, but you should add a more complete descripion in the 'What's New' entry (see above).
   
26. Given the number of times we use `log10(m_Luminsoity)`, I wonder if it's worth calculating that when `m_Luminosity` is calculated, and storing it in a class member variable (call it `m_Log10Luminosity` - similar to `m_Log10Metallicity`)?

That's about it - there are some minor formatting changes I'll either point out later or just fix myself at the final review.

I have not checked the functionality - I checked that the code runs and data is produced, but I have not checked if the contents of the data files are correct according to the new prescriptions added, or whether any existing functionality has been negatively impacted (wrt the correctness of the data produced).  I assume you have satisfied yourself that the data produced is correct.

jeffriley:
And...

> in program-options-list-defaults.rst, it's formatting some of the options lists with the bracket on the next line, but I'm not seeing why?

I think that's just the way Read the Docs formats - the line break just happens to be there.  You could play around with the text to shift the line break away from the bracket if it really bothers you.

jeffriley:
One more change:

Change

    constexpr double LSOLW = 4E26;

in constants.h to

    constexpr double LSOLW = 4.0E26;

jeffriley:
Wait...

>but, can't we rewrite that as:
>
>     return PPOW(m_Luminosity, -20.63 - 0.16 * m_MZAMS + 3.47);

No, that's wrong (I mentally put parentheses around `-20.63 - 0.16 * m_MZAMS + 3.47`).  Can we still avoid the log10?  Maybe...

jmerritt1:
Thanks for the very detailed and helpful suggestions.

Question about p_Teff vs. m_Temperature: most T dependent functions seem to be marked const and use m_Temperature, and then convert to K which isn't costly, or others (just Vink2001 before our changes, I believe) have not been marked const and have used p_Teff (then all further wrappers employing those functions also can't be marked const).  It would be nice to use one or the other throughout? Currently I've got everything using only m_Temperature. 

Additionally, anything that sets m_DominantMassLossRate can't be marked const either, which means Vink 2001, 2021, and anything that uses it (Vink2011, Sabhahit2023, most of the wrappers, etc.) It is my preference to only set this flag in higher wrapper functions, if there aren't any objections?

jmerritt1:
By number, here is where all the suggested changes stand in the latest commit:

1. done
2. skipped for now. We discussed different criteria involving gamma, L, M, etc. Not sure where this stands. 100 Msol is certainly not universally accepted. Adding an option may make sense. 
3. skipped for now (Sorry, I'm a bit lost here. What lines in Options.cpp?) (added the quote). 
4. done
5. done (Not sure where or if to add calibration range comment, only program options are in the online docs, not a description of each prescription)
6. done
7. done
8. done
9. done
10. done
11. done
12. done
13. Can't be marked const, uses Vink 2001. And rate2001 needs to be calculated before both clauses.
14. can't be marked const, uses Vink 2001.
15. Could use a double check to make sure I don't have utils::Compare backwards...
16. done
17. Could use a double check to make sure I don't have utils::Compare backwards...
18. done
19. Removed debugging code but left comments and constants, Can let Simon decide on this one. 
20. Marked parameter const, func can't be marked const, sets m_DominantMassLossRate. Employs Vink2001...
21.  Marked parameter const, func can't be marked const, sets m_DominantMassLossRate.
22.  Marked parameter const, func can't be marked const, uses functions that use Vink2001...
23. done, but could rename Vink, maybe to Belczynski? While we're at it we could also use a different name than Updated for the new defaults? 
24. skipped for now. I'm not sure that our current defaults are appropriate. OB and VMS should definitely not be NONE. I don't think we should use NJ90 as the default for RSG, however this was Andreas Sander's suggestion. 
25. skipped, will do.
26. Not sure yet, but this may be prudent?

jeffriley:
Thanks JD.


> 3. skipped for now (Sorry, I'm a bit lost here. What lines in Options.cpp?) (added the quote).

There are instructions for adding new options at the top of the file Options.cpp.  You missed step 9 (line 56 of Options.cpp).


> 13. Can't be marked const, uses Vink 2001. 
> 14. Can't be marked const, uses Vink 2001.

My bad, sorry - I didn't look deep enough.


> 15. Could use a double check to make sure I don't have utils::Compare backwards...
> 17. Could use a double check to make sure I don't have utils::Compare backwards...

Edited: sorry, misunderstood your comment :-) 

For (15) I probably wasn't clear (and I should have used `<=`).  I meant change this:

        if (logL < logL0){
            Mdot = 0.0;         // No mass loss for L < L0
        }
        else{
            // Equation 13 in Sander & Vink 2020
            double logMdot = alpha * log10(logL - logL0) + 0.75 * (logL - logL0 - 1) + logMdot10;
            Mdot = PPOW(10.0, logMdot);
        }

to this:

        if (utils::Compare(logL0, logL) <= 0) {
            // Equation 13 in Sander & Vink 2020
            double logMdot = alpha * log10(logL - logL0) + 0.75 * (logL - logL0 - 1.0) + logMdot10;
            Mdot = PPOW(10.0, logMdot);
        }

You don't need to set `Mdot = 0.0` in the `if` clause - it is already initialised to 0.0.  So reverse the condition and delete the `if` clause - you only need what was originally the `else` clause.

I think you've used utils::Compare() correctly for (16).


> 19. Removed debugging code but left comments and constants, Can let Simon decide on this one.

We shouldn't have commented code in the production codebase.  There are a couple of places we do for the BE binaries code that we removed, but that should also be fixed (it's on my list...).  We shouldn't be adding new commented code.

We should either (a) move this code:

    // Fitting constants - see Table 5 in Shenar et al. 2019
    // For H-rich WR stars
    // const double C1 = -6.26;
    // const double C2 =  0.66;
    // const double C3 = -0.11;
    // const double C4 =  1.16;
    // const double C5 =  0.81;

into the description of the function, and explain what it is for, or (b) uncomment it and add a new option to enable it.

Also, do we need to resolve the question here:

    double XHe = 1.0; // What to use for this?


(b) is probably preferable, but we (a) will also work.

The question is, what purpose does it serve being commented in the code - is it because you might want to implement it later?  Or do you just want to guide future developers?  If the former, and you don't want to implement it now, or  the latter, moving the commented code to the description (and explaining what it is for) will work.


> 20. Marked parameter const, func can't be marked const, sets m_DominantMassLossRate.

Was that there before?  I swear I checked for that - I must be hallucinating...

> 20. ... Employs Vink2001...

Yes, my bad again - I didn't look deep enough.


> 21. Marked parameter const, func can't be marked const, sets m_DominantMassLossRate.

Are you sure?  I can't see it.  I checked the functions it calls, and I can't see where any of them set m_DominantMassLossRate.  Maybe one or more of them should, but I can't see it - am I just missing it?


> 22. Marked parameter const, func can't be marked const, uses functions that use Vink2001...

Indeed it does - my bad again.


> 23. done, but could rename Vink, maybe to Belczynski? While we're at it we could also use a different name than Updated for the new defaults?

I agree we should use a name other than "UPDATED" (chatted to Simon about this earlier).

> ... could rename Vink, maybe to Belczynski?

Also chatted to Simon about this.  I think that works if it's not likely to confuse existing users/developers.  If it's just an issue for developers, we could clearly document in the description of the function what its former name was and why we changed it.


> 26. Not sure yet, but this may be prudent?

I think it would be worth doing - if you don't get to it I may just do it in my next PR.

jeffriley:
> Question about p_Teff vs. m_Temperature: most T dependent functions seem to be marked const and use m_Temperature,
>and then convert to K which isn't costly, or others (just Vink2001 before our changes, I believe) have not been marked const
>and have used p_Teff (then all further wrappers employing those functions also can't be marked const). It would be nice to 
> use one or the other throughout? Currently I've got everything using only m_Temperature.

I haven't looked, so am just responding to your question. 

First, `m_Temperature` vs `p_Teff`.  Our convention is to use `m_` to prefix class member variables, `p_` to prefix function parameters, and no prefix for local variables.  That way, if you are reading the code you can tell without having to search whether a variable is a class member variable, a parameter passed in to the function, or a local variable.

A function declared as `const` is effectively a "read only" function, and cannot modify the object upon which it is called (so calling a `const` class member function can't modify the class).  Using `m_Temperature` in a function doesn't preclude the function from being a `const` function - setting it, or any other class member variable, does.

So I'm unsure what you are asking.  If a function just uses `m_Temperature` to calculate a local variable in K, that doesn't mean the function can't be declared as `const` (or conversely must be declared `const`).  Mostly we would pass (say) `p_Teff` to a function instead of letting the function use `m_Temperature` because the function is, or needs to be, a `static` function (`static` functions don't have access to class member variables - to read or write).  \EDIT: The other reason to pass `p_Teff` as a parameter rather than have the function use 'm_Temperature` is if you might sometimes want the function to use a value for temperature that differs from `m_Temperature`.  \END EDIT

EDIT:
> It would be nice to use one or the other throughout?

I don't think interface consistency really matters here.  If the function should be using the class member variable for temperature, then it should use `m_Temperature` - there's no value in using (say) `p_Teff` as a parameter and having the calling function pass `m_Temperature` as the parameter (indeed, that allows the possibility that the caller may pass the wrong value when we know we want to use `m_Temperature`).  If the function needs to use a value for temperature that may differ from `m_Temperature`, the use (say) `p_Teff` as a parameter and have the calling function pass whatever value is required.

\END EDIT

> Additionally, anything that sets m_DominantMassLossRate can't be marked const either, which means Vink 2001, 2021, and
> anything that uses it (Vink2011, Sabhahit2023, most of the wrappers, etc.) It is my preference to only set this flag in higher
> wrapper functions, if there aren't any objections?

No objections from me.

SimonStevenson:
I just pushed a change that removes the commented text in the Shenar+19 WR mass-loss prescription. The issue was that the prescription depends on the surface hydrogen/helium abundances, which we don't really have access to in COMPAS. I've changed the implementation so that in BaseStar.cpp we use the H-rich fit, and in HeMS.cpp we use the H-poor fit (appropriate for helium stars). Hopefully this is ok.

jmerritt1:
> I agree we should use a name other than "UPDATED" (chatted to Simon about this earlier).
> 
> > ... could rename Vink, maybe to Belczynski?
> 
> Also chatted to Simon about this. I think that works if it's not likely to confuse existing users/developers. If it's just an issue for developers, we could clearly document in the description of the function what its former name was and why we changed it.

Our new idea for the naming is `BELCZYNSKI2010` in place of `VINK`/Default, and `FLEXIBLE2023` in place of `UPDATED`. How does this sound? We can also abbreviate the year everywhere if that looks better, like `BELCZYNSKI10`, etc.

SimonStevenson:
The motivation for 'flexible' is that JD added the ability to specify
different prescriptions for different stellar phases of massive star
evolution independently (like RSG, WR etc). This is different from the old
prescription that sets  prescriptions for several different phases.
Anything more descriptive, like 'phase based' seemed to long winded for an
option.

jeffriley:
> Our new idea for the naming is BELCZYNSKI2010 in place of VINK/Default, and FLEXIBLE2023 in place of UPDATED. 
> How does this sound? We can also abbreviate the year everywhere if that looks better, like BELCZYNSKI10, etc.

Sounds good.  I wouldn't abbreviate the year - I think 4-digit years are better.

jmerritt1:
I think we should be ready to go? Here are the outstanding suggestions:

-2. For now we can keep this 100Msun threshold hardcoded?
-23. Changed the naming to `BELCZYNSKI2010` and `FLEXIBLE2023`
-24. changed whats-new.rst
-25. done. 
-26. Have not added a logL variable, but I agree this may be a good idea in the future. 

Thanks again for all the help.

jeffriley:
Thanks JD - I'll get to this in the next few days.

jeffriley (approved):
@jmerritt1 @SimonStevenson 

Thanks for the changes.  I made a few small mods and did a bit of tidy-up (some on code unrelated to this PR).

It's failing the spell-check and one of the pytests at the moment - I've asked @avivajpeyi to look at that.  Once those tests pass I'm happy for you to merge into dev if you'r happy with the tweaks I made.

SimonStevenson:
Yeah we looked at the pytest thing before, its failing on YHe, claiming it should be 'the'. We need to add YHe to the dictionary of things the spell check ignores.

jeffriley:
I think `YHe` is in the dictionary.  Or maybe that has to be done first in a separate PR?  Hopefuly @avivajpeyi can tell us how to fix it.

avivajpeyi:
The build job is failing atm due to https://github.com/TeamCOMPAS/COMPAS/actions/runs/6661986147/job/18106373887?pr=996#step:6:353

```
Run example COMPAS job using the config example_bbh_compas_config.yaml
Commandline Options error: Unknown Mass Loss Prescription
Use option '-h' (or '--help') to see (descriptions of) available options
```

Im fixing the weird dictionary issue

jeffriley:
Added updated yaml files (if I could type the update would have read "Updated yaml files: changed mass-loss-prescription from VINK to FLEXIBLE2023") in both file:

   misc/examples/methods_paper_plots/detailed_evolution/example_bbh_compas_config.yaml
   compas_python_utils/preprocessing/compasConfigDefault
   
That should fix the pytests problem.

avivajpeyi:
This spellcheck error is rather irritating haha -- i propose we merge this in, and I'll make a new PR to figure out why this is failing (even though we've updated the dictionary)

jeffriley:
@avivajpeyi will fix the precommit test failure (spell-check problem).

The updated yaml files fixed the build issue.

@jmerritt1 @SimonStevenson please merge at your convenience (ignore the precommit check failure)

ilyamandel (changes requested):
Mostly minor comments, but some of these (e.g., 4, 5) should really be fixed

1.
precommit-checks.yaml:
Why do we need line 16 as a comment when it appears just above without being commented out?
`# - uses: actions/checkout@v4`
Comments on lines 18--21 also seem unnecessary or misplaced

2.
Given the change in default winds, has it been checked what the impact on our default binary used for standard tests is (see, will the following still make a merging BBH)?
`../../src/COMPAS --initial-mass-1 35 --initial-mass-2 31 -a 3.5 -n 1 --detailed-output --random-seed 0 --metallicity 0.001`

3. BaseStar.cpp, line 1828 -- the comment is a bit colloquial, should say "by which the star is..."
`double HD_limit_factor = m_Radius * std::sqrt(m_Luminosity) * 1.0E-5; // calculate factor by which you are above the HD limit`

4. BaseStar.cpp, line 1943: function intro comments don't match actual function (there's no argument "prescription" in the function)

```
 * double CalculateMassLossRateOBVink2001(const double prescription)
 *
 * @param   [IN]    prescription                      Effective temperature in K
 * @return                                      Mass loss rate for hot OB stars in Msol yr^-1
 */
double BaseStar::CalculateMassLossRateOBVink2001() const {
```

5. Similarly for BaseStar::CalculateMassLossRateOBVinkSander2021 on lines 1997--2002

ilyamandel (changes requested):
A few more:

6. Our function header comments generally include "* @param [IN]..." but this isn't done for double BaseStar::CalculateMassLossRateOB(const OB_MASS_LOSS p_OB_mass_loss), double BaseStar::CalculateMassLossRateRSG(const RSG_MASS_LOSS p_RSG_mass_loss), etc.  In fact, I didn't realise that the input was a prescription until reading the code.

7. Why is m_DominantMassLossRate no longer set in double BaseStar::CalculateMassLossRateOBVink2001() ?  [This is just a request to check; if this is intentional, no action required.]

8. Since 
```
// opacity
constexpr double OPACITY_CGS_TO_SI                      = 0.1;                                                      
// cm^2 g^-1 to m^2 kg^-1
```
is defined in constants.h, 
why isn't it consistently used in the code 
(e.g., line 1766 of BaseStar.cpp: 
`double kappa_e_SI = kappa_e / 10.0; // m^2/kg `
)?

9. enum class MASS_LOSS_TYPE in constants.h somewhat confusingly mixes mass loss as applied to different types of stars (VERY_MASSIVE, RED_SUPERGIANT, LBV) with specific prescriptions (say, VINK -- Vink wrote a lot of papers about mass loss from different kinds of stars, so not clear what this refers to in this context)

ilyamandel:
Last query for @SimonStevenson , before I forget: given the Slack discussion about the WR multiplicative factor (previous default of 0.1), can you please confirm that 
- running with the new default prescription won't artificially reduce WR by an extra factor of 10?
- running with the BELCZYNSKI2010 prescription will reproduce the immediately preceding default (including that factor of 0.1)?

jeffriley:
> precommit-checks.yaml:
> Why do we need line 16 as a comment when it appears just above without being commented out?
> # - uses: actions/checkout@v4
> Comments on lines 18--21 also seem unnecessary or misplaced

@avivajpeyi  - can you look at this?


> Given the change in default>  winds, has it been checked what the impact on our default binary used for standard tests is 
> (see, will the following still make a merging BBH)?

    ../../src/COMPAS --initial-mass-1 35 --initial-mass-2 31 -a 3.5 -n 1 --detailed-output --random-seed 0 --metallicity 0.001

I checked - it does produce a BBH.

> BaseStar.cpp, line 1828 -- the comment is a bit colloquial, should say "by which the star is..."
> double HD_limit_factor = m_Radius * std::sqrt(m_Luminosity) * 1.0E-5; // calculate factor by which you are above the HD limit

Thought I'd changed all those - this one now changed (latest commit)


> BaseStar.cpp, line 1943: function intro comments don't match actual function (there's no argument "prescription" in 
> the function)
> Similarly for BaseStar::CalculateMassLossRateOBVinkSander2021 on lines 1997--2002
...
> Our function header comments generally include "* @param [IN]..." but this isn't done for double
> BaseStar::CalculateMassLossRateOB(const OB_MASS_LOSS p_OB_mass_loss), double 
> BaseStar::CalculateMassLossRateRSG(const RSG_MASS_LOSS p_RSG_mass_loss), etc. In fact, I didn't realise that the 
> input was a prescription until reading the code.

Fixed in latest commit:

added
    * @param   [IN]    p_LBV_Prescription          Which LBV prescription to use
to intro comments for BaseStar::CalculateMassLossRateLBV()

removed
    * @param   [IN]    prescription                      Effective temperature in K
from intro comments for both BaseStar::CalculateMassLossRateOBVink2001() and BaseStar::CalculateMassLossRateOBVinkSander2021() 

added
    * @param   [IN]    p_OB_MassLoss              Mass loss prescription to use
to intro comments for BaseStar::CalculateMassLossRateOB()

added
    * @param   [IN]    p_RSG_MassLoss             Mass loss prescription to use
to intro comments for BaseStar::CalculateMassLossRateRSG() 

added
    * @param   [IN]    p_VMS_MassLoss             Mass loss prescription to use
to intro comments for BaseStar::CalculateMassLossRateVMS()

> Why is m_DominantMassLossRate no longer set in double BaseStar::CalculateMassLossRateOBVink2001() ? [This is just a
> request to check; if this is intentional, no action required.]

BaseStar::CalculateMassLossRateOBVink2001() is called from BaseStar::CalculateMassLossRateOB(), and m_DominantMassLossRate is set there - but BaseStar::CalculateMassLossRateOBVink2001() is also called
elsewhere in the code.

@jmerritt1, from your comment above:

> ... m_DominantMassLossRate ... It is my preference to only set this flag in higher wrapper functions ...

Can you just do a final check that m_DominantMassLossRate gets set when it should (e.g. scenario above - BaseStar::CalculateMassLossRateOBVink2001() is called from multiple places).


> Since
>
> // opacity
> constexpr double OPACITY_CGS_TO_SI = 0.1;   // cm^2 g^-1 to m^2 kg^-1
>
> is defined in constants.h, why isn't it consistently used in the code (e.g., line 1766 of BaseStar.cpp:
> double kappa_e_SI = kappa_e / 10.0; // m^2/kg 

Fixed  (latest commit).


> enum class MASS_LOSS_TYPE in constants.h somewhat confusingly mixes mass loss as applied to different types of stars
> (VERY_MASSIVE, RED_SUPERGIANT, LBV) with specific prescriptions (say, VINK -- Vink wrote a lot of papers about mass loss
> from different kinds of stars, so not clear what this refers to in this context)

@jmerritt1 and/or @SimonStevenson can you address this comment?

avivajpeyi:
RE precommit stuff -- sorry i was trying to hack on this to figure out why the precommit is falling, 

Deleting the commented stuff

avivajpeyi:
@ilyamandel , RE 2: 
> Given the change in default winds, has it been checked what the impact on our default binary used for standard tests is (see, will the following still make a merging BBH)?


The pytest that runs compas finds that we are getting the BBH. The pytest runs the[ 'example_bbh_compas_config'](https://github.com/TeamCOMPAS/COMPAS/blob/e4e10b47c9350170e01327b18cca5485a61213da/misc/examples/methods_paper_plots/detailed_evolution/example_bbh_compas_config.yaml): 

https://github.com/TeamCOMPAS/COMPAS/actions/runs/6686898243/job/18166850858#step:6:353
```
Run example COMPAS job using the config example_bbh_compas_config.yaml

COMPAS v02.40.00
A binary star simulator

Start generating binaries at Mon Oct 30 01:00:36 2023

0: DCO formed: (Main_Sequence_>_0.7 -> Black_Hole) + (Main_Sequence_>_0.7 -> Black_Hole)
1: DCO formed: (Main_Sequence_>_0.7 -> Black_Hole) + (Main_Sequence_>_0.7 -> Black_Hole)

Generated 2 of 2 binaries requested

Simulation completed

End generating binaries at Mon Oct 30 01:00:40 2023

Clock time = 3.57758 CPU seconds
Wall time  = 0000:00:03 (hhhh:mm:ss)
```


Note that this is just checking if a BBH is formed using the `example_bbh_compas_config.yaml` (I think Jeff updated this yaml to get the test to pass). 
The test is not doing anything fancy with checking if the masses of the BBH are in a specific range, or anything about the detailed evolution.

ilyamandel:
Hi @jmerritt1 , @SimonStevenson :

Looks like we are almost there, just three items remaining as far as I can tell and we can pull this in.  Do you think you can check this soon (would be good to merge this in now, before some other PRs that are waiting)?

A. 
@jmerritt1, from your comment above:

... m_DominantMassLossRate ... It is my preference to only set this flag in higher wrapper functions ...

Can you just do a final check that m_DominantMassLossRate gets set when it should (e.g. scenario above - BaseStar::CalculateMassLossRateOBVink2001() is called from multiple places).

B. enum class MASS_LOSS_TYPE in constants.h somewhat confusingly mixes mass loss as applied to different types of stars
(VERY_MASSIVE, RED_SUPERGIANT, LBV) with specific prescriptions (say, VINK -- Vink wrote a lot of papers about mass loss
from different kinds of stars, so not clear what this refers to in this context)

C. Given the Slack discussion about the WR multiplicative factor (previous default of 0.1), can you please confirm that

running with the new default prescription won't artificially reduce WR by an extra factor of 10?
running with the BELCZYNSKI2010 prescription will reproduce the immediately preceding default (including that factor of 0.1)?

SimonStevenson:
Thanks everyone for the input on this PR, it's great that it's almost there. 

> running with the new default prescription won't artificially reduce WR by an extra factor of 10?

Currently, it does reduce WR mass-loss rates by an additional factor of 10. I believe that all of JDs testing/results have been performed with this setup. I noticed this the other day. We will do some additional testing/comparisons of fWR=0.1 v fWR=1.0. I think that the new WR mass-loss prescription does not reduce mass-loss rates by as much as a factor of 10 for most high-mass He stars. 

I made a plot comparing the initial-final He star masses (at solar metallicity) for both the old and new prescriptions, with fWR=0.1 and fWR=1.0.
heliumStarInitialFinalMassLossRelation

Here we see that whilst for low-mass He stars (where we use Vink2017) the mass-loss rates are reduced, for high-mass solar metallicity He stars we actually see increased mass loss. 

> running with the BELCZYNSKI2010 prescription will reproduce the immediately preceding default (including that factor of 0.1)?

Currently this is not the case -- whatever value of fWR is set will be used. If the user doesn't specify a value, the default will be used (currently 0.1). 

Based on this, I'm not sure what the new default fWR should be (fWR=1.0 better reflects the implemented mass-loss prescriptions, fWR=0.1 allows for the production of Cyg X-1 like BH masses at solar metallicity).

jmerritt1:
Thanks everyone for your work on this. I've been travelling for two weeks but just got back online and caught up with the discussion.

> A. @jmerritt1, from your comment above:
> 
> ... m_DominantMassLossRate ... It is my preference to only set this flag in higher wrapper functions ...
> Can you just do a final check that m_DominantMassLossRate gets set when it should (e.g. scenario above - BaseStar::CalculateMassLossRateOBVink2001() is called from multiple places).

I've checked again that `m_DominantMassLossRate` is set in both the `BELCZYNSKI2010` (added one missing line for OB) and `FLEXIBLE2023` wrappers. For some of the giant branch specific prescriptions, it is set in the relevant stellar type files under the Hurley wrapper.

> B. enum class MASS_LOSS_TYPE in constants.h somewhat confusingly mixes mass loss as applied to different types of stars (VERY_MASSIVE, RED_SUPERGIANT, LBV) with specific prescriptions (say, VINK -- Vink wrote a lot of papers about mass loss from different kinds of stars, so not clear what this refers to in this context)

I changed this list to include broader types of mass loss, and not specific prescriptions. We are leaning towards only including types, eg. RSG, VMS, etc. I agree that `VINK` should be removed, since it should now no longer be set in the higher level wrappers in favor of OB.  `NIEUWENHUIJZEN_DE_JAGER`, `KUDRITZKI_REIMERS`, and `VASSILIADIS_WOOD` are sometimes set by the stellar type specific giant branch files. If distinguishing between these isn't an important diagnostic tool, we can change all these to `RSG`. Is there any reason we need the older prescription specific labels?

Regarding C. and fWR, I've been running with fWR=1.0 for everything so far, and hadn't realized the default .yaml had changed.

ilyamandel:
@SimonStevenson -- I am trying to match up your latest plot with the one from your paper with Teagan.  That one was in units of luminosity rather than mass on the abscissa -- but it seemed to show that at the highest luminosities (i.e., high masses), mass loss rates from newer models were actually below those from the old models with a rescaling factor of 0.1, right?

image

ilyamandel:
@jmerritt1 -- I like your suggestion of removing VINK and replacing all of NIEUWENHUIJZEN_DE_JAGER, KUDRITZKI_REIMERS, and VASSILIADIS_WOOD with RSG.  Thanks for doing that!

SimonStevenson:
@ilyamandel We're not using Vink2017 for high-mass (high luminosity) WR stars, only for low-mass (low-luminosity) stripped He stars. At solar metallicity, the transition is around L = 10^5 Lsol (which corresponds to an initial helium star mass somewhere in the range 5-10 Msol). We are using the prescription from Sander & Vink 2020 for high-mass He stars. This results in similar mass-loss rates to the old Belczynski+2010 prescription. 
compareVink2017SanderVink2020Solar

ilyamandel:
Thanks for the explanation, @SimonStevenson !  I wonder if we ultimately want two separate WR multiplier choices, for Vink 2017 at lower masses and for Sander & Vink 2020 at higher masses?  If we did, I might be tempted to use fWRlowmass = 1 and fWRhighmass=0.1 in my runs... But I think the default should be 1 for both, since these are the best available models, and so long as we use just one multiplier, let's keep the default at 1.

BTW, does the transition happen exactly at the point where the Vink 2017 and Sander & Vink 2020 curves cross over in your plot?  If not, wind mass loss rates would be discontinuous, which is a bit weird.

SimonStevenson:
Yes, we just use the maximum of the two prescriptions, so the transition is where they cross over.

ilyamandel:
OK, then my only two remaining requests are:
- remove VINK and replace all of NIEUWENHUIJZEN_DE_JAGER, KUDRITZKI_REIMERS, and VASSILIADIS_WOOD with RSG in MASS_LOSS_TYPE (and in the code where it's assigned)
- set the default Wolf-Rayet multiplier to 1.0 rather than 0.1 (and regenerate the default .yaml)

jmerritt1:
> @jmerritt1 -- I like your suggestion of removing VINK and replacing all of NIEUWENHUIJZEN_DE_JAGER, KUDRITZKI_REIMERS, and VASSILIADIS_WOOD with RSG. Thanks for doing that!

Ah actually I've forgotten we will need to add a label for evolved/giant stars that are also below the cutoffs for RSG, see the pink lines (KR) in the below figure for example. (colored by dominant mass loss rate)
Also, is there any opposition to using acronyms, `RSG`, `LBV`,  in place of full labels, since this is mainly a diagnostic tool?

image

jmerritt1:
And in the above figure, just want to make sure that this is the intended treatment for the gap between 12.5kK and 8kK, around HG, Simon and I came to the understanding that you'd like these to be treated as WR (see short purple line segments around HG)? We revisited the original conversation with Andreas Sander and this seems to be the intention?

ilyamandel:
Looks great!

jeffriley:
> and regenerate the default .yaml

Just in case you missed it, there's now an option to do that (--create-yaml-file, see https://compas.readthedocs.io/en/latest/pages/User%20guide/Program%20options/program-options-list-defaults.html#options-props-c)

jmerritt1:
> > and regenerate the default .yaml
> 
> Just in case you missed it, there's now an option to do that (--create-yaml-file, see https://compas.readthedocs.io/en/latest/pages/User%20guide/Program%20options/program-options-list-defaults.html#options-props-c)

Just tried this but have it writing the previous default, fWR=0.1. Where should I change this? options.cpp? (edit, just wasn't working because I forgot to recompile before generating).

jeffriley:
@jmerritt1 I don't think you pushed the new version of Options.cpp - with `m_WolfRayetFactor = 1.0`

Oh - and `compasDefaultConfig.yaml` should be in `compas_python_utils/preprocessing`, not `src`

jmerritt1:
> @jmerritt1 I don't think you pushed the new version of Options.cpp - with `m_WolfRayetFactor = 1.0`
> 
> Oh - and `compasDefaultConfig.yaml` should be in `compas_python_utils/preprocessing`, not `src`

Okay I think it's fixed now, and I've removed the one in src.

ilyamandel (approved):
Looks good -- thanks for all the hard work, everyone!

jmerritt1:
Having an issue with all of the new mass-loss settings. When running with default settings (all of the .yaml commented out), Run_Details reads: `mass-loss-prescription = 'VINK', DEFAULT_USED, STRING` 

Otherwise, using the newest versions of `compasConfigDefault.yaml` and `runSubmit.py`, and commenting out the new defaults for any of the following (or changing them to any other allowed option):     
`    --mass-loss-prescription: 'FLEXIBLE2023'                             
    --OB-mass-loss: 'VINK2021'
    --RSG-mass-loss: 'DECIN2023'  
    --VMS-mass-loss: 'SABHAHIT2023'    
    --WR-mass-loss: 'SANDERVINK2023'    `

yields a `Commandline Options error: Unknown Mass Loss Prescription`

Sorry, I didn't notice this until I tried to weigh in on the discussion of merger rates w.r.t new default winds. Let me know if this is not the right place to discuss or debug this. I know this week is Ozgrav so no worries if there isn't time to figure this out. Thanks.

ilyamandel:
@jmerritt1  -- I saw your new issue post via e-mail, but it seems to have disappeared.  I suggest posting new issues separately, since people will likely miss them in closed PRs.

PR #997, opened 2023-10-04
move ipynb to docs

### LINKED ISSUE #861: duplicates of .ipynb?
Hey -- it looks like there are duplicate ipynb present: 

[COMPAS/online-docs/notebooks](https://github.com/TeamCOMPAS/COMPAS/blob/dev/online-docs/notebooks/CosmicIntegration.ipynb)
[COMPAS/utils/Tutorial](https://github.com/TeamCOMPAS/COMPAS/blob/dev/utils/Tutorial/CosmicIntegration.ipynb) 

Initially it was a bit unclear to me which one I should be using -- but it looks like the Tutorial files have been copied over into the docs folder.

Maybe just keep the `Tutorial` folder in the `doc` dir? This may help future users from getting confused by this.

Examples of where this has been done in other repos: 

https://docs.exoplanet.codes/en/latest/

Screen Shot 2022-11-04 at 12 27 06 pm

Alternatively, maybe just include links to github where the notebooks are located in the `utils/Tutorials/` folder (rather than coping the files from tutorials --> docs/notebooks`)

avivajpeyi:
Speaking with @jeffriley , we think these can exist in one location rather than `cp`ing the files around.
Mostly moves ipynb around, but also fixes their `##` headings (now `#` only appears once, this is needed to play nice with the TOC)

Closes #861 

NOTE: atm only the `CosmicItegration.ipynb` is being checked/tested.

It would be good to test more of these notebooks to ensure that they still work/arnt hard-coded for some specific user.


jeffriley (approved):
Thanks @avivajpeyi . The changes look good to me, except for:

In `online-docs/pages/User Guide/Post-processing/hdf5/WorkingWithHDF5.ipynb`, you have:

    "This notebook is structed in the following:\n"

Should that be "... structured in the following way:"?

> It would be good to test more of these notebooks to ensure that they still work/arnt hard-coded for some specific user.

I have not done this for this review, but we could do it over time...

avivajpeyi:
Nice catch @jeffriley -- fixed the typo! (also fixed the path for the cosmic-integrator path for the pytest)

reinhold-willcox:
@avivajpeyi The challenge of making these flexible to any user is that we'd then have to upload an HDF5 file to the github page. This isn't debilitating, but we've restricted this for now because the files often break the github size limit. So we'd want to make sure that the files used in the examples are cut down to only include the relevant systems (e.g only including DCOs if that's what we're plotting).

avivajpeyi:
I think you can make a mock dataset or even get the code to download a small dataset example from zenodo. Uploading an hdf5 in the git repo isn't needed. We can put a large admonition stating that we're using a mock dataset and for legit data go to zenodo/create it using COMPAS with some ini.

The benefit of using a mock/small dataset is that we can thereby always ensure that the examples are functioning (as they should!)

See the comsic integrator notebook for an example of what I mean of using a mock dataset.

avivajpeyi:
Any more changes needed for this?

avivajpeyi:
@reinhold-willcox any more changes to this

jeffriley:
Merging. @reinhold-willcox let us know if you think there should be changes later.

reinhold-willcox:
@avivajpeyi nah all good, just letting you know why it was done this way

PR #1013, opened 2023-10-31
Add autogenerated detailed-evol plot to PR (if COMPAS-build workflow successful) 

avivajpeyi:
We have a CI tool that: 
- build COMPAS 
- Runs an example COMPAS job
- Runs some tests on the COMPAS output, eg
    - Make a 'detailed evolution plot'
    - Checks if the output H5 can still be read via the python-tools
    - Checks if some postprocessing scripts can be run on the H5
- Uploads the generated plots + logs to the action's artifacts

I thought it would be nice if we could add an extra CI tool that then takes the outputted data and provides us with a link to the plots/files, and generates a comment in PRs (something like the following).

It might make it easy to debug/compare detailed evolution plots from the `dev` branch and the PR's branch

This comment will only be in each PR one time (and will be updated, on new commits to the PR)

Screen Shot 2023-11-01 at 11 47 46 am

**Requires** the following to be merged in first
#1087


ilyamandel (approved):
Looks good to me -- thanks, @avivajpeyi !  Please feel free to merge if you think it's ready.

avivajpeyi:
Woops, sorry @ilyamandel -- i forgot about this! 

I'll have some time later this week to get through this.

avivajpeyi:
Ah the plot isnt uploading since the detailed-evolution plot isnt even being generated! Latex error still persists 🙄  
```
Generating detailed evolution plot

Traceback (most recent call last):
  File "/opt/hostedtoolcache/Python/3.9.18/x64/lib/python3.9/site-packages/matplotlib/texmanager.py", line 250, in _run_checked_subprocess
    report = subprocess.check_output(
  File "/opt/hostedtoolcache/Python/3.9.18/x64/lib/python3.9/subprocess.py", line 424, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
  File "/opt/hostedtoolcache/Python/3.9.18/x64/lib/python3.9/subprocess.py", line 505, in run
    with Popen(*popenargs, **kwargs) as process:
  File "/opt/hostedtoolcache/Python/3.9.18/x64/lib/python3.9/subprocess.py", line 951, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/opt/hostedtoolcache/Python/3.9.18/x64/lib/python3.9/subprocess.py", line 1837, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: 'latex'

The above exception was the direct cause of the following exception:
```

avivajpeyi:
@ilyamandel Do you know who worked the 'detailed evolution plot' functions? There are some things that can be done to prevent this latex error, but wanted to check with the author first (in case they'd already tried it).

ilyamandel:
Reinhold had done a lot of the work on it recently, if I recall correctly.
You can also check the git commit history.

reinhold-willcox:
@avivajpeyi I modify the file from time to time, but I only check that it works locally. My apologies if I've caused headaches with the git CI. But the error you posted references some latex issue, which is not familiar to me. We have previously used latex to get nicer fonts for the plot labels, but this has not been changed recently (to my knowledge).

avivajpeyi:
Yeah I had a look at the CI error a bit closer, it's to do with the Latex installation on the CI runner... Really irritating tbh. 

I wonder if it would be ok to not render the latex labels if latex isn't installed.

ilyamandel:
If there’s a way to check at runtime if latex is installed and render labels without if it’s not, that would be more than acceptable!

avivajpeyi:
Unfortunately, i cant test this feature unless the workflow is in the main repo branch! See https://stackoverflow.com/questions/77161818/workflow-not-showing-up-in-github-actions


Could we merge this new workflow file in, and Ill make a new PR to fix it up (i assume there _may_ be some errors with the commenting initially)

Issue #1022, opened 2023-12-11
Core dumped when using the GE20 prescription for critical mass ratios


nrsegovia:
**Describe the bug**

By using the following configuration on COMPAS v02.39.00
```
COMPAS --random-seed 15 --quiet TRUE --logfile-definitions gridDefinitions.txt --maximum-evolution-time 13700 --initial-mass-min 0.08  --common-envelope-alpha 1 --grid binaries.txt --rlof-printing TRUE --mass-transfer-accretion-efficiency-prescription FIXED --critical-mass-ratio-prescription GE20 -z 0.03 --mass-transfer-fa 0.5 --mass-transfer-angular-momentum-loss-prescription MACLEOD_LINEAR --mass-transfer-jloss-macleod-linear-fraction 0
```
I get the following error

```
terminate called after throwing an instance of 'std::bad_array_new_length'
terminate called recursively
Aborted (core dumped)
```
which produces rather small files, around 10% of what I expect. The same population under different prescriptions ( "NONE" and "HURLEY_HJELLMING_WEBBINK") works just fine. 

**To Reproduce**
Run COMPAS using the gridfile I have attached and the configuration specified above. The logfile configuration can be ignored, it seems.

**Additional context**
Jeff was able to reproduce the issue on v2.41.00

[gridfile.tar.gz](https://github.com/TeamCOMPAS/COMPAS/files/13630586/gridfile.tar.gz)

jeffriley:
Fixed in v02.41.01
PR #1025

ilyamandel:
Fixed by @jeffriley (thank you!), closing, with thanks to @nrsegovia for reporting!

PR #1023, opened 2023-12-11
Defect repair: Fix for issue #1022

jeffriley:
- Fix for issue #1022 - incorrect value being used for last entry in array in BaseStar::InterpolateGe20QCrit()


jeffriley:
Opened in error

PR #1025, opened 2023-12-11
Fix for issue #1022

### LINKED ISSUE #1022: Core dumped when using the GE20 prescription for critical mass ratios

jeffriley:
Fix for issue #1022 - an incorrect index was being used for the last entry of an array.

This PR includes two commits: the first contains the fix, the second a little bit of (related) code cleanup.

### DIFF
--- a/src/BaseBinaryStar.cpp
+++ b/src/BaseBinaryStar.cpp
@@ -1893,7 +1893,7 @@ void BaseBinaryStar::CalculateMassTransfer(const double p_Dt) {
         // NOTE: Critical mass ratio is defined as mAccretor/mDonor
         double qCrit = m_Donor->CalculateCriticalMassRatio(m_Accretor->IsDegenerate());
 
-        isUnstable = (m_Accretor->Mass()/m_Donor->Mass()) < qCrit;
+        isUnstable = (m_Accretor->Mass() / m_Donor->Mass()) < qCrit;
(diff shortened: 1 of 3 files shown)


ilyamandel (approved):
Looks good, thank you for the quick fix!

PR #1026, opened 2023-12-11
Fix for Issue#1022

jeffriley:
Fix for issue https://github.com/TeamCOMPAS/COMPAS/issues/1022 - an incorrect index was being used for the last entry of an array.

This PR includes two commits: the first contains the fix, the second a little bit of (related) code cleanup.


ilyamandel (approved):
All good! :)

PR #1028, opened 2023-12-14
Added power law mass ratio distribution

SimonStevenson:
I have added the possibility of using a power law mass ratio distribution (I had to add it for a project I'm working on).

This is a fairly small PR.

I've attached a plot comparing samples drawn with this new distribution (for a power law exponent of -2) to the theoretical pdf.

power_law_mass_ratio_dist

I think I've updated the changelog and docs, but please let me know if I missed somewhere.

### DIFF
--- a/online-docs/pages/whats-new.rst
+++ b/online-docs/pages/whats-new.rst
@@ -6,6 +6,10 @@ Following is a brief list of important updates to the COMPAS code.  A complete r
 
 **LATEST RELEASE** |br|
 
+**02.42.00 Nov 02, 2023**
+
+* Added power law mass ratio distribution (``--mass-ratio-distribution POWERLAW```) and associated option (``--mass-ratio-power```)
+
 **02.41.00 Nov 02, 2023**
 
 * Added a naive tides implementation.
--- a/online-docs/pages/User guide/Program options/program-options-list-defaults.rst
+++ b/online-docs/pages/User guide/Program options/program-options-list-defaults.rst
@@ -781,8 +781,9 @@ Default = Value is sampled if option not specified.
 
 **--mass-ratio-distribution** |br|
 Initial mass ratio distribution for :math:`q = \frac{m2}{m1}`. |br|
-Options: { FLAT, DUQUENNOYMAYOR1991, SANA2012 } |br|
-``FLAT`` is uniform in the mass ratio between ``--mass-ratio-min`` and ``--mass-ratio-max``, the other prescriptions follow Duquennoy & Mayor 1991 and Sana et al. 2012 |br|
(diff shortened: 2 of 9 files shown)


online-docs/pages/whats-new.rst
```diff
@@ -6,6 +6,10 @@ Following is a brief list of important updates to the COMPAS code.  A complete r
 
 **LATEST RELEASE** |br|
 
+**02.42.00 Nov 02, 2023**
+
+* Added power law mass ratio distribution (``--mass-ratio-distribution POWERLAW```) and associated option (``--mass-ratio-power```)
```
avigna, line 11: There are some extra commas here, right? I know its a comment, but just to be clean.

avigna (changes requested):
Looks good to me, just an extremely minor comment. Two questions:
1. The plot was produced with COMPAS, right?
2. The power law exponent is user defined and can go from -inf to int, right?

avigna:
I have reviewed this and effectively approved it (modulo a tiny edit). However, and besides this PR... wasn't the plan to eventually move sampling outside of the src code? Or is the plan to have hibrid version (use COMPAS sampling options vs create your own grids)?

ilyamandel (changes requested):
I fixed a couple of comment typos directly.

I agree with @avigna that we shouldn't add and maintain any more initial condition distributions -- can do this via grid files -- but since this option already exists, it's fine.

However, given the appearance of a new option, a new default .yaml file should be re-generated (this can be done automagically by running COMPAS).

jeffriley (changes requested):
Thanks Simon!

In addition to the comments made by @avigna and @ilyamandel : 

Code: 
   - The new option should be included in the `m_BSEOnly` vector in Options.h (see point 10 at the top of Options.cpp)
   - yaml.h needs to be updated to include the new option (see point 13 at the top of Options.cpp)
   - Do we need to sanity check the value entered for `--mass-ratio-power` (are there any values that are not allowed)?
 
Documentation: 

   - Point to the `--mass-ratio-power` option in the description of the available values for the `--mass-ratio-distribution` option (add text something like ("via the --mass-ratio-power option")
   - Make clear in the description of the `--mass-ratio-power` option that it refers only to the `--mass-ratio-distribution` option.

ilyamandel:
Oh, one more thought: perhaps clarify in the documentation that the power law option uses ``--mass-ratio-min`` and ``--mass-ratio-max`` in addition to the power law slope to determine the distribution.

ilyamandel:
@SimonStevenson -- ping on this one.

ilyamandel:
@SimonStevenson -- another ping on this one.

ilyamandel:
@SimonStevenson , the longer you wait, the more effort it will be to roll this in, as the code keeps evolving... ;-)

PR #1030, opened 2023-12-15
Reapply winds changes from v02.40.00

jeffriley:
Reapply winds changes from v02.40.00 - seem to have been backed-out in v02.41.00


SimonStevenson (approved):
I checked that this restores the FLEXIBLE2023 mass loss prescription, that the code compiles/runs and did a couple of spot checks that other recent changes are not affected. I didn't find anything missing.

PR #1032, opened 2023-12-15
fixes to the detailed plotter

reinhold-willcox:
...including improved robustness to the Roche Lobe, and masking for only end-of-timestep events


ilyamandel (approved):
Thank you, @reinhold-willcox , looks good!

PR #1052, opened 2024-01-05
BSE vs SSE investigation and changes

jeffriley:


jeffriley:
Will add details later

jeffriley:
@ilyamandel, @avigna, @reinhold-willcox, @SimonStevenson 

After a discussion on slack with @jmerritt1, I agreed to look into the BSE vs SSE evolution "issue", and see if I could explain why a wide binary with a small companion didn't evolve the same way a single star with the same initial attributes.

The short summary is that I made a few changes to the code to make single stars evolve the same way as the primary of a binary *when the same timesteps are used*.  To do that I:

1. Added code to read timesteps from a file, for both SSE and BSE mode.
2. Made a few small mods to how evolution happens in the SSE code - not big changes (order of calls etc.).
3. Quantised the timesteps.  I did this mainly to make sure we weren't losing precision when we added `dt` to `time`/`age`.  Doubles in c++ have 15 digits of precision (long double have 18 I think, so we could go to long double, but I don't think we need to), and since `dt`, `age`, and `time`, are stored in Myr, we use a lot of those digits just to store the values, then when we start adding (or multiplying by the timetep multiplier) we can walk a few digits off the end of the number.  Quantising helps stop that.  It also helps when storing timesteps in a timesteps file for use in a later run - quantising `dt` helps not lose precision when we store and re-read the numbers.  I define (in `constants.h`) `TIMSTEP_QUANTUM` as `1e-12Myr`, and then quantise `dt` to an integral number of `TIMESTEP_QUANTUM`s.  `ABSOLUTE_MINIMUM_TIMESTEP` and `NUCLEAR_MINIMUM_TIMESTEP` are adjusted accordingly.

I started out trying to put something together to show the differences between SSE and BSE, and provide a mechanism to see that in detail by forcing one or the other to use pre-defined timesteps, and by modifying the code to ensure that if the same timesteps are used BSE vs SSE, the evolution is the same regardless of mode.  But I think I might have actually fixed the problem...  I've run a few tests (I hasten to say, not many - really, about a dozen...), but I haven't seen one where BSE and SSE evolve to a different stellar type given the same initial conditions, even when they use their own calculated timesteps.  Whether that's true or not (that SSE and BSE are now consistent regardless of timestep calculation), this code at least represents a method to check evolution BSE vs SSE when the same timesteps are used - and we know this code will cause BSE and SSE to evolve the same way with the same timesteps - so if they actually evolve differently when they use their own (calculated) timesteps, the question is, why?  We know this code evolves consistently BSE vs SSE when the same timesteps are used.  If I haven't actually fixed the problem, and BSE and SSE evolve differently with the same initial conditions (for wide binaries) when each mode calculates its own timesteps, which is right?

This PR is a draft.  I haven't changed or added any documentation.  I don't know if we want to keep the `--timesteps-filename` option in the production code, or have it optionally compiled (like the profiling code).  I'm happy to leave it in, even though it might only be useful occasionally (I suppose that's true of a lot of the options, really).  If everyone is happy with the code here I'll go ahead and make sure it is tidy, and add/change documentation.

This could do with some extra testing.  Frankly, I've spent so long looking at this over the past 5 or 6 days that I'm just seeing the same things...  New eyes and testing would help.

I think there is still some work to do to rationalise the different variants of `UpdateAttributes...` and `AgeOneTimestep...` (I think that will please @reinhold-willcox) - they have been changed a few times and now don't perform according to their (in-code) documentation, and I think we could now reduce the number and complexity of those function (calls) - a future PR.

@ilyamandel I did do a little bit of code cleanup in this PR.  I didn't mean to - I forgot that I wasn't going to, and couldn't resist in a few places.  Sorry - it's not too much I hope :-)


The current dev version is v02.41.04.  This PR will change that to v02.42.00.  I ran some tests to compare and check - see below.

I performed two runs of 1 million binaries in each run: one for v02.41.04, and one for v02.42.00 (this PR), both with default COMPAS options except for the starting random seed, which was 0 in both cases.

Summary of results:

`v02.41.04`

    Number of binaries = 1000000, start seed = 0

    Number of DCOs     = 44990
    Number of BHBHs    = 40293
    Number of BHNSs    = 3822
    Number of NSNSs    = 875

    Number of BHBHs merging in a Hubble time = 2789
    Number of BHNSs merging in a Hubble time = 1210
    Number of NSNSs merging in a Hubble time = 253

    Runtime (seconds) = 15834


`v02.42.00`

    Number of binaries = 1000000, start seed = 0

    Number of DCOs     = 44970
    Number of BHBHs    = 40290
    Number of BHNSs    = 3811
    Number of NSNSs    = 869

    Number of BHBHs merging in a Hubble time = 2653
    Number of BHNSs merging in a Hubble time = 1194
    Number of NSNSs merging in a Hubble time = 249

Runtime (seconds) = 15613


The summaries shown above indicate that v02.42.00 is consistent with v02.41.04 in BSE mode, though not exactly the same - there were some minor modifications to the SSE and BSE code that might affect the evolution of some binaries.


I performed a run using COMPAS v02.41.04 for a single very wide binary with a massive primary and very small secondary, as follows:

    $ ./compas --mode bse -n 1 --random-seed 0 --initial-mass-1 140 --initial-mass-2 0.1 --semi-major-axis 1e20 --metallicity 0.000473 --logfile-type csv --detailed-output

    COMPAS v02.41.04
    A binary star simulator

    Start generating binaries at Fri Jan  5 17:27:47 2024

    0: Massless Remnant formed: (Main_Sequence_>_0.7 -> Massless_Remnant) + (Main_Sequence_<=_0.7 -> Main_Sequence_<=_0.7)

    Generated 1 of 1 binaries requested

    Simulation completed

    End generating binaries at Fri Jan  5 17:27:47 2024

    Clock time = 0.439283 CPU seconds
    Wall time  = 0000:00:00 (hhhh:mm:ss)


I then performed arun with COMPAS v02.41.04, for a single star with the same attributes of the primary in the run above:

    $ ./compas --mode sse -n 1 --random-seed 0 --initial-mass 140 --metallicity 0.000473 --logfile-type csv --detailed-output

    COMPAS v02.41.04
    A binary star simulator
    
    Start generating stars at Fri Jan  5 17:28:22 2024
    
    0: Simulation completed: RandomSeed = 0, Initial Mass = 140, Metallicity = 0.000473, Black_Hole   `<--- does NOT match primary of BSE run`
    
    Generated 1 of 1 stars requested
    
    Simulation completed
    
    End generating stars at Fri Jan  5 17:28:23 2024
    
    Clock time = 0.349725 CPU seconds
    Wall time  = 00:00:00 (hh:mm:ss)


The runs shown above show that for v02.41.04, for the systems shown, BSE mode and SSE mode do not agree wrt the evolution of the massive star.

The same runs were performed using v02.42.00:

    $ ./compas --mode bse -n 1 --random-seed 0 --initial-mass-1 140 --initial-mass-2 0.1 --semi-major-axis 1e20 --metallicity 0.000473 --logfile-type csv --detailed-output
    
    COMPAS v02.42.00
    A binary star simulator
    
    Start generating binaries at Sat Jan  6 14:16:34 2024
    
    0: Massless Remnant formed: (Main_Sequence_>_0.7 -> Massless_Remnant) + (Main_Sequence_<=_0.7 -> Main_Sequence_<=_0.7)
    
    Generated 1 of 1 binaries requested
    
    Simulation completed
    
    End generating binaries at Sat Jan  6 14:16:34 2024
    
    Clock time = 0.465042 CPU seconds
    Wall time  = 0000:00:00 (hhhh:mm:ss)

And the SSE run:

    $ ./compas --mode sse -n 1 --random-seed 0 --initial-mass 140 --metallicity 0.000473 --logfile-type csv --detailed-output
    
    COMPAS v02.42.00
    A binary star simulator
    
    Start generating stars at Sat Jan  6 14:17:41 2024
    
    0: Simulation completed: RandomSeed = 0, Initial Mass = 140, Metallicity = 0.000473, Massless_Remnant   `<--- matches primary of BSE run`
    
    Generated 1 of 1 stars requested
    
    Simulation completed
    
    End generating stars at Sat Jan  6 14:17:41 2024
    
    Clock time = 0.078528 CPU seconds
    Wall time  = 00:00:00 (hh:mm:ss)


The runs shown above show that for v02.42.00, for the systems shown, BSE and SSE modes agree wrt the evolution of the massive star.

I have added functionality to allow the specification of a timesteps file.  If a timesteps file is specified by the user, COMPAS will use the timesteps read from the timesteps file rather than calculating the timestep to be taken.  This allows checking of BSE vs SSE functionality for the same timestep values.

The new option is `--timesteps-filename`, and is valid for both SSE and BSE modes.  The file specified is expected to be an ascii file with a single floating point value on each line of the file (blank lines are not allowed - we could change the code to ignore blank lines and/or lines starting with '#' so users can add comments, but that's not in the code yet).  If a timesteps file is specified, COMPAS will read each line and use the velue read for a single timestep iteration.  If, the system being evolved finishes evolution before all timesteps in the file are consumed and warning will be issued.  If evolution of the system has not finished, but the timesteps from  the file have been exhausted, a warning will be issued.

Using v02.42.00 I performed the run,

    $ ./compas --mode bse -n 1 --random-seed 0 --initial-mass-1 140 --initial-mass-2 0.1 --semi-major-axis 1e20 --metallicity 0.000473 --logfile-type csv --detailed-output

and did the following:

1. Create a file, `timesteps-bse`, from the timestep values from the detailed output file for the BSE run (only those with record type '4' - indicating the timestep was complete).
2. Ran the BSE run again, but this time using the option `--timesteps-filename timesteps-bse`. This produced a run identical to the original (from which the timesteps were harvested), showing that BSE mode is self-consistent when run using the same timestep values   
3. Ran COMPAS in SSE mode, with the initial attributes of the star the same as the initial attributies of the primary from the BSE run, using the option `--timesteps-filename timesteps-bse`.  This produced a single star that, at each timestep, evolved exactly as the primary from the BSE run.

I then repeated that process in the reverse order: I ran an SSE run, stored the timesteps and used those timesteps to run a BSE run.  This produced a binary for which the primary, at each timestep, evolved exactly as the singles from the SSE run.

I repeated that process for a few different systems/single star, with differing masses (not just massive stars) - the results were the same each time - the caveat is that I have not done extensive, rigorous testing, but the testing I have done indicates that the process is robust (but that doesn't mean there's not a corner-case somewhere that will break it...).

One anomaly I found was systems like this:

    $ ./compas --mode bse -n 1 --random-seed 179673 --initial-mass-1 5.5 --initial-mass-2 0.1 --semi-major-axis 1e20 --metallicity 0.0142 --logfile-type csv --detailed-output
    
    COMPAS v02.42.00
    A binary star simulator
    
    Start generating binaries at Sat Jan  6 16:02:41 2024
    
    0: Allowed time exceeded: (Main_Sequence_>_0.7 -> Carbon-Oxygen_White_Dwarf) + (Main_Sequence_<=_0.7 -> Main_Sequence_<=_0.7)
    
    Generated 1 of 1 binaries requested
    
    Simulation completed
    
    End generating binaries at Sat Jan  6 16:02:44 2024
    
    Clock time = 2.05443 CPU seconds
    Wall time  = 0000:00:03 (hhhh:mm:ss)

I stored the timesteps from that BSE run and ran an SSE run as follows:

    $ ./compas --mode sse -n 1 --random-seed 179673 --initial-mass 5.5 --metallicity 0.0142 --logfile-type csv --detailed-output --timesteps-filename timesteps-bse
    
    COMPAS v02.42.00
    A binary star simulator
    
    Start generating stars at Sat Jan  6 16:23:07 2024
    
    0: User-provided timesteps not consumed: RandomSeed = 179673, Initial Mass = 5.5, Metallicity = 0.0142, Carbon-Oxygen_White_Dwarf
    
    Generated 1 of 1 stars requested
    
    Simulation completed
    
    End generating stars at Sat Jan  6 16:23:08 2024
    
    Clock time = 0.340247 CPU seconds
    Wall time  = 00:00:00 (hh:mm:ss)

Note the status: `User-provided timesteps not consumed`.  What happens here is that the binary continues evolving for a number of timesteps (~25) of size 500Myr after the primary evolves to a COWD - until we hit the time threshold and stop.  In SSE mode, evolution stops as soon as a CO is formed - hence the status message that not all the (BSE) timesteps were consumed.  The binary probably doesn't need to take quite so many long timesteps after the COWD is formed - if nothing changes after 500Myr we could probably stop there - but checking that condition at each timestep for every binary in the population being evolved would probably slow the run down more than it's worth (given the number of times this will pop up).

jeffriley:
I found some examples @jmerritt1 gave me and ran some tests:

Test1, `v02.41.04`

    $ ./compas --mode bse -n 1 --initial-mass-1 135.0 --initial-mass-2 0.1 --metallicity 0.000149 --eccentricity 0.0 --semi-major-axis 1e+20
    
    COMPAS v02.41.04
    A binary star simulator
    
    Start generating binaries at Sat Jan  6 17:38:15 2024
    
    0: Massless Remnant formed: (Main_Sequence_>_0.7 -> Massless_Remnant) + (Main_Sequence_<=_0.7 -> Main_Sequence_<=_0.7)
    
    Generated 1 of 1 binaries requested
    
    Simulation completed
    
    End generating binaries at Sat Jan  6 17:38:15 2024
    
    Clock time = 0.009638 CPU seconds
    Wall time  = 0000:00:00 (hhhh:mm:ss)
    
    $ ./compas --mode sse -n 1 --initial-mass 135.0 --metallicity 0.000149
    
    COMPAS v02.41.04
    A binary star simulator
    
    Start generating stars at Sat Jan  6 17:38:48 2024
    
    0: Simulation completed: RandomSeed = 1704523127, Initial Mass = 135, Metallicity = 0.000149, Black_Hole
    
    Generated 1 of 1 stars requested
    
    Simulation completed
    
    End generating stars at Sat Jan  6 17:38:48 2024
    
    Clock time = 0.014801 CPU seconds
    Wall time  = 00:00:00 (hh:mm:ss)

So BSE and SSE evolve to different remnants.

Now with `v02.42.00`

    $ ./compas --mode bse -n 1 --initial-mass-1 140.0 --initial-mass-2 0.1 --metallicity 0.000473 --eccentricity 0.0 --semi-major-axis 1e+20
    
    COMPAS v02.42.00
    A binary star simulator
    
    Start generating binaries at Sat Jan  6 17:40:20 2024
    
    0: Massless Remnant formed: (Main_Sequence_>_0.7 -> Massless_Remnant) + (Main_Sequence_<=_0.7 -> Main_Sequence_<=_0.7)
    
    Generated 1 of 1 binaries requested
    
    Simulation completed
    
    End generating binaries at Sat Jan  6 17:40:20 2024
    
    Clock time = 0.00833 CPU seconds
    Wall time  = 0000:00:00 (hhhh:mm:ss)

    $ ./compas --mode sse -n 1 --initial-mass 140.0 --metallicity 0.000473
    
    COMPAS v02.42.00
    A binary star simulator
    
    Start generating stars at Sat Jan  6 17:40:27 2024
    
    0: Simulation completed: RandomSeed = 1704523227, Initial Mass = 140, Metallicity = 0.000473, Massless_Remnant
    
    Generated 1 of 1 stars requested
    
    Simulation completed
    
    End generating stars at Sat Jan  6 17:40:27 2024
    
    Clock time = 0.005532 CPU seconds
    Wall time  = 00:00:00 (hh:mm:ss)

So in `v02.42.00` BSE and SSE evolvethe same remnant.

The other example @jmerritt1 gave me was `mass = 135.0` and `metallicity = 000149` - I ran the tests with the same result: `v02.41.04` produces different remnants BSE vs SSE, but `v02.42.00` produce the same remnant.

That's not conclusive - only a couple of tests - but it is positive.

jeffriley:
It just occurred to me that we could use the `--timesteps-filename` option to selectively zoom-in on events.  If we run COMPAS for a single star/system and save the timesteps from the detailed output file, we could insert short timesteps in selected places to zoom-in at those places, then re-run the same star/system using the modified timesteps file.  Could be useful, so maybe it is worth leaving that code in.

jeffriley:
And so back to the @jmerritt1 's original question that sparked this investigation:

> ... how much should should we trust these 50-70 Msol, low-Z systems which are only produced in SSE and can't be replicated with BSE?

Well, the two examples given by @jmerritt1 now result in massless remnants rather than black holes, so maybe be don't produce those systems with SSE any more, but if we do, then running COMPAS in BSE mode with the SSE timesteps and a wide binary should reproduce the star as the primary.

jeffriley:
This may be a separate issue, but I'll put it here just to record it.  @LiekeVanSon asked (on slack):

>  ... in both the SSE and BSE star right before 3.3Myr: why does the Menv drop to 0?

One thing I notice during this investigation is that during a common envelope event (not relevant to the wide binaries investigated here, just something I noticed in the code), the BSE code (`BaseBinaryStar::ResolveCommonEnvelopeEvent()`) calculates the envelope mass for each constituent star, and stores the values in `m_MassEnv1` and `m_MassEnv2`, and those are the values recorded in the `BSE_Common_Envelopes` file.  Meanwhile, the SSE code does not keep track of the envelope mass of the star, but calculates it on demand as `Mass() - CoreMass()` in `BaseStar::StellarPropertyValue()`, and that's what's recorded for each star in the `BSE_Detailed_Output` file.  I haven't looked to see if they are kept in sync, but at is possible, probably likely, that the records written to each of those files are written at different times during the timestep, so while the common envelopes value is recorded mid-timestep, the detailed output values are printed at various times throughout the timestep, and the value printed at the end of the timestep may not be in sync with the common envelope values.  We could check that by looking at all record types in the detailed output file.

So the answer to @LiekeVanSon 's questions that it just may be a matter of timing wrt when the value is printed in different files.

avigna:
Thanks for the check and the thorough and clear explanation Jeff. 

I think the new option `--timesteps-filename` is worth keeping. First of all, you already coded it and it work. Second, I think there might be cases where it can be useful. Third, Jakob Stegmann had asked me about such an option, so we know at least some external user would find it useful. I would probably go ahead and 
> change the code to ignore blank lines and/or lines starting with '#' so users can add comments
but long as it issues a warning/error when the file doesn't have the specified format, it should be fine. I leave it up to you (it can always be an enhancement if someone asks for it later).

And yes, using it to selectively zoom-in can be a good case of use.

avigna:
> And so back to the @jmerritt1 's original question that sparked this investigation:
> 
> > ... how much should should we trust these 50-70 Msol, low-Z systems which are only produced in SSE and can't be replicated with BSE?
> 
> Well, the two examples given by @jmerritt1 now result in massless remnants rather than black holes, so maybe be don't produce those systems with SSE any more, but if we do, then running COMPAS in BSE mode with the SSE timesteps and a wide binary should reproduce the star as the primary.

This might have to do with the changes/difference in winds, particularly on how massive the core is at core collapse: for massive stars, some of the will experience a pair-instability supernova, which leaves no remnant. A slightly more or slightly less massive star would lead a black hole.

In any case, as I think Jeff would agree, we are looking for consistency in this tests/PR, not the correct physics :)

avigna:
> This may be a separate issue, but I'll put it here just to record it. @LiekeVanSon asked (on slack):
> 
> > ... in both the SSE and BSE star right before 3.3Myr: why does the Menv drop to 0?
> 
> One thing I notice during this investigation is that during a common envelope event (not relevant to the wide binaries investigated here, just something I noticed in the code), the BSE code (`BaseBinaryStar::ResolveCommonEnvelopeEvent()`) calculates the envelope mass for each constituent star, and stores the values in `m_MassEnv1` and `m_MassEnv2`, and those are the values recorded in the `BSE_Common_Envelopes` file. Meanwhile, the SSE code does not keep track of the envelope mass of the star, but calculates it on demand as `Mass() - CoreMass()` in `BaseStar::StellarPropertyValue()`, and that's what's recorded for each star in the `BSE_Detailed_Output` file. I haven't looked to see if they are kept in sync, but at is possible, probably likely, that the records written to each of those files are written at different times during the timestep, so while the common envelopes value is recorded mid-timestep, the detailed output values are printed at various times throughout the timestep, and the value printed at the end of the timestep may not be in sync with the common envelope values. We could check that by looking at all record types in the detailed output file.
> 
> So the answer to @LiekeVanSon 's questions that it just may be a matter of timing wrt when the value is printed in different files.

This is relevant, and likely worth to fix... but should be separate issue/PR.

jeffriley:
> This might have to do with the changes/difference in winds, particularly on how massive the core is at core collapse: for
> massive stars, some of the will experience a pair-instability supernova, which leaves no remnant. A slightly more or slightly less
> massive star would lead a black hole.
>
> In any case, as I think Jeff would agree, we are looking for consistency in this tests/PR, not the correct physics :)

Indeed - and the last comment is the point - prior to this PR, SSE and BSE did not always evolve the same way given the same initial values - for a wide binary with a small companion - when we might have expected the primary of the binary to evolve as a single star evolves in SSE.  In some cases SSE would produce a black hole, wheres BSE might produce a massless remnant.  Now they both (seem to, pending more testing) evolve the same star.

I achieved this by making SSE evolve the same was as the primary in BSE - but maybe I should have gone the other way and made the primary in BSE evolve the same way as SSE.  How do we determine which is/would have been right?

avigna:
> > This might have to do with the changes/difference in winds, particularly on how massive the core is at core collapse: for
> > massive stars, some of the will experience a pair-instability supernova, which leaves no remnant. A slightly more or slightly less
> > massive star would lead a black hole.
> > In any case, as I think Jeff would agree, we are looking for consistency in this tests/PR, not the correct physics :)
> 
> Indeed - and the last comment is the point - prior to this PR, SSE and BSE did not always evolve the same way given the same initial values - for a wide binary with a small companion - when we might have expected the primary of the binary to evolve as a single star evolves in SSE. In some cases SSE would produce a black hole, wheres BSE might produce a massless remnant. Now they both (seem to, pending more testing) evolve the same star.
> 
> I achieved this by making SSE evolve the same was as the primary in BSE - but maybe I should have gone the other way and made the primary in BSE evolve the same way as SSE. How do we determine which is/would have been right?

Uncertain. Philosophically, BSE should evolve as SSE, because single stars make up binaries. In practice, BSE introduces new interactions and timescales that do not exist for SSE, and therefore it makes sense to gauge the sensitivities in BSE (effectively timestepping) and map to SSE for replication.

jeffriley:
> In practice, BSE introduces new interactions and timescales that do not exist for SSE

@avigna  While that is usually true, in the comparisons I was doing I had a very wide binary with a massive primary and a very small secondary - there should have been no interactions - the BSE code should have effectively been evolving the same single star as the SSE code.  And when I used the same timesteps for each, the difference came down to BSE and SSE having very slightly different algorithms for evolution.  I changed the SSE algorithm to look like the BSE algorithm (a couple of very minor tweaks), and now there are no differences when I compare the same vary wide binary to a single star.  However, I could have changed the BSE algorithm to look like the SSE algorithm - had I done that we would be getting very slightly different outcomes for binaries.  That's where my question came from - which one would be right, or doesn't it really matter that much?

EDIT: Unless you are saying the BSE algorithm is fine-tuned for binaries that do interact, and that the very wide binary with a massive primary and a small companion is a special case.  I suppose I could live with that - maybe :-)

avigna:
> Unless you are saying the BSE algorithm is fine-tuned for binaries that do interact, and that the very wide binary with a massive primary and a small companion is a special case. I suppose I could live with that - maybe :-)
Yes, this is exactly what I am saying.

The most general way I think of this problem is that we have a timestep algorithm for SSE and an additional one for BSE. So when you don't have BSE or the binary is wide, the timestep algorithm should match that of SSE. But when the BSE plays a role, the timestep for it does too.

jeffriley:
> The most general way I think of this problem is that we have a timestep algorithm for SSE and an additional one for BSE. 
> So when you don't have BSE or the binary is wide, the timestep algorithm should match that of SSE. But when the BSE plays
> a role, the timestep for it does too.

Ok, but I didn't change the timestep calculations.  The changes I made were to the way the star's attributes are calculated - not the prescriptions or fomulae, but the order in which some things occur.  It's a fine point, because order and timestep might well be related, but BSE and SSE still calculate their timesteps differently, but they now do things in the same order.

ilyamandel:
Hi @jeffriley,
Thank you for the very impressive work, and sorry it took me so long to read up on this.  I haven't quite finished parsing everything, but will aim to do so in the next 24 hours.   Overall, I really like this approach, and I think we should keep the quantized timestep option, including the ability to provide timesteps as an input file at runtime.
Thank you!

ilyamandel:
On the "what's the right choice of timesteps" issue, the answer is that we should ultimately check for convergence.  I.e., at some point, as we reduce timesteps, we should presumably get a unique answer.  When timesteps are too large, the answer may jump around -- and no choice of too-large timesteps is a priori better. 

Anyway, will now shift to reviewing the PR, since I think we all agree this is the right approach for now (and convergence is a long-term issue that we'll have to return to, but we don't need to do everything at once).

Issue #705, opened 2021-11-17
Timestep calculations in SSE and BSE code could produce inconsistent results when comparing BSE to SSE


jeffriley:
**Describe the bug**

In the SSE code, the first order timestep calculation is performed by BaseStar::CalculateTimestep().  The timestep calculated by BaseStar::CalculateTimestep() is multiplied by the timestep multiplier specified by the user (via the --timestep-multiplier program option), and clamped to a minimum value of NUCLEAR_MINIMUM_TIMESTEP (currently set to 0.0001 in constants.h) (or should be - see issue #704).

The SSE evolution driver code (the code that steps through the evolutionary process timestep-by-timestep to determine the attributes of the star as it ages) calls the function Star::EvolveOneTimestep() for each iteration of the loop (i.e. every timestep) until evolution stops.

Star::EvolveOneTimestep() calls Star::AgeOneTimestep() to age the star and update the star's attributes as a result of ageing a single timestep.  After the star's attributes have been changed (via Star::AgeOneTimestep()), Star::EvolveOneTimestep() checks how much the star changed over the timestep - specifically it checks the radial change of the star over the timestep.  If the radial change of the star over the timestep exceeds the threshold value of MAXIMUM_RADIAL_CHANGE (currently set to 0.01 in constants.h - 1% of the radius of the star), Star::EvolveOneTimestep() reverts the star to its state at the previous timestep, divides the timestep by 2, and tries the evolution again (by calling Star::AgeOneTimestep()).  This evolve->check->revert->evolve loop is performed repeatedly until either the radial change is below the MAXIMUM_RADIAL_CHANGE threshold, or the timestep reaches the minimum timestep threshold of the maximum of the dynamical timescale for the star (as calculated by BaseStar::CalculateDynamicalTimescale_Static()) and ABSOLUTE_MINIMUM_TIMESTEP (currently set to 100 seconds in constants.h (~3.1688x10^-12)), at which time the evolution of the star over the timestep is accepted and evolution continues with the next timestep.

[...]

PR #1053, opened 2024-01-09
BSE vs SSE changes - see https://github.com/TeamCOMPAS/COMPAS/pull/1052

jeffriley:
See https://github.com/TeamCOMPAS/COMPAS/pull/1052

Ready for review and merge into dev.


jeffriley:
Put back as draft while I fix the pytests...

jeffriley:
Ok, really ready for review and merge into dev now...

online-docs/pages/User guide/timestep-files.rst
```diff
@@ -0,0 +1,28 @@
...
+multiplier set by the ``timestep-multiplier`` option.  Furthermore, no check will be made after the timestep is taken to limit the
+radial expansion of the star being evolved.
+
+Timesteps files can have blank lines and comments. Comments begin with a hash/pound character ('#') - the hash/pound character and text 
+beyond it are ignored by COMPAS.
+
+The ``timesteps-filename`` option is valid both on the command line and in a grid file.  If the ``timesteps-filename`` option is specified
+in conjunction with the ``--number-of-systems`` (``-n``) option, or with option ranges and/or sets, the same timeteps file specified will
+be used for each star or system being evolved. On the other hand, if the ``timesteps-filename`` option is specified in a grid file, the
+timesteps file specified on a grid file line will be read and used for the star or system evolved by that grid file line.
```
avigna, line 29: > On the other hand, if the ``timesteps-filename`` option is specified in a grid file, the
> timesteps file specified on a grid file line will be read and used for the star or system evolved by that grid file line.

I'm not sure I understand how would this work in practice. @jeffriley would you mind sharing a tiny mock code/files to understand it better?

src/BaseBinaryStar.cpp
```diff
@@ -2407,10 +2407,37 @@ EVOLUTION_STATUS BaseBinaryStar::Evolve() {
...
-        double dt      = std::min(m_Star1->CalculateTimestep(), m_Star2->CalculateTimestep()) / 1000.0;                                     // initialise the timestep
-        int    stepNum = 1;                                                                                                                 // initialise step number
+
+        double dt;                                                                                                                          // timestep
+        if (usingProvidedTimesteps) {                                                                                                       // user-provided timesteps?
+            // get new timestep
+            //   - don't quantise
+            //   - don't apply timestep multiplier
+            // (we assume user wants the timesteps in the file)
+            dt = timesteps[0];
```
avigna, line 2432: Small confussion: doesn't `timesteps[0]` means you are always reading the same timestep in the vector?

src/BaseBinaryStar.cpp
```diff
@@ -2486,37 +2518,73 @@ EVOLUTION_STATUS BaseBinaryStar::Evolve() {
...
-                        }
+
+                        // check for reasons to stop evolution
                         else if (IsDCO() && m_Time > (m_DCOFormationTime + m_TimeToCoalescence) && !IsUnbound()) {                          // evolution time exceeds DCO merger time?
                             evolutionStatus = EVOLUTION_STATUS::DCO_MERGER_TIME;                                                            // yes - stop evolution
                         }
                         else if (m_Time > OPTIONS->MaxEvolutionTime()) {                                                                    // evolution time exceeds maximum?
                             evolutionStatus = EVOLUTION_STATUS::TIMES_UP;                                                                   // yes - stop evolution
                         }
+                        else if (!OPTIONS->EvolveDoubleWhiteDwarfs() && IsWDandWD()) {                                                      // double WD and their evolution is not enabled?
```
avigna, line 2536: Aren't these checks similar to those in line 2451?

src/CHeB.cpp
```diff
@@ -1214,7 +1214,7 @@ double CHeB::CalculateRemnantRadius() const {
  * @return                                      Core mass on the First Giant Branch in Msol
  */
 double CHeB::CalculateCoreMassOnPhase(const double p_Mass, const double p_Tau) const {
-    return std::min(((1.0 - p_Tau) * CalculateCoreMassAtHeIgnition(p_Mass)) + (p_Tau * CalculateCoreMassAtBAGB(p_Mass)), m_Mass);               //He mass capped at total mass (should become HeMS star)
+    return std::min(((1.0 - p_Tau) * CalculateCoreMassAtHeIgnition(p_Mass)) + (p_Tau * CalculateCoreMassAtBAGB(p_Mass)), m_Mass);               // We mass capped at total mass (should become HeMS star)
```
avigna: I think this should actually be He, as in Helium mass (and not We).

src/Log.h
```diff
@@ -378,9 +378,9 @@ class FormatVariantValue: public boost::static_visitor<string> {
...
     string operator()(const unsigned long int      v, const string fmtStr) const { string fmt = fmtStr; fmt = "%"  + fmt + "u"; return utils::vFormat(fmt.c_str(), v); } // also handles OBJECT_ID (typedef)
     string operator()(const unsigned long long int v, const string fmtStr) const { string fmt = fmtStr; fmt = "%"  + fmt + "u"; return utils::vFormat(fmt.c_str(), v); }
-    string operator()(const float                  v, const string fmtStr) const { string fmt = fmtStr; fmt = "%"  + fmt + "e"; return utils::vFormat(fmt.c_str(), v); }
-    string operator()(const double                 v, const string fmtStr) const { string fmt = fmtStr; fmt = "%"  + fmt + "e"; return utils::vFormat(fmt.c_str(), v); }
-    string operator()(const long double            v, const string fmtStr) const { string fmt = fmtStr; fmt = "%"  + fmt + "e"; return utils::vFormat(fmt.c_str(), v); }
+    string operator()(const float                  v, const string fmtStr) const { string fmt = fmtStr; fmt = "%"  + fmt + "f"; return utils::vFormat(fmt.c_str(), v); }
```
avigna, line 381: What is the difference between "e" and "f"?

src/Options.h
```diff
@@ -679,12 +683,14 @@ class Options {
             unsigned long int                                   m_RandomSeed;                                                   // Random seed to use
     
             double                                              m_MaxEvolutionTime;                                             // Maximum time to evolve a binary by
-            int                                                 m_MaxNumberOfTimestepIterations;                                // Maximum number of timesteps to evolve binary for before giving up
+            unsigned long int                                   m_MaxNumberOfTimestepIterations;                                // Maximum number of timesteps to evolve binary for before giving up
```
avigna, line 686: Why is it now unsigned long?

src/Star.cpp
```diff
@@ -330,8 +330,8 @@ STELLAR_TYPE Star::UpdateAttributesAndAgeOneTimestep(const double p_DeltaMass,
  *
  *
  * Checks whether the star:
- *    - is a massless remnant (checked after applying p_DeltaMass and p_DeltaMass0)
- *    - has become a supernova (checked after applying p_DeltaMass and p_DeltaMass0)
+ *    - is a massless remnant (checked after applying p_DeltaMass and p_DeltaMass0)  <--- JR: this is no longer true (it may have been once)
+ *    - has become a supernova (checked after applying p_DeltaMass and p_DeltaMass0)  <--- JR this is no longer true (it may have been once)  This causes different behaviour SSE vs BSE
```
avigna, line 337: Delete if no longer true?

src/BaseStar.cpp
```diff
@@ -3982,28 +3989,22 @@ STELLAR_TYPE BaseStar::UpdateAttributesAndAgeOneTimestep(const double p_DeltaMas
...
         if (stellarType == m_StellarType) {                                                 // still on phase?
             
-            UpdateAttributesAndAgeOneTimestepPreamble(p_DeltaMass, p_DeltaMass0, p_DeltaTime);      // apply mass changes and save current values if required
+            UpdateAttributesAndAgeOneTimestepPreamble(p_DeltaMass, p_DeltaMass0, p_DeltaTime); // apply mass changes and save current values if required
 
             if (p_ForceRecalculate                     ||                                   // force recalculate?
                 utils::Compare(p_DeltaMass,  0.0) != 0 ||                                   // mass change? or...
                 utils::Compare(p_DeltaMass0, 0.0) != 0 ||                                   // mass0 change? or...
-                utils::Compare(p_DeltaTime,  0.0)  > 0) {                                   // age/time advance?
+                               p_DeltaTime         > 0) {                                   // age/time advance? (don't use utils::Compare() here)
```
avigna: > (don't use utils::Compare() here)

Why?

src/changelog.h
```diff
@@ -1088,8 +1088,14 @@
 //                                      - Change to functionality (noted above) noted in 'What's New' online documentation page
 // 02.41.04     JR - Dec 30, 2023    - Defect repair:
 //                                      - Fix for issue #1048
+// 02.42.99     JR - Jan 04. 2024    - Enhancements, defect repair, a little cleanup
```
avigna: This should be 02.42.00, right?

src/BaseBinaryStar.cpp
avigna: Ahm it is because this is the constructor, right?

src/constants.h
```diff
@@ -395,11 +400,15 @@ enum class BSE_DETAILED_RECORD_TYPE: unsigned int {
     POST_CEE,                                                                                                       // 12 - record was logged immediately following a common envelope event
     POST_SN,                                                                                                        // 13 - record was logged immediately following a supernova event
     POST_MASS_RESOLUTION,                                                                                           // 14 - record was logged immediately following mass resolution (i.e. after winds mass loss & mass transfer complete)
-    POST_MASS_RESOLUTION_MERGER                                                                                     // 15 - record was logged immediately following a merger after mass resolution
+    POST_MASS_RESOLUTION_MERGER
```
avigna: Why was this deleted?

src/constants.h
```diff
@@ -2822,7 +2847,7 @@ typedef std::tuple<TYPENAME, std::string, std::string, int, int> PROPERTY_DETAIL
 // the logfiles - all keys present here should also be in the STAR_PROPERTIES #define and
 // STAR_PROPERTIES_LABEL
 const std::map<ANY_STAR_PROPERTY, PROPERTY_DETAILS> ANY_STAR_PROPERTY_DETAIL = {
-    { ANY_STAR_PROPERTY::AGE,                                               { TYPENAME::DOUBLE,           "Age",                             "Myr",              16, 8 }},
+    { ANY_STAR_PROPERTY::AGE,                                               { TYPENAME::DOUBLE,           "Age",                             "Myr",              24, 15}},
```
avigna, line 2852: What do these numbers represent?

avigna (changes requested):
Looks good to me. Some comments, mostly questions, and very minor changes might be needed. Always a pleasure to read your code, @jeffriley

jeffriley:
@avigna I'll answer your questions in a bit.  Putting this back in draft - there's a problem at the CHeB/HeMS boundary that I thought I had resolved, but apparently not...  Stay tuned.

ilyamandel:
@jeffriley , I am back and happy to add my review to @avigna 's (I think the changes are sufficiently significant that it's useful to have two reviewers), but maybe after you are done with the issue that you discovered and the ones @avigna brought up?  But let me know, happy to do things in any order.

jeffriley:
@ilyamandel @avigna I have made more changes, and believe I have fixed the problem that was happening at the ChEB/HeMS boundary - PR is now ready form final review (hopefully...).  @avigna answers to your questions below.


> I'm not sure I understand how would this work in practice. 
> @jeffriley would you mind sharing a tiny mock code/files to understand it better?

The option `--timesteps-filename` is a valid option for use in a grid file.  That means, for each system in a grid file, a timesteps file can be specified using the `--timesteps-filename` option.  

For example, if you were to run COMPAS twice to produce two wide binaries:

    ./compas -n 1 --random-seed 0 --initial-mass-1 45.0 --initial-mass-2 0.1 --semi-major-axis 1e20 --detailed-output --logfile-type csv
    ./compas -n 1 --random-seed 1 --initial-mass-1 25.0 --initial-mass-2 0.1 --semi-major-axis 1e20 --detailed-output --logfile-type csv

and then harvest dT from each of the resultant detailed output files (for record type 4 only) and produce two timesteps files called `timesteps1.txt` and `timesteps2.txt`, then construct a grid file named `grid.txt` as follows:

    --random-seed 0 --initial-mass-1 45.0 --initial-mass-2 0.1 -- semi-major-axis 1e20 --timesteps-filename timesteps1.txt
    --random-seed 1 --initial-mass-1 25.0 --initial-mass-2 0.1 -- semi-major-axis 1e20 --timesteps-filename timesteps2.txt

and then run COMPAS using the grid file:

    ./compas --grid grid.txt

the following output would be produced:

    COMPAS v02.42.00
    A binary star simulator
    
    Start generating binaries at Thu Feb  8 07:13:14 2024
    
    0: Allowed time exceeded: (Main_Sequence_>_0.7 -> Black_Hole) + (Main_Sequence_<=_0.7 -> Main_Sequence_<=_0.7)
    1: Allowed time exceeded: (Main_Sequence_>_0.7 -> Black_Hole) + (Main_Sequence_<=_0.7 -> Main_Sequence_<=_0.7)
    
    Generated 2 of 2 binaries requested
    
    Simulation completed
    
    End generating binaries at Thu Feb  8 07:13:14 2024
    
    Clock time = 0.047551 CPU seconds
    Wall time  = 0000:00:00 (hhhh:mm:ss)

That's a trivial example because we're just reproducing the same binaries.  However, if we change the grid file as follows:

    --initial-mass-1  5.0 --initial-mass-2 0.1 --semi-major-axis 1E20
    --initial-mass-1 10.0 --initial-mass-2 0.1 --semi-major-axis 1E20
    --initial-mass-1 15.0 --initial-mass-2 0.1 --semi-major-axis 1E20
    --initial-mass-1 20.0 --initial-mass-2 0.1 --semi-major-axis 1E20
    --initial-mass-1 25.0 --initial-mass-2 0.1 --semi-major-axis 1E20
    --initial-mass-1 30.0 --initial-mass-2 0.1 --semi-major-axis 1E20
    --initial-mass-1 35.0 --initial-mass-2 0.1 --semi-major-axis 1E20
    --initial-mass-1 40.0 --initial-mass-2 0.1 --semi-major-axis 1E20
    --initial-mass-1 45.0 --initial-mass-2 0.1 --semi-major-axis 1E20
    --initial-mass-1 50.0 --initial-mass-2 0.1 --semi-major-axis 1E20

and then run COMPAS:

    ./compas --grid grid.txt --timesteps-filename timesteps1.txt

the following output would be produced:

    COMPAS v02.42.00
    A binary star simulator
    
    Start generating binaries at Thu Feb  8 07:17:57 2024
    
    0: User-provided timesteps not consumed: (Main_Sequence_>_0.7 -> Naked_Helium_Star_MS) + (Main_Sequence_<=_0.7 -> Massless_Remnant)
    1: User-provided timesteps not consumed: (Main_Sequence_>_0.7 -> Naked_Helium_Star_MS) + (Main_Sequence_<=_0.7 -> Massless_Remnant)
    2: User-provided timesteps not consumed: (Main_Sequence_>_0.7 -> Naked_Helium_Star_Hertzsprung_Gap) + (Main_Sequence_<=_0.7 -> Massless_Remnant)
    3: User-provided timesteps not consumed: (Main_Sequence_>_0.7 -> Naked_Helium_Star_Hertzsprung_Gap) + (Main_Sequence_<=_0.7 -> Massless_Remnant)
    4: User-provided timesteps not consumed: (Main_Sequence_>_0.7 -> Naked_Helium_Star_Hertzsprung_Gap) + (Main_Sequence_<=_0.7 -> Massless_Remnant)
    5: User-provided timesteps not consumed: (Main_Sequence_>_0.7 -> Naked_Helium_Star_Hertzsprung_Gap) + (Main_Sequence_<=_0.7 -> Massless_Remnant)
    6: User-provided timesteps not consumed: (Main_Sequence_>_0.7 -> Naked_Helium_Star_Hertzsprung_Gap) + (Main_Sequence_<=_0.7 -> Massless_Remnant)
    7: User-provided timesteps not consumed: (Main_Sequence_>_0.7 -> Naked_Helium_Star_Hertzsprung_Gap) + (Main_Sequence_<=_0.7 -> Massless_Remnant)
    8: Allowed time exceeded: (Main_Sequence_>_0.7 -> Black_Hole) + (Main_Sequence_<=_0.7 -> Main_Sequence_<=_0.7)
    9: Allowed time exceeded: (Main_Sequence_>_0.7 -> Black_Hole) + (Main_Sequence_<=_0.7 -> Main_Sequence_<=_0.7)
    
    Generated 10 of 10 binaries requested
    
    Simulation completed
    
    End generating binaries at Thu Feb  8 07:17:57 2024
    
    Clock time = 0.11557 CPU seconds
    Wall time  = 0000:00:00 (hhhh:mm:ss)

where we can see that the 10 binaries produced had different initial masses, but all used the same timesteps file.  Would you want to do that?  Maybe not, but the example shows that the `timesteps-filename` option can be used with the command line and grid files in the same way other options are used - just highlighting the flexibility COMPAS gives users to combine options on both the command line and in grid files.

> Small confussion: doesn't timesteps[0] means you are always reading the same timestep in the vector?
> Ahm it is because this is the constructor, right?

It's because it's where the initial timestep is set - it's outside the loop.  We set each subsequent timestep inside the loop.

> Aren't these checks similar to those in line 2451?

They are, but these are after some evolution has happened.  This needs to be cleaned up at some stage (a (near) future project) - we can probably just make those checks once at the start (or end) of the evolution loop, but at the moment we do some evolution in between the checks at the end of the loop, and the checks near the start of the loop.  I just need to make sure we can move them all to the top - or bottom - of the loop safely.

> I think this should actually be He, as in Helium mass (and not We).

Indeed :-)  Fixed.

> What is the difference between "e" and "f"?

"e" will use scientific notation (mantissa/exponent) for the output string, "f" will not (it will just output the number in decimal notation).  I made the change so that dT would be output in decimal notation with 15 decimal places.  The change only affects CSV, TSV and TXT files (HDF5 files store the binary representation of the floating-point value).  (See also an answer below)

> Delete if no longer true?

Yes, but only after I rationalise those functions - again, a (very near) future project (high on my to-do list)

>> (don't use utils::Compare() here)
> Why?

Because dT can be as small as ~E-12.  We could use a tolerance there, but the global tolerance probably would not be small enough, so we would have to specify the tolerance at that call to Utils::Compare() (that was not possible before this PR, but now is).  However, as part of this PR, dT is quantised, so we shouldn't need a tolerance for dT.  (We don't actually use a global tolerance for floating-point comparisons yet - but we probably should).

> This should be 02.42.00, right?

Yes it should - fixed.  Good catch.

> Why was this deleted?

I don't know - it's in the code now.  Maybe accidentally and I put it back in - don't remember...

> What do these numbers represent?

They are the format specifiers for CSV, TSV, and TXT files - related to the "e" anf "f" above.  The first number specifies the field width, the second, the precision.  They are described briefly in `constants.h` at about line 2827.  A more detailed description is here: https://cplusplus.com/reference/cstdio/printf/

avigna:
Thanks for the replies Jeff! So far so good, today or tomorrow I will try take a look into the most recent changes you've made.

src/BaseBinaryStar.cpp
```diff
@@ -2407,10 +2407,37 @@ EVOLUTION_STATUS BaseBinaryStar::Evolve() {
...
+        double dt;                                                                                                                          // timestep
+        if (usingProvidedTimesteps) {                                                                                                       // user-provided timesteps?
+            // get new timestep
+            //   - don't quantise
+            //   - don't apply timestep multiplier
+            // (we assume user wants the timesteps in the file)
+            dt = timesteps[0];
+        }
+        else {                                                                                                                              // no - not using user-provided timesteps
+            dt = std::min(m_Star1->CalculateTimestep(), m_Star2->CalculateTimestep()) * OPTIONS->TimestepMultiplier() / 1000.0;             // calculate timestep - make first step small
```
avigna, line 2435: Do we need line 2435? Wouldn't line 2436 suffice and be more consistent with our current quantization of timesteps?

src/CHeB.cpp
avigna: I thought you have changed this to "He"

src/changelog.h
avigna: I also thought you have agreed to change this (am I doing a review on the wrong version of the files?)

avigna (commented):
I added a couple of very minor comments. Additionally, I was a bit confused that two minor changes didn't appeared as addressed. Otherwise, good to go.

jeffriley:
> I thought you have changed this to "He"

I did - I just didn't push that change ;-)  Now pushed.

src/BaseBinaryStar.cpp
avigna: @jeffriley this is my only pending item (unless you have made more edits in the last few hours), happy to approve the PR when this is answered/addressed :)

jeffriley:
Thanks @avigna 

> Do we need line 2435? Wouldn't line 2436 suffice and be more consistent with our current quantization of timesteps?

We need 2435, but we could probably combine 2435 and 2436 - I left them separate just to make it clearer.  Maybe that failed...

2435 calculates the timestep for the binary based on the constituent stars' calculated timesteps (we take the minimum) - but here we just divide that by 1000 because we want the very first timestep to be small so that we don't step over something important before we settle into the evolution.  We could, maybe, just choose an arbitrarily small timestep for the first timestep (i.e. don't bother asking the constituent stars what their calculated timestep is) - maybe we could just use the minimum timestep here?

avigna:
> Thanks @avigna
> 
> > Do we need line 2435? Wouldn't line 2436 suffice and be more consistent with our current quantization of timesteps?
> 
> We need 2435, but we could probably combine 2435 and 2436 - I left them separate just to make it clearer. Maybe that failed...
> 
> 2435 calculates the timestep for the binary based on the constituent stars' calculated timesteps (we take the minimum) - but here we just divide that by 1000 because we want the very first timestep to be small so that we don't step over something important before we settle into the evolution. We could, maybe, just choose an arbitrarily small timestep for the first timestep (i.e. don't bother asking the constituent stars what their calculated timestep is) - maybe we could just use the minimum timestep here?

Yes, I was even considering the possibility of using the quanta of the timestep (or a fixed multiple of it). That is probably too small compared to the other one, but the ad hoc "divide over 1000" seems awkard now. That said, the rationale on it is that you compute the "guessed" timestep according to SSE and then make it somewhat smaller... but probably not at small as the quanta. You could maybe add a comment about this (or not).

In any case, happy with this PR, now approved.

avigna (approved):

jeffriley:
@avigna We have the absolute minimum timestep defined in `constants.h` as

    constexpr double ABSOLUTE_MINIMUM_TIMESTEP              = 3.0 * TIMESTEP_QUANTUM;                                   // In Myr (=94.6728 seconds, given TIMESTEP QUANTUM)

We could use that as the first timestep.

@ilyamandel ready for your review now - if you still want to review

avigna:
Does this PR also addresses issue #705 ?

jeffriley:
> Does this PR also addresses issue https://github.com/TeamCOMPAS/COMPAS/issues/705 ?

I don't know that it fixes it, but I think it at least provides tools to help measure convergence, and why we're not getting convergence.  It's certainly worth revisiting https://github.com/TeamCOMPAS/COMPAS/issues/705 after this is merged.

ilyamandel (commented):
Thank you for this PR, @jeffriley !

I am listing things as comments because I am happy for you to merge once these are addressed to your satisfaction (no need for another review from me).  Some (like item 4) do not require any action; others (like item 5) should really be addressed via small cosmetic changes; yet others (like item 1) are up to you.

1. In the new timestep file documentation, it would be nice to explicitly state what will happen if the number of specified timesteps is less than the duration that the code wishes to evolve for (i.e., code still wants to run, but we've exhausted the user-specified time steps).  [BTW, I am not sure we need to show a warning when user timesteps aren't consumed; I think we are doping the right thing to show one if they are exhausted...]. Also, I think it would be useful to provide a short example of what the timestep file ought to look like in the corresponding documentation. 

2. It seems that there's repetition of checks between lines 2450+ and ~2540+ in the new BaseBinaryStar.cpp file.

3. BaseStar.cpp, line 2790, comment: "and the change to the line above" -- it won't be clear to whoever is reading the code in 2 months what the change was :). Ditto for "after the switch" comment on line 469 of Star.cpp.

4. Not a huge fan of forceRecalculate parameter, but neither are you :-), so this is just a reminder that it should eventually be resolved (not in this PR)

5. BaseStar.cpp, line 4097: "STELLAR_TYPE EvolveOnPhase()"  -- add description of newly added argument to comments.

6. Star.cpp, line 552: "If we remove this, we have to change BS accordingly" BS->BSE

7. Like @avigna , I am surprised by the deletion of the comment on line 403 of constants.h (explaining POST_MASS_RESOLUTION_MERGER)

8. Constants.h, line 568 "EMPTY_FILE, // filename is empty (contains no content)": filename -> file?

9. utils.cpp, line 1497: "if (!rec.empty() or rec[0] == '#')" -- should this read "if (!(rec.empty() or rec[0] == '#'))" ?

10. program-options-list-defaults.rst should include the new timestep-file option in the bottom section (listing options by type)

jeffriley:
Review comments addressed.

To do items (here as a reminder):

1. Rationalise "UpdateAttributes*" and "EvolveOneTimestep*" functions
2. Remove duplicated checks in BaseBinaryStar::Evolve() - will require rewriting of the function
3. Rewrite to remove p_ForceRecalculate parameter
4. Revisit p_ResolveEnvelopeLoss parameter - it's there to ensure BSE and SSE evolve the same way given the same timesteps, but is there a better way?

Issue #1033, opened 2023-12-16
BSE_Supernovae/Mass_Total@CO(SN) is occasionally zero


jeffriley:
**Describe the bug**
BSE_Supernovae/Mass_Total@CO(SN) is occasionally zero

**Label the issue**
`urgency_moderate` - This is a moderately urgent issue
`severity_moderate` - This is a moderately severe bug

**To Reproduce**
Using COMPAS v02.41.02, run

    ./compas -n 1 --metallicity 0.02  --random-seed 1702719075

**Expected behavior**
BSE_Supernovae/Mass_Total@CO(SN) should not be zero

**Screenshots**

**Versioning (please complete the following information):**
 - OS: Ubuntu 22.04
 - COMPAS v02.41.02

**Additional context**

jeffriley:
I believe this defect was introduced in this commit:

https://github.com/TeamCOMPAS/COMPAS/commit/4ca3ad4c88801b7ffb7a071a102637449435d545

as part of the work to add Accretion Induced Collapse (AIC) of ONeWD as another type of SN in COMPAS v02.31.08

Prior to the commit noted above, `ONeWD::ResolveSupernova()` called `GiantBranch::ResolveSupernova()`, which sets the variables

        m_SupernovaDetails.totalMassAtCOFormation  = m_Mass;
        m_SupernovaDetails.HeCoreMassAtCOFormation = m_HeCoreMass;
        m_SupernovaDetails.COCoreMassAtCOFormation = m_COCoreMass;
        m_SupernovaDetails.coreMassAtCOFormation   = m_CoreMass;

The commit noted above changed `ONeWD::ResolveSupernova()` to call `ONeWD::ResolveAIC()`.  At some stage (I haven't checked where yet), this was refined so that in the current release (v02.41.02), `WhiteDwarfs::ResolveSupernova()` (called by all WDs) calls `EvolveToNextPhase()`.  `ONeWD::EvolveToNextPhase()` calls `ONeWD::ResolveAIC()`.

The bottom line is that `ONeWD::ResolveAIC()`, so AIC SN events, doesn't set the `...AtCOFormation` variables that are set in `GiantBranch::ResolveSupernova()` - they are left at the default value of 0.0.  They need to be set.

[...]

PR #1055, opened 2024-01-16
Fix pre aic mass recording

reinhold-willcox:
Resolves issue #1033.


ilyamandel (commented):
@reinhold-willcox -- I am not sure I understand why the various core masses are set to zero on AIC; can you please explain your logic?
[Sorry for taking so long to review this because of travel...]

reinhold-willcox:
@ilyamandel I've readded in the correct masses, sorry for that (and the delay).

PR #1059, opened 2024-01-31
1058 bug calculating NS spin period in NS.cpp

### LINKED ISSUE #1058: Bug calculating NS spin period in NS.cpp
**Describe the bug**
In line 390 in NS.cpp, the calculation of initial spin period of a neutron star has a typo. 

It's suppose to be (2*pi/f):
double initialSpinPeriod      = _2_PI / m_PulsarDetails.spinFrequency;

However is is currently coded in as (pi^2/f):
double initialSpinPeriod      = PI_2 / m_PulsarDetails.spinFrequency;

This is causing errors in the NS spin period calculation during spin-down. 

**Label the issue**

`urgency_high`     - This is a very urgent issue and should be resolved as soon as possible

 - COMPAS v02.41.04

yuzhesong:


ilyamandel (approved):
Looks good -- thanks for the fix, @yuzhesong !

PR #1068, opened 2024-02-21
Fix for issue 1066

### LINKED ISSUE #1066: Omega sometimes not calculated correctly when tides enabled
**Describe the bug**
Omega sometimes not calculated correctly when tides enabled

**Label the issue**

**To Reproduce**
Run a single binary with the following initial attributes:

./COMPAS --number-of-systems 1 --random-seed 0 --enable-tides TRUE --initial-mass-1 150.0 --semi-major-axis 0.5 --initial-mass-2 1.0 --detailed-output --logfile-type csv 

**Expected behavior**
Omega calculated correctly

**Screenshots**

 - OS: Ubuntu 22.04]
 - COMPAS:  v02.42.00

**Additional context**

jeffriley:
The problem here was that the root finder was not correctly finding the root in BaseBinaryStar::OmegaAfterSynchronisation() because the search step size was too large.

Search step size here is problematic - too small and the root finder bumps up against maximum iterations; too large and this sort of thing happens. Here the guess for the root finder was very close to an inflection point on the curve, and the gradient between the guess and the upper search step was rising, but the lower search step and the guess straddled an inflection point, and the gradient between the lower search step and the guess was falling. The root finder stepped to the lower search step, didn't find the root, so stepped towards zero (because we told it the gradient was rising...), and eventually walked the root estimate all the way down to zero (or ~1E-318). The root finder didn't return an error, or throw an exception - it just returned a value of ~1E-318. I don't know if that's expected behaviour - I haven't found any documentation o n that behaviour.

The solution is to trap this outcome, reduce the search step size, and try finding the root again - iteratively (until we bump up against max iterations...).

### DIFF
--- a/src/BaseBinaryStar.cpp
+++ b/src/BaseBinaryStar.cpp
@@ -1924,8 +1924,7 @@ void BaseBinaryStar::CalculateMassTransfer(const double p_Dt) {
         else {                                                                                                                  // donor has no envelope
             massDiffDonor = MassLossToFitInsideRocheLobe(this, m_Donor, m_Accretor, m_FractionAccreted);                        // use root solver to determine how much mass should be lost from the donor to allow it to fit within the Roche lobe
             if (massDiffDonor <= 0.0) {                                                                                         // no root found - no mass can be lost from donor
-                // if donor cannot lose mass to fit inside Roche lobe,
-                // the only viable action is to enter CE phase
+                // if donor cannot lose mass to fit inside Roche lobe, the only viable action is to enter CE phase
                 m_CEDetails.CEEnow = true;                                                                                      // flag CE
             }
             else {                                                                                                              // have required mass loss
--- a/src/BaseBinaryStar.cpp
+++ b/src/BaseBinaryStar.cpp
@@ -1923,7 +1923,7 @@ void BaseBinaryStar::CalculateMassTransfer(const double p_Dt) {
         }
         else {                                                                                                                  // donor has no envelope
             massDiffDonor = MassLossToFitInsideRocheLobe(this, m_Donor, m_Accretor, m_FractionAccreted);                        // use root solver to determine how much mass should be lost from the donor to allow it to fit within the Roche lobe
-            if (massDiffDonor <= 0.0) {                                                                                         // no root found - no mass can be lost from donor
+            if (massDiffDonor <= 0.0) {                                                                                         // no root found
                 // if donor cannot lose mass to fit inside Roche lobe, the only viable action is to enter CE phase
                 m_CEDetails.CEEnow = true;                                                                                      // flag CE
             }
--- a/src/BaseBinaryStar.cpp
+++ b/src/BaseBinaryStar.cpp
@@ -1917,32 +1917,44 @@ void BaseBinaryStar::CalculateMassTransfer(const double p_Dt) {
         double envMassDonor    = m_Donor->Mass() - m_Donor->CoreMass();
         bool isEnvelopeRemoved = false;
 
-        if (utils::Compare(m_Donor->CoreMass(), 0) > 0 && utils::Compare(envMassDonor, 0) > 0) {                                // donor has a core and an envelope
-            massDiffDonor     = -envMassDonor;                                                                                  // set donor mass loss to (negative of) the envelope mass
+        if (utils::Compare(m_Donor->CoreMass(), 0) > 0 && utils::Compare(envMassDonor, 0) > 0) {                                // donor has a core and an envelope?
+            massDiffDonor     = -envMassDonor;                                                                                  // yes - set donor mass loss to (negative of) the envelope mass
             isEnvelopeRemoved = true;
         }
-        else{                                                                                                                   // donor has no envelope
-            massDiffDonor = -MassLossToFitInsideRocheLobe(this, m_Donor, m_Accretor, m_FractionAccreted);                       // use root solver to determine how much mass should be lost from the donor to allow it to fit within the Roche lobe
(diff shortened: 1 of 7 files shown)


ilyamandel (changes requested):
Thank you, @jeffriley !

I might have put the calculation of is_rising at the very start of the while loop (to avoid having the same code appear twice, before the while loop and at its end), but that's just a personal preference.

What worries me a bit more is whether a root that's too small is the only failure mode.  Could we ever go the other way and end up with roots that are too large?  My suggestion would be a bit different.  Let's check whether the root is *good enough* -- i.e., whether, with the best guess that the root-finder returns, the function is solved to an acceptable tolerance level (which is already determined by tol).  If it's not, then we know the root finder failed, and we need to iterate.

jeffriley:
Thanks @ilyamandel 

> I might have put the calculation of is_rising at the very start of the while loop ...

I put it where it is to ensure we only recalculated when we need to - but I think you're right - the only reason we iterate is if we change the factor, so putting the calculation of isRising at the start of the loop is safe (and better).

> Let's check whether the root is good enough ...

I agree that the condition as I have it is arbitrary.  I had thought of trying to determine if the step boundaries were straddling an inflection point, but anything I came up with was going to be too costly.  Are we guaranteed that if the solution found by the root finder is outside tolerance, changing the search step size will eventually result in a solution that is within tolerance?  That is, if our check is the solution returned by the root finder is out of tolerance, could we enter an infinite loop?  Should we add an absolute stopping condition (is there one?)?  Maybe we could put a limit on the number of iterations, and if that is reached we just fail (in the same way we fail if the root finder exceeded the maximum number of iterations)?

ilyamandel:
I agree, we should put in a maximum on how many times we go through the while loop.  I'd say 30 is a reasonable default number, since we end up with the (factor-1) dropping by 10^9 after 30 divisions by 2.

BTW, stupid question, but how does the variable "it" get updated -- it looks like it's being passed by value, not by reference?

jeffriley:
Ok, I'll use 30 iterations as the default maximum.

"it" is declared in bracket_and_solve_root() as a call-by-reference parameter (probably "(boost::uintmax_t) &it"), so the compiler takes care of passing the address rather than the value (different from passing a pointer).

jeffriley:
@ilyamandel while I'm adding this functionality to OmegaAfterSynchronisation(), should we consider a similar treatment for MassLossToFitInsideRocheLobe()?

veome22 (approved):
The omega solution finding works now :) 
The solution found by COMPAS can differ from mine by ~10%, which is explicitly computed over a range. But I am not too worried about this.

ilyamandel:
> should we consider a similar treatment for MassLossToFitInsideRocheLobe()?

Sure, let's do it! ;-)

jeffriley:
@ilyamandel

Changes made per review comments and related discussions:

   - Cleaned up root solvers OmegaAfterSynchronisation(), MassLossToFitInsideRocheLobe(), and Mass0ToMatchDesiredCoreMass(), and their respective functors.  Replaced Boost tolerance function with our own; added loop to retry with a smaller search step.

   - MassLossToFitInsideRocheLobe(), and Mass0ToMatchDesiredCoreMass() now return -1.0 if no acceptable root found

   - Calling code for MassLossToFitInsideRocheLobe() and Mass0ToMatchDesiredCoreMass() now handles -ve return:

      - if MassLossToFitInsideRocheLobe() returns -ve value (i.e. no root found), the binary immediately enters a CE phase
      - if Mass0ToMatchDesiredCoreMass() returns -ve value (i.e. no root found), an arbitrary value is used for core mass (see code for value)

I added descriptions for MassLossToFitInsideRocheLobe(), and Mass0ToMatchDesiredCoreMass(), and their respective functors. 
 Please update (or reply here and I'll update) the descriptions if necessary.  I couldn't remember the rationale for how -ve return from Mass0ToMatchDesiredCoreMass() is handled in HG.h, so if you could add that (here or in the code), that'd be great.  Also, how does the comment in HG.h "...but no more than total mass" reconcile with the choice of `2*std::min(m_Mass, MinimumCoreMass())`?

For the Mass0ToMatchDesiredCoreMass() root solver, I ran a test of 10,000 binaries, and the root solver did not return a -ve value.

For the MassLossToFitInsideRocheLobe() root solver, I ran a test of 1,000 binaries, and 41 binaries returned -ve values, and of those 30 merged anyway, so the results were not different.  Of the 11 remaining, 6 that were "Unbound Binaries" before the change instead merged due to CE, and 5 that evolved until they ran out of time instead merged due to CE.

@veome22 can you check that these new changes produce what you expect from tides?

jeffriley:
@ilyamandel @veome22 Root finder algorithms updated.  I'm much happier with the updated version - hopefully you are too :-)  Ready to review.

ilyamandel (approved):
Thank you, @jeffriley !

I made a few minor changes to comments directly.

I changed the doubling of the core mass on no root find to instead read

                // if no root found we keep m_Mass0 equal to the total mass
                m_Mass0 = m_Mass;
                
(as you said, we never actually hit this)

Line 553 of BaseBinaryStar.h: How about returning something like 1000.0*ROOT_ABS_TOLERANCE in this case?  The point is to show that we have failed to find a sensible solution.  I haven't made this change, but recommend something like it.

However, approving now and leave that final choice to you.

jeffriley:
> Line 553 of BaseBinaryStar.h: How about returning something like 1000.0*ROOT_ABS_TOLERANCE in this case? 

Done.  @veome22 just waiting on your final confirmation that tides work as you expect, and then I'll merge into dev.

veome22:
Just ran the updated code, looks good to me!

PR #1081, opened 2024-03-29
Ge qcrit nonconservative

reinhold-willcox:
Updated the table of critical mass ratio (qcrit) values, based on a new (unpublished) table from Hongwei Ge that account for non-conservative MT. The code now interpolates between these two extremes based on the accretion fraction, so that in the stability of mass transfer accounts for both delayed dynamical (in)stabilities, as well as variations to the accretion fraction. 

Notably, the table still does not include variations to the AM loss, so this prescription is only valid for the isotropic re-emission AM loss model, but its still a big step forward. 

I'm leaving this PR as a draft for now since I haven't done any testing, beyond confirming that it compiles, but if there is anthing missing or that needs to be changed, I can do that in parallel.


reinhold-willcox:
@jeffriley This PR is ready for checking. The code compiles, though there are many warnings (if these are undesired, let me know how I should go about addressing them). 

Below is a screenshot comparing COMPAS before and after the changes. Apologies that it's not very pretty. The old default (beta thermal) and beta=1 are in exact agreement by design, because we use to enforce that if Ge was used, beta should be fixed to 1. The new version does not have that restriction, which is why we see variation for different beta prescriptions. 

Notably, I had expected that new/beta1 should exactly equal old/default and old/beta1, since there the grid of critical mass ratios are identical. However, I also discovered a bug that occurs in some corner cases, where if the stellar radius in compas was larger than the largest supplied value from the Ge tables, the interpolator would return a `nan`. This is an unfortunate bug that affects the results of my 2023 paper, but it is at least corrected in this PR. 



image

jeffriley:
@reinhold-willcox you'll need to fix the warnings.  They are happening because you changed the signature of BaseStar::CriticalMassratio(), but that function is also declared in Star.h and Remnants.h - you need to change the signatures there to make the warnings go away (you can give the new parameter a default value if it is not required there).  The warnings indicate that you won't be getting runtime polymorphism (because the signatures are different).

reinhold-willcox:
@jeffriley Thanks! should be fixed now.

jeffriley (approved):

Just one question.  In `constants.h` you have:

"This is a subset of an unpublished table from Hongwei Ge, which replaces the previous table 3 from Ge et al. 2020."

I know you said that Hongwei gave you permission to use the table, but you mention the table is (as yet?) unpublished - do we have permission to publish the data (COMPAS code is public, so by including it in the source files we're effectively publishing it - we're at least making it publicly available).

I'll approve the PR and leave it to you to decide whether it should be merged now or whether we need o ask the question.

ilyamandel (changes requested):
In BaseStar::CalculateCriticalMassRatio(), you've added a new argument to the function; the comments need to be updated to describe this new argument

BaseStar::InterpolateGe20QCrit() needs an explanation of the arguments in the comments, too

`double interpolatedQCrit = p_massTransferEfficiencyBeta * interpolatedQCritNonc + (1-p_massTransferEfficiencyBeta)*interpolatedQCritFull;`
Isn't that backwards?  beta=1 corresponds to fully conservative MT, right?

Is Hongwei Ge OK with this unpublished work effectively being released through a public code?

jeffriley (changes requested):
> ...  the comments need to be updated to describe this new argument

Good catch @ilyamandel - I missed that.

@reinhold-willcox, in the line that @ilyamandel pointed out:

     double interpolatedQCrit = p_massTransferEfficiencyBeta * interpolatedQCritNonc + (1-p_massTransferEfficiencyBeta)*interpolatedQCritFull;

please change `1-p_massTransferEfficiencyBeta` to `1.0 - p_massTransferEfficiencyBeta`

reinhold-willcox:
@jeffriley @ilyamandel changes addressed, and I forward an email regarding permissions from Hongwei.

ilyamandel (approved):
Looks good -- thank you, @reinhold-willcox !

ilyamandel:
@jeffriley -- yours to approve (I've already merged the changelog.h changes)

jeffriley (approved):
Thanks @reinhold-willcox

PR #1083, opened 2024-04-04
Realistic Tides

veome22:
- Added 'realistic' tides option, which implements dynamical tides. Functionality enabled with new option `--enable-realistic-tides`.  Default is no tides.
- Dynamcial tides implementation follows Zahn, 1977, Eq. (5.5) , with the value of E_2 coming from Kushnir et al., 2017, Eq. (8).
- Secular evolution under the effect of tides follows Zahn, 1977, Eqs. (3.6) to (3.8)
- Added function `CalculateCoreRadius()`. Currently returns the same value as `CalculateRemnantRadius()` for Giant Branch stars, returns 0.0 otherwise
- Added new `CoreRadius()` member variable which can also be accessed in the Log file as CORE_RADIUS (Star property)


ilyamandel (commented):
Hi @veome22 ,

This looks very nice, though I haven't checked the implementation details carefully yet.  One general suggestion below:

I think --enable-tides and --enable-relastic-tides is becoming rather confusing.  @jeffriley should weigh in, but I would propose a tides option that could take on a few string values, such as "NONE", "INSTANTANEOUS" or "PERFECT", and "KAPIL2024" (to match the description in your upcoming paper -- some day we may have a different "realistic" version).

@jeffriley will have comments on code layout (e.g., your choice of one argument per line does not match the usual COMPAS code convention). :) 

Let's talk more about the core radius assumption tomorrow; and I know you wanted to add non l=m=2 components.

veome22:
> I think --enable-tides and --enable-relastic-tides is becoming rather confusing. @jeffriley should weigh in, but I would propose a tides option that could take on a few string values, such as "NONE", "INSTANTANEOUS" or "PERFECT", and "KAPIL2024" (to match the description in your upcoming paper -- some day we may have a different "realistic" version).

I agree completely. The current option name is quite confusing indeed. 

> @jeffriley will have comments on code layout (e.g., your choice of one argument per line does not match the usual COMPAS code convention). :)

I tried following the 'naive tides' function implemented by @jeffriley, but I am happy to improve the style. Also since I was trying to maintain encapsulation, I ended up passing a lot of the same arguments repeatedly, which makes for very long and unwieldy code. I would appreciate any suggestions on how to re-write the code in `BaseBinaryStar.cpp`.

veome22:
Hi @ilyamandel and @jeffriley, I have made the following updates to the code:

1. `CalculateImKlm()` now returns a `std::tuple <double, double, double, double>` with the 4 moments required to compute the secular evolution of the binary.
2. Tides are now enabled by setting a new option `--tides-prescription {NONE, PERFECT, or KAPIL2024}`. Default is `NONE`.
3. The main code in `BaseBinaryStar.cpp` is a bit cleaner I think. The ImKlm terms for each star are now computed only once per loop, and then passed as arguments to the secular evolution functions.

ilyamandel (changes requested):
Thank you, @veome22 !

I am almost ready to approve, but would like @jeffriley to review as well, particularly with regards to code layout choices (e.g., around line 2350 of BaseBinaryStar.cpp).

I do have a few comments:

There's no need to describe new variables in What's New (also, CoreRadius() is a function, not a member variable), but you should mention that the --enable-tides option has disappeared.

I am not a fan of keeping m_CoreRadius as a variable, because then we have to worry about keeping it up-to-date at all times.  I think it would be better to just compute the core radius on the fly when needed, since it's effectively what you are doing now, anyway.

Because CalculateImKlmTidal() is defined in GiantBranch.cpp, it will be inherited by things like HeMS; if you don't mean to do that, you'll have to redefine it appropriately along the inheritance hierarchy. 

Ditto for CalculateCoreRadius() -- though this could be OK if CalculateRemnantRadius() does the right thing... But I don't think it does at the moment [@jeffriley ?] :
double TPAGB::CalculateRemnantRadius() const {
    return HeWD::CalculateRadiusOnPhase_Static(m_CoreMass);
}

Why does the comment on line 2264 of Options.cpp refer to // Chemically Homogeneous Evolution ?

veome22:
Thanks @ilyamandel for going through the code! I'm putting a comment here for future reference.

- The `m_CoreRadius` variable was indeed unnecessary, it's been removed now.
- The class hierarchy for `CalculateImKlmTidal()` has been fixed.
- I've updated the documentation to reflect the removal of the `--enable-tides` option.
- I am going to merge the PickerUpdate branch into this one, so the `CalculateCoreRadius()` will be replaced by `CalculateConvectiveCore()` shortly. That should take care of the class hierarchy for the core radius.

veome22:
Hi @ilyamandel and @jeffriley!

The tides formalism is now complete 🙂 . Please take a look and let me know if there are any bugs / stylistic issues I should address. Thanks!

ilyamandel (commented):
Were the deletions on lines 363-366 of standard-logfiles-record-specification-stellar.rst intentional?

Added new function CalculateCoreRadius() in whats-new.rst

BaseBinaryStar, line 2406: why not use OrbitalAngularVelocity() to compute m_Omega?

In all comments, love number -> Love number (they are actually named after a guy whose last name was Love, not because they are lovely :) )

BaseStar.cpp, CalculateImKlmDynamical definition -- comments should describe all parameters

GiantBranch.cpp, line 1028 -- return type in comment should match DBL_DBL as in function definition

ilyamandel (changes requested):
Were the deletions on lines 363-366 of standard-logfiles-record-specification-stellar.rst intentional?

Do we actually use CalculateCoreRadius() anywhere?  Seems like we only use CalculateConvectiveCoreRadius() .  If so, let's remove CalculateCoreRadius() everywhere (including the CORE_RADIUS output variable).

BaseBinaryStar, line 2406: why not use OrbitalAngularVelocity() to compute m_Omega?

In all comments, love number -> Love number (they are actually named after a guy whose last name was Love, not because they are lovely :) )

BaseStar.cpp, CalculateImKlmDynamical definition -- comments should describe all parameters (check for other occurrences)

GiantBranch.cpp, line 1028 -- return type in comment should match DBL_DBL as in function definition

GiantBranch.cpp, around line 1130: I think you just want to write this as a one-line in GiantBranch.h: { return CalculateImKlmEquilibrium(p_Omega, p_SemiMajorAxis, p_M2); }

Alternatively (and I personally think this would be much cleaner and would avoid excessive code repetition), what about having BaseStar.cpp use the sum of the Dynamical and Equilibrium tides for CalculateImKlmTidal() (as done, e.g., in HG.cpp, MainSequence.cpp -- note repeated code), and then just have the stellar types correctly define whether the values of their Dynamical and Equilibrium coefficients (which can, of course, be zero as appropriate).

Regardless, classes automatically inherit methods from their parent -- so there's no need to overwrite a method unless you want it do so something different from the  parent.  At the moment, it seems that MainSequence uses the sum of Dynamical and Equilibrium tides; its child GiantBranch uses only Equilibrium tides; its child HG uses a sum of Dynamical and Equilibrium tides; and FGB goes back to using only Equilibrium tides.  Could replace four definitions with two by having MainSequence use the sum, and FGB use only Eq tides.  [But see above.]

HeMS.h, line 79 -- why does the comment say "Default is no tidal response"?  Remove?

MainSequence.cpp, line 655: DBL_DBL_DBL_DBL is a cleaner shortcut for the return type (check for other occurrences, e.g., BaseStar::CalculateImKlmDynamical())

It looks like we don't really need m_EnableTides and m_EnableRealisticTides -- we can just query relevant option value of TIDES_PRESCRIPTION; if so, let's remove these.

ilyamandel (changes requested):
BaseStar::CalculateImKlmDynamical(), line 3235: 

 if (utils::Compare(coreRadiusAU + convectiveEnvRadiusAU, radiusAU) >= 0.0) {
        return std::make_tuple(0.0, 0.0, 0.0, 0.0); 
        
For some stellar types, we are assuming that the entire envelope is convective and the rest of the star is a convective core, but in practice, these are always separated by a radiative intershell whose mass and radial thickness tend to zero (except for very low mass <0.35 Msun stars)...

Are Equilibrium tides meant to only have the (2,2) mode defined?

jeffriley (changes requested):
Thanks @veome22 - looks good overall.  Just a few comments in addition to @ilyamandel 's:

BaseBinaryStar.cpp, line 412

    if (OPTIONS->TidesPrescription() == TIDES_PRESCRIPTION::PERFECT && !m_Unbound) {

should be `else if` (saves some compute cycles...)

Maybe change the pattern there to something like this:

    if (!m_Unbound && OPTIONS->TidesPrescription() != TIDES_PRESCRIPTION::NONE) {
    	
        // if m_Omega == 0.0 (should only happen on the first timestep), calculate m_Omega here
        if (utils::Compare(m_Omega, 0.0) == 0) {
            m_Omega = OrbitalAngularVelocity(); 
        }
    
        if (OPTIONS->TidesPrescription() == TIDES_PRESCRIPTION::KAPIL2024) {
    
            // Change binary semi-major axis, eccentricity, and spin of each star based on realistic tidal torque
            // Adjust the binary orbital frequency to match semi-major axis.
            
            DBL_DBL_DBL_DBL ImKlm1 = m_Star1->CalculateImKlmTidal(m_Omega, m_SemiMajorAxis, m_Star2->Mass());
            DBL_DBL_DBL_DBL ImKlm2 = m_Star2->CalculateImKlmTidal(m_Omega, m_SemiMajorAxis, m_Star1->Mass());
    
    	....
    
        }
        else {
    
            // find omega assuming synchronisation
            // use current value of m_Omega as best guess for root
    
            m_Omega = OmegaAfterSynchronisation(m_Star1->Mass(), 
                                                m_Star2->Mass(), 
                                                m_Star1->CalculateMomentOfInertiaAU(), 
                                                m_Star2->CalculateMomentOfInertiaAU(), 
                                                m_TotalAngularMomentum,
                                                m_Omega); 
    
            if (m_Omega >= 0.0) {    
    
    	...
        }
    }

so we don't duplicate code.

(if we ever add another tides option we can change the inner `IF` to a `SWITCH`.

Why are `CalculateDSemiMajorAxisTidalDt()`, `CalculateDEccentricityTidalDt()`, and `CalculateDOmegaTidalDt()` declared in BaseBinaryStar.h?  I'm not convinced the compiler will inline those (it might I suppose, but ordinarily we'd put that much code on the cpp file rather than the header - if they're not inlined having them in the header just increases compile time for no runtime benefit).

In `CalculateDOmegaTidalDt()` in BaseBinaryStar.h, change `1/p_11` to `1.0 / p_11`

BaseStar::CalculateImKlmDynamical():
   - describe all parameters in the function description.
   - line 3235, utils::Compare() returns an integer, not a floating point number.
   - lines 3235/3237, we discourage multiple return points in functions - it's too easy for a reader of the code to miss a return: much better if the reader can assume entry at the top, exit at the bottom.  In this case, since the check is really just a sanity check and exit if we don't need to be there, we can let it slip by, but please rewrite so that the check immediately follows the calculations of the three local variables required for the check (not only is it better closer to the top of the function, there's no need to make the other calculations if we're just going to return anyway), and add a comment indicating that the function will return there and why.
   - `k10GravityEnv` is never calculated - it is only ever set to 0.0 (in two places) - is that deliberate?  If so, the second assignment to 0.0 can be removed.  (Maybe leave an expanded comment in if you want readers to know why it is not calculated).
   - lines 3254-3257, `Omega()` is just a getter for the class variable `m_Omega` which we have access to here - no need to incur the overhead of a function call - just use `m_Omega`.  And maybe declare a local variable for `2.0 * m_Omega` (no need to do the multiplication three times).
   - consider changing `2.0*p_Omega` to `p_Omega + p_Omega` (faster - if was an integer multiply it wouldn't be, but it is a fp multiply).  Similarly for `3.0*p_Omega`.

BaseStar::CalculateImKlmEquilibrium():
   - add `double r_out_2 = rOutAU * rOutAU;`
   - line 3381, change `double r_out_3 = rOutAU * rOutAU * rOutAU;` to `double r_out_3 = r_out_2 * rOutAU;`
   - line 3382, change `double r_out_5 = rOutAU * rOutAU * rOutAU * rOutAU * rOutAU;` to `double r_out_5 = r_out_3 * r_out_2;`
     (saves some multiply operations - doesn't see like much, but if this gets called many times then saving even a single multiply helps...)
   - line 3377, utils::Compare() returns an integer, not a floating point number.
   - line 3392 - same comment re use of `Omega()` as above
   - consider declaring a local variable for `omega_t/omega_conv` - it is currently calculated in several places.
   - ditto for `v_conv * l_conv`
   - ditto for `p_M2 / m_Mass`

BaseStar::CalculateEddyTurnoverTimescale() - are the added lines aligned?  (It might just be how my browser interprets tabs).

Return type `DBL_DBL` for GiantBranch::CalculateConvectiveEnvelopeMass()

Return type `DBL_DBL_DBL_DBL` for GiantBranch::CalculateImKlmTidal()
GiantBranch::CalculateImKlmTidal() - describe all the parameters in the function description

Return type `DBL_DBL_DBL_DBL` for HG::CalculateImKlmTidal()

Question re `CalculateImKlmEquilibrium()` - do we really need to know stellar type to calculate this, or is it some attribute that we can query?  i.e. can we move the function to the BinaryConstituentStar class?  Maybe not...

`compasConfigDefault.yaml` - maybe generate new yaml file (use `--create-yaml-file`).  The change you've made looks right, but the file header indicates it was created with `v02.40.00`, which doesn't have the new tides option...

I'll leave it here for now, and review again after you've made the changes requested.

veome22:
Hi @ilyamandel and @jeffriley!

All the suggested changes have been made :). Some of the highlights include:

1. Removed `CoreRadius()` everywhere, since we indeed now use only the Convective Core functions.
2. Changed the behavior of `CalculateImKlmTidal()` as Ilya suggested. It now always adds up the 2 tidal contributions, and each stellar type can decide which, if any, contribution is 0.
3. As of now, we always calculate Equilibrium + Dynamical tides for all stellar types. If some star has no radiative intershell, then it should be caught by the `CalculateImKlmDynamical()` function automatically. No more stellar type-specific tides. I think this is cleaner overall.
4. Other general cleanup.

@ilyamandel it turns out I overlooked the other (l,m) terms in Barker (2020). I will add (and test) the rest of the terms ASAP.

veome22:
All done!

ilyamandel (approved):
Congratulations, @veome22 !

Very happy to approve! 

@jeffriley , if you are good with this, we can merge.

jeffriley (approved):
Thanks @veome22 - all good.

PR #1086, opened 2024-04-08
Fix Bjorklund Eddington Factor

SimonStevenson:
Fix Bjorklund et al. OB mass loss rate.

The function CalculateMassLossRateBjorklundEddingtonFactor incorrectly used LSOL (in cgs) rather than LSOLW (in SI), resulting in incorrect Gamma values. This PR simply swaps one constant for the other.


ilyamandel (approved):
Looks great -- thank you, @SimonStevenson !

PR #1087, opened 2024-04-09
add check for latex before setting rcParams

avivajpeyi:
Checks if latex is installed. If not, turns latex off from the 'detailed-plotter' script. (See discussion in #1013 )

This seems to do the trick, woohoo! 
Screenshot from 2024-04-09 13-30-06


jeffriley (approved):
Looks good - thanks Avi!

PR #1090, opened 2024-04-14
Defect Repair: Issue 1084

### LINKED ISSUE #1084: Do not log clones
I am getting a lot of these on version 2.42.01 when running with --switch-log:

ERROR: in BinaryStar::bool BinaryStar::PrintSwitchLog(): Value out of bounds; Expected primary or secondary for BSE Switch Log, (ObjectId = 141131)

To reproduce, try
./COMPAS -n 10 --initial-mass-min 0.8 --random-seed 1970 --switch-log

@jeffriley identified this as an issue with logging clones, and we agreed that clones should not be logged in the output records (they are only used for temporary ease of access and will contain information that would be confusing to the typical user).  This issue is just a reminder to disable clone logging.

jeffriley:
Defect repair, some code cleanup:
- Defect repair: Issue #1084 - modified code to record desired persistence of objects so that cloned stars don't participate in logging etc.
- Removed some unused code (as a result of the defect repair)
- Some Code cleanup

@ilyamandel This turned out to be more involved than I had anticipated - quite a few files changed, but hopefully the review won't be too onerous.


ilyamandel (approved):
Hi @jeffriley,

Looks great, happy to approve!  A few comments for your consideration before you merge, and a couple of questions for my general education. ;-) 

BaseBinaryStar.h, line288:
`bool PrintSwitchLog(const bool p_PrimarySwitching) { return OPTIONS->SwitchLog() ? (LOGGING->ObjectSwitchingPersistence() == OBJECT_PERSISTENCE::PERMANENT ? LOGGING->LogBSESwitchLog(this, p_PrimarySwitching) : true) : true; }`

I struggled to parse this, in part because I don't remember what the arguments are supposed to do... but maybe a comment would be in order.

BaseStar.h, line 17:
Why is `class MainSequence; `needed?

Can you remind me why we need a BinaryStar in addition to BaseBinaryStar? [Aside: these are really binary types, not star types -- I always have to remind myself to ignore the Star in the name ;-) ]

GiantBranch.cpp, line 698:
`BaseStar *clone = Clone(static_cast<BaseStar&>(const_cast<GiantBranch&>(*this)));
`Why is the clone of type BaseStar, not of type GiantBranch?

GiantBranch::CalculateRadialExtentConvectiveEnvelope(), double GiantBranch::CalculateConvectiveEnvelopeMass() -- these and other related functions will have to be updated again in #1093 , but happy to roll yours in first. ;-)  [Actually, I think I reduced the number of functions where a clone is being made. ;-) ]

jeffriley:
@ilyamandel  ready for review

> BaseBinaryStar.h, line288:

Added explanatory comment.

> BaseStar.h, line 17: Why is class MainSequence; needed?

Added for testing and forgot to remove it - now removed.

> Can you remind me why we need a BinaryStar in addition to BaseBinaryStar?

BinaryStar is to BaseBinaryStar as Star is to BaseStar.  BinaryStar and Star are containers that contain pointers to BaseBinaryStar and BaseStar objects respectively.  We use that architecture to allow us to save state and revert to the saved state if necessary.  Currently we only do that for Star/BaseStar in the SSE code, but if we ever decide to save/revert binary stars (nomenclature discussed :-)) the infrastructure is there.  We could take it out if we decide we will never save/revert binary stars.

> GiantBranch.cpp, line 698: Why is the clone of type BaseStar, not of type GiantBranch?

We don't instantiate objects of type GiantBranch (or MainSequence, WhiteDwarfs, or Remnants).  The base class of all stars is BaseStar, so objects of the derived classes can be cast to BaseStar - and in that case since we didn't know the class of the underlying star we use BaseStar there.  However, the Clone() functions have been rewritten, and that code in GiantBranch was rewritten (by you) and we no longer use BaseStar there.

I have rewritten the Clone() functions, and removed some code (e.g. assignment operator overloads) that we don't use (I thought we would when I first refactored the code, but they are not necessary, especially with the Clone() functions).  I tested each use of the Clone() functions and I am satisfied the correct derived functions are being called where appropriate.  I have added comments to each use of the Clone() functions to explain to future readers of the code what is happening.  The declarations of the Clone() functions in the BaseStar class are documented, and that serves as documentation for the derived Clone() functions.

I ran a test of 1 million binaries for v02.43.03 and this new version (v02.43.04 if approved and merged before another PR).  I produced CSV logfiles, and a simple `diff` of the BSE_System_Parameters.csv and BSE_Double_Compact_Objects.csv files for each run shows that ~3120 of the 1 million binaries, and 71 of 3972 DCOs, differed between the runs.  The differences are probably due to the clones now calling the correct derived functions.  I have the CSV files available if you want to look at them.

ilyamandel (approved):
Thank you, @jeffriley , looks good!

PR #1092, opened 2024-04-15
Defect repair - Fix for issue #1074 - SSE Supernova records duplicated

### LINKED ISSUE #1074: Supernova being recorded twice in SSE_Supernovae output
**Describe the bug**
I am trying to look at a population of supernovae using COMPAS. I noticed that I was getting more supernovae than the number of (single) stars that I evolved, which seemed weird. Turns out each supernova appears twice in my output (example below).

**Label the issue**

**To Reproduce**
MWE:
./COMPAS --mode SSE --metallicity 0.0142 --maximum-evolution-time 500 --initial-mass-min 10.0 --remnant-mass-prescription FRYER2012 --kick-magnitude-distribution MAXWELLIAN --number-of-systems 100 --logfile-type CSV --quiet

This runs a small population of single stars. Inspect COMPAS_Output_X/SSE_Supernovae.csv and see that each supernova is listed twice (same seed, same time) with some differences in other parameters. 

**Expected behavior**
Each supernova record occurs only once.

 - OS: MacOS v12.4
 - COMPAS v02.42.01

jeffriley:
Fix for issue #1074 - SSE Supernova records duplicated

### DIFF
--- a/src/BaseStar.cpp
+++ b/src/BaseStar.cpp
@@ -2802,6 +2802,7 @@ void BaseStar::ResolveMassLoss(const bool p_UpdateMDt) {
         STELLAR_TYPE st = UpdateAttributesAndAgeOneTimestep(mass - m_Mass, 0.0, 0.0, false, false); // recalculate stellar attributes
         if (st != m_StellarType) {                                                                  // should switch?
             SHOW_WARN(ERROR::SWITCH_NOT_TAKEN);                                                     // show warning if we think we should switch again...
+            if (IsSupernova()) ClearSupernovaStash();                                               // we may have stashed SN details - need to clear them if we're not going to switch
         }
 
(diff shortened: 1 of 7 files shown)


ilyamandel (approved):
Yikes.  I'll approve, but this business of stashing info, delayed writing, and having to occasionally clear stashes looks very susceptible to future issues where someone forgets to clear a stash (I certainly would have).  I am not suggesting that I know a cleaner way, but I think this is primed to bite us again.

jeffriley:
I agree @ilyamandel.  The reason we stash and delay the write is because of the (COMPAS) SSE method of evolving and reverting if we decide we evolved too far - if we write directly to the file we might have written a record when we decide to revert, so we'd have to go and remove that record - easier to stash and delay the write.  I'm hoping the c'ear stash' problem will be obviated with the work to resolve the COMPAS vs Hurley sse differences.

PR #1093, opened 2024-04-15
Update to Picker+ (2024) fits for Hirai & Mandel 2-stage CE; exposure of functionality needed for tidal computations

ilyamandel:
Initial push; compiles, but yet to be tested.

### DIFF
--- a/src/BaseBinaryStar.cpp
+++ b/src/BaseBinaryStar.cpp
@@ -1599,7 +1599,7 @@ void BaseBinaryStar::ResolveCommonEnvelopeEvent() {
         double k1         = m_Star1->IsOneOf(COMPACT_OBJECTS) ? 0.0 : (2.0 / (lambda1 * alphaCE)) * m_Star1->Mass() * convectiveEnvelopeMass1 / m_Star1->Radius();
         double k2         = m_Star2->IsOneOf(COMPACT_OBJECTS) ? 0.0 : (2.0 / (lambda2 * alphaCE)) * m_Star2->Mass() * convectiveEnvelopeMass2 / m_Star2->Radius();
         double k3         = m_Star1->Mass() * m_Star2->Mass() / periastronRsol;                                         // assumes immediate circularisation at periastron at start of CE
-        double k4         = (endOfFirstStageMass1 * endOfFirstStageMass2);
+        double k4         = endOfFirstStageMass1 * endOfFirstStageMass2;
         double aFinalRsol = k4 / (k1 + k2 + k3);
         m_SemiMajorAxis   = aFinalRsol * RSOL_TO_AU;
 
--- a/src/BaseBinaryStar.cpp
+++ b/src/BaseBinaryStar.cpp
@@ -1583,16 +1583,21 @@ void BaseBinaryStar::ResolveCommonEnvelopeEvent() {
 
     // Two-stage common envelope, Hirai & Mandel (2022)
     else if ( OPTIONS->CommonEnvelopeFormalism() == CE_FORMALISM::TWO_STAGE ) {
-        double convectiveEnvelopeMass1  = m_Star1->CalculateConvectiveEnvelopeMass();
+        double convectiveEnvelopeMass1, maxConvectiveEnvelopeMass1;
+        std::tie(convectiveEnvelopeMass1, maxConvectiveEnvelopeMass1) = m_Star1->CalculateConvectiveEnvelopeMass();
         double radiativeIntershellMass1 = m_MassEnv1 - convectiveEnvelopeMass1;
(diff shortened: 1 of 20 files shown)


ilyamandel:
@veome22 -- I am flagging you because this will show you which functions I am exposing. 
@jeffriley & @ryosuke-hirai -- I'll eventually ask for your review, but need to test first.

ilyamandel:
@jeffriley , @veome22 , @ryosuke-hirai -- this runs and does sensible things; still testing, but removing Draft status so you can review (and Veome can see what the API will look like).

veome22:
Hi @ilyamandel! 

In line 30 of MainSequence.h, why does `CalculateConvectiveEnvelopeMass()` return (0.0, 0.0)? 
Based on Hurley (2000), shouldn't the initial envelope mass be $M_\*$ for $M_* < 0.35 M_\odot$, and gradually decrease for $0.35 M_\odot < M_* < 1.25 M_\odot$ for a Main Sequence star? (This is all in addition to the time evolution).

ilyamandel:
Good catch, @veome22!  Fixed.

ryosuke-hirai:
I don't quite understand why in various places there is something like
`RadialExtentConvectiveEnvelope = Radius - ConvectiveCoreRadius`
Judging from the name, the right hand side will give the depth of a radiative envelope.
I haven't followed how this is used, but is this just a confusing variable name?

ilyamandel:
@ryosuke-hirai : Indeed, for evolved giants (FGB, CHeB, HeGB) we assume that the radial extent of the radiative intershell is negligible.  We don't have fits for it, so it's the best we could do right now (and matches Hurley's recommendation).

ryosuke-hirai:
Ok, I now see that `ConvectiveCoreRadius` basically points to `RemnantRadius` for evolved stars. So it is just slightly confusing because the variable name is re-used for something else.

jeffriley:
@ilyamandel you pushed you Makefile - that's why the build is failing.

Also, re the Makefile, see slack https://compashq.slack.com/archives/CMCEQ0RS8/p1673941394742069

ilyamandel:
@veome22 , @jeffriley , @ryosuke-hirai -- thanks to all of you for great comments!
I've uploaded some fixes and would be grateful for a review so we can pull this in (if it's acceptable, of course).
I note that I already have some promising preliminary results, where the 2-stage CE is doing exactly what we expected: killing off the CE channel for BBH formation and enhancing it for DNS formation (see attached plot).

image (3)

jeffriley (changes requested):
Just a few comments @ilyamandel 

BaseBinaryStar.cpp
Line 1602 - Not a big deal, but you don't need the enclosing parentheses (I'll sometimes put parentheses where the compiler doesn't really need them just to make something easier for a human to parse, but I don't think you need them here).

BaseStar.cpp
CalculateConvectiveEnvelopeLambdaPicker(), args could be const, function could be const
Line 3836 - use class variable m_Log10Metallicity instead

FGB.h
Line 59 - small point: function calls incur overhead (to build and collapse the stack frame at least, though in this case a good optimising compiler would optimise the call away) - why use Radius() when m_Radius is available?  Radius() is just a getter for m_Radius - I'd be surprised if we ever changed it.

GiantBranch.cpp
Is it worth moving CalculateRadialExtentConvectiveEnvelope() to GiantBranch.h?
Line 1056 - change `1` to `1.0`

GiantBranch.h
Why is CalculateConvectiveEnvelopeLambdaPicker() declared here?  I don't see a definition in the .cpp file (am I just missing it?)

HG.cpp
Is it worth moving CalculateRadialExtentConvectiveEnvelope() to HG.h?

MainSeuqnce.cpp
Line 493 - change `0` to `0.0`

Star.h
CalculateConvectiveEnvelopeLambdaPicker(), args could be const, function could be const

ilyamandel:
Thank you very much for the review, @jeffriley !  All requested changes made.

veome22 (changes requested):
Hi @ilyamandel , the PR looks mostly great! Here are a few minor things I thought to change:

MainSequence.cpp
- `CalculateConvectiveEnvelopeMass()` currently sets `massEnvelope0 = 0.35` by default. For a star below 0.35 M_sun, the envelope mass would be greater than the total stellar mass.

TPAGB.h
- `CalculateConvectiveCoreRadius()` probably needs to check that `5.0 * CalculateRemnantRadius <= m_Radius` (also from Hurley (2000), last sentence of Sec. 6

ilyamandel:
@veome: thank you, very nicely spotted!

veome22 (approved):

jeffriley (approved):
Thanks @ilyamandel !

PR #1100, opened 2024-04-20
Don't allow ephemeral clones to operate the SSE SN stash

jeffriley:
@ilyamandel This is the last piece of no logging for clones - this prevents ephemeral clones from writing to or clearing the SSE SN stash.

Tiny bit of code cleanup - also changed `dynamic_cast` to `static_cast` in Star.cpp - dynamic is not required there.


ilyamandel (approved):
Thanks, @jeffriley !

Issue #978, opened 2023-06-12
Discontinuous radius evolution


ilyamandel:
We still have an issue with discontinuous radial evolution in COMPAS.  

Here's a trivial example, from

`./COMPAS -n 1 --mode SSE --initial-mass 50 --detailed-output
`
![image](https://github.com/TeamCOMPAS/COMPAS/assets/512213/ef03d52c-af61-41dd-8e52-4c148be62b35)

Note that this issue does not come up in Hurley's SSE/BSE.  

I believe this may be due to differences between the SSE paper and its implementation.

For example, around line 190–200 of hrdiag.f:

`rmin = rminf(mass)`

This uses mass (i.e., M_0 in the language of the paper) to calculate rmin; meanwhile, the paper states that M_0 is used for timescales and luminosities, but M (mt in the code) for radii. 

```
ry = ragbf(mt,lums(4),zpars(2))
rx = MIN(rmin,ry)
…
tau2 = tblf(mass,zpars(2),zpars(3))
if(tau2.lt.tiny) rx = ry
```

This means that, for massive stars, rmin isn’t used anyway — instead, it’s the AGB luminosity that’s used directly for the luminosity at the end of the Hertzsprung gap.  And that fundamentally explains how luminosities are consistent between HG and CHeB stars even though, according to the formulae in the paper (as I read them), one uses a function of mass and the other of mass and luminosity.  

In the end, it makes sense: the HG is just a smooth connection from the MS that precede it to the giant phases that follows.  We need to make sure that our code does the same.

reinhold-willcox:
Good catch! Have you talked to Jarrod about why there is a discrepency?

ilyamandel:
@reinhold-willcox -- yes, the discrepancy is there because the SSE paper doesn't quite cover match the SSE code in some cases; the trouble for me at the moment is parsing the SSE code (done for HG stars, but still struggling for CHeB stars, see Slack).

jeffriley:
@ilyamandel 

I have  made a couple of changes that I think help fix the problem - I can get COMPAS to look like Hurley sse for a single star (haven't tried others) with mass loss disabled.  Happy to try with mass loss enabled if you tell me how to configure sse for mass loss and tell me which option(s) in COMPAS match.

Branch on my fork - hopefully you have access (let me know if not).  Look for "~~ILYA~~" in the code to see the changes made.  Explanatory comments there.

https://github.com/jeffriley/COMPAS-1/tree/hurley-test

[...]

PR #1102, opened 2024-05-02
Fixes for issue 978 and 1075.

jeffriley:
Fixes for issue 978 and 1075.

@ilyamandel Code changes described in issue https://github.com/TeamCOMPAS/COMPAS/issues/978 .

I did a little code cleanup (sorry...).

You will also see lots (lots...) of changes for constants.h changing the string "14, 6" to "15.15". Those changes just change the precision of double values written to the csv (and tsv and txt) COMPAS log files. Variables stored in HDF5 files are full precision, so I decided we should do the same for the other logfile types, and the users can decide how many digits they want to use.

EDIT: I included the sse output files in the PR.


ilyamandel (approved):
Thank you for the wonderful work, @jeffriley !

Happy to approve, just leaving some minor comments behind for your consideration.  Note that I am not carefully reviewing the key change in HG::CalculateRadiusOnPhase(): I am satisfied that this follows the SSE code and has the expected behaviour of ensuring that the radius on the HG phase evolves continuously to its desired endpoint at the start of FGB/CHeB, which is the only real requirement here.

BaseStar.cpp, around lines 1130--1140, the following line is repeated and probably could be computed just once:
int Zind = utils::Compare(m_Metallicity, LAMBDA_NANJING_ZLIMIT) < 0 ? 0 : 1;

BaseStar.cpp, 1171, 1233: yes, I think the log's should be log10's, i.e., should use m_Log10Metallicity!  That would mean we are making a mistake in the lambda interpolation.  I think this is @themikelau 's code -- Mike, could you take a look?  [Note: I am happy to approve this PR without waiting for this to be resolved, this is just a reminder to follow up on this.]

Not sure why you changed all the comparison functions in Veome's tides code to use compare but left some normal comparisons in code you edited elsewhere (e.g., CHeB.cpp, lines 1380-1382), but I think it's fine.

HG.cpp, line 1192: 
"// The current mass would yield a core mass larger than *or equal to* the current core mass -- i.e., no unphysical core mass decrease would ensue" -- add the commented text since you changed the < to <=?

MainSequence.cpp, line 666: "double dtk = 1.0E-2 * timescales(tMS); // 0.02 of MS timescale (sse uses 0.05)" -- comment doesn't match code (1.0E-2 is 0.01, not 0.02)

constants.h, line 294: "constexpr double HIGH_MASS_THRESHOLD = 12.0; // value above which we consider stars to be high mass stars".  value -> "Mass (in solar masses)"

ilyamandel:
P.S. @themikelau confirmed, so could you please change the metallicity logs to log10 values?

> Yes! It should be log10. Apologies for the mistake. This should be changed in both double BaseStar::CalculateMassAndZInterpolatedLambdaNanjing and  double BaseStar::CalculateZInterpolatedLambdaNanjing

jeffriley:
Thanks @ilyamandel 

First, the pytests are failing in the COMPAS build process.  I believe the error is:

    KeyError: "Unable to synchronously open object (object 'BSE_Common_Envelopes' doesn't exist)"

but if I run COMPAS v02.44.01 with `runSubmit` and `example_bbh_compas_config.yaml` the HDF5 file produced contains a `BSE_Common_Envelopes` file.  Maybe @avivajpeyi or @reinhold-willcox could take a look at that?  I'll merge this PR anyway - the code compiles and runs ok - and seems to produce reasonable results.

As to your comments @ilyamandel :

> BaseStar.cpp, around lines 1130--1140, the following line is repeated and probably could be computed just once:
> int Zind = utils::Compare(m_Metallicity, LAMBDA_NANJING_ZLIMIT) < 0 ? 0 : 1;

I don't think that's right.  There are two instances of that line of code, but they are in different conditional clauses, so only one will be executed.

> BaseStar.cpp, 1171, 1233: yes, I think the log's should be log10's, i.e., should use m_Log10Metallicity!

Fixed.

> Not sure why you changed all the comparison functions in Veome's tides code to use compare but left some normal 
> comparisons in code you edited elsewhere (e.g., CHeB.cpp, lines 1380-1382), but I think it's fine.

I don't use utils::Compare() for age/time/timestep related comparisons - on the basis that any thresholds we (eventually) set for  utils::Compare() may need to be different for age/time/timestep related comparisons and other attributes like mass, radius etc.  My thinking is that the age/time/timestep values are sensitive to many more decimal places than the other attributes are likely to be, so I want to think about (maybe you to think about ;-)) whether the same threshold should be used.  Does that make sense?  Since we now use quantised timesteps, we are in a good position to change the code so that all our age/time/timestep values are stored as integer multiples of the TIMESTEP_QUANTUM (10^-12), and so all our age/time/timestep related comparisons could be integer comparisons - not a huge job (we'd just have to be careful), but worth thinking about.  That would fix/prevent any precision-related issues with age/time/timestep calculations.

I did, however, miss the mass related comparison in CHeB.cpp that you noted - now fixed. 

> HG.cpp, line 1192:
> "// The current mass would yield a core mass larger than or equal to the current core mass -- i.e., no unphysical core mass 
> decrease would ensue" -- add the commented text since you changed the < to <=?

Fixed.

> MainSequence.cpp, line 666: "double dtk = 1.0E-2 * timescales(tMS); // 0.02 of MS timescale (sse uses 0.05)" -- comment 
> doesn't match code (1.0E-2 is 0.01, not 0.02)

Oops - looked at the wrong digit :-)  Fixed.

> constants.h, line 294: "constexpr double HIGH_MASS_THRESHOLD = 12.0; // value above which we consider stars to be high
> mass stars". value -> "Mass (in solar masses)"

Fixed.

jeffriley:
Ok, @avivajpeyi pointed out that the failing test actually uses `fiducial bbh-config.yaml`, not `example_bbh_compas_config.yaml`.  Running with `fiducial bbh-config.yaml` and the grid file supplied produces no CEE, so no CE file.  This has happened before (though last time it was not enough DCOs being produced), and we "fixed" it by jiggling the initial attributes in the grid file.  Maybe we could do the same here - it looks like the changes to the code in this PR were enough to mean no CE happened in the binaries in the grid file.  I'm not going to hold the PR up for this, but we need someone to look at this test and associated grid file to see what needs to be changed to make the test run (@avivajpeyi , @reinhold-willcox , @themikelau ?).

@avivajpeyi kindly agreed to start documenting the tests in the developer guide of the online documentation.  Developers need to know what tests are going to be run against new code, and what those tests are actually testing, and what their input requirements are etc., so that we can ensure the new code actually at least produces the required inputs for the tests.

ilyamandel:
> > BaseStar.cpp, around lines 1130--1140, the following line is repeated and probably could be computed just once:
> > int Zind = utils::Compare(m_Metallicity, LAMBDA_NANJING_ZLIMIT) < 0 ? 0 : 1;
> 
> I don't think that's right. There are two instances of that line of code, but they are in different conditional clauses, so only one will be executed.

Sure, I meant it could be done before the conditional split, but it's fine as is.

ilyamandel (approved):

jeffriley:
> Sure, I meant it could be done before the conditional split, but it's fine as is.

You're absolutely right :-)  I'll move it out next time.

PR #1107, opened 2024-05-03
Change to CHeB end-of-phase core mass calculation

jeffriley:
@ilyamandel this PR changes the way core mass is calculated at CHeB end of phase. COMPAS now calculates core mass at phase end the same way the Hurley sse code does, but Hurley et al. 2000 eqs 66 & 67 seem to suggest it shouldn't make a difference - if Mc increases linearly on phase until it gets to Mc,BAGB at phase end, then eqs 66 & 67 should yield the save value at phase end - maybe the devil is in the detail somewhere (maybe the "is approximated" just before eq 66 is germane). Either way, this fix solves the problem of the 50 Msol star having that odd spike in Teff at the end of the CheB phase/start of the HeMS phase.


ilyamandel (approved):
Looks good!

PR #1116, opened 2024-05-09
Move compiler standard from c++11 to c++17

### LINKED ISSUE #984: Consider migrating to C++17 for compiling COMPAS (required by boost.math)
**Describe the bug**

After upgrading to boost 1.82, the following compiler warning appears:

warning: "The minimum language standard to use Boost.Math will be C++14 starting in July 2023 (Boost 1.82 release)"

As the warning indicates, as of the boost 1.82 release, the minimum C++ version for Boost.Math is C++14.  Currently we compile COMPAS with C++11.  We use Boost.Math is a few places throughout the code.  If we can't rewrite the code that uses Boost.Math so that it doesn't (we probably could, but probably don't want to...), we need to either advise users not to upgrade to boost v1.82 (or beyond), or ensure we can compile COMPAS with at least C++14 - I think we should look at migrating to C++17.

**Label the issue**

`urgency_low`      - This issue is not urgent (yet)

`severity_minor`    - This is a minor issue with minimal impact (for now)

**To Reproduce**
1. Upgrade boos to v1.82
2. Compile COMPAS using C++11
3. See compile warning

**Expected behavior**

We can compile, link, and run COMPAS using the boost.math functionality even after upgrading to boost v1.82 (we'll have to move there eventually)

**Screenshots**
n/a

 - OS: Ubuntu 22.04, MacOS Ventura 13.4.1
 - COMPAS latest

**Additional context**
n/a

jeffriley:
1. changed compiler standard from c++11 to c++17 in Makefile - see issue #984
   (Tested ok with Ubuntu v20.04, g++ v11.04, and boost v1.74; and macOS v14.1.1, clang v15.0.0, and boost v1.85.)
2. added check for boost version to allow for deprecated filesystem option
3. added `requirements.in` file to online docs to specify requirements for latest

@ilyamandel:

1. is straightforward
2. is the #defines in Log.h and changed code in Log.cpp.  We currently use boost::copy_file, but since c++17 the std::filesystem::copy_file is available.  For now, untyil eveyone moves to c++17, I'm leaving boost::copy_file in Log.cpp - but after some (maybe extended) grace period I'll plan to replace it with std::filesystem::copy_file
3. The online docs haven't been building for the past three months...  The problem is that the theme we use isn't compatible with sphinx 7 and above.  I'm hoping the 'requirements.in' file will fix it - if not, I'll try again...

### DIFF
--- a/online-docs/conf.py
+++ b/online-docs/conf.py
@@ -21,7 +21,7 @@ sys.path.insert(0, os.path.abspath('../compas_python_utils/'))
 # -- Project information -----------------------------------------------------
 
 project = 'COMPAS'
-copyright = '2021, 2022, 2023 The Authors'
+copyright = '2021, 2022, 2023, 2024 The Authors'
 author = 'TeamCOMPAS'
 
 
@@ -29,7 +29,7 @@ author = 'TeamCOMPAS'
 
 # If your documentation needs a minimal Sphinx version, state it here.
 #
-needs_sphinx = '4.3.2'
+#needs_sphinx = '4.3.2'
 
 # Add any Sphinx extension module names here, as strings. They can be
 # extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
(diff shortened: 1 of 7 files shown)


ilyamandel (approved):

PR #1118, opened 2024-05-09
Trying to get the online docs to build

jeffriley:
The online docs build has been failing for the past three months.  The reason is that sphinx > 7 is not compatible with the rtd-theme we are using.  There is an updated rtd-theme that fixes the problem - I'm hoping that adding the yaml and requirements files here will cause the build to pull in the new rtd-theme.


ilyamandel (approved):
Looks good to me, thank you, @jeffriley !

PR #1121, opened 2024-05-09
Online docs build issues - Changed theme to default (Alabaster)

jeffriley:
The online docs building is failing (has been for three months...) because the theme (readthedocs theme `sphinx_rtd_theme`) is not compatible with the latest version of sphinx. Apparently there is an updated theme, but I can't figure out how to get it into our readthedocs project.  To resolve the problem now I've changed our documentation theme to the shpinx default `Alabaster` - I don't mind the look and feel of it, and if we leave it as the default hopefully we won't have these build problems again (presumably the sphinx people will always make their default theme compatible with the latest version of sphinx).  If people really want the old rtd theme maybe we can look at getting the updated version...

I also change a couple of headings in one of the files - they clashed with headings in another file.

Hopefully this fixes the problem...

### DIFF
--- a/online-docs/conf.py
+++ b/online-docs/conf.py
@@ -21,7 +21,7 @@ sys.path.insert(0, os.path.abspath('../compas_python_utils/'))
 # -- Project information -----------------------------------------------------
 
 project = 'COMPAS'
-copyright = '2021, 2022, 2023, 2024 The Authors'
+copyright = '2024 The Authors'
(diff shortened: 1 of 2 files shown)


avivajpeyi:
Looks good to me! BTW we can always get the docs to be built from github-actions instead of readthedocs if you want more flexibility (wont be tied down to rtd)

avivajpeyi (approved):

Issue #1114, opened 2024-05-08
MandelMueller prescription overwrites maximum neutron star mass option


SimonStevenson:
**Describe the bug**
The default --maximum-neutron-star mass is 2.5 Msun. However, if you use the Mandel&Mueller prescription, it has its own hardcoded parameter MULLERMANDEL_MAXNS which is 2.0 Msun by default

This is kind of similar to https://github.com/TeamCOMPAS/COMPAS/issues/1000 where this prescription overrides some other COMPAS input option. 

Issue identified by @stellarphysicsandpopulations 

**Label the issue**
`urgency_low`      - This issue is not urgent
`severity_minor`    - This is a minor bug with minimal impact

**To Reproduce**
See GiantBranch::CalculateRemnantMassByMullerMandel

**Expected behavior**
Either we should update the documentation/warn the user, or we should make this prescription use the user specified maximum neutron star mass, or something along those lines.

**Versioning (please complete the following information):**
 - OS: MacOS 12.4
 - any recent version of COMPAS e.g., COMPAS v02.44.05

ilyamandel:
Removed MULLERMANDEL_MAXNS  in lieu of just using the choice provided by the user via --maximum-neutron-star in PR #1123 .   I agree that a single max NS choice for all prescriptions is sensible.  I'd personally prefer to set it to something we now know to be more realistic (2.2 Msun?) by default, but that's not urgent.  Closing.

Issue #1088, opened 2024-04-12
Stroopwafel produces no hits


nsteinle:
StroopwafelInterface.py runs (after a few edits explained below) but does not find any hits and so does not perform better than usual COMPAS run at producing a population of specific DCO type. An example samples.csv output file from Stroopwafel is attached. This seems to be independent of the input parameters. 
[samples.csv](https://github.com/TeamCOMPAS/COMPAS/files/14960571/samples.csv)

Issue 'severity' and 'urgency' (but not actually sure!):
`urgency_moderate` - This is a moderately urgent issue
`severity_moderate` - This is a moderately severe bug

1. Go to StroopwafelInterface.py and use it to obtain ZAMS parameters for BNSs that merge within t_universe, but first have to edit it to fix logic errors.
2. Pass the following variables (debug, mc_only, run_on_hpc, output_filename, output_folder, userunSubmit) to the function main() to fix error.
3.  Move the function configure_code_run() to be inside of the main() function (ie b/w steps 1 and 2) to fix error.
4. Run it with the default config.yaml file.
5. Examine the csv output  - the column with 'is_hit' has all zero entries meaning no hits.
6. Convert that csv output into a Grid file and run the binaries with COMPAS to confirm that no additional BNSs are produced 

I expected stroopwafel to produce a number of ZAMS binaries that evolve into BNSs, and then I could run those binaries in COMPAS forward evolution to get a BNS population that is much larger than what a normal COMPAS run gives in the same amount of time. This would be nice to save on computational resources and time. I do not expect that the edits I made in steps 2 and 3 above will be the cause of this issue but I very well might be doing something wrong here.

Versioning:
 - OS: Ubuntu-22.04
 - COMPAS: v02.39.00

Any help would be greatly appreciated!

FloorBroekgaarden:
Thanks Nathan, I will try to look at this today, but latest on Monday! Thanks for your clear message! 
Floor

ilyamandel:
@FloorBroekgaarden, @nsteinle  , has this been resolved?

nsteinle:
@ilyamandel Thanks for checking! It has not been resolved

ilyamandel:
Dear @nsteinle ,

[...]

Issue #1076, opened 2024-03-11
Issue with mass transfer onto neutron star


yuzhesong:
**Describe the bug**
When attempting to fix the issues with neutron star recycling in the current version, @SimonStevenson and I have discovered that in some situations, the NS would accrete 0.25 Msol in one timestep of 1000 years, which is unphysical, since the accretion rate onto a neutron star should be Eddington limited. 

@SimonStevenson and I have narrowed down the possible issue in the function
void BaseBinaryStar::CalculateMassTransfer(),
where m_FractionAccreted is calculated using CalculateMassAcceptanceRate(). 

It seems like that the code is calculating the mass transfer rate in this timestep described above as thermal mass loss rate, while it is much shorter than thermal time scale. 

**Label the issue**

`urgency_moderate` - This is a moderately urgent issue. While this creates some issues, these cases are relatively rare and don't seem to pose problems to the majority of simulations.

**To Reproduce**
I have attached a yaml file (in the format of txt as GitHub doesn't support uploading yaml file) that should be able to reproduce this case
[config_accretion_bug.txt](https://github.com/TeamCOMPAS/COMPAS/files/14552274/config_accretion_bug.txt)

**Versioning (please complete the following information):**
 - COMPAS v02.41.05

SimonStevenson:
I think I have a solution for this. I think the issue is that for these low-mass companion stars, the duration of the mass transfer should be long as the stars have long thermal timescales tau_th (tens of Myrs), which are much longer than the timestep dt we were taking during the mass transfer (in this case, the initial tilmestep is as small as dt = 0.001 Myr, since we are trying to resolve the early pulsar evolution). 

To attempt to solve this, I have added a clause that if the star's thermal timescale is longer than the current timestep, we only accrete a fraction (dt / tau_th) of the mass that the donor star wants to donate. 

This seems to have the desired effect of avoiding a large amount of mass being accreted in one timestep. 

When I first did this, the system ended up merging after a few time steps. I managed to track this down to the tolerance `ROOT_ABS_TOLERANCE` used in `MassLossToFitInsideRocheLobe`, which by default is 1E-10. I noticed that the solution that was found was good to a few 1E-10. Relaxing the tolerance to 1E-9 solved this issue. 

![m1vtime](https://github.com/TeamCOMPAS/COMPAS/assets/6952112/42761fb6-cff5-4b09-8e22-3317cb5b765d)
[...]

Issue #1016, opened 2023-11-10
High mass black holes receiving no kick, even when using options that should allow for kicks


SimonStevenson:
**Describe the bug**
I noticed that when using something like the following set of options:

./COMPAS -n 1000000 --metallicity-distribution LOGUNIFORM --evolve-unbound-systems FALSE --chemically-homogeneous-evolution NONE --allow-rlof-at-birth FALSE --pulsational-pair-instability-prescription FARMER --kick-magnitude-distribution MAXWELLIAN  --black-hole-kicks REDUCED

that black holes above ~30 Msun still get no kicks

So far, I think this may be a conflict between --black-hole-kicks and --pulsational-pair-instability-prescription FARMER

**Label the issue**
Please label the 'severity' and 'urgency' of this issue. You can choose:

`urgency_moderate` - This is a moderately urgent issue

`severity_moderate` - This is a moderately severe bug

**To Reproduce**
Run a small population. Look at black hole kicks Drawn_Kick_Magnitude(SN) and 'Applied_Kick_Magnitude(SN)' in BSE_Supernova

**Expected behavior**
If user allows black holes to have kicks, they should have kicks

**Screenshots**

![applied_kicks_v_mass](https://github.com/TeamCOMPAS/COMPAS/assets/6952112/41e0c05d-a806-4811-9952-0073b98da406)
![drawn_kicks_v_mass](https://github.com/TeamCOMPAS/COMPAS/assets/6952112/e6d50905-1dca-42a0-8bbc-22eb5c587444)

**Versioning (please complete the following information):**
MacOS v 12.4
COMPAS v02.40.00

**Additional context**

ilyamandel:
@SimonStevenson -- I don't think it's a bug, more like an intentional choice: PISNe and PPISNe don't result in natal kicks (see, e.g., BaseStar::CalculateSNKickMagnitude() ).  Do you have an alternative proposal?

SimonStevenson:
[...]

PR #1123, opened 2024-05-13
Changes to timesteps, mass transfer timescale, other minor tweaks

ilyamandel:
Added options --radial-change-fraction and --mass-change-fraction, as approximate desired fractional changes in stellar radius and mass on phase when setting SSE and BSE timesteps

The recommended values for both parameters are 0.005, but the default remains 0, which reproduces previous timestep choices

Mass transfer from main sequence donors (including HeMS) can now proceed on nuclear timescales -- approximated as the radial expansion timescales -- if equilibrium zetas are greater than Roche lobe zetas

Removed the fixed constant MULLERMANDEL_MAXNS; instead, OPTIONS->MaximumNeutronStarMass() is used for consistency (see issue #1114)

Corrected return units of CalculateRadialExpansionTimescale() to Myr

Updated documentation

Minor code cleanup


ilyamandel:
May address #1076 (but not tested specifically for accretion onto NS, so not declaring victory yet).

ilyamandel:
Should address #889 , leaving open for now as a reminder to test explicitly.

ilyamandel:
Closing #1114 .

ilyamandel:
Closing #470 -- arbitrarily smooth HR diagrams can now be produced with this PR.

ilyamandel:
I hoped to address #24 with this one (convergence checks and adaptive time steps), but only partially succeeded.

Convergence is almost guaranteed (though not strictly ensured, since these are estimated changes for future time steps, not guaranteed changes in past time steps) by setting --mass-fractional-change and --radial-fractional-change to desired values, such as 0.005.  Of course, that does carry a significant computational cost, so adaptive time stepping is still desirable.  

I did attempt to increase the time step adaptively (by a factor of 2 over previous time step) if not counter-indicated by radial or mass change timescales, but that did sometimes lead to over-runs (e.g., SSE starting with a 6 Msun star could leave a massless remnant).  Leaving this out for this PR -- a project for the future.

jeffriley (changes requested):
Thanks @ilyamandel.  Just a few comments/suggestions:

1. You should add the new options to yaml.h.  yaml.h is the template for the yanl file production (via the --create-yaml-file option).  As you have seen, it's not absolutely necessary to update yaml.h - the code will add the new options to the end of the yaml file - but updating yaml.h with the new options locates them in (hopefully) a logical part of the yaml file so people can locate them easily.
2. The new options should both be in the m_RangeExcluded and m_SetExcluded vectors (at the top of Options.h - full Instructions for adding a new option are at the top of Options.cpp - can you check that you did all the steps required?)
3. Remove commented line in BaseStar::UpdateAttributesAndAgeOneTimestepPreamble()
4. Since we no longer use the LimitTimestep() function it should be removed - unless you're thinking you mightreinstate it?  Even then, you can get it from a previous version.  Should the documentation explicitly state the 1% mass loss limit is being replaced by a user-defined limit via --mass-change-fraction? Should that be made clearer in What's New?
5. It looks like the timestep you calculate in BaseStar::CalculateTimestep() is the timestep required to meet the change requirement(s) in isolation. So there is no interaction/combination of the two new options, and the timestep calculations in BaseStar::CalculateTimestep() don't take any limiting of mass loss (e.g. photon tiring) into account?  Should we document that? 
6. Should there be maximum allowed values for the --mass-change-fraction and --radial-change-fraction options?  If not, what's the effect of very large values?  I suppose we just lose up to the maximum the star is capable of losing - so the value given by the user is a maximum? (also relates to photon tiring limit I think).
7. How does the --timestep-multipler option interact with each of the --mass-change-fraction and --radial-change-fraction options?
8. Ok - just saw the "approxinate goal" comment in the code (and the text of the option descriptions in the docs).  Maybe "default" timesteps rather than "previously suggested" timesteps (in the docs decriptions)?
9. Should we have a page in the documentation describing the effect of the new options, and the --timestep-multiplier option, and maybe some of the above?  Maybe a "Timestepping" page in the User Guide somewhere?
10. How does --mass-change-fraction interact with the constant MAXIMUM_MASS_LOSS_FRACTION (wherever it is used in the code)?  Is the maximum mass loss fraction still defined by MAXIMUM_MASS_LOSS_FRACTION (so is the value for the new option limited by MAXIMUM_MASS_LOSS_FRACTION)? So should we actually limit the allowed value for the option?
11. How does --radial-change fraction interact with MAXIMUM_RADIAL_CHANGE, and so Star::EvolveOneTimestep() (for SSE mode)? Ditto for this - is the new option limited by MAXIMUM_RADIAL_CHANGE? So should we actually limit the allowed value for the option?

jeffriley:
> It looks like the timestep you calculate in BaseStar::CalculateTimestep() is the timestep required to meet the change 
> requirement(s) in isolation. So there is no interaction/combination of the two new options

No, that's wrong - there is interaction.  Does one take precendence over the other?  You possibly limit by the radial fraction first, then possibly by the mass fraction - results could be different if you switched those, couldn't they?  Maybe not...

ilyamandel:
Thank you for the review, @jeffriley !

1, 2, 3, 4 addressed.

The second part of 4, 10, 11 (mass and radius change fractions) are not impacted: these are hard limits calculated post factum, while I am adjusting timestep estimates before taking the timesteps.

To make these educated guesses (that's all they are), I use mass and radius evolution scales estimated based on the previous timestep.  If these were impacted by photon tiring, then that's automatically included, I intentionally don't delve into the details.  (That's my answer to 5.)

6: very large values for --mass-change-fraction or --radial-change-fraction, these end up being ignored when estimating the next timestep (i.e., similar behaviour as when the user picks zero).  

7: TimestepMultiplier is applied after the timestep is estimated, so effects are cumulative.  Clarified in documentation.

8: I didn't like "previously suggested", either, but defaults might eventually change, so I didn't want to mention default.  I've changed to "A value of 0 means that this choice is ignored and timestep estimates from earlier COMPAS version are used."

9: Maybe eventually, but I'll wait until we have an even more permanent solution to #24 [this is better than nothing, but it's not the final version].

10, 11: See above.

Ready for re-review.

ilyamandel:
Updated `pre-processing-sampling.rst` to not mention Stroopwafel for now (see #1088 ).

ilyamandel:
Addressed #1016: added option --natal-kick-for-PPISN; if set to true, PPISN remnants receive the same natal kick as other CCSN, otherwise (default) they receive no natal kick.

jeffriley (approved):
Thanks @ilyamandel - looks good

PR #1125, opened 2024-05-15
Updates for nuclear timescale mass transfer

ilyamandel:
- Create a new function, CalculateNuclearMassLossRate(), to compute the nuclear mass loss rate rather than CalculateRadialExpansionTimescale(), which can be unreliable during mass transfer
- Update BaseBinaryStar::CalculateMassTransfer() to use this function and to correctly evaluate m_AccretionFraction and the corresponding zetaRocheLobe

### DIFF
--- a/src/BaseBinaryStar.cpp
+++ b/src/BaseBinaryStar.cpp
@@ -1911,7 +1911,7 @@ double BaseBinaryStar::CalculateZetaRocheLobe(const double p_jLoss, const double
     double q            = donorMass / accretorMass;
     double cbrt_q       = std::cbrt(q);
 
-    double k1 = -2.0 * (1.0 - (p_beta * q) - (1.0 - beta) * (gamma + 0.5) * (q / (1.0 + q)));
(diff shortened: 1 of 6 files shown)


jeffriley (changes requested):
Looks good - just a couple of things:

In 

    double BaseBinaryStar::CalculateZetaRocheLobe(const double p_jLoss, const double beta)

`beta` should be `p_Beta` (also in .h file)

In

    return (p_StellarTypePrev == p_StellarType && utils::Compare(p_RadiusPrev, p_Radius) )
            ? (p_DtPrev * p_RadiusPrev) / fabs(p_Radius - p_RadiusPrev)
            : -1.0;

`utils::Compare()` returns an integer (-1 for <, 0 for =, 1 for >).  In c++ false == 0, and true == 1, but anything that is !false evaluates to true (there's a difference between something being equal to true and evaluating to true).  So a non-zero return will evaluate to true, but I don't think it's the clearest way of doing it...

ilyamandel:
Thank you, @jeffriley -- changes applied.

jeffriley (approved):
All good

PR #1126, opened 2024-05-15
Tides fix

veome22:
This PR corrects the frequency-dependent effective viscosity formulae for Equilibrium tides. This should (correctly) reduce the strength of Equilibrium tides by several orders of magnitude.


ilyamandel (approved):
Thank you, @veome22 -- looks good!

Issue #1127, opened 2024-05-15
Random systems are hanging


LiekeVanSon:
**Describe the bug**
A seemingly very random selection of runs with the current dev are hanging 

**Label the issue**
`urgency_moderate` - 
`severity_minor` - not super severe because only a selection of systems are impacted

**To Reproduce**
Run e.g. one of the following systems:

`$COMPAS_ROOT_DIR/src/COMPAS --metallicity 0.02 --wolf-rayet-multiplier 0.1 --mass-loss-prescription 'BELCZYNSKI2010'  --random-seed 5618 --initial-mass-1 43.88659326585003 --initial-mass-2 14.870895873364844 --semi-major-axis 9.379762986914274 --kick-magnitude-random-1 0.7497928488800597 --kick-phi-1 5.478971354711673 --kick-theta-1 1.4825278969647475 --kick-mean-anomaly-1 5.897831553803832 --kick-magnitude-random-2 0.6223669359402453 --kick-phi-2 1.9495891464049313 --kick-theta-2 1.3964441818333058 --kick-mean-anomaly-2 3.4693`

`$COMPAS_ROOT_DIR/src/COMPAS --metallicity 0.0003 --wolf-rayet-multiplier 0.1 --mass-loss-prescription 'BELCZYNSKI2010'   --random-seed 3011 --initial-mass-1 10.233399464598648 --initial-mass-2 4.166543854596874 --semi-major-axis 3.576671648178455 --kick-magnitude-random-1 0.35430223238159886 --kick-phi-1 1.450991365682787 --kick-theta-1 2.1388719072110836 --kick-mean-anomaly-1 1.0844552894213055 --kick-magnitude-random-2 0.943741385882113 --kick-phi-2 5.640506528904249 --kick-theta-2 0.7851673046553287 --kick-mean-anomaly-2 0.3038872942438468`

Very arbitrary changes, like increasing the random seed with 1, or removing any of the flags, will resolve the issue and make the system run without a problem. 

**Expected behavior**
These systems should run like any other

**Versioning (please complete the following information):**
 - OS: Rocky Linux release 8.9
 - COMPAS v02.46.01

ilyamandel:
Thanks, @LiekeVanSon -- agree this is a major bug to be fixed urgently!

ilyamandel:
Resolved, just preparing the PR.

ilyamandel:
Fixed in #1128.

PR #1128, opened 2024-05-15
Urgent bug repair -- infinite loop

ilyamandel:
Repaired a bug in GiantBranch::CalculateRemnantMassByMullerMandel() that could cause an infinite loop; see #1127 .


jeffriley (approved):
All good (I made a minor change to changelog.h - notes issue repaired by this PR).  Thanks @ilyamandel

PR #1130, opened 2024-05-16
Fixes for issues #744 and #873

### LINKED ISSUE #744: Hurley gbParams p and q values for He giants

### LINKED ISSUE #873: Remove superfluous p_NoCheck parameter of the ResolveEnvelopeLoss() function(s)
The function ResolveEnvelopeLoss() modifies a star's attributes after it loses its envelope - if it in fact lost its envelope. ResolveEnvelopeLoss() first checks whether the envelope has been lost - the check is some variation of core mass >= mass depending upon stellar type - if that is true then the envelope is gone and the attributes of the star are updated. We don't want to update the attributes if the star hasn't actually lost its envelope, even if the updates would be benign (in terms of affecting the attributes of the star) - the calculations are not trivial, so we don't want to negatively affect performance by doing non-trivial calculations that are not necessary.

The p_NoCheck parameter of the ResolveEnvelopeLoss() function allows callers of the function to force the calculations to happen - that is, if p_NoCheck is passed as TRUE, no check is performed, and the star's attributes are updated without checking whether it still has an envelope. p_NoCheck is (currently) only passed as true in three places in the code:

    GiantBranch::CalculateRadialExtentConvectiveEnvelope()
    HG::CalculateRadialExtentConvectiveEnvelope()
    Star::ResolveEnvelopeLossAndSwitch()

In the first two cases we clone the star and then call ResolveEnvelopeLoss(true) to force the calculations so we can pick up the newly calculated radius of the star.

In the third case we assume we already know the envelope is gone, so don't bother checking again (this could be an issue in the current code - see issue #874 ).

The p_NoCheck parameter exists because the legacy code did the envelope existence check, and bypassed it (differently, but still bypassed it) in the cases described above. The first two cases where we bypass the check, we're just trying to figure out the new radius of the star after losing its envelope - we don't want to update it yet, we just want to know what the value would be. When I ported the code I wondered if bypassing the check was actually required. If we call ResolveEnvelopeLoss() and always perform the check for envelope loss, then in the event of no envelope loss the radius of the star would be unchanged, and we would use that unchanged radius in the two cases above (that have p_NoCheck = true). I planned to check if that actually made a difference, but I didn't get back to it until now.

As described above, the p_NoCheck parameter to ResolveEnvelopeLoss() overrides the check at the top of the function that checks whether the star actually has lost its envelope - if p_NoCheck is TRUE, the check is not performed. What that means is that if p_NoCheck is TRUE, the loss of the star's envelope is resolved (i.e. attributes of the star are recalculated) whether it has lost it's envelope or not.

The check at the top of the *::ResolveEnvelopeLoss() functions is to determine if in fact the star has lost its envelope - no point resolving the loss of an envelope if the envelope was not actually lost. That check is implemented differently for different stellar types. Here are the current checks by stellar type:

    MS    : m_Mass <= 0
    HG    : m_CoreMass >= m_Mass
    FGB   : m_CoreMass >= m_Mass
    CHeB : m_CoreMass >= m_Mass
    EAGB : m_HeCoreMass >= m_Mass
    TPAGB: m_CoreMass >= m_Mass
    HeMS : m_Mass <= 0.0
    HeHG : m_COCoreMass >= m_Mass

Other stellar types inherit as usual.

I ran a test of 1,000,000 binaries, and found that in all cases whenever ResolveEnvelopeLoss() was called with p_NoCheck = TRUE, the envelope lost check at the top of the function was also satisfied - i.e. every call to ResolveEnvelopeLoss() with p_NoCheck = TRUE would have resulted in the code being executed anyway because the envelope lost check was also satisfied. It would seem then that the p_NoCheck parameter is superfluous and can safely be removed.

Based on the result of the test of 100,000 binaries described above, my recommendation is to remove the p_NoCheck parameter for all variants of the function (and callers).

jeffriley:
@reinhold-willcox, @ilyamandel 

Issue #873 : this PR changes the name of the `p_NoCheck` parameter to `ResolveEnvelopeLoss()` - weed it, we now better understand why, and the new name is consistent with other parameter names in COMPAS.

Issue #744: See issue for details, but the bottom-line is that @reinhold-willcox is right, and I have changed COMPAS to match Hurley's sse - but I was not able to generate a HeGB star to test the changes...  I have tested the changes for HeMS and HeHG, but not HeGB.  If either of you are able to generate an HeGB star, then you can test it or tell me how to generate the star and I'll test it.  I see no reason why the code shouldn't work for HeGB, but as it is now, it is untested there.

There is a little bit of code cleanup - sorry.


ilyamandel:
@jeffriley: This produces HeGB stars:

./COMPAS -n 1 --mode BSE --initial-mass-1 2.1 --initial-mass-2 2 --semi-major-axis 1

ilyamandel:
(Even this one, which I suggested off the top of my head yesterday, would have produced one HeGB star ;-) 
./COMPAS -n 1 --mode BSE --initial-mass-1 5 --initial-mass-2 3 --semi-major-axis 1 )

jeffriley:
Huh. I wrote down 10 + 5, which didn't work... ok, I'll give this a try
later.

ilyamandel (approved):
Happy to approve, but two minor points for your consideration:

---

Remnants.h:

STELLAR_TYPE ResolveEnvelopeLoss(bool p_Force = false) { return BaseStar::ResolveEnvelopeLoss(p_Force); } // Default to BaseStar

The function is declared as being virtual in BaseStar, so I am not actually sure what this would resolve to -- an infinite loop?  I think it might be safe to just do
{ return m_StellarType; }

---

// JR: Added this - depends on B, D, p, q & Mx - recalculate if any of those are changed

Do we need this comment on line 65 of HeMS.cpp?

ilyamandel:
> Huh. I wrote down 10 + 5, which didn't work... ok, I'll give this a try later.
> […](#)

Then I probably said the wrong thing -- sorry!

jeffriley:
Putting this back in draft.  Managed to create an HeGB system - HR diagrams differ current dev vs this fix (more than expected).  Will fix that problem then open this for review.

jeffriley:
Ok, ready for review.

For the following binary:

    ./COMPAS --random-seed 0 -n 1 --mode BSE --initial-mass-1 5.6 --initial-mass-2 5.1 --semi-major-axis 2

COMPAS v02.46.04 produces the following HR diagram for the primary:

image

and the changes in the PR produce the following HR diagram:

image

PR #1132, opened 2024-05-18
Significant updates to mass transfer functionality

ilyamandel:
These updates should really enable slow case A mass transfer:

- Equilibrium zeta and radial response of MS stars to mass loss are now calculated using CalculateRadiusOnPhase() rather than by cloning
- MassLossToFitInsideRocheLobe() and associated functor updated, work more efficiently, no longer artificially fail, and also use CalculateRadiusOnPhase()
- ROOT_ABS_TOLERANCE increased to avoid artificial failures on round-off errors
- code cleanup and bug repairs elsewhere

### DIFF
--- a/online-docs/pages/User guide/COMPAS output/standard-logfiles-record-specification-stellar.rst
+++ b/online-docs/pages/User guide/COMPAS output/standard-logfiles-record-specification-stellar.rst
@@ -1349,7 +1349,7 @@ the other is printed in any file, but not both. If both are printed then the fil
    * - COMPAS variable:
      - BaseStar::m_Mdot
    * - Description:
-     - Mass loss rate (\ :math:`M_\odot yr^{−1}`).
+     - Mass loss rate in winds (\ :math:`M_\odot yr^{−1}`).
    * - Header Strings:
      - Mdot, Mdot(1), Mdot(2), Mdot(SN), Mdot(CP)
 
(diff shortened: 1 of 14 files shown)


ilyamandel:
Behaved well on most of my tests such as MS->MS mass transfer
`./COMPAS -n 1 --initial-mass-1 2.5 --initial-mass-2 2.1 --semi-major-axis 0.05 --detailed-output --logfile-type CSV |more
`
but misbehaving on accretion onto NS 
`./COMPAS -n 1 --detailed-output --evolve-pulsars True --random-seed 6944 --common-envelope-alpha 1.0 --pulsar-minimum-magnetic-field 0.1 --common-envelope-mass-accretion-prescription 'MACLEOD'  --pulsar-birth-magnetic-field-distribution 'UNIFORM'  --pulsar-birth-spin-period-distribution 'UNIFORM' `
To be investigated further.

ilyamandel:
Limited nuclear timescale mass transfer to accrete only the smaller of the desired total MT and rate*dt on a timestep of size dt (since timescale to end of MS phase is, by construction, comparable to the nuclear MT timescale, not doing so risks transferring more mass than necessary; furthermore, it does not account for the change in the nuclear timescale as mass is lost).

jeffriley (approved):
Looks good @ilyamandel 

Just a few (related) questions:

You have changed the value of `ROOT_ABS_TOLERANCE`.  `ROOT_ABS_TOLERANCE` is used in other root finders in the code, and in `utils::BracketTolerance()` which is used in other root finders - did you explicitly check how the new tolerance affects those? Does the change to the abs tolerance indicate we should think about whether a change to the relative tolerance is needed?

If you run with warnings enabled you'll see warnings if the other root finders are affected (which I'm sure you know because I know you often run with warnings enabled :-))

I'll approve the PR, and you can merge at your leisure if the change in `ROOT_ABS_TOLERANCE` isn't an issue.

ilyamandel:
Thank you, @jeffriley .  Yes, I believe this should be fine elsewhere, too.

PR #1133, opened 2024-05-20
Fix to CalculateRadiusOnPhase()

ilyamandel:
Renamed the version of CalculateRadiusOnPhase() that takes in mass and tau as arguments into CalculateRadiusOnPhaseTau() to avoid clash with the version that takes in mass and luminosity as arguments

More importantly, somehow, the following change did not make it into #1132 (!) -- but is reflected here:

Limited nuclear timescale mass transfer to accrete only the smaller of the desired total MT and rate*dt on a timestep of size dt (since timescale to end of MS phase is, by construction, comparable to the nuclear MT timescale, not doing so risks transferring more mass than necessary; furthermore, it does not account for the change in the nuclear timescale as mass is lost).


jeffriley (commented):
All good -  thanks @ilyamandel

jeffriley (approved):

PR #1136, opened 2024-05-22
split up macleod linear AM loss fraction for degenerate and non-degen…

### LINKED ISSUE #1096: Different accretor AM response for degenerate and non-degenerate companions
AM loss is a major uncertainty in non-conservative (stable) MT models. Currently we have a few different options for this, all of which allow only for a fixed value for all stable MT events (taken either from the accretor or the L2 point, or some linear interpolation between them). However, it is likely the case that the physical mechanism for AM loss differs between degenerate and non-degenerate accretors. Even though we may not know what the mechanism is, it would be useful to split up these cases, and allow users to specify distinct AM loss models for degenerate and non-degenerate accretors. 

I'm happy to take care of this when I have some time, though if anyone else is interested I am happy to consult / advise.

reinhold-willcox:
…erate accretors


jeffriley (approved):

ilyamandel:
@reinhold-willcox , @jeffriley -- I didn't have time to check this yesterday, but there's a bug in this PR that needs to be fixed.

The default values in the code and in the YAML file do not match the values in the documentation:

    m_MassTransferJlossMacLeodLinearFractionDegen                   = 0.0;
    m_MassTransferJlossMacLeodLinearFractionNonDegen                = 0.0;
    
vs.
    
    **--mass-transfer-jloss-macleod-linear-fraction-degen** |br|
    Specific angular momentum interpolation fraction for degenerate accretors, linear between 0 and 1 corresponding to the accretor and L2 point. |br|
    Used when ``--mass-transfer-angular-momentum-loss-prescription = MACLEOD_LINEAR``, ignored otherwise. |br|
    Default = 0.5

Could you please fix this discrepancy?  I believe in this case the documentation describes the expected defaults, so the code needs to be updated and the YAML file needs to be regenerated.

Also, since you've removed one of the previously existing options, --mass-transfer-jloss-macleod-linear-fraction, this may merit an entry in the whatsnew.

jeffriley:
I'll fix this - I'm about to push some other changes.  I should have
noticed this in my review - I rushed it...

Jeff

--
Dr Jeff Riley
Monash University, Australia
Ph: +61 (0)407 811 282
***@***.***

Issue #998, opened 2023-10-09
output container doesn't work if target dir doesn't exist


reinhold-willcox:
**Describe the bug**
This is a small one. If you run compas with `-o new/output/directory`, but that directory doesn't already exist, COMPAS ignores the command and uses the default CWD. 

**Label the issue**
`urgency_low`      - This issue is not urgent
`severity_minor`    - This is a minor bug with minimal impact

**To Reproduce**
Run `COMPAS -o directory/that/doesnt/exist`

**Expected behavior**
COMPAS should try to create the directory (if possible), or print a statement saying that it can't (if there are user privilege issues). 

**Versioning (please complete the following information):**
 - COMPAS v02.39.01

jeffriley:
Huh. Never noticed...  Shouldn't be too hard to fix.

jeffriley:
@reinhold-willcox 

Actually... Now that I've looked at this I remember I wrote it this way deliberately - it's not described in the documentation (oops), but the way the code is we expect the output directory to exist, and if it doesn't we just use the CWD.  If we don't require the output directory to exist, the logging code has to create it before creating files in it - which it can do, but that could mean if there is an error later we've created a directory that could be empty, and I didn't want that to happen.  The workaround is to ~~`touch`~~ *make* the directory (with `mkdir`) before running `COMPAS`, and `COMPAS` will run ok.  But if that's too unwieldy, I could change the code so that the logging code creates the directory before using it - it just may be that if something goes wrong (EDIT: or there are no stars/binaries to be written to the logfiles) there may be a circumstance where `COMPAS` leaves behind an empty directory.  Messy, but probably not the end of the world...

Thoughts?

ilyamandel:
I think making the directory is probably the preferred outcome in this case; it's the user's responsibility to clean up such directories. :-)

PR #1139, opened 2024-05-24
Minor fixes

jeffriley:
- Changed functionality of `output-path` option to create missing directories in the path (see issue #998)
- Fixed incorrect default values for options `--mass-transfer-jloss-macleod-linear-fraction-degen` and `--mass-transfer-jloss-macleod-linear-fraction-non-degen`
- Changed BaseStar::UpdateAttributesAndAgeOneTimestepPreamble() so timescales are not recalculated when we know dT = 0
- Added documentation for log file record type
- Added "Quick Links" to documentation
- Updated "What's New"


ilyamandel (commented):
In standard-logfiles-record-types.rst, there's a typo, rand should be range in
"The record type property is an unsigned integer that can take any value in the rand :math:`0..4294967295`. "
(Sorry to ask you to fix this, but it must be your branch rather than a COMPAS branch because I can't edit directly :) )

ilyamandel (commented):
In changelog.h, I think you are giving credit (or blame) to the wrong person: "RW - May 24, 2024"

ilyamandel:
I saw your comment on my comment in BaseStar::CalculateZetaEquilibrium().  Will respond to this one separately -- maybe you can remove both comments (yours and mine) for now, and I will investigate offline.

Happy to approve (though I realise it's a draft for now :) ).

PR #1140, opened 2024-05-24
Minor fixes

### LINKED ISSUE #998: output container doesn't work if target dir doesn't exist

jeffriley:
- Changed functionality of `output-path` option to create missing directories in the path (see issue #998)
- Fixed incorrect default values for options `--mass-transfer-jloss-macleod-linear-fraction-degen` and `--mass-transfer-jloss-macleod-linear-fraction-non-degen`
- Changed BaseStar::UpdateAttributesAndAgeOneTimestepPreamble() so timescales are not recalculated when we know dT = 0
- Added documentation for log file record type
- Added "Quick Links" to documentation
- Updated "What's New"


ilyamandel (approved):
Thank you, @jeffriley -- looks good!

Issue #1138, opened 2024-05-24
AIC labels appear to be misapplied


ilyamandel:
We sometimes label things as AIC when there was no accretion onto the ONeWD...  E.g.

`./COMPAS -n 1 --random-seed 1716511244 --detailed-output
`
I believe this is a bug -- ONeWD::IsSupernova() just checks whether the mass is above the ECSN threshold, and it seems that it sometimes is without accretion...

ilyamandel:
OK, here's what we are currently doing: 

In GiantBranch::ResolveElectronCaptureSN(), we ask if the star should go through an ECSN, and if we decide it shouldn't, we make it an ONeWD with a mass equal to the minimum of the current CO core mass and the Chandrasekhar mass, MCH.

Then, in ONeWD, we say it should evolve to the next phase (and go through an AIC) if IsMassAboveEcsnThreshold() evaluates to true -- but that compares the mass to MECS, rather than MCH.  

Thus, if we form an ONeWD with a mass between MECS and MCH (because it didn't meet our ECSN criteria), it will immediately go through AIC, even without accreting.  I don't think this makes sense, and I think there are two possible solutions:

A) The maximal mass of an ONeWD is set to MECS rather than MCH at formation, AIC happens once mass exceeds MECS (as coded up now).

B) The maximal mass of an ONeWD is set to MCH at formation (as is done now), but AIC should only happen once the mass of the ONeWD exceeds MCH.

In both cases, mass growth, i.e., accretion, is required to create AIC.

I will follow option B) for now to resolve this, but happy to change to option A) if there's a preference for that.

ilyamandel:
Closed by #1144 (note that the resolution drastically reduces the rate of AICs).

PR #1144, opened 2024-05-25
Address AIC issue 

ilyamandel:
- Resolves AIC issue #1138 (see discussion in that issue)
- Other minor fixes


jeffriley (approved):
Looks good - just one question: do we need `WhiteDwarfs::IsMassAboveEcsnThreshold()` anymore? 

I'll approve - you can resolve the question and merge at your leisure.

ilyamandel:
Good point, @jeffriley -- and while removing it, I noticed that we have IsMassAboveChandrasekhar() already coded up, so I simplified the code by using that. ;-)

PR #1145, opened 2024-06-11
Dynamical Tides fixes

veome22:
This PR contains a few miscellaneous corrections to the implementation of tides in `--tides-prescription KAPIL2024`.

The following is a set of comparisons of the various Dynamical tides terms, explaining what has changed. 
The binary being tested has the following parameters: `--number-of-systems 1 --random-seed 1 --initial-mass-1 50 --initial-mass-2 50 --semi-major-axis 10.5 --eccentricity 0.1 --chemically-homogeneous-evolution NONE --tides-prescription KAPIL2024`.

## Comparison of all changes
Before going into details, here is a plot showing the effect of the various fixes in the PR. I am only plotting the absolute value of the (2,2) dynamical tide term here for clarity. Overall, the strength of Dynamical tides on the Giant Branch (where there is a convective envelope and a radiative intershell) will decrease by ~4 orders of magnitude.
image
Depending on how close the binary is to synchronization, the sign of dynamical tides can alternate across the different fixes, which we will see in the subsequent plots.

Here are the effects of each fix:

### Before any fixes
image

### IW Sign Fix
The sign of the Inertial Wave dissipation term has been corrected to always have the same sign as `(2*omega_orb - 2*omega_spin)`, since it corresponds to the (l=2, m=2) moment.
This fix affects the sign of dynamical tides and ensures synchronization.
image

### $\beta$ and $\gamma$ for tri-layered stars
The definitions of $\beta$ and $\gamma$ follow Ahuir+ 2021, Eqs. (118) and (119), respectively. However, these equations assume a bi-layered star where M = M_rad + M_conv env. This becomes problematic for evolved stars, where there is also a substantial M_conv core. The new definitions of $\beta$, $1-\beta$, and $\gamma$ explicitly use the mass of the radiative shell and convective envelope to get around this issue.
This fix makes dynamical tides weaker by ~1 order of magnitude.
image
 
### IW $\epsilon$ fix
I discovered that the definition of $\epsilon$ was incorrect. It should be based on Eq. (42) of Ogilvie. (2013).
This fix makes IW dynamical tides weaker by ~2 orders of magnitude. The sign of dynamical tides for this binary behaves differently than the others, likely because the tides are weak enough to maintain synchronicity. However, there are non-trivial contributions from Equilibrium tides (not shown here) as well.
image

### DIFF
--- a/src/BaseStar.cpp
+++ b/src/BaseStar.cpp
@@ -3314,7 +3314,7 @@ DBL_DBL_DBL_DBL BaseStar::CalculateImKlmDynamical(const double p_Omega, const do
     double w12 = ((p_Omega) - two_OmegaSpin);
     double w22 = ((p_Omega + p_Omega) - two_OmegaSpin);
     double w32 = ((p_Omega + p_Omega + p_Omega) - two_OmegaSpin);
-
+        
     if (utils::Compare(coreRadiusAU, 0.0) > 0 && utils::Compare(coreMass, 0.0) > 0) {                   // No GW dissipation from core boundary if no convective core
         double beta2Dynamical           = 1.0;
         double rhoFactorDynamcial       = 0.1;
@@ -3358,7 +3358,7 @@ DBL_DBL_DBL_DBL BaseStar::CalculateImKlmDynamical(const double p_Omega, const do
         double one_minus_alpha   = 1.0 - alpha;
         double beta              = radIntershellMass / m_Mass;
         double one_minus_beta    = envMass / m_Mass;
-        
+
         double alpha_2           = alpha * alpha;
         double alpha_3           = alpha_2 * alpha;
         double alpha_5           = alpha_3 * alpha_2;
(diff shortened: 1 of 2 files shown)


jeffriley (approved):
All good from my perspective!  Thanks @veome22

ilyamandel:
@veome22 , @jeffriley --

Sorry, too late to review this, and the changes are amazing, but I really don't like this line:

double one_minus_beta    = envMass / m_Mass;

This creates a variable one_minus_beta whose value, in general, may not be equal to 1 - beta (it won't be whenever there is a tri-layer structure).  This is potentially going to create confusion for future readers / developers of this code.  I strongly suggest renaming variables, or at the very least adding clear comments that this may not have the value that one might expect from the variable name.  Though, really, changing the variable name would be best.

veome22:
@ilyamandel thanks for the suggestion. Please see the fix in : https://github.com/TeamCOMPAS/COMPAS/pull/1146

PR #1146, opened 2024-06-12
Renamed variable in dynamical tides to avoid confusion

veome22:
A small variable name change, as requested by @ilyamandel .

Please let me know if I should also add this to the changelog.

### DIFF
--- a/src/BaseStar.cpp
+++ b/src/BaseStar.cpp
@@ -3357,7 +3357,7 @@ DBL_DBL_DBL_DBL BaseStar::CalculateImKlmDynamical(const double p_Omega, const do
         double alpha             = radiusIntershellAU / radiusAU;
         double one_minus_alpha   = 1.0 - alpha;
         double beta              = radIntershellMass / m_Mass;
-        double one_minus_beta    = envMass / m_Mass;
+        double envMass_over_totalMass    = envMass / m_Mass;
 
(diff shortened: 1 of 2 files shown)


jeffriley (commented):
Ok, since Ilya pointed out the variable...  why do we need it at all?  It's only in one equation isn't it?  Why not just use "envMass / m_Mass" in that equation?  I'm all for making things clearer and reducing cpu, but I don't see that we actually gain anything from declaring a new variable here.

veome22:
@jeffriley I agree, it only shows up once. I had defined `one_minus_beta` along with the other variables before I realized this. I can remove the definition.

And just to clarify-- does this change need a note in the changelog?

jeffriley:
> And just to clarify-- does this change need a note in the changelog?

Yes, we should.  I know it feels like it's not necessary in this case, but if we don't, then two users might have different code, but both labelled with the same version - that's not right...  Just note it as "code cleanup".

ilyamandel:
Thank you, @veome22 , looks good!  I agree with Jeff's suggestion of removing the extra variable you don't really need and updating the changelog.  @jeffriley , I'll be traveling back to Australia over the next couple of days, so I hope you can approve.

ilyamandel (approved):
Looks good!

PR #1147, opened 2024-06-19
Minor fixes to mass transfer

ilyamandel:
- Corrected check for nuclear timescale (slow case A) mass transfer
- Reduced MAXIMUM_MASS_TRANSFER_FRACTION_PER_STEP to 0.0001 to improve accuracy of orbital separation updates following mass transfer
- Corrected temperature units in Picker formula for Tonset used in the calculation of the convective envelope mass
- Code cleanup


jeffriley (approved):
Looks good - thanks @ilyamandel.

Up to you, but I think we could probably ditch `massAplusMassD` in BaseBinaryStar::CalculateMassTransferOrbit() and just use `massA + massD` in its place.

I'll approve and you can merge - with our without removing `massAplusMassD`

PR #1150, opened 2024-06-22
Mass transfer ODE

ilyamandel:
Replaced fixed-step, first-order integrator for orbital change after mass transfer with an adaptive-step, higher-order ODE integrator for improved speed and accuracy


jeffriley (approved):
Looks good!

PR #1154, opened 2024-07-01
Winds Patch 1

jmerritt1:
Two quick winds fixes:

1. Changed the VERY_MASSIVE_MINIMUM_MASS threshold to use m_Mass (current), rather than m_ZAMS.
2. Lowered VINK_MASS_LOSS_MINIMUM_TEMP from 12.5 to 8kK, to eliminate the short interval during CHeB when WR winds were active between the RSG and OB temperature ranges, at low Z.


jeffriley (changes requested):
Thanks JD.  A couple of things:

You need to update changelog.h and increment the version number.

Take the explanatory comments out of the code:

"// Changed from ZAMS to m_Mass (current)"
"(changed from 12.5kK to 8kK to fix WR/CHeB issue)"

and put them in changelog to explain the changes if you want - in a couple of versions time nobody will know what the code looked like before, so explaining changes in the code will be irrelevant after a short while - better to explain them in changelog

jmerritt1:
Ah, yes thanks for the reminder.

jeffriley (approved):
All good - thanks JD

jeffriley:
JD - looks like you didn't merge with the latest dev?  There are conflicts in changelog.h - you'll need to resolve those and then change the version number again (remember, both in the comments and in the version string at the bottom of the file)

PR #1155, opened 2024-07-02
fixed bug in detailed evolution plotter

reinhold-willcox:
...in the van den heuval diagrams. Previously, these were not showing differences in before / after values of properties that should have changed in a timestep. This was due to not correctly masking for the Record Type (==4 for the end of a timestep). This does not affect the standard evolution plots, only the van den heuval diagrams. 

Also, no differences are shown for the end state, the format is identical for the ZAMS state.


jeffriley (approved):
I'm happy to approve this, but just so you know, as well as type 4 records (TIMESTEP_COMPLETED"), both SSE and BSE detailed output files have "INITIAL_STATE" (type 1) and "FINAL_STATE" (type 5) records.

Merge when ready.

PR #1157, opened 2024-07-03
New error handling

jeffriley:
@ilyamandel there is a lot going on in this PR:

1. implementation of more coherent and robust error handling
2. added source files (all are .h file, so the makefile does not need to change)
3. (this will be controversial) deprecation of some program options, and some program option values
4. added debug functionality to show stack trace and halt the program - see the discussion and implementation of the SIGUSR2 signal handler in main.cpp
5. fixed what I believe was a defect in utils::SolveKeplersEquation() that was causing erroneous "out-of-bounds" warnings for the eccentric anomaly
6. code cleanup
7. I have not changed the online docs - I wanted to push this so you could start reviewing it - I will update the docs as you complete your review
8. I updated changelog.h, but I haven't fleshed it out yet - I won't do that until after your review so I know what's staying, what's being dropped, and what's changing
9. I have not created a new default yaml file - I won't do that until after your review
10. I advanced the version number to v03.00.00. I think the new error handling philosophy is enough of a change to the COMPAS functionality to warrant bumping the major number (let me know if you don't agree)

Description:

1. The new error handling philosophy is described at the top of the Errors.h file.
2. The added source files are mostly the result of me separating out sections of the constants.h file. IMO it had become too unwieldy, and sectioning it out seemed to be the reasonable thing to do. I broke constants.h into 5 separate file:
    a. constants.h - pretty-much just contains constants now
    b. typedefs.h - an existing file, but I moved the enum class declarations and associated label maps, except those that pertain directly to logging, from constants.h to typedefs.h
    c. LogTypedefs.h - new file containing logging-related type definitions, including things like definitions of the default record composition for the various log files
    d. ErrorCatalog.h - new file containing the COMPAS error catalog - this where symbolic names for errors are defined, and contains the mapping from symbolic name to error string
    e. EnumHash.h - contains the hash function for enum class types

Other added files are (only 1):

   a. stellarUtils.h - similar to utils.h but includes the stellar classes (and header files) so that functions are aware of the stellar types (not possible in utils.h because that file is includes in the stellar classes - the includes would become circular)

3. As I was implementing the new error handling and doing some code cleanup, I noticed that some of our program options had been implemented with names inconsistent with existing option names. As well, some of the option values (imo) didn't represent what the value actually meant (I may well be corrected on some, or all, of those - happy to revert my changes there if that's the case). I know we don't like to change too much that is user-facing, but I think it is important we are consistent with the user-facing stuff...

I have changed some option names, and the names of some option values - effectively deprecating some options and option values. I have left the deprecated options and option values in the code, alongside their replacements (if applicable). I have written code (in the Options class) to notify users when they use the deprecated options and option values that they are indeed deprecated and will be removed soon, and where applicable showing the replacement option name or option value name. The code doesn't require there be a replacement, and where's there's not one, that part of the deprecation notice is not displayed.

The options I deprecated are:

     1. `--mass-transfer`                       , replaced by `--use-mass-transfer` (to be consistent with `--use-mass-loss`)
     2. `--luminous-blue-variable-prescription` , replaced by `--LBV-mass-loss-prescription` (to be consistent with other 'prescription'-type options
     3. `--OB-mass-loss`                        , replaced by `--OB-mass-loss-prescription` (to be consistent with other 'prescription'-type options
     4. `--RSG-mass-loss`                       , replaced by `--RSG-mass-loss-prescription` (to be consistent with other 'prescription'- ype options
     5. `--VMS-mass-loss`                       , replaced by `--VMS-mass-loss-prescription` (to be consistent with other 'prescription'- ype options
     6. `--WR-mass-loss`                        , replaced by `--WR-mass-loss-prescription` (to be consistent with other 'prescription'- ype options
     7. `--kick-direction`                      , replaced by `--kick-direction-distribution` (to be consistent with other 'distribution'-type options
     8. `--mass-transfer-thermal-limit-accretor`, replaced by `--mass-transfer-thermal-limit-accretor-multiplier` (for consistency and to better describe the option)
     9. `--black-hole-kicks`                    , replaced by `--black-hole-kicks-mode` (for consistency and to better describe the option) 
    10. `--chemically-homogeneous-evolution`    , replaced by `--chemically-homogeneous-evolution-mode` (for consistency and to better describe the option)

I know changing these could be problematic, but if we're ever going to (and I think we should), now is the time to do it. I should note that the deprecated options will still work in the deprecation notice period, and even after they are removed many will continue to work because Boost only requires sufficient characters to match to uniquely identify an option.

The option values deprecated are:

     1. `--critical-mass-ratio-prescription`   , value `NONE` replaced by `ZERO`
     2. `--LBV-mass-loss-prescription`         , value `NONE` replaced by `ZERO`
     3. `--logfile-type`                       , value `NONE` replaced by `ZERO`
     4. `--luminous-blue-variable-prescription`, value `NONE` replaced by `ZERO`
     5. `--OB-mass-loss`                       , value `NONE` replaced by `ZERO`
     6. `--OB-mass-loss-prescription`          , value `NONE` replaced by `ZERO`
     7. `--RSG-mass-loss`                      , value `NONE` replaced by `ZERO`
     8. `--RSG-mass-loss-prescription`         , value `NONE` replaced by `ZERO`
     9. `--stellar-zeta-prescription`          , value `NONE` replaced by `ZERO`
    10. `--VMS-mass-loss`                      , value `NONE` replaced by `ZERO`
    11. `--VMS-mass-loss-prescription`         , value `NONE` replaced by `ZERO`
    12. `--WR-mass-loss`                       , value `NONE` replaced by `ZERO`
    13. `--WR-mass-loss-prescription`          , value `NONE` replaced by `ZERO`

The motivation for changing these is that (imo) ZERO is a better description of the option value. For example, NONE implies no (whatever) prescription, when what we really mean is that if the user chooses NONE we'll apply a value of 0.0 - to me that's better described by a prescription value of ZERO than NONE.
As I noted above though, I'm happy to be corrected and revert any or all of these.

As an example of the deprecation functionality, running COMPAS with the following commandline:

    ./compas --mode bse --random-seed 0 --logfile-type csv --fp-error-mode off --remnant-mass-prescription schneider2020 --mass-transfer --grid gridfile.txt --OB-mass-loss-prescription NONE --critical-mass-ratio-p none

and gridfile.txt:

    --initial-mass 12.3 --metallicity 0.02
    --initial-mass 1.3 --metallicity 0.012
    --initial-mass 22.3 --metallicity 0.015
    --initial-mass 32.5 --metallicity 0.01 --vms-mass-loss VINK2011
    --initial-mass 2.3 --metallicity 0.02
    --initial-mass 112.3 --metallicity 0.014
    --initial-mass 125.3 --metallicity 0.01 --rsg-mass-loss NONE

    COMPAS v03.00.00
    A binary star simulator
    
    DEPRECATION NOTICE: option '--mass-transfer' has been deprecated and will soon be removed.  Please use '--use-mass-transfer' in future.
    DEPRECATION NOTICE: option value 'NONE' for option '--critical-mass-ratio-prescription' has been deprecated and will soon be removed.  Please use 'ZERO' in future.
    DEPRECATION NOTICE: option value 'NONE' for option '--OB-mass-loss-prescription' has been deprecated and will soon be removed.  Please use 'ZERO' in future.
    
    Start generating binaries at Mon Jun 24 19:44:15 2024
    
    0: Allowed time exceeded: (Main_Sequence_>_0.7 -> Neutron_Star) + (Main_Sequence_>_0.7 -> Carbon-Oxygen_White_Dwarf)
    1: Double White Dwarf formed: (Main_Sequence_>_0.7 -> Carbon-Oxygen_White_Dwarf) + (Main_Sequence_>_0.7 -> Helium_White_Dwarf)
    2: Double White Dwarf formed: (Main_Sequence_>_0.7 -> Carbon-Oxygen_White_Dwarf) + (Main_Sequence_>_0.7 -> Carbon-Oxygen_White_Dwarf)
    DEPRECATION NOTICE: option '--VMS-mass-loss' has been deprecated and will soon be removed.  Please use '--VMS-mass-loss-prescription' in future.
    3: Double White Dwarf formed: (Main_Sequence_>_0.7 -> Carbon-Oxygen_White_Dwarf) + (Main_Sequence_>_0.7 -> Carbon-Oxygen_White_Dwarf)
    4: Stars merged: (Main_Sequence_>_0.7 -> Naked_Helium_Star_MS) + (Main_Sequence_>_0.7 -> Naked_Helium_Star_MS)
    5: Double White Dwarf formed: (Main_Sequence_>_0.7 -> Carbon-Oxygen_White_Dwarf) + (Main_Sequence_>_0.7 -> Oxygen-Neon_White_Dwarf)
    DEPRECATION NOTICE: option '--RSG-mass-loss' has been deprecated and will soon be removed.  Please use '--RSG-mass-loss-prescription' in future.
    DEPRECATION NOTICE: option value 'NONE' for option '--RSG-mass-loss' has been deprecated and will soon be removed.  Please use 'ZERO' in future.
    6: Double White Dwarf formed: (Main_Sequence_>_0.7 -> Carbon-Oxygen_White_Dwarf) + (Main_Sequence_>_0.7 -> Carbon-Oxygen_White_Dwarf)
    
    Generated 7 of 7 binaries requested
    
    Simulation completed
    
    End generating binaries at Mon Jun 24 19:44:15 2024   

As I was documenting this PR I realised there are a couple of program option property types (noted in enum class PROGRAM_OPTION and associated label map in LogTypedefs.h that will be deprecated but that Options::ShowDeprecations() won't display. Not sure yet whether just noting those on the 'What's New' page is enough, or whether I should write code to show the deprecation notice if they are specified (they are for the logfile-definitions file). I'm leaning toward the latter.

Also, when testing the new --fp-error-mode option I found a divide-by-zero in HG::Initialise. CalculateTimescales() calls GiantBranch::CalculateTimescales(), which has:

    double p1_p = p1 / gbParams(p);

Code cleanup:
There is a lot of code cleanup - sorry. I have identified problems or questions that need an answer by including the text "\*\*ilya\*\*" at the appropriates places, so that you can search for that string and answer, or at least offer an opinion. The rest you could probably just take a cursory look. I think downloading the source and just glancing at each file is probably going to easier than looking at the diffs...

And finally, I know we discussed removing the BE Binary code at some stage - what did we decide?

I think that's everything - if I remember something else I'll add it. I confess I rushed a bit today to get this submitted this evening - I'm conscious that you don't have a lot of time to review this.

I meant to add:

1. I used "soon" in the deprecation notice to give us some flexibility with when we actually remove a deprecated option or value. We could use a fixed date instead to give users a better idea of how long they have to change their configurations. Perhaps something like "... will be removed no earlier than Dec 31, 2024", or "... will be removed on or after Dec 31, 2024".
2. If a user specifies both a deprecated option, and its replacement, on the same commandline or gridline, the value for the non-deprecated option is used.
3. As you could probably tell from the sample output, notices for deprecated options on the commandline appear immediately after the splash message, and notices for deprecated options on gridlines are shown as they are encountered.


reinhold-willcox:
I'm ok with changing input parameters and their values, though this is a significant change. Should this be included as part of the switch over to v3?

jeffriley:
Thanks @reinhold-willcox .  This PR already rolls the version to v03 (the change to the error handling is the main driver for that - it's a big enough change both for developers and users to warrant bumping the major number).  I think this will probably be merged into dev before the Non-ZAMS initialisation PR and, if so, that PR should then roll the version to v04 - vice versa if this gets merged after the Non-ZAMS initialisation.

ilyamandel:
>  I think this will probably be merged into dev before the Non-ZAMS initialisation PR and, if so, that PR should then roll the version to v04 - vice versa if this gets merged after the Non-ZAMS initialisation.

Minor: I don't think we need to bump the major version number for the Non-ZAMS initialisation.  There were more significant changes during v02.

ilyamandel:
@jeffriley -- I am reviewing your description for now, not yet the code.  Because I am doing so during breaks at a conference, I may be adding comments slightly piecemeal, apologies for that.  

You write:

>> The motivation for changing these is that (imo) ZERO is a better description of the option value. For example, NONE implies no (whatever) prescription, when what we really mean is that if the user chooses NONE we'll apply a value of 0.0 - to me that's better described by a prescription value of ZERO than NONE.

That makes sense in principle, but I don't understand it for some of the options.  If I literally mentally replace ZERO with 0.0, what would a --logfile-type of 0.0 mean?

Elsewhere (I did look at the code in BaseStar.cpp), I don't actually see any mention of ZETA_PRESCRIPTION::ZERO in the new code; it doesn't seem that your intention was to set zeta to be 0.0 in this case, so, again, in what sense are we applying a value of 0.0?

I haven't checked all other examples, but similar logic may apply.

General point: I am not opposed to changes in input options in principle, and I like the approach to making such changes gradual via deprecating them and giving notice to the user, but I'd prefer to only do that when the new version is clearly an improvement over the old one, and I'm not convinced by the NONE->ZERO change.

Oh, and I am OK with keeping deprecation warnings with the "soon" wording. :)

ilyamandel:
>> Also, when testing the new --fp-error-mode option I found a divide-by-zero in HG::Initialise.

Yes, @veome22 noticed that some HG parameters are NaN in the first time step and this creates issues for tides.  I see that you solved this by including CalculateTimescales(); on line 54 of HG.h, which presumably will solve the NaNs issue.  I might actually steal that solution for a quick PR that I promised Veome, since this one might take a bit longer.  Hope you won't mind! :)

ilyamandel:
>> And finally, I know we discussed removing the BE Binary code at some stage - what did we decide?

I don't remember what we decided -- but yes, I think we should remove it.  One can identify Be XRBs in post-processing.

jeffriley:
@ilyamandel 

> That makes sense in principle, but I don't understand it for some of the options. If I literally mentally replace ZERO with 0.0,
> what would a --logfile-type of 0.0 mean?

I agree - but I don't think I changed `NONE` to `ZERO` for logfile-type - I just removed `NONE` (the code never checked for `NONE`, so we now don't allow it (the user has to use the default type (`HDF5`), or choose another valied type).

> I don't actually see any mention of ZETA_PRESCRIPTION::ZERO

Same - just removed `NONE`, didn't replace it with `ZERO` - `NONE` was never used in the code (we never set zeta to 0.0)

(When I say removed, I mean deprecated - will remove).

Also, I have almost finished updating the online documentation for these changes - next couple of days.

ilyamandel (changes requested):
program-options-list-defaults.rst: 

```
**--critical-mass-ratio-prescription** |br|
Which critical mass ratio stability prescription to use (if any).
Options: { NONE, CLAEYS, GE20, GE20_IC, HURLEY_HJELLMING_WEBBINK } |br|
``ZERO``    uses 0.0, and defaults to the zeta prescription for stability. |br|
```

NONE -> ZERO in the options list, but I do think this should remain NONE; in the description, it just defaults to the zeta prescription, there's no sense in which it is 0.0.

--luminous-blue-variable-prescription
Should this become --LBV-mass-loss-prescription?

``HURLEY`` : Hurley, Pols, Tout (2000) |br|
-->
``HURLEY`` : Hurley, Pols, Tout (2000) LBV winds only for LBV stars |br|


`ZERO turns off mass loss for giant stars (stellar types CHeB, FGB, AGB, TPAGB) below the RSG_MAXIMUM_TEMP. Also available are Vink and Sabhahit (2023), Beasor et al. (2020), Decin et al. (2023), Yang et al. (2023), Kee et. al (2021), and Nieuwenhuijzen and de Jager (1990, previous default).`

AGB->EAGB; I'd remove the mention of previous default, here and in the OB-mass-loss-prescription.

ilyamandel:
Re logfile type, I was going off your comment above rather the code.  Glad to hear we agree. :-)

> 3. `--logfile-type`                       , value `NONE` replaced by `ZERO`

jeffriley:
> NONE -> ZERO in the options list, but I do think this should remain NONE; in the description, it just defaults to the zeta 
> prescription, there's no sense in which it is 0.0. 

Ok, I'll take a look.


> --luminous-blue-variable-prescription
> Should this become --LBV-mass-loss-prescription?

It has...  I have left the deprecated options in - they will be removed after some period of time.  But the new prescription is there, and I'm pretty sure it is `--LBV-mass-loss-prescription`


> AGB->EAGB; I'd remove the mention of previous default, here and in the OB-mass-loss-prescription.

Will do - I didn't notice that or I would have removed it.

jeffriley:
> --logfile-type , value NONE replaced by ZERO

Oops - autopilot probably...

ilyamandel (commented):
In BaseBinaryStar.cpp, 
`double BaseBinaryStar::CalculateDEccentricityTidalDt(const DBL_DBL_DBL_DBL p_ImKlm, const double p_M1, const double p_R1, const double p_M2, const double p_Omega, const double p_SemiMajorAxis, const double p_Eccentricity)`
I see that you are now using m_Omega, m_Eccentricity, etc. -- but still passing in p_Radius, p_Mass2, etc.  What is the logic behind using the binary parameters but passing in the stellar ones?  One could, in principle, just specify whether one computing the impact of star1 or from star2, right?  [Similarly in BaseBinaryStar::CalculateDOmegaTidalDt, BaseBinaryStar::CalculateDSemiMajorAxisTidalDt]

ilyamandel:
> It has... I have left the deprecated options in - they will be removed after some period of time. But the new prescription is there, and I'm pretty sure it is `--LBV-mass-loss-prescription`

I don't see --LBV-mass-loss-prescription in that documentation file, nor in the list of options at its end; please check.

ilyamandel:
Question for me regarding `m_OrbitalVelocityPreSN = -nan("");` in BaseBinaryStar.cpp: 

Yes, this is only used for printing.  I propose using 0.0 for this value in this case (unbound binary) and adding an appropriate comment to the documentation.  

I would remove m_uK, the UK() function, and the corresponding printing option altogether -- this is trivial to compute in post-processing if/when someone needs it and we don't need to waste time on this in the code.

jeffriley:
> I don't see --LBV-mass-loss-prescription in that documentation file, nor in the list of options at its end; please check.

I haven't finished updating the the documentation files yet - should be done by the end of the week.

> ... I see that you are now using m_Omega, m_Eccentricity, etc. -- but still passing in p_Radius, p_Mass2, etc. ...

True.  I don't think I changed what Veome had implemented (I think I just changed the names of the parameters.  It is called twice in BaseBinaryStar::ProcessTides(), with different parameters.  

> What is the logic behind using the binary parameters but passing in the stellar ones?

Well, it was more not passing in the binary attributes because we can just use the class member variables inside the function.  But yes, we could retrieve the stellar parameters in BaseBinaryStar::CalculateDEccentricityTidalDt() if we know wrt which star we're calculating the eccentricity.  I can clean that up if you like.

ilyamandel (commented):
BaseBinaryStar.cpp:

`        case MT_ANGULAR_MOMENTUM_LOSS_PRESCRIPTION::CIRCUMBINARY_RING: {                                                            // based on the assumption that a_ring ~= 2*a*, Vinciguerra+, 2020 `

Change comment to 
// based on the assumption that a_ring = 2*a, Vinciguerra+, 2020

ilyamandel (commented):
BaseBinaryStar.cpp:

`if (utils::Compare(m_MassLossRateInRLOF,donorMassLossRateNuclear) == 0) // if transferring mass on nuclear timescale, limit mass loss amount based to rate * timestep (thermal timescale MT always happens in one timestep)`

superfluous 'based' in comment (my fault!)

`if (OPTIONS->CirculariseBinaryDuringMassTransfer()) { // circularise binary to the periapsis separation?`

Comment should just say "circularise binary" -- it may not be to periapsis, that's decided by the next line.

			    // ALEJANDRO - 23/11/2016 - Bug fix for systems which enter MT being eccentric.
			    // Previous values have to be the ones for periastron as later orbit is modified according to previous values.
			    // If you don't do this, you end up modifying pre-MT pre-circularisation orbit
			    // JR: todo: check that this is proper functionality, or just a kludge - if kludge, resolve it **Ilya**
			    
This is proper functionality; the issue here is that BaseBinaryStar::ResolveMassChanges() uses m_SemiMajorAxisPrev to compute new semi-major axis (see ` m_SemiMajorAxis = m_SemiMajorAxisPrev + m_aMassLossDiff + m_aMassTransferDiff` ) and similarly in CalculateWindsMassLoss().  The trouble is that we are basically updating the semi-major axis twice per timestep: due to mass loss / mass transfer and due to circularisation.  So I think it's fine as is, and I'd just remove the long comment and say something like 

// Update previous timestep values to those of the circularised binary to serve as a baseline for future updates.	    

ProcessTides() -- should we have a consistent naming scheme for functions that change the orbit in various ways (because of tides, because of winds, because of mass transfer, etc.)?

                    /*ILYA*/
`                    //if ((m_Star1->IsOneOf({ STELLAR_TYPE::MASSLESS_REMNANT }) || m_Star2->IsOneOf({ STELLAR_TYPE::MASSLESS_REMNANT })) || dt < NUCLEAR_MINIMUM_TIMESTEP) {`
                    
I left in these comments, sorry, should be removed...           

```
        // if we leave the following check in, almost all binaries will have error = floating_point_error
        // in the BSE_SYSTEM_PARAMETERS log file, but their evolution status will not indicate an error
```
        
I think we should leave the floating_point_error in the BSE_SYSTEM_PARAMETERS log file for the user's information, but I don't quite understand why almost all binaries will have it -- I would hope that almost none would...

OK, got to the end of BaseBinaryStar.cpp, have to stop here for the day.

jeffriley:
@ilyamandel 

>        // if we leave the following check in, almost all binaries will have error = floating_point_error
>        // in the BSE_SYSTEM_PARAMETERS log file, but their evolution status will not indicate an error
> I think we should leave the floating_point_error in the BSE_SYSTEM_PARAMETERS log file for the user's information, but 
> I don't quite understand why almost all binaries will have it -- I would hope that almost none would...

Ok, my comment was a little brief...

At that location in the code, we have successfully completed the evolution of the binary - the evolution status recorded in the system parameters file will not indicate "evolution stopped because an error occurred" and, on the face of it, the user has every right to assume they can trust the results.  If there had been an error of any kind thrown during evolution, we'd be in one of the catch blocks rather than at the location of the comment in question.  However, that does not guarantee that there was not a floating-point error during evolution - just that we didn't trap it and throw an error when it occurred.  We can check if a floating-point error occurred since the last time we cleared the floating-point error flags (which we do at the start of the evolution of every binary).  If we conduct that check here at the location of the comment, and find that a floating-point error did occur during the evolution of the binary, the only conclusion we can draw is that the user used "--fp-error-mode OFF" to turn off floating-point error traps (this is the default at the moment btw - because if we turn it ON more than 90% of systems fail with a floating-point error), but there was in fact a floating-point error encountered during evolution.  If we remove that check in the code, the error value written to the system parameters file (in addition to the evolution status) will be 0 (or no error).  If we leave that check in, then the error value written to the system parameters file will be non-zero, and indicate a floating-point error occurred.  That would mean we have evolution status = "OK" (i.e. didn't fail because of an error), but a non-zero value for error.  I think that is more informative - we'd be saying to the user "You turned fp traps off, but there actually was an fp error - maybe treat the results with caution", rather than just setting error = 0 and effectively saying to the user "all good - no problems - the results are trustable".

> I don't quite understand why almost all binaries will have it -- I would hope that almost none would...

My comment that almost all binaries will have error = floating_point_error refers to the code as it is (and as it would be after merging this PR).  As noted above, with the most recent version (current dev), more than 90% of systems report a floating-point error - until we identify and fix each of the places in the code that result in a floating-point error, that will continue to be true (it will presumably drop a little each time we find and identify a problem...).

jeffriley:
@ilyamandel

- I have updated the online docs - except the What's New page - I'll do that in the next day or two.
- I have not fleshed out the changes in changelog.h yet - just dot points.  I'll tidy that up when I update What's New
- I may have addressed some of your comments above with a few minor fixes, bit I will now start addressing them properly.  I'll respond here when I have done that with how I addressed them.

ilyamandel:
> I think that is more informative - we'd be saying to the user "You turned fp traps off, but there actually was an fp error - maybe treat the results with caution", rather than just setting error = 0 and effectively saying to the user "all good - no problems - the results are trustable".

Thank you for the explanation, @jeffriley  -- I agree with your recommendation.

> As noted above, with the most recent version (current dev), more than 90% of systems report a floating-point error

Wow!

jeffriley:
@ilyamandel , I have addressed you comments to date.

> If I literally mentally replace ZERO with 0.0, what would a --logfile-type of 0.0 mean?

My mistake.  I didn't actually replace NONE with ZERO here - I removed NONE.  It doesn't need a deprecation notice - NONE
was only ever present in the code and never presented to users as an option, so just removing it is safe.


> Elsewhere (I did look at the code in BaseStar.cpp), I don't actually see any mention of ZETA_PRESCRIPTION::ZERO in
> the new code; it doesn't seem that your intention was to set zeta to be 0.0 in this case, so, again, in what sense are
> we applying a value of 0.0?

Also my mistake.  I didn't actually replace NONE with ZERO here - I removed NONE. It doesn't need a deprecation notice -
NONE was only ever present in the code and never presented to users as an option, so just removing it is safe.


> ... but I'd prefer to only do that when the new version is clearly an improvement over the old one, and I'm not convinced by
> the NONE->ZERO change.

For all the option valued that were changed from NONE -> ZERO, or just --logfile-type and --stellar-zeta-prescription?

The full list of options where NONE is deprecated and to be replaced by ZERO is:

    --critical-mass-ratio-prescription
    --LBV-mass-loss-prescription
    --luminous-blue-variable-prescription (option also deprecated in favour of --LBV-mass-loss-prescription)
    --OB-mass-loss                        (option also deprecated in favour of --OB-mass-loss-prescription)
    --OB-mass-loss-prescription
    --RSG-mass-loss                       (option also deprecated in favour of --RSG-mass-loss-prescription)
    --RSG-mass-loss-prescription
    --VMS-mass-loss                       (option also deprecated in favour of --VMS-mass-loss-prescription)
    --VMS-mass-loss-prescription
    --WR-mass-loss                        (option also deprecated in favour of --WR-mass-loss-prescription)
    --WR-mass-loss-prescription

In each of those cases, I believe NONE (and now ZERO) really does mean use 0.0, and I think ZERO makes more sense
(but see note on --critical-mass-ratio-prescription below).


> **--critical-mass-ratio-prescription** |br|
> Which critical mass ratio stability prescription to use (if any).
> Options: { NONE, CLAEYS, GE20, GE20_IC, HURLEY_HJELLMING_WEBBINK } |br|
> ``ZERO``    uses 0.0, and defaults to the zeta prescription for stability. |br|
> NONE -> ZERO in the options list, but I do think this should remain NONE;
> in the description, it just defaults to the zeta prescription, there's no sense in which it is 0.0.

I don't really understand what "uses 0.0, and defaults to the zeta prescription for stability" really means.

In BaseStar::CalculateCriticalMassRatio(), the return value (qCrit) defaults to 0.0, and Remnants::CalculateCriticalMassRatio()
always returns 0.0.  I don't see it defaulting to anything else. As far as I can tell CalculateCriticalMassRatio() is only called in
BaseBinaryStar::CalculateMassTransfer(), and the return value is only used to determine if mass transfer is unstable.
I can certainly remove NONE (and ZERO), and just throw an error if the value is not in 
    { CLAEYS, GE20, GE20_IC, HURLEY_HJELLMING_WEBBINK }.

What about Remnants::CalculateCriticalMassRatio()? I suppose we could just leave it returning 0.0?


> ZERO turns off mass loss for giant stars (stellar types CHeB, FGB, AGB, TPAGB) below the RSG_MAXIMUM_TEMP. 
> Also available are Vink and Sabhahit (2023), Beasor et al. (2020), Decin et al. (2023), Yang et al. (2023),
> Kee et. al (2021), and Nieuwenhuijzen and de Jager (1990, previous default).
> AGB->EAGB; I'd remove the mention of previous default, here and in the OB-mass-loss-prescription.

Fixed.


>> In BaseBinaryStar.cpp,
>> double BaseBinaryStar::CalculateDEccentricityTidalDt(...)
>> I see that you are now using m_Omega, m_Eccentricity, etc. -- but still passing in p_Radius, p_Mass2, etc.
>> What is the logic behind using the binary parameters but passing in the stellar ones? One could, in principle,
>> just specify whether one computing the impact of star1 or from star2, right?
>> [Similarly in BaseBinaryStar::CalculateDOmegaTidalDt, BaseBinaryStar::CalculateDSemiMajorAxisTidalDt]
>
> Well, it was more not passing in the binary attributes because we can just use the class member variables
> inside the function. But yes, we could retrieve the stellar parameters in 
> BaseBinaryStar::CalculateDEccentricityTidalDt() if we know wrt which star we're calculating the eccentricity.
> I can clean that up if you like.

Fixed.


>> I don't see --LBV-mass-loss-prescription in that documentation file, nor in the list of options at its end;
>> please check.
>
> I haven't finished updating the the documentation files yet - should be done by the end of the week.

Documentation updates now complete (except, as noted, for What's New page - tbc)


> Question for me regarding m_OrbitalVelocityPreSN = -nan(""); in BaseBinaryStar.cpp:
>
> Yes, this is only used for printing. I propose using 0.0 for this value in this case (unbound binary)
> and adding an appropriate comment to the documentation.
>
> I would remove m_uK, the UK() function, and the corresponding printing option altogether -- this is  trivial to compute in
> post-processing if/when someone needs it and we don't need to waste time on this in the code.

Done.


> BaseBinaryStar.cpp:
>
> ... // based on the assumption that a_ring ~= 2*a*, Vinciguerra+, 2020
>
> Change comment to
> // based on the assumption that a_ring = 2*a, Vinciguerra+, 2020

Done


> BaseBinaryStar.cpp:
> 
> if (utils::Compare(m_MassLossRateInRLOF,donorMassLossRateNuclear) == 0)
> // if transferring mass on nuclear timescale, limit mass loss amount based to rate * timestep
> (thermal timescale MT always happens in one timestep)
>
> superfluous 'based' in comment (my fault!)

Fixed.


> if (OPTIONS->CirculariseBinaryDuringMassTransfer()) { // circularise binary to the periapsis separation?
>
> Comment should just say "circularise binary" -- it may not be to periapsis, that's decided by the next line.

Fixed.


>> // ALEJANDRO - 23/11/2016 - Bug fix for systems which enter MT being eccentric.
>> // Previous values have to be the ones for periastron as later orbit is modified according to previous values.
>> // If you don't do this, you end up modifying pre-MT pre-circularisation orbit
>> // JR: todo: check that this is proper functionality, or just a kludge - if kludge, resolve it **Ilya**
>
> This is proper functionality; the issue here is that BaseBinaryStar::ResolveMassChanges() uses m_SemiMajorAxisPrev
> to compute new semi-major axis (see  m_SemiMajorAxis = m_SemiMajorAxisPrev + m_aMassLossDiff + m_aMassTransferDiff)
> and similarly in CalculateWindsMassLoss(). The trouble is that we are basically updating the semi-major axis twice
> per timestep: due to mass loss / mass transfer and due to circularisation. So I think it's fine as is, and I'd just
> remove the long comment and say something like
>
> // Update previous timestep values to those of the circularised binary to serve as a baseline for future updates.

Fixed.


> ProcessTides() -- should we have a consistent naming scheme for functions that change the orbit in various ways
> (because of tides, because of winds, because of mass transfer, etc.)?

Yes, probably.  I've tried to have consistent names for functions that do similar things (Calculate*(), Evaluate*(), Resolve*(),
Determine*(), etc.).  Probably not urgent we resolve that in this PR - do you have suggestions for a consistent naming scheme
for functions that change the orbit in various ways?


> /*ILYA*/
> //if ((m_Star1->IsOneOf({ STELLAR_TYPE::MASSLESS_REMNANT }) || m_Star2->IsOneOf({ STELLAR_TYPE::MASSLESS_REMNANT })) || dt < NUCLEAR_MINIMUM_TIMESTEP) {
>
> I left in these comments, sorry, should be removed...

Fixed.

jeffriley:
Oh, and

>> As noted above, with the most recent version (current dev), more than 90% of systems report a floating-point error
>
> Wow!

Yeah.  I suspect (maybe hope...) that the problem is in just a few places that affect almost every system.

Download the code and run with `--fp-error-mode debug` - you probably won't get through the first system.

ilyamandel:
I am responding to some of the *Ilya* comments, and leaving a few for later (e.g., comment related to https://github.com/TeamCOMPAS/COMPAS/issues/1178 )

ilyamandel (approved):
I made some changes to enable compilation on MacOS, to address several queries, to correct the Hurley remnant mass prescription CalculateRemnantMass_Static() to handle black hole formation, etc.  Happy to approve and merge!

ilyamandel:
P.S. For q crit, the reason I reverted the change was that we were not checking that the prescription is 0.0, we were saying if the prescription is NONE, don't use q critical, instead use zeta. 

    else if (OPTIONS->QCritPrescription() != QCRIT_PRESCRIPTION::ZERO) {                                                        // determine stability based on critical mass ratios
        // NOTE: Critical mass ratio is defined as mAccretor/mDonor
        double qCrit = m_Donor->CalculateCriticalMassRatio(m_Accretor->IsDegenerate(), m_FractionAccreted);
        isUnstable   = utils::Compare((m_Accretor->Mass() / m_Donor->Mass()), qCrit) < 0;
    }
    else {                                                                                                                      // determine stability based on zetas
        isUnstable   = (utils::Compare(m_ZetaStar, m_ZetaLobe) < 0);
    }

PR #1162, opened 2024-07-04
Fix for issue #1160:

### LINKED ISSUE #1160: Weird error: cannot create first entry in BSE_System_Parameters output when using Eccentricity in logfile and a grid.
**Describe the bug**
This is a weird one, occuring only in a specific corner case. When I use `--logfile-definitions` with a logfile that includes eccentricity in `BSE_SYSPARMS_REC`, and I also include a gridfile, I get an error `ERROR: Unable to create HDF5 dataSet Eccentricity for file BSE_System_Parameters`. However, this error only shows up for the first simulated binary, which means this first system does not show up in the `BSE_System_Parameters` output. 

Weirdly, despite this error, all the other systems which appear in BSE_System_Parameters have eccentricity values (all set to 0, as expected). Even more weirdly, the error goes away if I do not use a gridfile. 

**Label the issue**
`urgency_moderate`      - This issue is not urgent
`severity_moderate`    - This is a minor bug with minimal impact

**To Reproduce**
Here's a MWE:

```
#logfile.txt
 BSE_SYSPARMS_REC += {
     BINARY_PROPERTY::ECCENTRICITY,
 }

#grid.txt
--initial-mass-1 10
--initial-mass-1 9

COMPAS --logfile-def logfile.txt --grid grid.txt 
```

If you run this, you will see the error only before the first output entry, and `BSE_System_Parameters` will contain only the second entry in the grid. 

**Expected behavior**
There should be no error, and the system should be present in the BSE_System_Parameters.

 - COMPAS v02.49.04

jeffriley:
Fix for issue #1160: added prefix "PO_" to all program option header strings to differentiate from stellar/binary properties.
(Looks like there might be some code cleanup in there too - sorry, I had forgotten I did that - but it's only a few lines)


ilyamandel (approved):
Changes look great, I just fixed one tiny comment typo (not of your making, @jeffriley ).  Leaving this for you to merge in in case you want to merge things in a particular order.  I'll look at the pulsar PR next -- perhaps you'll want to pull that in first, then merge into into this and update the changelog to make things more efficient?

PR #1183, opened 2024-08-07
Continue evolution of main sequence merger products beyond the main sequence

ilyamandel:
This update allows for the evolution of MS merger products beyond the end of the MS.

Also fixes extraneous prints and cleans up some typos.

### DIFF
--- a/online-docs/pages/Developer guide/Headers/constants-dot-h.rst
+++ b/online-docs/pages/Developer guide/Headers/constants-dot-h.rst
@@ -9,6 +9,6 @@ Any new constants being added to COMPAS should be added here.
 
 Refer to ``EnumHash.h`` for the definition of the type alias ``COMPASUnorderedMap``.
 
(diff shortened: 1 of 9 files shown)


brcekadam (commented):
Hi Ilya! Looks good, there's just one thing I noticed. In BaseBinaryStar.cpp line 2522, the comment says "avoid CEE if CH+CH **or** one star is a massless remnant", but && operator is used. Should there be || instead?

ilyamandel:
@brcekadam : Thank you for the careful check!  In the place you queried, the code literally says the following:

Execute Common Envelope evolution if (CE or Merger) AND it's not a CHE binary AND it's not a binary containing a massless remnant [the ! in C++ means NOT].  So the comment is accurate: do CE if neither CHE nor massless = avoid CE if either CHE or massless. ;-)

brcekadam:
Oh right, thank you for the explanation! I’ll go ahead and approve the changes :-)

brcekadam (approved):

PR #1186, opened 2024-08-21
Fix for Issue 1184

jeffriley:

PR #1188, opened 2024-08-22
Fix for Issue 1182

### LINKED ISSUE #1182: Switch log issue
The switch logs are no longer working as of version 03.00.00.

I believe the problem is with object IDs.  

Adding the following print statements in BinaryStar::PrintSwitchLog() shows that the objectIdSwitching (which correctly matches the value set in Star::SwitchTo() ) is not the object ID of either the primary or secondary.  So this must have to do with changes in object ID assignments, but I have not been able to diagnose further yet.

    OBJECT_ID primaryObjectId   = m_BinaryStar->Star1()->ObjectId();
    OBJECT_ID secondaryObjectId = m_BinaryStar->Star2()->ObjectId();
    OBJECT_ID objectIdSwitching = LOGGING->ObjectIdSwitching();

    std::cout<<primaryObjectId<<std::endl;
    std::cout<<secondaryObjectId<<std::endl;
    std::cout<<objectIdSwitching<<std::endl;

jeffriley:
Fixed this time...

### DIFF
--- a/src/BinaryConstituentStar.h
+++ b/src/BinaryConstituentStar.h
@@ -16,19 +16,13 @@ class BinaryConstituentStar: virtual public Star {
 
 public:
 
-    BinaryConstituentStar() : Star() {
-        m_ObjectId          = globalObjectId++;
-        m_ObjectPersistence = OBJECT_PERSISTENCE::PERMANENT;
-    };
+    BinaryConstituentStar() : Star() { };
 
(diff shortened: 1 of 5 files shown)


ilyamandel (approved):
Thanks, @jeffriley -- looks good!  
Please merge in #1186 (including changelog file), then feel free to pull into dev. :-)

PR #1190, opened 2024-08-22
Reinstate correctly functioning fp-error code for Linux; disable fp_error code for MacOS (will reinstate later)

jeffriley:
Reinstates the correct fp error handling code for Linux, and disables fp error handling for MacOS.  I will reinstate the MacOS code later when I figure out how to make it work properly for both INTEL and ARM architectures (it sort-of works for INTEL, but has some anomalies that need to be resolved; ARM will need some more work).

Silently ignores the `--fp-error-mode` option for MacOS.

### DIFF
--- a/src/changelog.h
+++ b/src/changelog.h
@@ -1255,9 +1255,13 @@
 //                                      - Continue evolution of main sequence merger products beyond the main sequence
 //                                      - Remove spurious print statement
 //                                      - Typo fixes
+// 03.00.03   JR - Aug 22, 2024     - Defect repairs:
+//                                      - Reinstate correctly functioning code for floating-point error handling for Linux
+//                                      - Disable floating-point error handling for MacOS - until I can figure out how to
+//                                        make it work properly for both INTEL and ARM architectures.
 
 
-const std::string VERSION_STRING = "03.00.02";
+const std::string VERSION_STRING = "03.00.03";
 
 
 # endif // __changelog_h__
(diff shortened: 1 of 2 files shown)


ilyamandel (commented):
First of all, apologies once again for my previous very crude changes that broke the functionality -- I just wanted a quick solution that would allow code to compile on all OS.  Thank you for these fixes.  I look forward to a Mac-friendly version, too. ;-)

On line 869, in the comment, "yes" seems slightly confusing (since it's a few lines away from the nearest if).  

Why is typecasting to void needed on lines 572 and 573 (but not 869 and 870)?

PRs #1186 and #1188 will need to be rolled in to this one, happy to approve once that's done.

jeffriley:
Thanks @ilyamandel 

> First of all, apologies once again for my previous very crude changes that broke the functionality -- I just wanted a quick
> solution that would allow code to compile on all OS.

No worries - understood.  I should have checked compilation on MacOS before I pushed the code...

> On line 869, in the comment, "yes" seems slightly confusing (since it's a few lines away from the nearest if).

Fixed (removed "yes" - it was a leftover from a copy&paste)

> Why is typecasting to void needed on lines 572 and 573 (but not 869 and 870)?

Technically casting the functional return to void is not required - the compiler is smart enough to handle it.  I like to add it explicitly so that future readers of the code know that I am deliberately discarding the return value, and that it's not a potential cause of failure.  I missed lies 869 and 870 -  I have now added "(void)" to those lines.

ilyamandel (approved):

PR #1191, opened 2024-08-23
fix pip install

avivajpeyi:


avivajpeyi:
Hey @jeffriley / @ilyamandel  -- i think something changed with the way the CI file has to be written, which is why the pip install was failing. 

The pip install now works, the current error is from some other pytest issue.

ilyamandel (approved):
Thanks for this fix, approving and merging.

PR #1194, opened 2024-08-23
Fix for Issue 1179 + code cleanup

### LINKED ISSUE #1179: Unsupported option PULSAR_BIRTH_SPIN_PERIOD_DISTRIBUTION::FIXED
We have the following code snippet in  NS.cpp:

        case PULSAR_BIRTH_SPIN_PERIOD_DISTRIBUTION::FIXED:                                                      // FIXED  constant value as used in default model in Oslowski et al 2011 https://arxiv.org/abs/0903.3538
            SHOW_WARN_STATIC(ERROR::UNSUPPORTED_PULSAR_BIRTH_SPIN_PERIOD_DISTRIBUTION, "Using spin = 0.0");     // show warning
            pSpin = 0.0;
            break;
            
As @jeffriley points out, it's odd to give the user the option "FIXED" here to then say it's not supported.  We should either support it or remove the option.

Ditto for PULSAR_BIRTH_MAGNETIC_FIELD_DISTRIBUTION::FIXED.

jeffriley:
- Remove unsupported option value (FIXED) for options PULSAR_BIRTH_SPIN_PERIOD_DISTRIBUTION and PULSAR_BIRTH_MAGNETIC_FIELD_DISTRIBUTION

- Remove BeBinary-related code (mostly already commented)


ilyamandel (changes requested):
Thanks, @jeffriley !

The one change request is for documentation: 
https://compas.readthedocs.io/en/latest/pages/User%20guide/Program%20options/program-options-list-defaults.html#options-props-p 
should be updated to no longer advertise FIXED for the two distributions you mentioned.

However, I'd really like to roll in #1080 first, since @Adam-Boesky is about to go offline -- so could you please check that one (it's waiting on you since you have a change request)?  

---
@yuzhesong -- flagging this one for your attention, no need to include this in #1082 as we discussed on Slack earlier.

ilyamandel (approved):
Thank you, @jeffriley -- looks good!

Issue #1152, opened 2024-06-27
Remove `example_bbh_compas_config` and use just the `compasConfigDefault.yaml+CLI` args in the CI


avivajpeyi:
People will forget to update both compasConfigDefault.yaml and example_bbh_compas_config. 
Lets update the CI to use compasConfigDefault.yaml (default settings) but only over-write the initial conditions as command-line arguments.

<details><summary>Details</summary>
<p>

From convo between Ilya+Avi: 
```
Avi: 
> ... I noticed some other errors in the CI logs:
https://github.com/TeamCOMPAS/COMPAS/actions/runs/9589100365/job/26442327180#step:8:27

Run example COMPAS job using the config example_bbh_compas_config.yaml
Commandline Options error: option '--mass-transfer-jloss-macleod-linear-fraction' is ambiguous and matches '--mass-transfer-jloss-macleod-linear-fraction-degen', and '--mass-transfer-jloss-macleod-linear-fraction-non-degen'

Use option '-h' (or '--help') to see (descriptions of) available options
python_version = 3
compas_executable_override /home/runner/work/COMPAS/COMPAS/src/COMPAS
grid_filename Grid_demo.txt
[...]

PR #1195, opened 2024-08-23
Fix pytest errors

avivajpeyi:
Fixing  two errors:
1. test_plot_detailed_evol: Easy, pthing issue
2. FastCosmicIntegrator: I think this has to do with a change in the 'example bbh' file not producing BBHs?


avivajpeyi:
Ah, @ilyamandel and I have spoken about (2) in the past: 
#1152

TLDR: 
Instead of using a 'comfig.yaml', Ilya suggests we use the `compasConfigDefault.yaml (default settings) but only over-write the initial conditions as command-line arguments... 

Now the ticky bit for me will be getting the command-line arguments such that we produce a BBH... Hmm

ilyamandel:
@avivajpeyi : exactly! 
The command line arguments that are my testing default are
`./COMPAS -n 1 --initial-mass-1 35 --initial-mass-2 31 -a 3.5 --random-seed 0 --metallicity 0.001 --detailed-output
`
This should be enough of an input for plotting the van den Heuvel diagrams, etc.
If you want to do this via a grid file and have more than one BBH for cosmic integration, you can just repeat this twice (without --detailed-output)

ilyamandel:
@avivajpeyi -- see, e.g., https://github.com/TeamCOMPAS/COMPAS/actions/runs/10543006208/job/29210464333

FAILED py_tests/test_fast_cosmic_integration.py::test_fast_cosmic_integration - IndexError: too many indices for array: array is 0-dimensional, but 1 were indexed
FAILED py_tests/test_fast_cosmic_integration.py::test_compas_output_has_dcos - IndexError: too many indices for array: array is 0-dimensional, but 1 were indexed
FAILED py_tests/test_plot_detailed_evolution.py::test_plotter - FileNotFoundError: [Errno 2] No such file or directory: '/van_den_heuvel_figures/2.png'

The first two failings are presumably because we hadn't been producing DCOs (see above) -- but the last one seems to be an issue that may have been introduced in your latest fixes?

ilyamandel:
Hi @avivajpeyi ,
Gentle ping on this one. ;-)

ilyamandel (approved):
Looks amazing -- thank you, @avivajpeyi !

PR #1205, opened 2024-08-27
Add Hendriks+ PPI prescription

SimonStevenson:
Add the pulsational pair instability prescription from Hendriks+23 (https://arxiv.org/abs/2309.09339).
This will allow us to sample over their parameter DeltaMPPI,COshift

Example usage:

./COMPAS --mode SSE --initial-mass r[10,240,1] --metallicity 0.0001 --wolf-rayet-multiplier 0.1 --pulsational-pair-instability-prescription HENDRIKS --PPI-upper-limit 80.0 --PISN-lower-limit 80.0 --add-options-to-sysparms never --PPI-CO-Core-Shift-Hendriks 5.0

Example plots:

MCO_v_Mrem
MHe_v_Mrem


ilyamandel (changes requested):
Hi @SimonStevenson ,

This looks really good, I just simplified the code in GiantBranch a tiny bit, please take a look.

One question: do we really want to change the maximum mass that users can run COMPAS with to 300 solar masses?  I thought our previous choice was based on the fact that we were increasingly uncomfortable with SSE track extrapolations to higher masses.

And two requests: 
- You want to update the documentation, and probably whatsnew, to reflect the appearance of the new option.
- Can you update ../compas_python_utils/preprocessing/compasConfigDefault.yaml  with the new default YAML file to include this option? (You can auto-generate it)

ilyamandel (changes requested):
Also, please update changelogs.h (and make sure the latest dev is merged in).

jeffriley (changes requested):
Thanks Simon.  A few things:

1. Update changelog (as discussed)
2. Update documentation - document new option and new choice for existing PPI prescription option (`HENDRIKS`), plus the relationship/dependence between them
3. Create new default yaml file
4. Add "What's New" entry to announce new option (and new value for existing option)
5. Question: should you add sanity checks for the new optiion value?  Can it really be any real value, or are there limits (high and low)?

EDIT: further to (5), it looks like you clamp the mass to 0.0 in the code.  Should there be an upper limit to which mass is clamped in the code (as well as any limits placed on the option value)?  The limits should be documented - both any limits you might decide to put on the option value, and the upper and lower limits you clamp the mass to in the code.

jeffriley:
I think this:

    m_Mass = std::min(m_HeCoreMass - DeltaMPPI, 0.0); // Remnant mass should be non-negative

should be

    m_Mass = std::max(m_HeCoreMass - DeltaMPPI, 0.0); // Remnant mass should be non-negative

SimonStevenson:
Thanks for the quick reviews! 

- Merged in dev
- Updated changelog 
- Updated What's New
- Updated documentation (program-options-list-defaults.rst and standard-logfiles-record-specification-options.rst)
- Generated a new compasConfigDefault.yaml
- Added basic sanity checking for pair instability options
- Fixed min -> max as mentioned by @jeffriley above

I changed the maximum mass that COMPAS will simulate so that I could make the plots above to compare with the ones in Hendriks+. I didn't really mean to leave that change in, so I've changed it back, but perhaps we could make that an option? Whilst it is certainly true that we don't really trust the SSE models at 150--300 Msun, I think that being able to explore what the code predicts there can still be useful (Hendriks+ are using binary_c and extrapolating to 300 Msun).

jeffriley (approved):
Looks good, thanks Simon.  Minor nit pick:

> COMPLAIN_IF(m_PulsationalPairInstabilityCOCoreShiftHendriks < -38.0, 
> "CO Core Shift parameter (--PPI-CO-Core-Shift-Hendriks) should be >-38.0")

Given the condition, comment should actually be 

    "CO Core Shift parameter (--PPI-CO-Core-Shift-Hendriks) should be >= -38.0"

I'll approve - over to @ilyamandel for his review

ilyamandel (approved):
Thanks for the min->max catch!

I don't actually understand where the value of 38.0 comes from... But happy to trust you. :)

SimonStevenson:
Gah, 38 is actually a typo, and should be 34.8 [EDIT: I think I was right the first time, please ignore this]

Re 38: In the Hendriks+ model, the onset of PPI mass-loss occurs at 34.8 Msun. You can see this in their paper, and in the PPIOnset variable. So the idea was just to not allow the user to shift more than this value (which would make the onset go negative). It's pretty arbitrary, and physically meaningful values are closer to 0 (maybe -10 to +10). I'll fix the typo.

SimonStevenson:
hmm or not, now I've confused myself, here's the relevant text:
Screen Shot 2024-08-28 at 5 34 36 pm
So they do assume that the onset occurs after MCO = 38 Msun. Please ignore my comment above about it being 34.8.

ilyamandel:
OK, but saying that the CO core is > 38 is a different statement than saying that the shift has to be >-38, right?

Maybe what should happen is that the term in parentheses,  PPIOnset in the code, should be non-negative?  But that again is a different condition...

Also, note that, in the current implementation, PPISN would become PISN if the remnant mass is <= 0, not <= 10 as the Hendriks+ text suggests.

[Again, I haven't looked at this closely, so perhaps that's what you intended -- if so, no objections.]

SimonStevenson:
>> Also, note that, in the current implementation, PPISN would become PISN if the remnant mass is <= 0, not <= 10 as the Hendriks+ text suggests.

Good point, I'll take a look at that tomorrow. We probably don't want to make 1 Msun black holes.

SimonStevenson:
OK I've added the check if the remnant mass is greater than 10 Msun. Here's an updated plot with that condition (I checked they become type 15 MR as well). 

MCO_v_Mrem

Think this should be ready to go once someone checks this (1 line change)

jeffriley (approved):
All good - thanks Simon

Issue #502, opened 2021-01-07
(Giant) Core He Burning stars have WR winds at low Z


LiekeVanSon:
**Describe the bug**
At Z = 0.001, for fiducial pythonSubmit, stars with MZAMS [15-30] have WR winds as their dominant Mass loss rate when they pass the VINK_MASS_LOSS_MINIMUM_TEMP  on their way to becoming red giants (and still have an envelope). 

This only seems to occur for low metallicities (at least, it doesn't happen for Z = Zsun or Z = 0.01)

**Label the issue**
`urgency_moderate` - This is a moderately urgent issue
`severity_moderate` - This is a moderately severe bug

**To Reproduce**
Run the attached pythonSubmit.txt (changed to .txt so I could upload it to this bug report). Also attached are required COMPAS_Output_Definitions and a grid (SingleStar_grid) to reproduce attached plots (coloured by Dominant_Mass_Loss_Rate)

This pythonSubmit is basically the  Fiducial one exccept for: luminous_blue_variable_prescription = 'HURLEY_ADD'
But the same behaviour occurs when setting luminous_blue_variable_prescription = 'BELCZYNSKI'

[COMPAS_Output_Definitions.txt](https://github.com/TeamCOMPAS/COMPAS/files/5782645/COMPAS_Output_Definitions.txt)
[SingleStar_grid.txt](https://github.com/TeamCOMPAS/COMPAS/files/5782648/SingleStar_grid.txt)
[pythonSubmit.txt](https://github.com/TeamCOMPAS/COMPAS/files/5782664/pythonSubmit.txt)

![HR_Z0 001](https://user-images.githubusercontent.com/57514659/103914769-9ff3a600-510a-11eb-84a7-eac1ed4308f9.png)

**Expected behaviour**
These are massive stars with an envelope so I think they should have NieuwenHuijzen & de Jager as dominant mass loss.

**Screenshots**
Detailed evolution for a 15 and 30 Msun example are attached:

_30 Msun_
![30Msun_envelopeCcore](https://user-images.githubusercontent.com/57514659/103914887-bf8ace80-510a-11eb-8a66-97f351bd6e2d.png)
![30MsunWinds](https://user-images.githubusercontent.com/57514659/103914882-be59a180-510a-11eb-94f1-184056d3fa5b.png)

_15 Msun_
[...]

PR #1206, opened 2024-08-28
Winds patch 2

jmerritt1:
-Changed the logic applying RSG winds to include a stellar type check for HG or `GIANTS`, preventing GB winds from being applied to high-mass(20-40), high-Z HG stars in the RSG parameter space. This issue came up in the fix for WR winds being applied during CHeB [(Issue 502)](https://github.com/TeamCOMPAS/COMPAS/issues/502).

-Changed all FLEXIBLE2023 wind prescription naming to MERRITT2024, as suggested in Ilya's comment on the paper draft. 

-Will need to generate a new default .yaml, and change examples.


jeffriley (changes requested):
Changes look ok, thanks JD.

As you noted, you'll need to create a new default yaml file (use `./compas --create-yaml-file filename.yaml`, and move `filename.yaml` into the appropriate directory to replace the default yaml file).  You'll also need to merge dev and resolve any conflicts.

Interestingly, the online documentation for `--mass-loss-prescription` does not mention `BELCZYNSKI2010` or `FLEXIBLE2023` (now replaced by `MERRITT2024`).  That needs to be updated (in file `program-options-list-defaults.rst`) - I'm surprised we let that through last time...

Ordinarily I'd be concerned about changing option values that people might have in configuration files etc., especially the default value, but since it was never documented maybe it is ok to just change it here and hope it doesn't break anyone's configuration files.  You might want to add a `What's New` entry to announce that `FLEXIBLE2023` has changed to `MERRITT2024` (even though it wasn't changed in the documentation for `--mass-loss-prescription`).

EDIT: I forgot to add that I'm not completely happy with this construct:

    (IsOneOf(GIANTS) || m_StellarType == STELLAR_TYPE::HERTZSPRUNG_GAP))

(because I don't like code that picks out a specific stellar type - if we can avoid that we should) but it's ok for now - one day when I get time I might try to rationalise the initialiser lists (and hopefully create a list that covers this circumstance).

ilyamandel (commented):
In the documentation, NONE vs ZERO:

Options: { ZERO, HURLEY, BELCZYNSKI2010, MERRITT2024 } |br|
``NONE``          : No wind mass loss |br|

ilyamandel (commented):
From changelog:

"// - Related to issue #502: added HG to allowed RSG stellar type check, preventing GB winds from being applied during HG."

Aren't you now *allowing* GB winds to be applied during HG?

jmerritt1:
> From changelog:
> 
> "// - Related to issue #502: added HG to allowed RSG stellar type check, preventing GB winds from being applied during HG."
> 
> Aren't you now _allowing_ GB winds to be applied during HG?

We are allowing RSG winds to be applied, where appropriate, rather than the older GB (which, depending on the stellar type may take the greatest of NJ, VW, KR, WR winds).

jeffriley (changes requested):
Changes look good thanks JD.  Just one thing: you pushed your Makefile - you need to remove it.

jeffriley (approved):
All good :-)

ilyamandel:
> In the documentation, NONE vs ZERO:
> 
> Options: { ZERO, HURLEY, BELCZYNSKI2010, MERRITT2024 } |br| `NONE` : No wind mass loss |br|

@jeffriley , @jmerritt1 -- this request seems to have been missed; just mentioning it here so I don't forget to add it in another PR.

PR #1208, opened 2024-08-28
Issue 1170 - new option to add WD binaries to BSE DCO file

### LINKED ISSUE #1170: program option to create log of WD binaries
It would be nice to have a program option to (whether or not to) include WDs in the double compact objects log. Or there can be separate log file altogether with WD-WD, WD-NS and WD-BH binaries. Along with this, individual WDs can be allowed to evolve in their respective binaries.

jeffriley:
Tested with run of 10000 binaries - seems to work as expected.


ilyamandel (approved):
Leaving it to you and @SimonStevenson to figure out who should merge first, since you are both in Aspen -- the other will need to merge in dev into their PR. :)

jeffriley:
I'll wait for SImon and JD's PRs to be merged.  Might also need to recreate a new default yaml after merging dev.

SimonStevenson:
#1205 and #1206 are now merged, so you can go ahead and resolve the conflicts above and then we can get this in.

PR #1216, opened 2024-09-05
Improved treatment of chemically homogeneously evolving stars

SimonStevenson:
For a while I have had some code that implements some new features aimed at improving the treatment of chemically homogeneously evolving stars within COMPAS. 

This is meant to help address the overproduction of BBHs via CHE that was found in Stevenson & Clarke 2022 and Riley & Mandel 2023.

1. Increasing main sequence lifetimes and luminosities of CH stars to match detailed models. I made some polynomial fits to some BOoST models from Szecsi et al. 2015, 2020
logLuminosityRatioTwoPanelIZW18
lifetimeRatioTwoPanelIZW18

2. Transition from OB mass loss to WR mass loss towards the end of the main sequence when the surface helium abundance becomes high. This is following an implementation from Yoon et al. 2006 that is also used by Marchant et al. and du Buisson et al. 

3. Rotationally enhanced mass loss for rapidly rotating stars following Langer (1998) 
4. Scaling the vinf parameter with metallicity as in Vink et al. 2001 

Here is a comparison of the primary mass distribution  (pdf/CDF) using the default COMPAS model (with all changes set to False) and a run using the updated model (commands to reproduce below).

```
./COMPAS --mode BSE --chemically-homogeneous-evolution-mode PESSIMISTIC --enhance-CHE-lifetimes-luminosities FALSE --scale-CHE-mass-loss-with-surface-helium-abundance FALSE --enable-rotationally-enhanced-mass-loss FALSE --scale-terminal-wind-velocity-with-metallicity-power 0.0 -n 1000000 --metallicity-distribution LOGUNIFORM --initial-mass-min 10.0 > default.out &
./COMPAS --mode BSE --chemically-homogeneous-evolution-mode PESSIMISTIC --enhance-CHE-lifetimes-luminosities TRUE --scale-CHE-mass-loss-with-surface-helium-abundance TRUE --enable-rotationally-enhanced-mass-loss TRUE --scale-terminal-wind-velocity-with-metallicity-power 0.0 -n 1000000 --metallicity-distribution LOGUNIFORM --initial-mass-min 10.0 > enhanced.out &
```

This is using the default parameters for FastCosmicIntegration.py as follows:

```
compas_fast_cosmic_integrator --path ./COMPAS_Output/COMPAS_Output.h5
```

m1_cdf_comparison_CHE
m1_hist_comparison_CHE

This reduces the BBH merger rate by a factor of ~2.

[EDIT: Updated plot with larger run to make it smoother]


ilyamandel:
Note the precheck failure -- it flags homogenously -> homogeneously :-)

ilyamandel (changes requested):
In BaseStar.h, we still have CalculateEddingtonCriticalRate() -- should this now use the new functionality instead, making use of the surface He abundance of the donor to compute the opacity?  

There's also utils::MESAZAMSHeliumFractionByMetallicity() -- I think this should be combined with the new functionality (i.e., leave only one function for computing the initial He mass fraction, no need for two that do identical things).

ilyamandel (changes requested):
In CH.h:

double CalculateHeliumAbundanceCoreOnPhase() const { return BaseStar::CalculateHeliumAbundanceCoreOnPhase(); };

and other similar cases:

I don't understand what's happening here.  If you want all stellar types that descend from BaseStar.h to do the same thing, there's no reason to include such code in the .h -- they will do this automatically.  If you don't, then presumably some descendant of BaseStar upstream from CH should do something different, and CH should go back to this operation -- but I don't see this happening.

Also, if double CalculateHeliumAbundanceCoreOnPhase() always just returns the value stored in the class variable, without doing any calculations (that's what the BaseStar version does), i.e., is just a getter, having "Calculate" in the function name is a bit of a misnomer.

But is that class variable -- in this case, m_HeliumAbundanceCore -- really being updated properly?

Ahh, wait, I do see it's being recalculated in CH.cpp -- so why define it to have a different return value in CH.h?

ilyamandel (commented):
Should CH::UpdateAgeAfterMassLoss() use the MainSequence version of CalculateLifetimeOnPhase, not the adjusted CH version?  Why?

ilyamandel (changes requested):
Even though you may not currently envision using functions like  HeliumAbundanceSurface() or HydrogenAbundanceCore() for stellar types like HeWD or HeGB or COWD, etc., since you've defined them for all stars (in BaseStar), future developers may reasonably expect them to return sensible things for any stellar type.  I am not convinced they currently would do that, so this needs to be fixed, to prevent future errors.

ilyamandel:
Oh, since we made another pull recently, you'll need to merge in dev.

BTW, despite the barrage of comments, I should say that this is a really nice and very thorough PR.  Really nicely done, @SimonStevenson !

SimonStevenson:
Thanks for the comments @ilyamandel ! I'll get to work on addressing these.

jeffriley:
@ilyamandel 

> In CH.h:
>
> double CalculateHeliumAbundanceCoreOnPhase() const { return BaseStar::CalculateHeliumAbundanceCoreOnPhase(); };
>
> and other similar cases:
>
> I don't understand what's happening here. If you want all stellar types that descend from BaseStar.h to do the same
> thing, there's no reason to include such code in the .h -- they will do this automatically. If you don't, then presumably 
> some descendant of BaseStar upstream from CH should do something different, and CH should go back to this 
> operation -- but I don't see this happening.

The reason for this is that when a virtual function is overridden in a derived class by a function with a signature that is not declared in the base class, _all variants_ of the function are hidden by the overriding function.  This doesn't normally affect us (but has in the past).  Here @SimonStevenson has declared a new variant of the `CalculateHeliumAbundanceCoreOnPhase()` function in `CH.h`:

    CalculateHeliumAbundanceCoreOnPhase(const double p_Tau) const;

This variant is not declared in `BaseStar.h`, which contains only:

    virtual double CalculateHeliumAbundanceCoreOnPhase() const { return m_HeliumAbundanceCore; }

Declaring `CalculateHeliumAbundanceCoreOnPhase(const double p_Tau) const;` in `CH.h` hides the declaration in `BaseStar.h`, so compiling (without the declaration of `CalculateHeliumAbundanceCoreOnPhase()` in `CH.h`) results in the infamous compiler warning:

    warning: 'CH::CalculateHeliumAbundanceCoreOnPhase' hides overloaded virtual functions [-Woverloaded-virtual]
    double          CalculateHeliumAbundanceCoreOnPhase(const double p_Tau) const;
                        ^
    ./BaseStar.h:538:33: note: hidden overloaded virtual function 'BaseStar::CalculateHeliumAbundanceCoreOnPhase' declared here: different number of parameters (0 vs 1)
    virtual double              CalculateHeliumAbundanceCoreOnPhase() const                                             { return m_HeliumAbundanceCore; } 

We can get around that in two ways:

(a) we can declare `CalculateHeliumAbundanceCoreOnPhase(const double p_Tau)` in `BaseStar.h`, or
(b) we can declare (as @SimonStevenson) has done, declare `CalculateHeliumAbundanceCoreOnPhase()' in `CH.h`

In retrospect, given your comment re the use of these functions for other stellar type, maybe we should have chosen (a).  Perhaps @SimonStevenson (or I) can make those changes.  Then we would not need the declaration of `CalculateHeliumAbundanceCoreOnPhase()' in `CH.h`.

ilyamandel (commented):
@jeffriley :

You are right that I was careless and conflated two different function declarations in CH.h:

 double          CalculateHeliumAbundanceCoreOnPhase(const double p_Tau) const;
    double          CalculateHeliumAbundanceCoreOnPhase() const { return BaseStar::CalculateHeliumAbundanceCoreOnPhase(); };
    
So ignore the particular comment I made.

That said, I am still very puzzled by the CalculateHeliumAbundanceCoreOnPhase() [no argument] version.  Unless I am missing something, it returns the value of m_HeliumCoreAbundance... and is used to set the value of m_HeliumCoreAbundance in BaseStar.cpp. ;-)  I don't see where this function is actually defined to calculate the He core, or how the current code will ensure that the value of m_HeliumCoreAbundance will be kept accurate.  Ditto for other similar functions, I was just using this as an example. 
 Probably I am missing something, of course.

jeffriley:
Well, I don't know :-). I haven't actually reviewed the code yet - I just helped fix the compiler warning (which I think we could do a bit better as described above).

I see what you're saying - I'll talk to @SimonStevenson

SimonStevenson:
I am copying what is done for e.g., CalculateLuminosityOnPhase or CalculateHeCoreMassOnPhase. The function is defined as virtual in BaseStar with no arguments, and simply returns the relevant parameter. Then, I am overriding this function in the derived class (in this case, CH.h).

I had not originally implemented these functions for MS, evolved stars or remnants. I have now added a rudimentary version of this, but note that this is really not the main point of this PR, and these variables are not currently used for any calculations for these stellar types. A future PR should look to improve the tracking of these abundances for these stars. 

Here are some plots for both CH and non-CH stars showing X and Y as a function of time (made using a custom log file definitions file, which should make @jeffriley happy :) ).

1 Msun non-CHE single star 

Abundances_v_time_1M

For the non-CH star, it burns H->He in its core on the main-sequence, but the surface abundances are unchanged. Then the central helium is burned during the CHeB phase. I set the abundances to 0 for the remnants. 

CHE star

CHE_abundances_v_time_1

For the CH star, the surface and core abundances are the same during the CH/MS phase, as expected as the star is homogeneous. Then the core helium is burned during the HeMS phase, and the surface abundances remain unchanged. 

I have only focused on single stars + CHE stars. I haven't looked at any binary interactions; it is likely that stars that lose/gain mass will not currently have sensible surface abundance properties. Fixing this is on the to do list, but feels a bit like scope creep for this PR. 

I have removed utils::MESAZAMSHeliumFractionByMetallicity() and replaced the calls that used it with a Getter called InitialHydrogenAbundance; please check that maintains the spirit of the edited code.

ilyamandel:
Thank you for these plots and related updates, @SimonStevenson !  I agree that this wasn't the goal of this PR, but it's nice not to have functions that provide obviously wrong information (you might remember they do, but other developers or users could rely on them without knowing), so it's nice to fix this.  

Just FYI, because of how inheritance works, you did not need to update, say, ONeWD.h with functions that do exactly the same thing as in COWD.h that it inherits from.  It's not wrong, so not objections from me (unless @jeffriley objects), just unnecessary.

ilyamandel:
> Should CH::UpdateAgeAfterMassLoss() use the MainSequence version of CalculateLifetimeOnPhase, not the adjusted CH version? Why?


I asked this question earlier and didn't get an answer.  I just wanted to check that this was an intentional choice to use MainSequence::CalculateLifetimeOnPhase() in CH::UpdateAgeAfterMassLoss() rather than a typo.

ilyamandel (commented):
>     double                              MESAZAMSHeliumFractionByMetallicity(double p_Z);


should be removed from utils.h

ilyamandel:
> In BaseStar.h, we still have CalculateEddingtonCriticalRate() -- should this now use the new functionality instead, making use of the surface He abundance of the donor to compute the opacity?

I think this hasn't been resolved.  Should we do this later?  Or do you think this shouldn't be changed at all?

SimonStevenson:
I reran the population test overnight and here is the updated plot:

m1_cdf_comparison_CHE

SimonStevenson:
> > Should CH::UpdateAgeAfterMassLoss() use the MainSequence version of CalculateLifetimeOnPhase, not the adjusted CH version? Why?
> 
> I asked this question earlier and didn't get an answer. I just wanted to check that this was an intentional choice to use MainSequence::CalculateLifetimeOnPhase() in CH::UpdateAgeAfterMassLoss() rather than a typo.

I think this is currently ok as I then multiply this by the relevant ratio if the user has chosen to do so. I had not implemented a seperate function for CalculateLifetimeOnPhase for CH stars. I could do that, and move that functionality into that function, and then just call that function here if you think that is better.

SimonStevenson:
> > ```
> > double                              MESAZAMSHeliumFractionByMetallicity(double p_Z);
> > ```
> 
> 
> should be removed from utils.h

Must have missed this, now removed.

SimonStevenson:
> > In BaseStar.h, we still have CalculateEddingtonCriticalRate() -- should this now use the new functionality instead, making use of the surface He abundance of the donor to compute the opacity?
> 
> I think this hasn't been resolved. Should we do this later? Or do you think this shouldn't be changed at all?

Yeah sorry, I wasn't really sure what to do with this. Is it the helium abundance, or the hydrogen abundance that is relevant?  Hurley et al. 2002 have 
Screen Shot 2024-09-13 at 11 17 23 am
where X is the surface hydrogen abundance.

Do I just need to add another argument (the surface hydrogen abundance of the donor, X) to this function and to CalculateMassAcceptanceRate replacing the factor of 1.7 with (1+X)?

ilyamandel:
> > > In BaseStar.h, we still have CalculateEddingtonCriticalRate() -- should this now use the new functionality instead, making use of the surface He abundance of the donor to compute the opacity?
> > 
> > 
> > I think this hasn't been resolved. Should we do this later? Or do you think this shouldn't be changed at all?
> 
> Yeah sorry, I wasn't really sure what to do with this. 

The assumption for the Eddington limit is that opacity is driven by scattering of photons off free electrons, and that everything is ionised, so what really matters is the number of electrons per unit mass.  Technically, metals also matter, but are usually ignored, so the big difference is that H has 1 electron per nucleon and He has 2 electrons per 4 nucleons, or 0.5/nucleon.  Thus, opacity is twice lower per unit mass of He than H, which means photon pressure is twice weaker for the same mass accretion rate, so the Eddington limit would be twice as high for a pure He star than for a pure H star.  You can see this in the Hurley formula you quoted: if X=1, the limit is twice lower than if X=0.  So, yes, I think the right thing to do is to just pass the donor's surface hydrogen abundance in.  But I am happy for this to be left as a future modification if you prefer, so will leave this choice to you and happy to approve the PR as it stands.  

Thank you!

ilyamandel (approved):

ilyamandel:
In your court, @SimonStevenson -- I am not merging in case you want to make that change to CalculateEddingtonCriticalRate() as part of this PR, but leaving the decision to you.

SimonStevenson:
Thanks Ilya, I think @jeffriley had some changes he wanted to make, so holding off merging until he has had a chance to add those

ilyamandel:
Sounds good, @SimonStevenson .  Just a heads up that I've approved Lieke's PR in the meantime, so depending on whether @jeffriley approves that one or this one first, he will need to do another dev pull and merge to ensure that things like the changelog and version number are accurate.

SimonStevenson:
I've merged in dev again and fixed the conflicts with the version number (making this the most recent version). @jeffriley let me know when you've had a chance to look at this/make any changes you want so that we can get this merged in.

jeffriley (approved):
All good - made a few minor changes.  Thanks @SimonStevenson

SimonStevenson:
Awesome, thanks Jeff!

Issue #1219, opened 2024-09-06
Option --black-hole-kicks-mode (aka --black-hole-kicks) ignored


jeffriley:
The following COMPAS runs all produce exactly the same output:

    ./compas --black-hole-kicks-mode zero --initial-mass-1 50 --random-seed 0
    ./compas --black-hole-kicks-mode reduced --initial-mass-1 50 --random-seed 0
    ./compas --black-hole-kicks-mode fallback --initial-mass-1 50 --random-seed 0
    ./compas --black-hole-kicks-mode full --initial-mass-1 50 --random-seed 0

The reason is that this call

    if (thisSNevent == SN_EVENT::CCSN) {                                                        // core-collapse supernova event this timestep?
        vK = ReweightSupernovaKickByMass(vK, m_SupernovaDetails.fallbackFraction, m_Mass);      // yes - re-weight kick by mass of remnant according to user specified black hole kicks option, if relevant (default is no reweighting)
    }

in `BaseStar::CalculateSNKickMagnitude()` is always defaulting to calling `BaseStar::ReweightSupernovaKickByMass()`, which just returns `vK` as passed.  The reason `BH::ReweightSupernovaKickByMass()` is never called is that at the time of the call the switch of stellar type has not been done, and the star is not a remnant.

I think the simplest fix is to make `BH::ReweightSupernovaKickByMass()` static, and only call it (in `BaseStar::CalculateSNKickMagnitude()`) if the predicted stellar type of the remnant is `BLACK_HOLE`.

This problem was introduced in COMPAS v02.44.04 in the fix for issue #1027 

`urgency_high`     - This is a very urgent issue and should be resolved as soon as possible
`severity_major`    - This is a severe bug

Run COMPAS as described above.

**Expected behavior**
The different values for `--black-hole-kicks-mode` produce different outcomes.

**Versioning**
 - OS: MacOS Sonoma 14.6.1
 - COMPAS v03.01.06

SimonStevenson:
I am also wondering what happens by default when this is used in 'FALLBACK' mode with our default 'MULLERMANDEL' `--remnant-mass-prescription`. Does that set the value of the `p_FallbackFraction`? Is that parameter set to 0 by default, but the 'MULLERMANDEL' prescription is already internally accounting for fallback?

ilyamandel:
MULLERMANDEL has a slightly different way of accounting for partial fallback, so I don't think this should impact that prescription.

[...]

PR #1221, opened 2024-09-06
Fix for issue 1219: Option --black-hole-kicks-mode (aka --black-hole-kicks) ignored

jeffriley:
- Fix for issue #1219: Option --black-hole-kicks-mode (aka --black-hole-kicks) ignored
- Fixed some stray typos


ilyamandel (approved):
Looks good, @jeffriley !

Issue #1211, opened 2024-08-30
Do we need --mass-loss-prescription NONE (now ZERO)?


jeffriley:
Users have the following choices for the option `--mass-loss-prescription`: 

    {NONE, ZERO, HURLEY, BELCZYNSKI2010, MERRITT2024}

(`NONE` is deprecated in favour of `ZERO` and will be removed at the end of 2024).

In the interests of reducing redundant/unnecessary options/option values, given that we can turn mass loss off with `--use-mass-loss false`, do we really need `NONE`/`ZERO` as options for `--mass-loss-prescription`?  Running COMPAS with `--use-mass-loss false` produces the same results as running it with `--mass-loss-prescription NONE` or `--mass-loss-prescription ZERO`.

It seems a little odd to enable mass loss and then set the option to have no mass loss - why not just disable mass loss?

Versioning:
 - COMPAS v03.01.06

ilyamandel:
No objections to removing NONE/ZERO since we have the --user-mass-loss false option... 
But my preference would actually be to reduce the total number of command-line possibilities, so I'd rather keep the option --mass-loss-prescription ZERO (since we have to have --mass-loss-prescription anyway), and instead remove --use-mass-loss false.

LiekeVanSon:
We also have `--overall-wind-mass-loss-multiplier` which could be set to 0 to produce the same result in yet another way. 

Also related I wonder what the behavior of DOMINANT_MASS_LOSS_RATE is for all different ways of setting the mass loss to zero? I also noticed that the documentation for dominant mass loss rate is now outdated, as it should reflect # enum class MASS_LOSS_TYPE: int { NONE, GB, LBV, OB, RSG, VMS, WR};

jeffriley:
> But my preference would actually be to reduce the total number of command-line possibilities, so I'd rather keep the
> option --mass-loss-prescription ZERO (since we have to have --mass-loss-prescription anyway), and instead remove 
> --use-mass-loss false.

Happy to do that - I thought that might be considered too radical :-). We can deprecate `--use-mass-loss` and eventually remove it.

> We also have --overall-wind-mass-loss-multiplier which could be set to 0 to produce the same result in yet another way.

Good point.  Should we limit the multiplier to > 0 and <= some upper limit?

> Also related I wonder what the behavior of DOMINANT_MASS_LOSS_RATE is for all different ways of setting the mass 
> loss to zero?

I can check that.

[...]

PR #1223, opened 2024-09-07
Fixed application of WolfRayetFactor

LiekeVanSon:
Small changes to fix the application of OPTIONS->WolfRayetFactor() to all WR prescriptions (currently it only applies to part of the WR mass loss prescriptions in an inconsistent way)

I tried to only apply WolfRayetFactor() at the highest level possible without applying WR factor twice

(somewhat related to issue #1211; there is no 'NONE' option for WR mass loss, making WolfRayetFactor() more essential than the other wind multiplication factors)


jeffriley (changes requested):
Thanks Lieke.  I haven't looked at the code changes yet - I'll do that later.  Just a note to say that you need to describe the changes in changelog.h, and change COMPAS version number.

ilyamandel (commented):
Hi @LiekeVanSon ,

I haven't traced through where the WR factors are applied carefully, but I trust you.  

However, from the point of future development, I find it rather confusing that some functions (with similar names) now apply this factor and some don't.  E.g., CalculateMassLossRateWolfRayetZDependent() and CalculateMassLossRateHeliumStarVink2017() do not, while CalculateMassLossRateWolfRayetShenar2019() and CalculateMassLossRateBelczynski2010() do.  This seems like an error-prone situation for future development.  Is there a way to make this more uniform?

jmerritt1:
Thanks Lieke, we agree that moving the WR factor up a level to the wrapper functions is much clearer and less error prone.

LiekeVanSon:
Hi All, sorry for the slow response!

 I agree with your worries that implementing the WR factor in lower levels is not what we want. 
 I removed OPTIONS->WolfRayetFactor()  from the Shenar 2019 function, but I believe the problem is that CalculateMassLossRateMerritt2024 using WR_MASS_LOSS_PRESCRIPTION::BELCZYNSKI2010 points back to the wrapper function of the overall wind prescription BELCZYNSKI2010. 

 So I think we _have_ to apply the WR factor directly in BELCZYNSKI2010, and cannot do it again in the MERRITT2024 + WR:BELCZYNSKI2010 case


I now apply the WR factor as follows in the   "HURLEY" , "BELCZYNSKI2010" , and "MERRITT2024" prescriptions as below:

**HeMS::CalculateMassLossRateHurley()** 
-  **_OPTIONS->WolfRayetFactor()** *_ CalculateMassLossRateWolfRayet(0.0)

**HeMS::CalculateMassLossRateBelczynski2010()**
- **_OPTIONS->WolfRayetFactor()** *_ CalculateMassLossRateWolfRayetZDependent(0.0);

**HeMS::CalculateMassLossRateMerritt2024()**
- case WR_MASS_LOSS_PRESCRIPTION::SANDERVINK2023:{
 MdotWR = **_OPTIONS->WolfRayetFactor()** *_ std::max(Mdot_Sander2023, Mdot_Vink2017);
 
- case WR_MASS_LOSS_PRESCRIPTION::SHENAR2019: {
MdotWR = **_OPTIONS->WolfRayetFactor()** *_  std::max(Mdot_Shenar2019, Mdot_Vink2017);

- WR_MASS_LOSS_PRESCRIPTION::BELCZYNSKI2010: {
MdotWR = CalculateMassLossRateBelczynski2010(); // OPTIONS->WolfRayetFactor() is applied in Belczynski2010 function 


let me know if this makes sense to you all :)

ilyamandel (approved):
Thank you, @LiekeVanSon !

ilyamandel:
@jeffriley -- waiting for your approval.

jeffriley:
Sorry - I was en route to Australia I think...

jeffriley (approved):
Thanks @LiekeVanSon !

PR #1228, opened 2024-09-25
Fix to issue affecting tides and 2-stage CEs

ilyamandel:
Use m_Mass0 rather than m_Mass in GiantBranch::CalculateConvectiveEnvelopeMass() to avoid negative mass radiative intershells as consequences of artificially low BAGB core masses


jeffriley (approved):
All good

PR #1229, opened 2024-09-26
Fix for PR #1216

brcekadam:
There were missing lines in Options::OptionValue(const T_ANY_PROPERTY p_Property) after PR #1216. This was causing a weird error when I was specifying the initial semi-major axis as a range.


SimonStevenson (approved):
Awesome, thanks for spotting and fixing this Adam! The changes look good to me and the tests pass so I'm approving this.

Issue #1180, opened 2024-07-27
Intermediate-mass stars may have unrealistic convective envelope masses and radii


ilyamandel:
GiantBranch::CalculateConvectiveEnvelopeMass() calculates the mass of the convective envelope using the Picker, Hirai, Mandel (2024) formalism for all GB stars, even though the formalism was calibrated for massive stars only.  We should check whether these fits yield reasonable results for intermediate-mass stars and if not, revert to Hurley+ fits below the mass applicability threshold of Picker+ fits, at least for now.

The radial extent of the convective envelope is calculated in HG.h CalculateRadialExtentConvectiveEnvelope() on the basis of the current and maximum convective envelope core mass.  I suspect a poor fit to the latter is particularly problematic for intermediate mass stars, and is leading to artefacts such as extremely low convective envelope radii / high densities. For now, I will cap the maximum convective envelope mass for intermediate mass stars to the total envelope mass (since such stars have a negligible radiative intershell late in their evolution). 

However, more careful investigation is required, so creating an issue for this.

veome22:
Here are a few tests I ran for for additional context, comparing the convective envelope masses and radii between the Hurley+(2000) and the Picker, Hirai, Mandel (2024) models.

The dashed lines correspond to the onset of HG and FGB. 

Generally speaking, below initial masses of 7 $M_\odot$, the convective envelope grows too abruptly for HG stars in the Picker+ (2024) model. The Hurley model seems to do a better job here. Both models are somewhat consistent with the envelope mass at the base of the FGB, although of course the Picker+(2024) model produces a continuous curve.

At and above initial masses of 7 $M_\odot$, the Picker+ (2024) model predicts no convective envelope mass on the Giant Branch. While the Picker+(2024) model is consistent with this along the HG, the Hurley+ (2000) model results in a significant convective envelope by the end of the HG. 

|  Mass  | Radius |
| ------------- | ------------- |
|   <img width="1084" alt="image" src="https://github.com/user-attachments/assets/b2da5304-ed37-4f3a-9760-177037f8fdf3"> | <img width="959" alt="image" src="https://github.com/user-attachments/assets/f2b8aaa4-2e8f-4f8b-93b6-53419d070fe9"> |
[...]

PR #1230, opened 2024-09-29
Multiple changes to the prescriptions for the masses and radii of convective and radiative layers in stars

ilyamandel:
- Picker+ 2024 prescription for the convective envelope mass corrected so that T_onset is always 0.1 dex hotter than T_min in order to avoid artefacts due to differences between MESA and Pols SSE tracks

- A range of changes to ensure that convective core and convective envelope masses and radii vary smoothly wherever possible, including improvements to convective core mass and radius on the main sequence, on the Helium MS, TPAGB now have fully convective envelopes, etc.

- All naked helium stars have purely radiative envelopes, until we develop better models

- Minor fixes to code and documentation elsewhere

### DIFF
--- a/src/EAGB.cpp
+++ b/src/EAGB.cpp
@@ -855,7 +855,7 @@ double EAGB::CalculateRemnantRadius() const {
     double R1, R2;
     std::tie(R1, R2) = HeGB::CalculateRadiusOnPhase_Static(m_HeCoreMass, CalculateRemnantLuminosity());
 
-    return std::min(R1, R2);
+    return std::min(m_Radius, std::min(R1, R2));
 }
 
 
--- a/src/GiantBranch.cpp
+++ b/src/GiantBranch.cpp
@@ -679,6 +679,32 @@ double GiantBranch::CalculateRemnantRadius() const {
 #undef massCutoffs
 }
 
+/*
+ * Calculate the radial extent of the convective outer envelope
+ *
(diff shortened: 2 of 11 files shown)


jeffriley (approved):
Looks good, thanks @ilyamandel 

I made a few minor formatting changes, and made the signature uniform for the Clone() function in all classes.  I tested my changes and I believe they produce the same results as your changes.

I approve the PR, but will leave open for other reviewers/tests - you can merge at your leisure :-)

veome22:
Hi @ilyamandel! Just for the sake of posterity, I am noting here that the updated envelope code fixes the bugs from before. The convective envelope mass and radius for a He burning star are now reported as ~0, and the radiative envelope can be correctly inferred from the remaining components. 
Here is a very quick plot I made for the mass evolution of a star that starts out as a 12.5 $M_\odot$ star
image

ilyamandel:
Thank you for the check, @veome22 !

I don't think a 12.5 Msun star should self-strip through winds -- was this one in a binary? -- but I think this is enough to close #1180 , so I'll do that now.

veome22:
> Thank you for the check, @veome22 !
> 
> I don't think a 12.5 Msun star should self-strip through winds -- was this one in a binary? -- but I think this is enough to close #1180 , so I'll do that now.

Yes, this is a binary system with masses 17.5 + 12.5. The mass shown is for the secondary. I did not run an SSE system.

PR #1232, opened 2024-10-03
Add tidal timestep limit, minor tides enhancements

veome22:
- Added limits to the BSE `ChooseTimestep()` function, such that KAPIL2024 tides don't make binary parameters change by more than 1% in a given timestep.
- Updated the dynamical tides model to ignore GW dissipation if there is a convective core as well as a convective envelope, following a suggestion from Jim Fuller.
- Added a fractional threshold of 1E-4 for the mass and radius of relevant stellar structural components (cores, envelopes), below which tides do not act. This is necessary to avoid numerical instabilities, such as dividing by very small values.
- Updated the online documentation to suggest using `--chemically-homogeneous-evolution-mode NONE` alongside `--tides-prescription KAPIL2024`. This is a stopgap until CHE can be made compatible with tides.


ilyamandel (commented):
Hi @veome22 ,
This looks great!  A couple of small requests and one question.
Requests: 
- Define LOWEST_FRACTION (with a more descriptive name, such as LOWEST_SHELL_FRACTION_FOR_TIDES or something of that sort) in constants.h .
- Define the 0.01 maximal change when computing timesteps in constants.h as well -- this is all to make it easier to find/change these parameters in the future, or even make them user-defined.
- Remind me (sorry, in a bit of a rush this morning), but what does m_Omega even mean in the context of BaseBinaryStar -- aren't there two different Omega's for the two component stars?

veome22:
Hi @ilyamandel !
>     * Remind me (sorry, in a bit of a rush this morning), but what does m_Omega even mean in the context of BaseBinaryStar -- aren't there two different Omega's for the two component stars?

`m_Omega` is the binary orbital frequency. In principle I could require that the spin of each star shouldn't change much between timesteps (relative to their respective spins), but when component spins are very small, this is too restrictive of a condition. So I chose to take the timestep such that any given star won't spin up by more than 1% of the orbital frequency instead. This still ensures that, for instance, stars don't get spun up beyond synchronization, while also allowing slow rotators to evolve on reasonable timescales.

ilyamandel (approved):
Looks great, @veome22 -- thank you for the changes!

PR #1235, opened 2024-10-05
Changed order of binary and stellar evolution in BaseBinaryStar::Evolve()

jeffriley:
Changed order of binary and stellar evolution in BaseBinaryStar::Evolve().  The evaluation of the binary is now done before stellar evolution.

@ilyamandel there is some code cleanup in this PR.  Only the changes to:

. changelog.h
. LogTypedefs.h
. standard-logfiles-record-types-bse-detailed-output.rst
. BaseBinaryStar::Evolve() (lines 2723..2970) are relevant to the changed functionality

The rest of the changes are code cleanup and do not change functionality.

I have run a quick test of 1000 binaries using the commandline

    ./compas --random-seed 0 -n 1000 --metallicity-distribution LOGUNIFORM --chemically-homogeneous-evolution-mode NONE --scale-terminal-wind-velocity-with-metallicity-power 0.0 --wolf-rayet-multiplier 1.0 --mass-change-fraction 0.0075 --radial-change-fraction 0.0075 --pulsational-pair-instability-prescription HENDRIKS --PPI-CO-Core-Shift-Hendriks 0.0 --PPI-upper-limit 80.0 --PISN-lower-limit 80.0 --common-envelope-formalism TWO_STAGE --tides-prescription KAPIL2024

which created 44 DCOs, whereas before this change none were created.  I have not checked any other consequences of the changes.

### DIFF
--- a/online-docs/pages/User guide/COMPAS output/standard-logfiles-record-types-bse-detailed-output.rst
+++ b/online-docs/pages/User guide/COMPAS output/standard-logfiles-record-types-bse-detailed-output.rst
@@ -33,3 +33,5 @@ Following is a list of the BSE Detailed Output file record type numbers and corr
    Record was logged immediately following mass resolution (i.e. after winds mass loss & mass transfer complete)
 #. POST_MASS_RESOLUTION_MERGER |br|
    Record was logged immediately following a merger after mass resolution
+#. PRE_STELLAR_TIMESTEP |br|
+   Record was logged immediately prior to stellar timestep (i.e. the evolution of the constituent stars for a single timestep)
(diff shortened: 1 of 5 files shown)


ilyamandel (commented):
// JR: note, this will (probably) fail if option --fp-error-mode is not OFF (the calculation that resulted in m_SemiMajorAxis = inf will (probably) result in a trap)

@jeffriley -- I think in the future, we should move away from using infinite or negative semi-major axes, or eccentricities outside the [0,1] range, as proxies for unbound binaries. :)

ilyamandel (approved):
Looks good -- let's try this! ;-)

jeffriley:
> @jeffriley -- I think in the future, we should move away from using infinite or negative semi-major axes, or eccentricities outside the [0,1] range, as proxies for unbound binaries. :)

I agree entirely!

PR #1237, opened 2024-10-07
Changed the prescription for Tonset in the Picker+ models to take advantage of improved metallicity-dependent fits

ilyamandel:


jeffriley (approved):

PR #1238, opened 2024-10-10
Fixes to SN resolution after USSNe and SwitchLog printing

ilyamandel:
- Reverted IsCCSN() to include USSN following a change in 3.00.00 that inadvertently led to no binary orbit updates following USSNe
 - Include a call to EvolveBinary(0.0) on initialisation of evolved stellar types; this ensures that Switch logs include consistent data for the new stellar type


jeffriley (approved):
All good - thanks @ilyamandel

PR #1242, opened 2024-10-14
New qcrit tables

reinhold-willcox:
New MT stability tables for qCrit from [Ge et al. 2024] (https://ui.adsabs.harvard.edu/abs/2024arXiv240816350G/abstract)


jeffriley (changes requested):
Looks good, thanks @reinhold-willcox.  Just a few things:

You need to update `changelog.h` and increment the version number.

In BaseStar::InterpolateGe20QCrit():

Remove

    double logZ = log10(m_Metallicity);

and use class variable `m_Log10Metallicity` instead

Change

    double logZlo = log10(0.001);

to

    double logZlo = -3;    // log10(0.001)

So we don't evaluate the `log10` every time the function is called.  Unfortunately declaring it `constexpr` won't work there.

Consider changing    

    double interpolatedQCrit = qcritPerMetallicity[1] + (logZ - logZhi)*(qcritPerMetallicity[1]-qcritPerMetallicity[0])/(logZhi-logZlo);
    return interpolatedQCrit;

to

    return qcritPerMetallicity[1] + (m_Log10Metallicity - logZhi) * (qcritPerMetallicity[1] - qcritPerMetallicity[0]) / (logZhi - logZlo);

No need to create a variable that is only used to return a value on the next line.  The compiler will collapse those two lines anyway, but I think changing it to just a `return` statement is clearer in the code (less clutter).

You could also remove

    double logZhi = LOG10_ZSOL;

and just use `LOG10_ZSOL` in the equation.  I know why you've declared it - just not sure it adds that much to clarity - up to you (the compiler will probably do exactly that anyway).

reinhold-willcox:
Good points, thanks @jeffriley. I kept Zhi for clarity, but made all the other changes.

jeffriley (approved):

PR #1251, opened 2024-10-23
Fix for Issue 1246 and others

### LINKED ISSUE #1244: Typo / inconsistency in documentation
There is a small inconsistency / typo in the documentation (program option page)

At three places, instead of MullerMandel, the word MandelMueller is used.

### LINKED ISSUE #1246: Slow performance compared to older versions
Since v3.00.00 I see a strange performance issue. If I give 5 or 6 parallel runs of COMPAS with the same arguments (varying only metallicity), I find that some runs finish in 5-6 hours, while others take up to 12.5 hours.

In earlier versions, I could give up to 8 parallel runs (with the same command as used for later versions) and all would complete in under 3 hours. I have sufficient memory and storage as I suppress the creation of most of the spacious log files.

Also, the run (for a given metallicity) pauses very briefly at certain times (perhaps it encounters some floating point errors). In earlier versions it went smooth throughout.

I use the following command. I only vary the metallicity and output container values across runs.

--mode 'BSE' --initial-mass-min 5.0 --eccentricity-distribution 'THERMAL' --number-of-systems 1000000 --switch-log False --detailed-output False --logfile-type 'CSV' --allow-rlof-at-birth False --logfile-supernovae '' --logfile-common-envelopes '' --logfile-rlof-parameters '' --metallicity 0.0105 --random-seed 0 --output-container Z.0105

Version (last checked):

COMPAS: 3.05.02

Likely cause:

Elaborate error handling scheme

OS: MacOS
Memory: 18GB
Storage: >100GB free

jeffriley:
- Fix for issue #1246 - performance degradation
   - Restructured deprecations code to revert performance degradation introduced in v03.00.00
- Added OptionDefaulted() function to Options class - see documentation there.
- Reverted change to `utils::SolveKeplersEquation()` made in v03.00.00 - no idea what I was thinking...
- Changes to documentation per issue #1244.

The main reason for this PR is to fix the performance degradation introduced in v03.00.00 (issue #1246).

The root cause of the performnace degradation was the treatment of deprecated option and values introduced in v03.00.00.  We give users a grace period between announcing deprecated options and/or option values, so users can continue to specify deprecated options and option values until the grace period is over and we remove the deprecated options and/or option values.  Because during that grace period users can specify either the deprecated option, or its replacement (if applicable - some options will just be removed, while others will be replaced with different names), we need to check which option value to retrieve.  In versions prior to this PR, we performed that check by calling the function OPTIONS->OptionSpecified() to determine which of the options was specified (the deprecated option, or the replacement).  The OPTIONS->OptionSpecified() function determines if an option was specified by traversing the list of specified options provided by Boost (Boost creates the list durng its parsing of the options).  Traversing that list doing string comparisons takes some time, and doing it every timestep for one or more deprecated options is what introduced the performance degradation (mea culpa - it just didn't occur to me at the time - I wrote the OPTIONS->OptionSpecified() function a long time ago...).

To try to resolve the problem, I wrote a new function (OPTIONS->OptionDefaulted()) which is good proxy for OPTIONS->OptionSpecified() (if an option was not defaulted, then it was specified), but only if the option name provided to OPTIONS->OptionDefaulted() is a valid option name.  It should be, since the person writing the code needs to specify it, but that can't be guaranteed (so caveat utilitor).  Determining if an option was defaulted is less onerous that determining if it was specified (by a factor of about 7), but as it turns out not enough to fix the performance degradation.  However, it is useful and I have left it in the code.

Fortunately, COMPAS parses the options provided by the user before we pass them to Boost for parsing - so that we can expand (e.g) ranges and sets, etc. (things that Boost knows nothing about).  I have now added code to the COMPAS parsing of the options to detect and replace (where necessary) deprecated options and option values.  This means that the COMPAS code beyond the initial parsing of options does not need to determine which version of the option/value was specified by the user - by that time, because we detect and replace deprecated options during the COMPAS pre-parse of options, we know that the replacement will be specified if one exists, or the deprecated option will be specified where no replacement exists, obviating the need to determine which.  This has resolved (reverted) the performance degradation.

I also changed remaining calls to OPTIONS->OptionSpecified() to OPTIONS->OptionDefaulted() - carefully checking that in each case the option string passed is a valid option.  These calls are outside the evolution loops - they are at the creation of the star or binary, so only performed once per system evolved.  Still, every little bit helps.

Also addressed in this PR is issue #1244, and a poor change to `utils::SolveKeplersEquation()` made in v03.00.00 (that shouldn't have had much impact, but needed to be fixed).

### DIFF
--- a/online-docs/pages/User guide/Program options/program-options-list-defaults.rst
+++ b/online-docs/pages/User guide/Program options/program-options-list-defaults.rst
@@ -83,13 +83,13 @@ Default = FALSE
 :ref:`Back to Top <options-props-top>`
 
 **--black-hole-kicks** |br|
-Black hole kicks relative to NS kicks (not relevant for `MANDELMUELLER` ``--remnant-mass-prescription``). |br|
+Black hole kicks relative to NS kicks (not relevant for `MULLERMANDEL` ``--remnant-mass-prescription``). |br|
 Options: { FULL, REDUCED, ZERO, FALLBACK } |br|
 Default = FALLBACK |br|
 DEPRECATION NOTICE: this option has been deprecated and will soon be removed. Please use ``--black-hole-kicks-mode`` in future.
 
 **--black-hole-kicks-mode** |br|
-Black hole kicks relative to NS kicks (not relevant for `MANDELMUELLER` ``--remnant-mass-prescription``). |br|
+Black hole kicks relative to NS kicks (not relevant for `MULLERMANDEL` ``--remnant-mass-prescription``). |br|
 Options: { FULL, REDUCED, ZERO, FALLBACK } |br|
 Default = FALLBACK
 
@@ -961,7 +961,7 @@ FALSE indicates PPISN remnants will receive no natal kicks. |br|
 Default = TRUE  
 
 **--neutrino-mass-loss-BH-formation** |br|
-Assumption about neutrino mass loss during BH formation (works with `FRYER2012` or `FRYER2022` ``--remnant-mass-prescription``, but not `MANDELMUELLER`). |br|
+Assumption about neutrino mass loss during BH formation (works with `FRYER2012` or `FRYER2022` ``--remnant-mass-prescription``, but not `MULLERMANDEL`). |br|
 Options: { FIXED_FRACTION, FIXED_MASS } |br|
 Default = FIXED_MASS
 
--- a/online-docs/pages/whats-new.rst
+++ b/online-docs/pages/whats-new.rst
@@ -4,7 +4,17 @@ What's new
 Following is a brief list of important updates to the COMPAS code.  A complete record of changes can be found in the file ``changelog.h``.
 
 
-**LATEST RELEASE** |br|
+**03.07.01 Oct 23, 2024**
+
+Resolved (and reverted) performance degradation introduced in v03.00.00.
+
+**03.07.00 Oct 16, 2024**
+
+Added new critical mass ratio tables from Ge et al. 2024.
+
+**03.06.00 Oct 14, 2024**
+
+Incorporated the Maltsev+ (2024) prescription for supernova remnant masses.
 
 **03.03.00 Sep 24, 2024**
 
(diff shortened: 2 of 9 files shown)


jeffriley:
Test runs comparing `v03.07.00` vs `v03.07.01`:

    COMPAS Version     = v03.07.00
    Run start          = Wed Oct 23 09:53:32 2024
    Run end            = Wed Oct 23 11:36:16 2024
    Run clock time     = 6161.54 CPU seconds
    Run wall time      = 0001:42:43 (hhhh:mm:ss)
    
    Mode               = BSE
    Number of systems  = 250000
    
    Total initial mass = 5069727.49 Msol (all systems)
    Total final mass   = 292437.52 Msol (DCOs only)
    
    Number of DCOs     = 11141
    
        Number of BHBH = 10371
        Number of BHNS = 677
        Number of NSNS = 93

        Number of BHBH merging in a Hubble time = 872
        Number of BHNS merging in a Hubble time = 285
        Number of NSNS merging in a Hubble time = 65
    
    Evolution status counts:
    
        TIMES_UP               : 62543
        STEPS_UP               : 891
        STELLAR_MERGER         : 73464
        DCO                    : 11141
        WD_WD                  : 71656
        UNBOUND                : 30305


    COMPAS Version     = v03.07.01
    Run start          = Wed Oct 23 08:32:41 2024
    Run end            = Wed Oct 23 09:46:44 2024
    Run clock time     = 4441.56 CPU seconds
    Run wall time      = 0001:14:03 (hhhh:mm:ss)
    
    Mode               = BSE
    Number of systems  = 250000
    
    Total initial mass = 5069727.49 Msol (all systems)
    Total final mass   = 292437.52 Msol (DCOs only)
    
    Number of DCOs     = 11141
    
        Number of BHBH = 10371
        Number of BHNS = 677
        Number of NSNS = 93
    
        Number of BHBH merging in a Hubble time = 872
        Number of BHNS merging in a Hubble time = 285
        Number of NSNS merging in a Hubble time = 65
    
    Evolution status counts:
    
        TIMES_UP               : 62543
        STEPS_UP               : 891
        STELLAR_MERGER         : 73464
        DCO                    : 11141
        WD_WD                  : 71656
        UNBOUND                : 30305

From the test results above we see that the run results are identical except for the run times - `v03.07.01` provides ~28% reduction in runtime over `v03.07.00`.

SimonStevenson (approved):
Thanks for identifying and fixing this issue Jeff. I've looked through the changes and they look good.

ilyamandel:
Thank you, @jeffriley !

PR #1268, opened 2024-10-30
Fix for error parsing deprecated option properties in logfile-definitions file

### LINKED ISSUE #1261: Some more typos
In the Program Option webpage,

--circulariseBinaryDuringMassTransfer

should be

--circularise-binary-during-mass-transfer

and

--common-envelope-lambda-nanjing-use_rejuvenated-mass

should be

--common-envelope-lambda-nanjing-use-rejuvenated-mass

In main.cpp, line 645, it should be "as it is))" instead of "as it))" similar to line 315

jeffriley:
Fix for error "ERROR:  in COMPAS_VARIABLE Options::OptionValue(const T_ANY_PROPERTY) const: Unexpected program option" when options are added to the system parameters file (--add-options-to-sysparms).  Root cause was a couple of deprecated options left in the list of options available for printing to log file.  Fix removes those options.  Also added code to the 
logfile-definitions parsing function so that deprecated options properties are replaced as required.

This defect prevents the use of grid files.  Until this fix is applied, a workaround is to specify `--add-options-to-sysparms never`

### DIFF
--- a/online-docs/pages/User guide/Program options/program-options-list-defaults.rst
+++ b/online-docs/pages/User guide/Program options/program-options-list-defaults.rst
@@ -120,7 +120,7 @@ of Chemically Homogeneous Evolution in COMPAS |br|
 Options: { NONE, OPTIMISTIC, PESSIMISTIC } |br|
 Default = PESSIMISTIC |br|
 
-**--circularise-binary-during-nass-transfer** |br|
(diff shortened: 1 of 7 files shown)


ilyamandel (approved):
Thank you, @jeffriley !

PR #1271, opened 2024-10-31
Tides bug fixes

veome22:
A few bug fixes related to the `KAPIL2024` tides prescriptions:

- Absolute values of timesteps in the ChooseTimestep() function, required to take sensible timesteps when the tidal dissipation is 'negative' (the binary spin exceeds orbital frequency).
- In Dynamical tides, made IW and GW dissipation from the convective envelope valid as long as either the mass OR the radial extent of the envelope exceeds the threshold. Required to avoid 0 tides in the presence of an envelope.

### DIFF
--- a/src/BaseBinaryStar.cpp
+++ b/src/BaseBinaryStar.cpp
@@ -2548,21 +2548,23 @@ double BaseBinaryStar::ChooseTimestep(const double p_Multiplier) {
         double DEccentricity2Dt_tidal  = CalculateDEccentricityTidalDt(ImKlm2, m_Star2);
                                                     
         double DOmega1Dt_tidal         = CalculateDOmegaTidalDt(ImKlm1, m_Star1);
-        double DOmega2Dt_tidal         = CalculateDOmegaTidalDt(ImKlm2,  m_Star2);
+        double DOmega2Dt_tidal         = CalculateDOmegaTidalDt(ImKlm2, m_Star2);
                                                                 
         // Ensure that the change in orbital and spin properties due to tides in a single timestep is constrained (to 1 percent by default)
         // Limit the spin evolution of each star based on the orbital frequency rather than its spin frequency, since tides should not cause major problems until synchronization. 
-        double DaDt_tidal              = std::abs(TIDES_MAXIMUM_ORBITAL_CHANGE_FRAC * m_SemiMajorAxis * std::min(std::abs(1.0 / DSemiMajorAxis1Dt_tidal), std::abs(1.0 / DSemiMajorAxis2Dt_tidal)) * YEAR_TO_MYR);
-        double DeDt_tidal              = std::abs(TIDES_MAXIMUM_ORBITAL_CHANGE_FRAC * m_Eccentricity * std::min(std::abs(1.0 / DEccentricity1Dt_tidal), std::abs(1.0 / DEccentricity2Dt_tidal)) * YEAR_TO_MYR);
-        double DOmegaDt_tidal          = std::abs(TIDES_MAXIMUM_ORBITAL_CHANGE_FRAC * m_Omega * std::min(std::abs(1.0 / DOmega1Dt_tidal), std::abs(1.0 / DOmega2Dt_tidal)) * YEAR_TO_MYR);
+        double Dt_SemiMajorAxis1_tidal = (DSemiMajorAxis1Dt_tidal == 0.0 ? dt : std::abs(TIDES_MAXIMUM_ORBITAL_CHANGE_FRAC * m_SemiMajorAxis / DSemiMajorAxis1Dt_tidal) * YEAR_TO_MYR);
+        double Dt_SemiMajorAxis2_tidal = (DSemiMajorAxis2Dt_tidal == 0.0 ? dt : std::abs(TIDES_MAXIMUM_ORBITAL_CHANGE_FRAC * m_SemiMajorAxis / DSemiMajorAxis2Dt_tidal) * YEAR_TO_MYR);
+        double Dt_SemiMajorAxis_tidal  = std::min(Dt_SemiMajorAxis1_tidal, Dt_SemiMajorAxis2_tidal);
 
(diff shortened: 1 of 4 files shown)


jeffriley (commented):
I have some questions:

> Absolute values of timesteps in the ChooseTimestep() function, required to take
> sensible timesteps when the tidal dissipation is 'negative' (the binary spin exceeds
> orbital frequency).

Is just taking the absolute value the right/best thing to do?  I know we want a positive number, but is that just a kludge?  Should the binary spin ever exceed the orbital frequency?  Are we just hiding the real problem?  Ok if not - just asking :-)

I didn't like this code when I first saw it, and I still don't like it:

    if (std::isnan(DaDt_tidal)) DaDt_tidal = std::numeric_limits<double>::infinity();
    if (std::isnan(DeDt_tidal)) DeDt_tidal = std::numeric_limits<double>::infinity();
    if (std::isnan(DOmegaDt_tidal)) DOmegaDt_tidal = std::numeric_limits<double>::infinity();

What might the consequences be of these values being set to inifinity.  Will they be used as infinity somewhere else in the code?  Will the ever be written to the logfiles as infinity.  If the answer to either of those questions is yes, we should look for a better solution.

veome22:
> Is just taking the absolute value the right/best thing to do? I know we want a positive number, but is that just a kludge? 

The tidal dissipation term can be negative, but the timesteps should always be positive. In our models, negative tides are possible, although not very plausible (e.g. for highly eccentric systems). I imagine if we want to 'fix' this behavior, we should do it when calculating tides, not in the timestep determination step.



> I didn't like this code when I first saw it, and I still don't like it:
> 
> ```
> if (std::isnan(DaDt_tidal)) DaDt_tidal = std::numeric_limits<double>::infinity();
> if (std::isnan(DeDt_tidal)) DeDt_tidal = std::numeric_limits<double>::infinity();
> if (std::isnan(DOmegaDt_tidal)) DOmegaDt_tidal = std::numeric_limits<double>::infinity();
> ```

I am on board with finding a better method. I should say that `DaDt_tidal`, etc. are not meant to be used outside this function (see `ProcessTides()`). The problem here was that `NaN`s were not handle reliably by `std::min()`. Do you know if there is a good way of comparing against NaNs?

jeffriley:
> Do you know if there is a good way of comparing against NaNs?

Well, I would be asking why we're getting NaNs in the first place?  A NaN is usually the result of (e.g.) taking the sqrt of a negative number, or log of a number <= 0.  Why would we be doing that, and can we either prevent it, or at least catch it and deal with it, at the source?

veome22:
> > Do you know if there is a good way of comparing against NaNs?
> 
> Well, I would be asking why we're getting NaNs in the first place? A NaN is usually the result of (e.g.) taking the sqrt of a negative number, or log of a number <= 0. Why would we be doing that, and can we either prevent it, or at least catch it and deal with it, at the source?

Oh that's the result of a division by zero within the ChooseTimestep() function. The tidal terms themselves should never be NaN. I had thought of setting the timescales to zero when the tidal terms are zero, but that will mess up the min() calculation. Any suggestions?

jeffriley:
> Oh that's the result of a division by zero within the ChooseTimestep() function.

The only way you get a NaN from a divide-by-zero is if the dividend is 0.0 - anything else gives -inf or +inf.  Where in ChooseTimestep() is the NaN coming from?  Dividing by zero isn't a reasonable thing to do, and if we're doing it somewhere we should be asking why and how to avoid it.

veome22:
Oh yes, sorry I should have clarified. The tidal terms (dx/dt) can be zero if no tides act on a given stellar or binary property (e.g. black holes have no tides, or de/dt = 0 for circular binaries). In ChooseTimestep(), we compute the maximum timestep dt to take given a fractional change dx'. For this, we need to divide dx' by dx/dt. So if dx/dt is zero (which it is allowed to be), we just need a smarter way of saying that the appropriate timestep contribution would be to do nothing, and just rely on other mechanisms (stellar processes, other tidal quantities, GWs, etc.). Perhaps I can add a bunch of if statements to catch these cases and set their timestep to dt from the processes above?

jeffriley:
> Perhaps I can add a bunch of if statements to catch these cases and set their timestep to dt from the processes above?

That would be preferable to dividing by 0.0

jeffriley:
>>Is just taking the absolute value the right/best thing to do? I know we want a positive number, but is that just a kludge?
>
> The tidal dissipation term can be negative, but the timesteps should always be positive. In our models, negative tides are
> possible, although not very plausible (e.g. for highly eccentric systems). I imagine if we want to 'fix' this behavior, we should
> do it when calculating tides, not in the timestep determination step.

Ok, the tidal dissipation term can be negative, but is the taking absolute value of the timestep you've calculated using a negative value the right thing to do?  Does the calculation of the timestep using a negative tidal dissipation really result in just the negative of the timestep we should be taking?  If the tidal dissipation term is negative, should we use a different calculation for the timestep?

veome22:
> > > Is just taking the absolute value the right/best thing to do? I know we want a positive number, but is that just a kludge?
> > 
> > 
> > The tidal dissipation term can be negative, but the timesteps should always be positive. In our models, negative tides are
> > possible, although not very plausible (e.g. for highly eccentric systems). I imagine if we want to 'fix' this behavior, we should
> > do it when calculating tides, not in the timestep determination step.
> 
> Ok, the tidal dissipation term can be negative, but is the taking absolute value of the timestep you've calculated using a negative value the right thing to do? Does the calculation of the timestep using a negative tidal dissipation really result in just the negative of the timestep we should be taking? If the tidal dissipation term is negative, should we use a different calculation for the timestep?

I hope the new code makes the logic clearer. But in general, the change in a quantity (such as semi-major axis) can be negative from tides. When this happens, we want to preserve the sign when applying the change to the semi-major axis, but ignore the sign when deciding the length of the time step. As such, I don't see a problem with allowing negative tidal terms, and taking the absolute value in the `ChooseTimestep()` function.

jeffriley (approved):
Looks good - thanks @veome22

Issue #1270, opened 2024-10-31
PERFECT Tides finding incorrect roots


veome22:
**Issue**
When using `--tides-prescription PERFECT`, the root finder seems to fail consistently for binaries with initial separations of ~30 AU, and masses (10+5).

**To Reproduce**
Run: 
```
./COMPAS --random-seed 0 -n 1  --initial-mass-1 10 --initial-mass-2 5 --semi-major-axis 20  --eccentricity 0.0 --chemically-homogeneous-evolution-mode NONE --tides-prescription PERFECT
```

**Expected behavior**
When solving for the cubic polynomial roots by brute force, I obtain a sensibly small value of $\Omega$. However, the value from the root solver implemented in `OmegaAfterSynchronisation()` of BaseBinarStar.h returns a very large value, which leads to an instant stellar merger. 
![image](https://github.com/user-attachments/assets/88d9a897-f99c-4b88-b409-2dac52d6559d)

Over a population of binaries, this produces a vast under-density of compact objects with ZAMS semi-major axes around 10-100 AU.

**Versioning:**
 - COMPAS v03.07.02

ilyamandel:
This can happen when the initial guess is close to the truth and one of the root brackets from bracket-and-solve already provides a sufficiently good solution.  A check for this has now been added in #1276 , so this problem should be gone.

PR #1276, opened 2024-11-07
Fix for issue #1270

ilyamandel:
Fix for issue #1270 : root finder functions now check if either of the bracket edges provides a sufficiently good solution, which sometimes happens when the initial guess is very close to the truth


jeffriley (approved):
All good - thanks @ilyamandel

PR #1279, opened 2024-11-17
Stellar rotation enhancement

ilyamandel:
- Switch to using angular momentum rather than omega as the basic tracker of stellar rotation
- Stars that evolve without mass loss do not change angular momentum, but may change omega as the moment of inertia changes
- Stars that lose mass through winds or mass transfer keep the same rotation rate (omega) and lose angular momentum -- corresponds to completely inefficient angular momentum transport assumption
- Mass gain through accretion brings in a specific angular momentum equal to that of a Keplerian orbit at the accretor's radius
- Associated code clean-up

### DIFF
--- a/src/BaseBinaryStar.cpp
+++ b/src/BaseBinaryStar.cpp
@@ -1666,10 +1666,10 @@ void BaseBinaryStar::ResolveCommonEnvelopeEvent() {
     }
 
     // if stars are evolving as CHE stars, update their rotational frequency under the assumption of tidal locking if tides are not enabled
-    if(OPTIONS->TidesPrescription() == TIDES_PRESCRIPTION::NONE) {
+    if (OPTIONS->TidesPrescription() == TIDES_PRESCRIPTION::NONE) {
         double omega = OrbitalAngularVelocity();                                                                        // orbital angular velocity
-        if (m_Star1->StellarType()==STELLAR_TYPE::CHEMICALLY_HOMOGENEOUS) m_Star1->SetOmega(omega);
-        if (m_Star2->StellarType()==STELLAR_TYPE::CHEMICALLY_HOMOGENEOUS) m_Star2->SetOmega(omega);
+        if (m_Star1->StellarType() == STELLAR_TYPE::CHEMICALLY_HOMOGENEOUS) m_Star1->SetOmega(omega);
+        if (m_Star2->StellarType() == STELLAR_TYPE::CHEMICALLY_HOMOGENEOUS) m_Star2->SetOmega(omega);
     }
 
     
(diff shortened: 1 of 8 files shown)


ilyamandel:
@veome22 -- this one is for you, should avoid the artificial growth in the angular momentum of a star as its moment of inertia increases. :) I look forward to your tests!
Also sets us up for future improvements, such as magnetic braking.

jeffriley (dismissed):
Thanks @ilyamandel . Ok from my perspective.  Awaiting @veome22's tests and review.  I'll approve if you want to merge before that's done.

ilyamandel:
Thank you, @jeffriley !

Please don't pull yet.   I was going to make the spin-down on mass loss more accurate in a future update, but I just thought of a way to do so now, will try to implement that before the pull.

ilyamandel:
A few more updates:

- Stars that lose the entire envelope keep their pre-envelope-loss rotational frequency
- Chemically homogeneous evolution is now checked for at binary initialisation and stars are assigned the orbital frequency if that exceeds the CHE threshold
- However, subsequently, only CHE stars are artificially kept in co-rotation with binary (ignoring angular momentum conservation) only if TIDES_PRESCRIPTION::NONE is used
- Clean-up of BaseBinaryStar::ResolveMassChanges(): if m_Mass variable has already been updated (because ResolveEnvelopeLoss() has been called), no need to update attributes again

@jeffriley , sorry to ask, but this will need to be reviewed again. :-)

ilyamandel:
For completeness:

Without mass change or tides, a star will keep the same angular momentum, not the same omega.  This is a fix to the previous behaviour, where, incorrectly, omega stayed constant even as the moment of inertia evolved.  So now the moment of inertia is the key variable: we don't even store omega any more, only compute it on the fly when needed.

When the star loses mass in winds or through mass transfer, the assumption is that the mass is lost from the outer shell -- i.e., the change in angular momentum is dL = (2/3) dm r^2 omega, where dm is the mass change (negative), r and omega are the pre-mass-loss radius and angular frequency.  The exception is when we lose the entire envelope as a consequence of thermal or dynamical timescale mass transfer, I now set the remaining core to rotate at the same angular frequency as the preceding star (logic: not enough time for angular momentum transport during this rapid phase).  [Note: this means things aren't quite right for the special case of rapid (thermal timescale) case A MT: as radius and omega gradually reduce during this phase, this model probably over-estimates the amount of angular momentum loss.  To keep things safe, if |dL|>|L|, the final angular momentum is just set to zero.

On mass gain (through accretion), mass is assumed to be gained from a Keplerian circular orbit at the star's equator, so dL = dm \sqrt{G m r}, where m and r are the pre-gain mass and radius.

jeffriley:
@ilyamandel I made some changes to your code - can you check that you're ok with the changes?

Also, in lines `BaseBinaryStar` lines 2421 and 2447 of my changed code (in `ResolveMassChanges()`), given the comment

`add angular momentum of accreting material...`

should the `=` be `+=` ? (both lines)

EDIT: If the `=` stands (i.e. not `+=`) then I think the construct should change - why calculate something only to recalculate it (should be if/else?)

jeffriley (approved):
Thanks @ilyamandel.  Note I made some changes.  I'm ok with the new code pending the question I asked re `=` vs `+=`.

Will approve now - feel free to wait for @veome22 or to merge at your convenience

PR #1281, opened 2024-11-18
ignore latex if detailed plotter is run on github actions

reinhold-willcox:
Added flag to detailed plotter to turn off latex when run on github actions


ilyamandel (approved):
Thank you, @reinhold-willcox and @avivajpeyi !

PR #1282, opened 2024-11-19
Multiple rotation-related fixes to 03.08.01 (units, initialisation, min->max typo, retain omega on envelope loss)

ilyamandel:
Multiple rotation-related fixes to 03.08.01 (units, initialisation, min->max typo, retain omega on envelope loss)


jeffriley (approved):
All good

PR #1283, opened 2024-11-20
Spin retention after envelope loss

veome22:
Minor bug fix for the behavior of stellar spins to be retained after envelope loss. This is complementary to @ilyamandel's recent updates to stellar rotation models.

The previous behavior was to (inadvertently) retain the full angular momentum after envelope loss, which would spin up the remnant star very fast instead of losing the envelope's angular momentum. 

Solved this by storing the rotational frequency at the beginning of the CE phase, and then assigning the spin back to the stars after envelope loss.


ilyamandel:
Hi @veome22 ,

I thought my edits to 

void Star::ResolveEnvelopeLossAndSwitch() {
    double omega = Omega();
    (void)SwitchTo(m_Star->ResolveEnvelopeLoss(true));
    SetOmega(omega);
}

fixed precisely this issue; are you saying they do not?

jeffriley (approved):
Looks fine to me - thanks @veome22.  Just need dev pulled in, or fix the conflict by hand.

veome22:
> Hi @veome22 ,
> 
> I thought my edits to
> 
> void Star::ResolveEnvelopeLossAndSwitch() { double omega = Omega(); (void)SwitchTo(m_Star->ResolveEnvelopeLoss(true)); SetOmega(omega); }
> 
> fixed precisely this issue; are you saying they do not?

Hi @ilyamandel ! Yes, I am saying that those changes did not fix the problem. The mass of the stars must change at some point before ResolveEnvelopeLossAndSwitch() (maybe in  ResolveCommonEnvelopeAccretion()). So the Omega() function returns the angular momentum conserved omega after mass loss, rather than the initial spin before CE.

ilyamandel:
@veome22 -- yes, you are right, the mass is changed in ResolveCommonEnvelopeAccretion(), sorry I missed that!

Issue #1231, opened 2024-10-02
Thermal/nuclear timescale mass transfer in main-sequence donors


brcekadam:
**Describe the bug**
I have been looking at how case A mass transfer affects the core mass of the donor at TAMS. I noticed that starting out with the same initial masses, but increasing the initial semi-major axis can result in very different core masses of the donors as well as different main sequence lifetimes. The difference is particularly visible with the winds turned off. The outcome of the system oscillates between two situations:

**1.** The MS donor transfers mass on thermal timescale (in one time step) and then continues to transfer mass on nuclear timescale (over multiple time steps) until it becomes a HG star. Then the second thermal timescale mass transfer episode follows, and the donor becomes a HeMS star. 
**2.** The MS donor transfers mass on thermal timescale, star evolves into HG star right after this mass transfer episode, and nuclear timescale mass transfer is skipped. Second thermal timescale mass transfer follows, and the stars becomes a HeMS after. This results in shorter MS lifetime and higher core mass at TAMS compared to the previous situation.

This could suggest that there is an issue with the estimate of whether mass transfer can occur stably over a nuclear timescale.

**To Reproduce**
`./compas --initial-mass-1 40 --initial-mass-2 30  --semi-major-axis r[0.2,100,0.002] --retain-core-mass-during-caseA-mass-transfer FALSE  --mass-loss-prescription ZERO`

![no_winds_retain_core_mass_FALSE_page-0001](https://github.com/user-attachments/assets/9f07e314-b215-4851-860c-a1ffcc362c13)

ilyamandel:
Hi @brcekadam ,
Finally looking at this.  I can reproduce the issue, but I find (even more surprisingly!) that the next MT phase, when there's only a single MT phase on the MS, is from a ChEB donor, not an HG donor.  I think there's an issue with timescales; investigating...

ilyamandel:
Yikes.  This was actually a more serious issue than I thought. :( 
The problem was in MainSequence::UpdateAgeAfterMassLoss().  We changed the star's age on mass loss, as we should have, but didn't recalculate its timescales.  So the star could get older than its MS lifetime, which meant on the very next time step we'd falsely conclude that it was too old to be an MS star, and perhaps even an HG star, skipping way into the future.  Will fix in a moment.
@jeffriley , @SimonStevenson -- linking you here because this means that funny things were happening on case A MT...

ilyamandel:
[...]

PR #1284, opened 2024-11-25
Recalculate timescales when updating stellar age after mass loss

ilyamandel:
Addresses #1231 and fixes a significant bug in the process.


jeffriley (approved):
Thanks @ilyamandel

Issue #1285, opened 2024-11-25
Misbehaving nuclear timescale mass transfer


ilyamandel:
Update to nuclear timescale mass transfer, following the mass transfer discussion at the retreat and some follow-up tests.

1. We currently apply nuclear timescale MT only to MS donors (other donors are assumed to always transfer mass on a thermal timescale).  We assume a MS donor can transfer mass on a nuclear timescale if its equilibrium zeta = dlnR/dlnM  is larger than the RL zeta.  In practice, to evaluate dlnR, we compute the radius of MS stars with two nearby masses at the same fractional MS age tau.  (The logic is that, in our models, MS stars retain the same tau on mass loss -- see Hurley+ (2000), section 7.1)

2. The actual rate of nuclear timescale mass loss is set to be the current donor mass divided by its radial expansion timescale, where the latter is given as R / \dot{R}, with \dot{R} estimated by slightly aging (a copy of) the star and evaluating the change in radius.  

The problem with this approach is that the nuclear timescale mass loss rate from the donor may not keep up with the evolution of the Roche lobe.  In other words, the donor may expand to be filling the Roche lobe by a significant fraction.  Here's a mild example, but it can be worse.  (Obtained with 
`./COMPAS --metallicity 0.008 --initial-mass-1 1 --initial-mass-2 0.7 -n 1 --semi-major-axis 0.01251 --detailed-output` ) 

The best solution I see is to allow the nuclear timescale mass loss rate to be larger than the estimate in item 2 -- in fact, to be high enough that, at the end of the nuclear mass transfer time step, the donor fits inside its Roche lobe (that's what we do for thermal timescale "fast" MT from MS donors).  In fact, the more I think about it, the more arbitrary I find the current approach of estimating the nuclear mass transfer timescale using the radial expansion timescale as a proxy; not sure why I assumed that in the first place?  Anyway, I'll revise that now, but if folks have other thoughts/suggestions, please let me know...

![image](https://github.com/user-attachments/assets/fa5b8760-b6ff-4326-b307-e4ee163e5856)

ilyamandel:
OK, problem finally solved!  It was actually a combination of problems.  Firstly, the nuclear mass transfer rate was not chosen sensibly; it is now set by the requirement that the star ends the time step just filling its Roche lobe.  Secondly, there was an issue with the root finder for fitting into the RL that led to artificial failures to find a root -- this could have led to artificial CEs / mergers on mass transfer in previous simulations.

[...]

PR #1287, opened 2024-11-25
Enhancements to nuclear timescale MT, several fixes, addresses issue #1285

ilyamandel:
See discussion and plots in #1285 .

- The nuclear timescale mass transfer rate is now set by the requirement that the star ends the time step just filling its Roche lobe (addresses issue #1285)
- Fix an issue with the root finder for fitting into the RL that led to artificial failures to find a root
- Fix issue (likely introduced in 03.08.00) with the accretor not gaining mass appropriately


jeffriley (approved):
All good - thanks @ilyamandel

ilyamandel:
Thanks, @jeffriley .  Given that this has multiple fixes for pretty severe bugs (e.g., the secondary was not accreting when it should have been), I'll go ahead and merge. Folks tagged in this are aware and can comment at their convenience if they wish (also tagging @veome22 , since he will probably want to use v. 3.09.00 for future tests).

PR #1290, opened 2024-11-27
SN kick fixes

### LINKED ISSUE #1241: Companion ejection velocity seems to be a factor of 2 too large
**Describe the bug**
The ejection velocity of secondary stars seems to be a factor 2 larger than one would expect when comparing to their pre-superova orbital velocities. 

**Label the issue**

**To Reproduce**
As an example, run a grid of binaries with m1=20, m2=15, porb=100d, Z=0.02, each time with a different random kick. Then compare these vaues from the BSE_SUPERNOVAE table for the first supernova: `ComponentSpeed(CP)` to `Orb_Velocity<SN * (Mass_Total@CO(SN) / (Mass_Total@CO(SN) + Mass(CP)))`. The ratio of these two values is ~2 on average.

**Expected behavior**
I'd expect that
image

**Screenshots**
Here's a histogram of the ratio that I'd expect to be centred around 1
image

 - OS: Ubuntu 22.04
 - COMPAS: v03.01.10

**Code snippet for processing data**
```
    df_dict = {}

    with h5.File("COMPAS_Output.h5") as f:
        print(f.keys())
        print(f["BSE_Supernovae"].keys())

        for k in f["BSE_Supernovae"].keys():
            df_dict[k] = f["BSE_Supernovae"][k][...]
    df = pd.DataFrame(df_dict).set_index("SEED", drop=False)
    df = df.drop_duplicates(subset="SEED", keep="first")

    print(f'Disruption rate: {100 - ((df["Unbound"] == 0).sum() / len(df) * 100)}%')

    v_orb_2 = df["Orb_Velocity<SN"] * (df["Mass_Total@CO(SN)"] / (df["Mass(CP)"] + df["Mass_Total@CO(SN)"]))
    v_orb_2 = v_orb_2.iloc[0]
    print(f"v_orb_2 = {v_orb_2}")

    natal_kicks = df.rename(columns={"ComponentSpeed(CP)": "vsys_2_total",
                                     "Applied_Kick_Magnitude(SN)": "natal_kick",
                                     "Unbound": "disrupted"})

    plt.hist(natal_kicks["vsys_2_total"][natal_kicks["disrupted"] == 1] / v_orb_2, bins=80, range=(0, 4))
    plt.xlabel(r"$v_{\rm 2, postSN} / v_{\rm 2, preSN}$ (km/s)")
    plt.show()
```

reinhold-willcox:
Fixes to the vector3d function that were resulting in anomalous component velocities. 

The fixes pertain to
- Incorrect indexing. For `vec = {x, y, z}`, `vec[i]` was retrieving the value at `{x, y, z}[i - 1]`, cyclically.
- Incorrect rotation. The rotation algebra was wrong so the magnitude of the vector under rotation was doubling.

The bugs were introduced in June, so any results between then and now that depend on any vector quantities (or their magnitudes) should be double checked / rerun.


jeffriley (approved):
Thanks @reinhold-willcox .  Those changes were me - my fault.  Again, this highlights the need for better regression testing...

jeffriley:
Actually, this code:

    // Operator overloads
    double   operator [] (const size_t p_i) const { return (*const_cast<Vector3d*>(this))[p_i]; }
    double&  operator [] (const size_t p_i) {

        THROW_ERROR_IF(p_i < 0 || p_i > 2, ERROR::INDEX_OUT_OF_RANGE);                                  // this is a code defect

             if (p_i == 1) return m_x;
        else if (p_i == 2) return m_y;
        else               return m_z;
    }
    
was this in the original code:

    double&  operator [] (const size_t p_i) {

        // check that supplied index is in range.
        // if index is not in range, issue a warning and use index modulo 3
        size_t index = p_i;
        if (index > 2) {
            SHOW_WARN(ERROR::INDEX_OUT_OF_RANGE, "Using index modulo 3");                               // index out of range
            index %= 3;                                                                                 // use index modulo 3
        }
        
             if (index == 1) return m_x;
        else if (index == 2) return m_y;
        else                 return m_z;
    }
    
so the checks:

             if (index == 1) return m_x;
        else if (index == 2) return m_y;

were there in the original code.  I think changing them to 0 and 1 is correct - but the problem has been around longer than v03.00.00

jeffriley:
Ok, looks like the problem with the index checking was introduced in v02.43.04 (by me...)

Issue #1265, opened 2024-10-29
Switch stellar type to massless remnant on merger in common envelope


ilyamandel:
This will ensure that mergers are recorded in switch logs, and also that the post-merger stellar type is correctly recorded in the BSE_Common_Envelopes output

ilyamandel:
Resolved in #1291 , closing.

Issue #1258, opened 2024-10-25
floating-point-underflow in MainSequence::CalculateRadiusOnPhase()


jeffriley:
Running COMPAS with:

    ./compas -n1 --random-seed 1724542090 --kick-magnitude-distribution ZERO --critical-mass-ratio-prescription NONE --common-envelope-alpha 1.6265542767303343 --wolf-rayet-multiplier 0.8084912681292984 --initial-mass-function KROUPA --mass-ratio-distribution DUQUENNOYMAYOR1991

results in a `floating-point underflow` in in `MainSequence::CalculateRadiusOnPhase()` in the following statement:

    double tau_40 = tau_10 * tau_10 * tau_10 * tau_10;

because the value of  `tau` is 1.16432e-09, and `tau_10` is 4.57833e-90.  `tau` is calculated in this statement:

    double tau = p_Time / timescales(tMS);                // Hurley et al. 2000, eq 11

where `timescales(tMS)` is 490656. 

`p_Mass` = 0.26822, `p_Time` = 0.000571279, `p_RZAMS` = 0.2779, and  `timescales(tBGB)` = 516480

To Reproduce
Run COMPAS with:

    ./compas -n1 --random-seed 1724542090 --kick-magnitude-distribution ZERO --critical-mass-ratio-prescription NONE --common-envelope-alpha 1.6265542767303343 --wolf-rayet-multiplier 0.8084912681292984 --initial-mass-function KROUPA --mass-ratio-distribution DUQUENNOYMAYOR1991

Expected behavior
Production code does not cause `floating-point underflow`

**Versioning

OS: Ubuntu 22.04
COMPAS v03.07.01

Also, typo in this comment could be fixed at the same time the issue is addressed:

    // pow() is slow - use multipliaction where it makes sense

ilyamandel:
Fixed in #1291 , closing.

Issue #1255, opened 2024-10-25
-inf (divide-by-zero) in  BaseStar::CalculateMassLossRateWolfRayetTemperatureCorrectionSander2023()


jeffriley:
Running COMPAS with:

    ./compas -n1 --minimum-secondary-mass 5

results in `-inf` in  `BaseStar::CalculateMassLossRateWolfRayetTemperatureCorrectionSander2023()` in the following statment:

    double logMdotUncorrected  = log10(p_Mdot);     // uncorrected mass-loss rate

because the parameter `p_Mdot` = 0.0

    double BaseStar::CalculateMassLossRateWolfRayetTemperatureCorrectionSander2023(const double p_Mdot) const {
    
        const double teffRef = 141.0E3;                                 // reference effective temperature in Kelvin
        const double teffMin = 100.0E3;                                 // minimum effective temperature in Kelvin to apply correction
    
        double teff                = m_Temperature * TSOL;              // get effective temperature in Kelvin
        double logMdotUncorrected  = log10(p_Mdot);                     // uncorrected mass-loss rate
        double logMdotCorrected    = 0.0;
    
        // Only apply to sufficiently hot stars
        if (utils::Compare(teff, teffMin) > 0) {
            logMdotCorrected = logMdotUncorrected - 6.0 * log10(teff / teffRef);
        }
        else {
            logMdotCorrected = logMdotUncorrected;
        }
    
        return PPOW(10.0, logMdotCorrected);
    }

The naive solution is to add the following sanity check as the first line of the function:

    if (p_Mdot <= 0.0) return 0.0;                                  // don't use utils::Compare() here

so we just return a mass loss rate of 0.0 if the rate passed in `p_Mdot` is 0.0

`p_Mdot` can be passed as 0.0 from `HeMS::CalculateMassLossRateMerritt2024()` when `CalculateMassLossRateWolfRayetSanderVink2020()` is called with `p_Mu` < 1.0 (as it is in `HeMS::CalculateMassLossRateMerritt2024()`) (this is the case that caused the failure that prompted opening of this issue).
There may be other ways for the problem to manifest 

@jmerritt1, @SimonStevenson, can you confirm that the suggestion above is the best solution?

**To Reproduce**
Run COMPAS with:

    ./compas -n1 --minimum-secondary-mass 5

**Expected behavior**
Production code does not cause `-inf`

**Versioning
 - OS: Ubuntu 22.04
 - COMPAS v03.07.01

[...]

Issue #1286, opened 2024-11-25
Incorrect brief evolutionary outputs printed to screen


ilyamandel:
The brief outcome of the evolution printed to the screen is sometimes incorrect.  For example, the following 

./COMPAS -n 1 --initial-mass-1 1 --initial-mass-2 1 --semi-major-axis 0.03 --detailed-output --logfile-type CSV --evolve-main-sequence-mergers

leads to the output

0: Stars merged: (Main_Sequence_>_0.7 -> Helium_White_Dwarf) + (Main_Sequence_>_0.7 -> Helium_White_Dwarf)

In fact, the stars merge earlier, as giants, not as HeWDs.

jeffriley:
@ilyamandel the output is correct - it's only reporting what happened during the evolution of the binary.

The problem, I think, is that we are not flagging the merger early enough in `BaseBinaryStar::ResolveCommonEnvelopeEvent()`

We set the stellar merger flag here:

    if (utils::Compare(m_SemiMajorAxis, 0.0) <= 0 || utils::Compare(m_Star1->Radius() + m_Star2->Radius(), m_SemiMajorAxis * AU_TO_RSOL) > 0) {
        m_Flags.stellarMerger = true;
    }

but that is after we've allowed both stars to call `ResolveEnvelopeLossAndSwitch()`.  I think we should set the stellar merger flag earlier, here:

    if (isDonorMS || (!envelopeFlag1 && !envelopeFlag2)) {                                                              // stellar merger
        m_MassTransferTrackerHistory = MT_TRACKING::MERGER; 
        m_Flags.stellarMerger        = true;
    }
    else if ((m_Star1->DetermineEnvelopeType()==ENVELOPE::RADIATIVE && !m_Star1->IsOneOf(ALL_MAIN_SEQUENCE)) ||
             (m_Star2->DetermineEnvelopeType()==ENVELOPE::RADIATIVE && !m_Star2->IsOneOf(ALL_MAIN_SEQUENCE)) ) {        // check if we have a non-MS radiative-envelope star
        if (!OPTIONS->AllowRadiativeEnvelopeStarToSurviveCommonEnvelope() && OPTIONS->CommonEnvelopeFormalism()!=CE_FORMALISM::TWO_STAGE) { // stellar merger
            m_CEDetails.optimisticCE     = true;
            m_MassTransferTrackerHistory = MT_TRACKING::MERGER;
            m_Flags.stellarMerger        = true;
        }
    }

[...]

PR #1291, opened 2024-11-27
A range of fixes: using core radii for CE survival, fixes to #1255, #1258, #1265

ilyamandel:


jeffriley:
@ilyamandel I just pushed some changes that add to your fix.  There is also a question for you at the end of `BaseBinaryStar::ResolveCommonEnvelopeEvent()` - need to decide what to do about printing to the CE file if there was a stellar merger.

ilyamandel:
Thanks, @jeffriley -- I hadn't meant to ask you to do my work for me, but I appreciate it. ;-)
And yes, I think we should store that there was a merger, so all good!
Now, to ensure that we indicate a merger in the Switch Log as well, I had a proposal to switch both stars to massless remnants on merger.  I've now implemented that, resolving https://github.com/TeamCOMPAS/COMPAS/issues/1265... But that means we are now outputting Massless remnants in the screen output message (the CE file is fine, it's written earlier).  Thoughts?

jeffriley:
No worries @ilyamandel  - as soon as I read that we were still reporting HeWD I knew why, and the fix was easy (and really should already have been in - continuing evolution after a stellar merger was an oversight).

The switchlog issue.  When I coded the output all I wanted to indicate was:

(a) the starting stellar type for each star
(b) the ending stellar type for each star
(c) the reason evolution stopped

In the specific case of the example in issue https://github.com/TeamCOMPAS/COMPAS/issues/1286, that worked until you put the switches to MR in :-)  

By switching the stars to MR after the stellar merger, you've actually continued the evolution and changed the "ending" stellar type.  But both stars becoming MR is not really physical, is it?  You're just doing that so you get an entry in the switchlog?  If we evolved merger products, something different would happen and we'd still get a switch, right?  Just not to both MR.  What I'm getting at is that I think you're doing something unphysical just so you get an entry in the switchlog, and if that's true then trying to manipulate the output to account for that is a bit of a kludge - maybe we should figure out the right way (or a better way) of getting an entry in the switchlog upon stellar merger?  Maybe we should always evolve past a stellar merger and switch the stars to what they should become upon merger - but that would mean we don't output what they were just prior to the merger...  Do we want to change the output so in the case of mergers we print the stellar types of the stars just prior to the merger as well as the final stellar types?  Whatever we do, users have to be confident that what we output is consistent (i.e. we shouldn't change the meaning of what's output just for stellar mergers - whatever we do should be additional info).  Hopefully that makes sense.

online-docs/pages/User guide/Program options/program-options-list-defaults.rst
```diff
@@ -1122,10 +1122,11 @@ Enable mass loss due to pulsational-pair-instability (PPI). |br|
 Default = TRUE
 
 **--pulsational-pair-instability-prescription** |br|
-Pulsational pair instability prescription. |br|
+Pulsational pair instability prescription (only relevant when using ``--pulsational-pair-instability``. |br|
```
dumbscholar: closing parenthesis missing

ilyamandel: Fixed.

ilyamandel:
@jeffriley -- I take your point, and will think some more about what is a sensible "short-hand" output to the screen.  For now, I think what's in this PR is the best solution until we actually implement the evolution of merger products (other than MS merger products).  Ready for your review!

jeffriley:
@ilyamandel what about we add a new column to the BSE switchlog file ("isMerger"), and create a function that can be called from anywhere that logs a merger record to the BSE switchlog file?  It would list the from and to stellar types as the same (because no switch), and they would be whatever the stellar types were at the time of the call to log the record.  The "isMerger" field would be "false" for ordinary switch logs, and "true" for the merger logs.  The function to log the merger would be a no-op if the switchlog is not enabled.  Thoughts?  I could do this in the next day or so.

ilyamandel:
I like that, @jeffriley !  We can take out lines 3088-3090 of the new BaseBinaryStar.cpp file -- either now or in your next PR, your call.  

P.S. I am going offline now, good night. ;-)

jeffriley (approved):
Thanks @ilyamandel

Issue #1245, opened 2024-10-18
NaN Values for 'Mass_Core@CO(SN)'  with Latest COMPAS (version 03.07.00)


lee0517-hub:
**Describe the bug**
When using the latest version of COMPAS (version 03.07.00), I encountered an issue where the data for 'Mass_Core@CO(SN)' is NaN for some binaries.

**Label the issue**
urgency_high

**To Reproduce**
./COMPAS --critical-mass-ratio-prescription NONE  --kick-magnitude-distribution ZERO  --random-seed  1724542090 --common-envelope-alpha 1.6265542767303343  --wolf-rayet-multiplier 0.8084912681292984  --number-of-systems 1  --initial-mass-function KROUPA  --mass-ratio-distribution DUQUENNOYMAYOR1991  --semi-major-axis-distribution SANA2012  --minimum-secondary-mass 5  --metallicity-distribution LOGUNIFORM  --tides-prescription PERFECT  --remnant-mass-prescription FRYER2022  --common-envelope-allow-radiative-envelope-survive False  --common-envelope-allow-immediate-RLOF-post-CE-survive False

**Expected behavior**

**Screenshots**
![2024-10-18](https://github.com/user-attachments/assets/96689345-4aa1-4931-a85e-4ec119127543)

**Versioning (please complete the following information):**
 - OS: [e.g. Ubuntu 22.04]
 - COMPAS [03.07.00]

**Additional context**

jeffriley:
The immediate cause of this problem is that in `HeGB::CalculateCoreMassOnPhase_Static()`, `p_time` is greater than `tx`, but it is also greater than `tinf2`, so we end up taking the sqrt of a negative number, hence the `NaN`.

The calculation is given by Hurley et al., 2000 eq 39, modified per the discussion following eq 84.  I believe our implementation of the equation is correct, so we need to determine why `p_Time` is > `tinf2`.  Hurley doesn't mention caveats, so I assume he expected `tinf2` would always be greater than `p_Time` (`t` in eq 39).  Preventing the `NaN` naively is easy, but I'm not sure that would be physical... Needs more investigation.

lee0517-hub:
It seems that this bug has only recently appeared. I am not certain from which version it started, but I can confirm that the issue does not exist in version v03.00.05. I hope this information helps and saves you some time in the debugging process.

jeffriley:
Problem was introduced in `v03.05.02`.

@ilyamandel 

[...]

PR #1298, opened 2024-11-30
Bug fixes in tides, helium giant updates, white dwarf radii and McBGB calculations

ilyamandel:
- corrected treatment of rotation to retain pre-mass-loss spin frequency, not angular momentum, on complete envelope removal during stable mass transfer
- fixed issue with updating helium giants that manifested as supernovae with nan core mass (see #1245 )


jeffriley (approved):
Thanks @ilyamandel

Issue #1257, opened 2024-10-25
inf (divide-by-zero) in NS::CalculateAndSetPulsarParameters, NS::CalculateSpinDownRate(), and NS::SpinDownIsolatedPulsar()


jeffriley:
Running COMPAS with:

    ./compas -n1 --random-seed 0

results in `inf` in `NS::CalculateAndSetPulsarParameters` in the following statement:

    m_PulsarDetails.spinFrequency = _2_PI / (m_PulsarDetails.spinPeriod * SECONDS_IN_MS);

because `CalculateBirthSpinPeriod()` in this statement:

    m_PulsarDetails.spinPeriod = CalculateBirthSpinPeriod();                // spin period in ms

returns 0.0 because the default `PULSAR_BIRTH_SPIN_PERIOD_DISTRIBUTION` is `ZERO`

    void NS::CalculateAndSetPulsarParameters() {
    
        m_PulsarDetails.magneticField     = PPOW(10.0, CalculateBirthMagneticField()) * GAUSS_TO_TESLA;         // magnetic field in Gauss -> convert to Tesla
        m_PulsarDetails.spinPeriod        = CalculateBirthSpinPeriod();                                         // spin period in ms
    
        m_PulsarDetails.spinFrequency     = _2_PI / (m_PulsarDetails.spinPeriod * SECONDS_IN_MS);
        m_PulsarDetails.birthPeriod       = m_PulsarDetails.spinPeriod * SECONDS_IN_MS;                         // convert from ms to s 
        
        m_MomentOfInertia_CGS             = CalculateMomentOfInertiaCGS();                                      // in CGS g cm^2
    	
        // Note we convert neutronStarMomentOfInertia from CGS to SI here
        m_PulsarDetails.spinDownRate      = CalculateSpinDownRate(m_PulsarDetails.spinFrequency, m_MomentOfInertia_CGS, 
        m_PulsarDetails.magneticField, m_Radius * RSOL_TO_KM);  
        m_PulsarDetails.birthSpinDownRate = m_PulsarDetails.spinDownRate; 
        m_AngularMomentum_CGS             = m_MomentOfInertia_CGS * m_PulsarDetails.spinFrequency;              // in CGS g cm^2 s^-1
    }

The naive solution is to change the statement:

    m_PulsarDetails.spinFrequency = _2_PI / (m_PulsarDetails.spinPeriod * SECONDS_IN_MS);
    
to:

[...]

PR #1315, opened 2024-12-19
Updated MSP treatments. 

yuzhesong:
Replaces #1082 

Includes pulsar accretion treatments calculated with BOOST ODE integrator and other fixes, including solution to #1002


ilyamandel (changes requested):
Options.cpp:

        if (!DEFAULTED("neutron-star-accretion-in-ce")) {                                                                         // neutron star accretion in common envelope
            std::tie(found, m_NeutronStarAccretionInCE.type) = utils::GetMapKey(m_NeutronStarAccretionInCE.typeString, NS_ACCRETION_IN_CELabel, m_NeutronStarAccretionInCE.type);
            COMPLAIN_IF(!found, "Unknown Neutron Star Equation of State");
        }
        
Why is the complaint about an unknown NS EoS?  This is about accretion, not EoS...   

Same file, 

        case _("neutron-star-accretion-in-ce")                      : POPULATE_RET(NS_ACCRETION_IN_CELabel);                        break;
        case _("neutron-star-equation-of-state")                    : POPULATE_RET(NS_EOSLabel);                                    break;
        
Format / capitalisation of labels different from other labels?

changelog.h:
- comments should be edited for English grammar
- "deltaAngularMomentumByPulsarAccretion()" -- that's not the name of the function, check capitalisation
- "- Update to NS::UpdateMagneticFieldAndSpin(), specifically adding" <- adding what?

A new option has been added.  That means:
-  the program options documentation needs to be updated accordingly
- a new default YAML should be created
- whats-new.rst should probably be updated

NS.cpp:

- comments for DeltaAngularMomentumByPulsarAccretion need to be adjusted for English grammar (e.g., "Calculate the magnetic field... and also returns"), correct capitalisation of function name, correct number and list of return variables in the tuple (the comment sometimes lists three, sometimes four -- the function returns three), correct ordering and list of input parameters, etc.

Do we need to pass parameters like p_Mass?  This does not appear to be a static function, and a neutron star known its own mass -- would the argument passed in ever be different than m_Mass?  Ditto for some of the other variables.

Why p_MassGainPerTimeStep?  Isn't the important thing just the mass gain?

double newPulsarMagneticField = (initialMagField_G - magFieldLowerLimit_G) * exp(-p_MassGainPerTimeStep / p_Kappa) + magFieldLowerLimit_G;

If initialMagField_G was, say, zero to begin with, it will end up having a non-zero value but still one less than magFieldLowerLimit_G.  I don't understand the logic...

More to come later.

ilyamandel:
Aside: @yuzhesong , if this replaces #1082 , can we delete that one?

yuzhesong:
@ilyamandel 

Thanks for the review! Making changes to the codes...

To some of your points: 

> Do we need to pass parameters like p_Mass? This does not appear to be a static function, and a neutron star known its own mass -- would the argument passed in ever be different than m_Mass? Ditto for some of the other variables.

NS::DeltaAngularMomentumByPulsarAccretion() is used for the integration for NS in RLOF as a dynamic ODE, so certain parameters needs to be updated. mass, magnetic field, spin frequency etc are all updated throughout all iterations. 


> Why p_MassGainPerTimeStep? Isn't the important thing just the mass gain?

p_MassGainPerTimeStep is the mass gain for that specific time step. If the name is confusing, I can change it to p_MassGain. 


> double newPulsarMagneticField = (initialMagField_G - magFieldLowerLimit_G) * exp(-p_MassGainPerTimeStep / p_Kappa) + magFieldLowerLimit_G;
> 
> If initialMagField_G was, say, zero to begin with, it will end up having a non-zero value but still one less than magFieldLowerLimit_G. I don't understand the logic...

I don't quite understand the issue with this? It's following Eq. 12 in arxiv:1912.02415 to calculate the decay of magnetic field from mass accretion. There wouldn't be any physical situation in a NS's evolution for the magnetic field to be low enough, certainly not zero. The magnetic field needs to be evolved from a certain value, either decays through spin down or mass accretion. The birth value is drawn from a distribution specified by the program options, and the lower limit of magnetic field should be always lower than that entire distribution. 



> Aside: @yuzhesong , if this replaces #1082 , can we delete that one?

Yes, #1082  can be deleted

jeffriley (changes requested):
Placeholder - I have some changes I'd like made.

ilyamandel:
Just a quick note that I have to go offline now (else I'll have to swim over to Tasmania as the ferry will leave without me), so I'll have to return to the review, which I haven't quite finished, in January of 2025.  Happy New Year!

jeffriley (changes requested):
In addition to @ilyamandel's comments:

In `void NS::CalculateAndSetPulsarParameters()` you have the following code:

    m_PulsarDetails.spinPeriod    = CalculateBirthSpinPeriod();                                         // spin period in ms
    m_PulsarDetails.spinFrequency = _2_PI / (m_PulsarDetails.spinPeriod * SECONDS_IN_MS);

`CalculateBirthSpinPeriod()` can return a value of 0.0 by design (e.g. when `OPTIONS->PulsarBirthSpinPeriodDistribution()` == `ZERO`, which is actually the default), which results in a divide-by-zero error - and sets 
`m_PulsarDetails.spinFrequency` to INF, and we then use that value as a parameter to `NS::CalculateSpinDownRate()`, propagating the error...

At a minimum there should be a check here for spin period = 0.0, but if the user has specified `pulsar-birth-spin-period-distribution ZERO` (or left it at the default value), it seems to me we don't need to call `NS::CalculateAndSetPulsarParameters()` at all,
 
`NS::SpinDownIsolatedPulsar()` also divides by `m_PulsarDetails.spinFrequency` - do we even need to call `NS::SpinDownIsolatedPulsar()` if `m_PulsarDetails.spinFrequency` == 0.0?  At least put a check at the top of the function to check for zero spin and exit immediately.

Ditto for `NS::UpdateMagneticFieldAndSpin()` - can wet put a check at the top of the function to check for zero spin and exit immediately?

How much of the pulsar code can we actually avoid (and should avoid to prevent divide-by-zero errors) if we know the star is not spinning?

yuzhesong:
@jeffriley 

Added solution to issue #1257

ilyamandel:
@yuzhesong :

> p_MassGainPerTimeStep is the mass gain for that specific time step. If the name is confusing, I can change it to p_MassGain.

That sounds like a good solution.

>> double newPulsarMagneticField = (initialMagField_G - magFieldLowerLimit_G) * exp(-p_MassGainPerTimeStep / p_Kappa) + magFieldLowerLimit_G;

>> If initialMagField_G was, say, zero to begin with, it will end up having a non-zero value but still one less than magFieldLowerLimit_G. I don't understand the logic...

> I don't quite understand the issue with this? It's following Eq. 12 in arxiv:1912.02415 to calculate the decay of magnetic field from mass accretion. There wouldn't be any physical situation in a NS's evolution for the magnetic field to be low enough, certainly not zero. The magnetic field needs to be evolved from a certain value, either decays through spin down or mass accretion. The birth value is drawn from a distribution specified by the program options, and the lower limit of magnetic field should be always lower than that entire distribution.

I would not assume that initialMagField_G is lower than magFieldLowerLimit_G.  You never know in advance what silly choices someone else might make. :-) A safer solution could be as follows:

double newPulsarMagneticField = initialMagField_G < magFieldLowerLimit_G ? magFieldLowerLimit_G : (initialMagField_G - magFieldLowerLimit_G) * exp(-p_MassGainPerTimeStep / p_Kappa) + magFieldLowerLimit_G;

ilyamandel (changes requested):
double p = R_CM_6 * R_CM_6 / (mass_g * mass_g * mDot);

in NS::DeltaAngularMomentumByPulsarAccretion_Static()

This is not correct and does not match Eq. (10) of https://www.arxiv.org/pdf/1912.02415 (see powers of M and Mdot).

@yuzhesong , typos happen, but this is going very slowly because typos seem to be particularly common in this code and we are all busy...  I suggest that you check that your code is yielding correct results (as compared to some alternative code that you write independently, say, in Python or Matlab) for a range of parameter choices -- otherwise, we'll be reviewing this for a while. ;-) 

[Aside: if you haven't done it yet, I would also personally recommend re-deriving quantities like the Alfven radius yourself; I find that once you've derived something and therefore understood the scalings, copy-paste typo errors when copying equations are less likely.]

yuzhesong:
@jeffriley has helped with solving the BOOST ODE integrator problem and code cleanup. It is now ready for review and hopefully OK to be merged into dev now. @ilyamandel @SimonStevenson

SimonStevenson:
I got this compiler error trying to compile. The automatic compile test actually succeeded, so this must be a me problem.
```
./constants.h:295:18: error: constexpr variable 'ALFVEN_CONST' must be initialized by a constant expression
```

jeffriley:
Worked on my Linux system. I'll take a look when I get home.

jeffriley:
@SimonStevenson Try again when you're ready.  Apparently c++ on osx doesn't allow std:pow in constexpr statements until c++26.  I moved the definition of ALFVEN_CONST out of constants.h and put it in NS.h - that should work until we move to c++26 and then we can put it back in constants.h :-)

SimonStevenson:
Awesome, can confirm this now works for me on MacOS. Thanks Jeff!!

SimonStevenson (approved):
This looks good to me. I have run the code using the same LMXB binaries I was using for testing the mass transfer fix (command below). These binaries successfully produce MSPs. 
masses_v_time_example
P_v_time_example
PPdot_example_MSP_formation

```
./COMPAS -n 1 --mode BSE --detailed-output TRUE --initial-mass-1 9.0 --initial-mass-2 1.0 --orbital-period 750.0 --common-envelope-alpha s[1.4,1.5,2.0,3.0,4.0,5.0,7.0,10.0,15.0,20.0] --kick-magnitude-1 0.0 --remnant-mass-prescription FRYER2012 --random-seed 0 --evolve-pulsars TRUE --pulsar-birth-magnetic-field-distribution LOGNORMAL --pulsar-birth-spin-period-distribution NORMAL
```

jeffriley (approved):
I approve my changes :-)  Over to you @ilyamandel

ilyamandel (changes requested):
NS::CalculateMomentOfInertiaCGS():
>> double m_r = p_Mass / MSOL_TO_G / p_Radius / KM_TO_CM;

I think what was meant was 
>> double m_r = (p_Mass / MSOL_TO_G) / (p_Radius / KM_TO_CM);

If so, the current calculation is off by 10 orders of magnitude.

[Also, please remove the tap.arizona.edu link from the comment -- it is no longer working.]

jeffriley (changes requested):
Just noticed NS::SpinDownIsolatedPulsar() needs some guards against divide-by-zero - I thought we caught them all, but apparently not.  I'll fix that and push a change later.

jeffriley:
> I think what was meant was
>
>>double m_r = (p_Mass / MSOL_TO_G) / (p_Radius / KM_TO_CM);

Ah - I think that was me in my cleanup.  I'll fix that when I fix NS::SpinDownIsolatedPulsar()

ilyamandel (changes requested):
double CalculateSpinDownRate(const double p_Omega, const double p_MomentOfInteria, const double p_MagField, const double p_Radius)

The comments cite Eq. 2 of https://arxiv.org/pdf/1912.02415, which provides Omega dot.  Instead, the code attempts to first compute Pdot, then re-convert it back to Omega dot.  It is therefore needlessly complex, harder to check against the original equation, and artificially introduces the scope for additional issues (e.g., divide by zero if p_Omega=0 on input -- no such issues if the original equation is implemented directly).  

Also, it seems that you are setting sin^2 alpha from that equation to be 1, i.e., assuming the maximum spin down rate, corresponding to a magnetic field dipole that is perpendicular to the spin axis.  Such an assumption should be reflected in the comments.

In general, the implemented equations should match the references in the comments.

SimonStevenson:
I would prefer that we just only work with P, Pdot and drop all references/uses of Omega/Omegadot. P, Pdot are more commonly quoted observationally (we typically want to make a PPdot diagram, not a Omega/Omegadot diagram) and the initial property that is drawn is spin period, not omega.

ilyamandel (changes requested):
NS.cpp, line 328:
 
    if (utils::Compare(m_PulsarDetails.spinPeriod, 0.0) == 0 || utils::Compare(m_PulsarDetails.magneticField, 0.0) == 0) {  // spin period or magnetic field 0.0?
                                                                                                                            // yes - set all values to 0.0
                                                                                                                            
                                                                                                                            
It seems that if the initial magnetic field is zero, the spin is set to zero regardless of the chosen spin.  Is that intended / documented?

Second question: if the spin period is zero, the code sets the spin frequency to zero.  I am not a fan.  I suggested this before, and will suggest again: the frequency is a well defined quantity which can have a value of zero, meaning no spin, and goes up to finite values (in practice, under 1 kHz). A period of zero is meaningless -- it just should not happen.  Options like 
PULSAR_BIRTH_SPIN_PERIOD_DISTRIBUTION::ZERO 
are not meaningful: there is no such beast as an object with a spin period of zero (the limit as the spin period approaches zero from above corresponds to increasingly faster spin periods, but there's a limit to how fast any NS can spin).  We should set the frequency, store the frequency, and work with frequencies only.

Third question: if the user chose PULSAR_BIRTH_MAGNETIC_FIELD_DISTRIBUTION::ZERO, 
it seems that the magnetic field is actually set to 1 G, not zero.  Is that intended / documented?

ilyamandel:
@SimonStevenson : I wrote just the opposite before I read your comment. ;-)
All right, I can be convinced that we should work with P -- but then let's be consistent and only work with P throughout, as you say.  
And, for continuity, a non-spinning NS should have P = infinity, not zero.

ilyamandel (changes requested):
NS.cpp, line 375:

// calculate isolated decay of the magnetic field for a neutron star
    // see Equation 6 in  arXiv:0903.3538v2       
    m_PulsarDetails.magneticField = NS::NS_MAG_FIELD_LOWER_LIMIT + (initialMagField - NS::NS_MAG_FIELD_LOWER_LIMIT) * std::exp(-p_Stepsize / NS::NS_DECAY_TIME_SCALE); // update pulsar magnetic field in cgs.
    
    
If the initial magnetic field is lower than NS::NS_MAG_FIELD_LOWER_LIMIT, this will increase the magnetic field rather than decaying it.  This seems unphysical.  We should either set the magnetic field initially to be at least NS::NS_MAG_FIELD_LOWER_LIMIT or not increase it if it is below this value.

Also, the comment mentions "isolated decay" [???] -- I think "isolated" probably belongs next to "neutron star", not next to "decay"

-----

Same function, line 386:
>>double Psquared = 2.0 * constant2 * (term1 - term2 - term3) + (initialSpinPeriod * initialSpinPeriod);

I have no idea how 8*pi (from the original equation) multiplied by 4 pi^2 (from the conversion of 1/Omega^2 into P^2) became 2 * 8 * pi^2.  

Same for the implementation of Eq. 4 of arXiv:0903.3538v2 later in this function.  And why not just use CalculateSpinDownRate(), instead of re-implementing it from scratch here?

Sorry, I'll stop checking here -- it's just too difficult for me to confirm the equations.  Maybe I am now making mistakes in the mental arithmetic, but please implement the equations that you refer to, then, if need be, convert the results into the quantities you are interested in, with an explanation in the comments.

jeffriley:
I just pushed some minor changes - including one unrelated to this PR (a fix for the declarations of `BaseStar::CalculateLambdaLoveridgeEnergyFormalism()` (see BaseStar.h and BaseStar.cpp) which I stumbled across - not sure of the impact (haven't tested/thought about it)).  My review of the current NS code follows.

jeffriley (changes requested):
In `NS::CalculateBirthSpinPeriod()`    

    case PULSAR_BIRTH_SPIN_PERIOD_DISTRIBUTION::UNIFORM: {                                                  // UNIFORM distribution between minimum and maximum value as in Oslowski et al 2011 https://arxiv.org/abs/0903.3538 (default Pmin = and Pmax = )
                                                                                                            // and also Kiel et al 2008 https://arxiv.org/abs/0805.0059 (default Pmin = 10 ms and Pmax 100 ms, section 3.4)
        double maximum = OPTIONS->PulsarBirthSpinPeriodDistributionMax();
        double minimum = OPTIONS->PulsarBirthSpinPeriodDistributionMin();

        if ((utils::Compare(maximum, minimum) < 0.0) ||
            (utils::Compare(maximum, 0.0) < 0.0)) {
            // Initial distribution check. If maximum is set below minimum or below 0,
            // default to maximum = 100 ms and minimum = 10 ms
            maximum = 100.0;
            minimum = 10.0;
        } else if (utils::Compare(minimum, 0.0) < 0.0) {
            // Initial distribution check. If minimum is below 0, 
            // it defaults to maximum/10
            minimum = maximum / 10.0;
        }

These checks, and resetting of the minimum and maximum values, should be done in the Options code.  Doing them here introduces a few problems:

   (a) we're using magic numbers in the code that aren't documented - why 100 and 10?
   (b) we're changing user input limits (via the options) without telling the user - they are happily operating under the mistaken notion that the limits they entered are being enforced.  Not only that, we seem to be changing them to some arbitrary value (minimum = 10.0 in one case; minimum = maximum / 10.0 in another).
   (c) we're doing this every time we call this function (possibly every timestep), when it really only needs to be done once - in the Options code.
   
The checks here should be moved to the Options code (`Options::OptionValues::CheckAndSetOptions()` in `Options.cpp`), and that way we can (a) inform the user they've input values that are not self-consistent (i.e. minimum above maximum), and (b) have the user input consistent values rather than us set arbitrary values in the code.  And we get the added benefit of a performance benefit (albeit possibly small).

EDIT: forgot to add that the "magic" numbers `100` and `10`, if we use them in the Options code, should be added to `constants.h` with symbolic names so that anyone reading or changing the code in the future knows what they are and why they have those values.

Similar constructs in `NS::CalculateBirthMagneticField()`, slightly complicated by the use of `NS::NS_MAG_FIELD_LOWER_LIMIT` (but we can actually calculate that as a local variable in the options code for the compare), should be treated as described above.

In `NS::CalculateBirthSpinPeriod()`

        case PULSAR_BIRTH_SPIN_PERIOD_DISTRIBUTION::INF:                                                        // INFINITY if pulsar not spinning 
            pSpin = std::numeric_limits<float>::infinity();
            break;

At a minimum, the function documentation should indicate that the spin period might be returned as `inf` so callers know to check for that rather than use (and so propagate) `inf` in their calculations.

But I have to say that I am not a fan of `PULSAR_BIRTH_SPIN_PERIOD_DISTRIBUTION::INF`, and setting the spin period to `inf`.  I've seen the discussion between @ilyamandel  and @SimonStevenson , and I understand the motivation to use period rather than frequency, but I have to say it seems (to me) to be far more natural to indicate the pulsar is not spinning by setting the frequency to 0.0 rather than the period to infinity.  I also think the coding would be more intuitive (for me) to work with a zero frequency rather than an infinite period.  We can still output the period, but work internally with a zero frequency.  Just saying...

In `NS::CalculateAndSetPulsarParameters()`:

We set `m_PulsarDetails.spinPeriod = CalculateBirthSpinPeriod()`, which may result in `m_PulsarDetails.spinPeriod` being =`inf`.  Then we check if either `m_PulsarDetails.spinPeriod == 0.0` or `m_PulsarDetails.magneticField == 0.0` as a sanity check, and exit the function with zeros for everything except spin period, which is set to `inf`.

What happens if `CalculateBirthSpinPeriod()` returns` inf`?  Will that only happen if `magfield == 0.0`?  If not, we will use `spinPeriod = inf` in the calculations in the `else` part of the conditional.


`NS::UpdateMagneticFieldAndSpin()` - again, we should state in the function documentation that the spin period might be returned as `inf` so that callers can deal with that correctly.

`NS::SpinDownIsolatedPulsar()` can we put a sanity check at the start of the function to exit without change if the NS is not spinning?

yuzhesong:
@jeffriley the latest commit should address all your comments.

jeffriley (changes requested):
I have made some minor changes:

- A little tidy-up
- I explicitly set spin period to `std::numeric_limits<float>::infinity()` in both `NS::CalculateAndSetPulsarParameters()` and `NS::UpdateMagneticFieldAndSpin()`.  That's what the comments say we do, so we should do what we advertise we do.  Setting spin period to `_2_PI / m_PulsarDetails.spinFrequency` does not guarantee infinity if `m_PulsarDetails.spinFrequency` is not `0.0`, and the conditional does not guarantee `m_PulsarDetails.spinFrequency` is `0.0`
- I deprecated the value of `ZERO` for `--pulsar-birth-spin-period-distribution` in favour of `NOSPIN`
- I changed whats-new a little
- EDIT: I removed some older, deprecated, options and option values (it was time...)

@yuzhesong could you change the following documentation files to reflect the new logfile default records:

`standard-logfiles-default-record-specifications-SSE-pulsars`
`standard-logfiles-default-record-specifications-BSE-pulsars`
`standard-logfiles-default-record-specifications-BSE-detailed`

And finally, a question:

On line 513 of `NS.cpp` (in `NS::UpdateMagneticFieldAndSpin()`) you have:

    m_PulsarDetails.magneticField = m_PulsarDetails.magneticField;   

That line does nothing. Did you mean to use `initialMagField`?  If so, the declaration and initialisation of `initialMagField` needs to be moved.  But I don't think `m_PulsarDetails.magneticField` has been changed at this stage, so unsure what the motivation is here...  Should you be setting it to `NS::NS_MAG_FIELD_LOWER_LIMIT` maybe?

jeffriley (approved):
Thanks @yuzhesong - all good.

Over to you I think @ilyamandel

ilyamandel (changes requested):
NS::CalculateSpinDownRate()
//See Equation 5 in https://arxiv.org/abs/2406.11428
OK, but that equation has a factor of sin^2 alpha.  The code does not.  You are making an additional assumption about the angle between the magnetic dipole and the spin axis.  It should be stated as a comment (as I noted previously).

ilyamandel (changes requested):
NS::CalculateBirthSpinPeriod()

        case PULSAR_BIRTH_SPIN_PERIOD_DISTRIBUTION::NOSPIN: {                                                   // Set pulsar spin frequency to 0 if pulsar not spinning. 
            double pSpinFrequency = 0.0;
            pSpin = _2_PI/pSpinFrequency;
            } break;
            
            
Wait, we are knowingly dividing by zero?  I am not convinced this is a good idea, see the first answer in
https://stackoverflow.com/questions/42926763/the-behaviour-of-floating-point-division-by-zero
I though @jeffriley suggested setting the period to infinity in this case, which would have made me more comfortable.

ilyamandel (changes requested):
NS::CalculateAndSetPulsarParameters()

m_PulsarDetails.spinPeriod = CalculateBirthSpinPeriod(); // spin period in ms

Contrary to this, the comments of CalculateBirthSpinPeriod()
say
* @return Birth spin period of Pulsar in s
I believe the latter is correct, the former is not.

jeffriley:
> I though @jeffriley suggested setting the period to infinity in this case, which would have made me more comfortable.

I think the example in the link is flawed - I think the poster is doing integer division.

In C++ implementations that implement the IEEE floating-point standard, in ordinary operation, the division of a finite non-zero floating-point value by 0[.0] is well-defined and results in +infinity if the value is > zero, -infinity if the value is < zero, and NaN if the value is = 0.0, and in each case program execution continues uninterrupted.  Integer division by 0 is undefined and results in a floating-point exception and the process is halted.


I did recommend explicitly setting the value to infinity in other places because we couldn't be guaranteed we'd be dividing by 0.0.  Here we are guaranteed we are dividing by 0.0, so the result should be well-defined - but to be consistent maybe we should just settle on explicitly setting the value to infinity when we know that's actually what we want.

EDIT - actually, no, I think you are right.  Dividing a float by integer 0 should (according to the standard) give +/- inf - but if the poster is not seeing that then maybe there's a problem/misunderstanding somewhere.  Maybe we should just set it explicitly so we are sure :-)

EDIT: I did test this when I implemented the error handling code (that works on Linux...), and dividing a floating-point value by 0.0 sets +/- inf appropriately.  I can't remember if I tested dividing a floating-point valued by integer 0.

ilyamandel (commented):
NS::CalculateAndSetPulsarParameters()
line 335

if (utils::Compare(m_PulsarDetails.spinFrequency, 0.0) == 0 || utils::Compare(m_PulsarDetails.magneticField, 0.0) == 0) {

I do not see where m_PulsarDetails.spinFrequency gets set above this line: only m_PulsarDetails.spinPeriod has been set.  Why are we using a variable that has not yet been set to make decisions (and, possibly, unset the spin period)?

jeffriley:
> I do not see where m_PulsarDetails.spinFrequency gets set above this line

Is it not possible the value on entry to the function is set?  (I haven't checked - but it is a class variable)

ilyamandel (changes requested):
Sorry, that was meant to be a change request, not a comment:

NS::CalculateAndSetPulsarParameters()
line 335

if (utils::Compare(m_PulsarDetails.spinFrequency, 0.0) == 0 || utils::Compare(m_PulsarDetails.magneticField, 0.0) == 0) {

I do not see where m_PulsarDetails.spinFrequency gets set above this line: only m_PulsarDetails.spinPeriod has been set. Why are we using a variable that has not yet been set to make decisions (and, possibly, unset the spin period)?

I also still don't understand (and didn't see a response above to my previous change request on this) why a magnetic field of zero forces a change in the spin period.

ilyamandel:
> Is it not possible the value on entry to the function is set?

If the frequency is set, we should not be calling 
m_PulsarDetails.spinPeriod = CalculateBirthSpinPeriod(); 
-- we should be using the frequency to set the period.  The current behaviour does not appear to be self-consistent or to match the desired goals.

ilyamandel:
> EDIT: I did test this when I implemented the error handling code (that works on Linux...), and dividing a floating-point value by 0.0 sets +/- inf appropriately. I can't remember if I tested dividing a floating-point valued by integer 0.

But why not just use the following (we already do in various other places in the code) instead of trying to divide by zero?  Seems safer...
std::numeric_limits<double>::infinity();

yuzhesong:
@ilyamandel 

> I also still don't understand (and didn't see a response above to my previous change request on this) why a magnetic field of zero forces a change in the spin period.

Essentially we consider if a neutron star with zero spin frequency, infinite spin period, or zero magnetic field to be non-spinning in this situation, so we force all the spin-related parameters to return 0 for the neutron star under this situation. This is accounting for the option of setting magnetic field at birth to ZERO in the program option.

jeffriley:
I agree, which is why I suggested it in other places. I think Robert wanted
to emphasise the relationship, but upon reflection I think setting it
explicitly is better.

yuzhesong:
> > EDIT: I did test this when I implemented the error handling code (that works on Linux...), and dividing a floating-point value by 0.0 sets +/- inf appropriately. I can't remember if I tested dividing a floating-point valued by integer 0.
> 
> But why not just use the following (we already do in various other places in the code) instead of trying to divide by zero? Seems safer... std::numeric_limits::infinity();

@ilyamandel i'm happy with doing it that way.

ilyamandel (changes requested):
NS.cpp, lines 448, 449:

    double omegaK = std::sqrt(G_CGS * p_Mass / magneticRadius) / magneticRadius;                // Hz
    double vDiff  = omegaK - p_SpinFrequency;                                                   // Hz
    
Although Hz and rad/s are dimensionally equivalent, by convention, angular velocities are always quoted in rad/s, Hz is reserved for frequencies.

===========

Line 508:

        double jAcc            = m_MomentOfInertia_CGS * std::sqrt(G_CGS * massG / (radiusCM * radiusCM * radiusCM)) * p_MassGain / G_TO_KG / massG;    
        m_AngularMomentum_CGS += jAcc;  
        
I didn't quite understand what SURFACE means, but it looks like it just means "keep angular velocity fixed".  In that case, what's the point of computing fDot on line 520 -- should it not be identically zero?

=====

ilyamandel:
> @ilyamandel
> 
> > I also still don't understand (and didn't see a response above to my previous change request on this) why a magnetic field of zero forces a change in the spin period.
> 
> Essentially we consider if a neutron star with zero spin frequency, infinite spin period, or zero magnetic field to be non-spinning in this situation, so we force all the spin-related parameters to return 0 for the neutron star under this situation. This is accounting for the option of setting magnetic field at birth to ZERO in the program option.


That's not accurate: ZERO and 0.0 are not the same here.  I already commented on this in a previous review.

        case PULSAR_BIRTH_MAGNETIC_FIELD_DISTRIBUTION::ZERO:                                                    // ZERO
            log10B = 0.0;
            break;
            
 This sets the log of the B field to zero.  That means the B field is set to be 1.0 Gauss, *not* 0.0.
 
 Meanwhile, the code in question looks at the comparison
 
 utils::Compare(m_PulsarDetails.magneticField, 0.0) .
 
 1.0 \neq 0.0, so with the option ZERO, this will in fact not match.

yuzhesong:
> > @ilyamandel
> > > I also still don't understand (and didn't see a response above to my previous change request on this) why a magnetic field of zero forces a change in the spin period.
> > 
> > 
> > Essentially we consider if a neutron star with zero spin frequency, infinite spin period, or zero magnetic field to be non-spinning in this situation, so we force all the spin-related parameters to return 0 for the neutron star under this situation. This is accounting for the option of setting magnetic field at birth to ZERO in the program option.
> 
> That's not accurate: ZERO and 0.0 are not the same here. I already commented on this in a previous review.
> 
> ```
>     case PULSAR_BIRTH_MAGNETIC_FIELD_DISTRIBUTION::ZERO:                                                    // ZERO
>         log10B = 0.0;
>         break;
> ```
> 
> This sets the log of the B field to zero. That means the B field is set to be 1.0 Gauss, _not_ 0.0.
> 
> Meanwhile, the code in question looks at the comparison
> 
> utils::Compare(m_PulsarDetails.magneticField, 0.0) .
> 
> 1.0 \neq 0.0, so with the option ZERO, this will in fact not match.

Ah you're right. The magnetic field distribution setting to ZERO should return ZERO magnetic field not 1.0. I will change this.

yuzhesong:
Just a note: @SimonStevenson and I were discussing this earlier today, and we thought that these unphysical behaviours (0 spin frequency, infinity spin period or 0 magnetic field) really shouldn't exist, because why record pulsar evolution into a file when the pulsars are not evolving... 

Would you @ilyamandel and @jeffriley oppose to remove these setups? It would actually automatically nullify these discussions. We can still have the checks in the calculations just in case something goes wrong but we really should not encourage the setting of non-spinning pulsars here.

jeffriley:
> we really should not encourage the setting of non-spinning pulsars here.

I agree with that.  If we provide the "NOSPIN" option, aren't we really just saying we will treat all neutron stars as vanilla neutron stars, and none as pulsars (effectively turning off pulsar evolution in COMPAS)?  Isn't that the same (or should it be the same) as the `--evolve-pulsars` option?  Also, zero spin frequency is unphysical for all stars isn't it (maybe not in COMPAS...)?  

Here's my naive take on it.  All pulsars are neutron stars, but not all neutron stars are pulsars - that's right, isn't it?  Maybe we should have implemented pulsars as their own class, inheriting from the NS class - and maybe even their own stellar (sub?)type.  If we actually have a pulsar class (and maybe stellar type), then we could define the spin threshold at which a NS spins up and becomes a pulsar (presumably until they spin fast enough there are no jets?), and the conditions under which it spins down and there are no longer jets and it is no longer a pulsar (should it then just become a NS again?).  Should we consider a rewrite (really shouldn't be that hard), or is that something for the future?

yuzhesong:
> All pulsars are neutron stars, but not all neutron stars are pulsars - that's right, isn't it?

This is correct. But, I think in practice (or observation) the differentiation between a NS and a pulsar is if it's observed to pulsate, and there are many reasons why the pulsation is not observed (beaming fraction, beaming angle and direction, spin down luminosity is too low, too far away, to name a few). Plus, in theory, due to the conservation of angular momentum, all neutron stars born from supernovae should have a spin period and it should be relatively low (fast spinning at birth). 

> then we could define the spin threshold at which a NS spins up and becomes a pulsar (presumably until they spin fast enough there are no jets?), and the conditions under which it spins down and there are no longer jets and it is no longer a pulsar (should it then just become a NS again?).

So for all that above I personally would just recommend us getting rid of the no spin and zero magnetic field options, instead of creating an additional class just to accommodate the no-spinning cases. If people don't need the pulsar informations, they don't need to set --evolve-pulsar=True. All other stellar/binary properties of the neutron star systems (mass, radius, orbits, mass transfer, etc) should already be recorded even if --evolve-pulsar=False. But there are chances I might be missing something here? @SimonStevenson correct me here if I'm wrong.

jeffriley:
> instead of creating an additional class just to accommodate the no-spinning cases

No, maybe I didn't explain clearly.  Pulsars are types of neutron stars aren't they.  That lends itself to having a pulsar class that inherits from the NS class.  Then all the code that relates specifically to pulsars, but not to plain neutron stars, resides in the pulsar class - and then, presumably, we can make assumptions in the pulsar class because we wouldn't be there if those assumptions were not true.  And then when the pulsars spin down they go back to being plain neutron stars?  And we don't have to worry about the checks in the pulsar code.

> All other stellar/binary properties of the neutron star systems (mass, radius, orbits, mass transfer, etc) should already be 
> recorded even if --evolve-pulsar=False. 

That's right - that would be the `NS` class, as opoosed to the `Pulsar` class.

> If people don't need the pulsar informations, they don't need to set --evolve-pulsar=True. 

True, and that was my point - or question.  What's the difference between `--evolve-pulsars False` and `--evolve-pulsars True` + `--pulsar-birth-spin-period-distribution NOSPIN` - aren't they effectively the same thing?  Or is that simplifying too much?

yuzhesong:
I think I get your point now. But - 





> What's the difference between `--evolve-pulsars False` and `--evolve-pulsars True` + `--pulsar-birth-spin-period-distribution NOSPIN` - aren't they effectively the same thing? Or is that simplifying too much?

Logically yes they are the same thing. What the code is doing right now is that: 

- `--evolve-pulsars False` means we end up with a neutron star and that's it. It doesn't call on anything in NS.cpp to calculate spin/magnetic field or anything. 
- `--pulsar-birth-spin-period-distribution NOSPIN` creates the SSE/BSE_Pulsar_Evolution file and records everything (except for mass/radii/orbital properties) to be zero. Which really is unnecessary. 

> > All other stellar/binary properties of the neutron star systems (mass, radius, orbits, mass transfer, etc) should already be
> > recorded even if --evolve-pulsar=False.
> 
> That's right - that would be the `NS` class, as opoosed to the `Pulsar` class.

These are in the BaseStar or BaseBinaryStar classes, so maybe not necessary to have an additional class just for them when the star is a NS? 

So I guess my point is that we already have all the necessary parameters for a neutron star calculated and stored as stellar properties. Do we really need another class for NSs that are not pulsars?

jeffriley:
> Logically yes they are the same thing

So that's my point - why would anybody enable pulsars and the set the birth spin period to zero?  They wouldn't, right? So we should remove `NOSPIN`.

> These are in the BaseStar or BaseBinaryStar classes

All (or most) stellar properties are recorded in the base class, and are inherited by the derived classes.

> , so maybe not necessary to have an additional class just for them when the star is a NS?

I'm not sure I understand that comment - we already have a class for when a star is a NS - it's the NS class...

> Do we really need another class for NSs that are not pulsars?

No, that's not what I'm saying.  We already have the NS class - that will stay, and that is for all neutron stars.  The (new) pulsar class would be for neutron stars that are also pulsars.  The Pulsar class would inherit from the NS class - it's just an OO way of looking at, and implementing, it...

jeffriley:
Ok, so @yuzhesong has convinced me - either all neutron stars in COMPAS are pulsars, or none are - there is no physical distinction that would allow us to decide if/when a NS is actually a pulsar or not.  So no need for a separate pulsar class.

yuzhesong:
> Line 508:
> 
> ```
>     double jAcc            = m_MomentOfInertia_CGS * std::sqrt(G_CGS * massG / (radiusCM * radiusCM * radiusCM)) * p_MassGain / G_TO_KG / massG;    
>     m_AngularMomentum_CGS += jAcc;  
> ```
> 
> I didn't quite understand what SURFACE means, but it looks like it just means "keep angular velocity fixed". In that case, what's the point of computing fDot on line 520 -- should it not be identically zero?

@ilyamandel SURFACE here means pulsar accretes material directly onto its surface from the common envelope, so the jAcc here calculates the increase in angular momentum by adding p_MassGain mass at the co-rotation angular velocity at NS surface, while the NS is spinning at its own spin frequency. So fDot here shouldn't be zero. 

Other than that, following the discussion above, I have deprecated the "ZERO" option for --pulsar-birth-magnetic-field-distribution and "NOSPIN" (originally "ZERO") option for --pulsar-birth-spin-period-distribution. I also removed all the guards in NS.cpp as there is no need for them if we're not dealing with non spinning NSs any more.

ilyamandel:
> @ilyamandel SURFACE here means pulsar accretes material directly onto its surface from the common envelope, so the jAcc here calculates the increase in angular momentum by adding p_MassGain mass at the co-rotation angular velocity at NS surface, while the NS is spinning at its own spin frequency. So fDot here shouldn't be zero.

Just to repeat what I said on the call: in that case, line 475 of NS.cpp is not correct.  The specific Keplerian angular momentum at the surface is just sqrt(G*M_NS*R_NS), and j_Acc should be p_MassGain times this quantity.  It should not depend on the NS moment of inertia.

> Other than that, following the discussion above, I have deprecated the "ZERO" option for --pulsar-birth-magnetic-field-distribution and "NOSPIN" (originally "ZERO") option for --pulsar-birth-spin-period-distribution. I also removed all the guards in NS.cpp as there is no need for them if we're not dealing with non spinning NSs any more.

OK!

But you accept a draw of a zero spin period; if you really don't want to allow them (I know, set of measure zero, but still), you'll want to change < to <= on line 167 on NS.cpp:

do { pSpin = RAND->RandomGaussian(sigma) + mean;} while (iterations++ < PULSAR_SPIN_ITERATIONS && utils::Compare(pSpin, 0.0) < 0);

to

do { pSpin = RAND->RandomGaussian(sigma) + mean;} while (iterations++ < PULSAR_SPIN_ITERATIONS && utils::Compare(pSpin, 0.0) <= 0);

ilyamandel (commented):
Not sure we really need this around line 225 of NS.cpp: even if lof10B is negative, the actual B field is always positive...


            // this should terminate naturally, but just in case we add a guard
            std::size_t iterations = 0;
            do { log10B = RAND->RandomGaussian(sigma) + mean;} while (iterations++ < PULSAR_MAG_ITERATIONS && utils::Compare(log10B, 0.0) < 0);
            if (iterations >= PULSAR_MAG_ITERATIONS) THROW_ERROR(ERROR::TOO_MANY_PULSAR_MAG_ITERATIONS);
            
            
Would it be more useful to replace             
utils::Compare(log10B, 0.0) < 0
with utils::Compare(log10B, log10(NS::NS_MAG_FIELD_LOWER_LIMIT) < 0
to ensure that the B field is always set to be above NS::NS_MAG_FIELD_LOWER_LIMIT ?

ilyamandel (changes requested):
The frequency derivative is not computed correctly on line 487 of NS.cpp:

double fDot = jAcc / p_Stepsize / m_MomentOfInertia_CGS;

There are two reasons why this isn't right.  First, the change in frequency is not the change in angular momentum divided by the moment of inertia.  One can add angular momentum and keep the frequency the same as it was.  Secondly, this doesn't account for the change in moment of inertia (which may have changed because mass was added).  

Fortunately, there's an easy way to do the calculation correctly.  Replace the following lines

m_PulsarDetails.spinFrequency = m_AngularMomentum_CGS / m_MomentOfInertia_CGS;
        m_PulsarDetails.spinPeriod    = _2_PI / m_PulsarDetails.spinFrequency;
        double fDot                   = jAcc / p_Stepsize / m_MomentOfInertia_CGS;
        
with

double previousSpinFrequency = m_PulsarDetails.spinFrequency;
m_PulsarDetails.spinFrequency = m_AngularMomentum_CGS / m_MomentOfInertia_CGS;
        m_PulsarDetails.spinPeriod    = _2_PI / m_PulsarDetails.spinFrequency;
        double fDot                   = (m_PulsarDetails.spinFrequency - previousSpinFrequency) / p_Stepsize;

ilyamandel (approved):
Woohoo, we are done! ;-)

PR #1318, opened 2025-01-07
Tides Eccentricity Behavior Fix

veome22:
Updated the behavior of tides in the `KAPIL2024` model, where eccentric binaries could sometimes be spun up past synchronization due to the positive spin contribution of non-circular terms, i.e. $(l,m) \neq (2,2)$ . 

See the plot below for an example of the spin evolution for a binary with the following parameters ($e=0.7$):
```
./COMPAS/src/COMPAS --random-seed 10 -n 1 --initial-mass-1 45.0 --initial-mass-2 18.0 --semi-major-axis 105 --eccentricity 0.7 --chemically-homogeneous-evolution-mode NONE --tides-prescription KAPIL2024 --mass-change-fraction 0.001 --radial-change-fraction 0.001
```

Due to the large eccentricity, the $d\Omega /dt$ term for the primary towards the beginning of CHeB is positive despite its spin exceeding the orbital frequency. 
image

After the update, the spin contribution from tides reverts to the (2,2)-only contribution whenever the primary spin exceeds orbital frequency, while using the full expression when the spin is lower. This behavior better preserves synchronization throughout the binary evolution.
image

### DIFF
--- a/src/BaseBinaryStar.cpp
+++ b/src/BaseBinaryStar.cpp
@@ -1081,11 +1081,11 @@ double BaseBinaryStar::CalculateDOmegaTidalDt(const DBL_DBL_DBL_DBL p_ImKlm, con
     double R1_AU       = radiusStar * RSOL_TO_AU;
     double R1_over_a   = R1_AU / m_SemiMajorAxis;
     double R1_over_a_6 = R1_over_a * R1_over_a * R1_over_a * R1_over_a * R1_over_a * R1_over_a;
-    double e2_term = (m_Eccentricity * m_Eccentricity) *  ((ImK12 / 4.0) - (5.0 * ImK22) + (49.0 * ImK32 / 4.0));
+    double e2_spin_term = (m_Eccentricity * m_Eccentricity) *  ((ImK12 / 4.0) - (5.0 * ImK22) + (49.0 * ImK32 / 4.0));
 
     // if the star is rotating (super) synchronously AND quadratic 'e' terms cause the star to spin up further, ignore the higher order terms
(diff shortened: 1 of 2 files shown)


ilyamandel (changes requested):
@veome22 , thanks for making this change.

You'll need to merge in the latest dev.

Perhaps more importantly, is there a copy-paste error from 
BaseBinaryStar::CalculateDOmegaTidalDt()
to 
BaseBinaryStar::CalculateDSemiMajorAxisTidalDt() ?

e2_term_spin has the value given to e2_term above, rather than the correct value for this function.  But the correct value is given to e2_term, and e2_term_spin doesn't seem to be used, so maybe this has no impact?  If so, please delete it, if not, please clarify.

veome22:
@ilyamandel thank you for the review!
I have merged the dev changes so the PR should be good to go.

> Perhaps more importantly, is there a copy-paste error from
> BaseBinaryStar::CalculateDOmegaTidalDt()
> to
> BaseBinaryStar::CalculateDSemiMajorAxisTidalDt() ?

This is not a typo-- I have chosen to ignore $O(e^2)$ terms across all tidal equations based on the spin synchronization condition only. So if the star is synchronized, and tides would make it spin faster, then we want to ignore the non-(2,2) terms across $d\Omega/dt$ AND $da/dt$, ensuring that their signs are always opposite to one another. This way the angular momentum transfer from the binary to the star should remain self-consistent.

I have expanded the comments to hopefully explain this choice better. Hope this makes sense!

ilyamandel (approved):
Hi @veome22 ,
Thanks for the clarification -- looks good to me!

PR #1320, opened 2025-01-07
Fix for issue 1317

### LINKED ISSUE #1317: Supernova from main sequence merger product not recorded in BSE_Supernovae
**Describe the bug**
When using `--evolve-main-sequence-merger TRUE` to evolve main sequence merger products, I am finding that even if the merger product subsequently undergoes a supernova and forms a NS/BH, that event is not recorded in BSE_Supernovae

**Label the issue**
`urgency_low` - This is a moderately urgent issue
`severity_moderate`    - This is a minor bug with minimal impact

**To Reproduce**
For example, a single binary that merges can be run with 
```
./COMPAS -n 1 --initial-mass-1 7.0 --initial-mass-2 1.5 --orbital-period 1.0 --evolve-main-sequence-merger TRUE --detailed-output TRUE
```

example_MS_merger_masses
example_MS_merger_types

**Expected behavior**
An entry in BSE_Supernovae should be generated for the merger product/all supernovae that occur.

 - OS: MacOS v12.4
 - COMPAS v03.10.04

[EDITED TO REMOVE OTHER SYSTEM]

jeffriley:
Fix for issue 1317 - SN event not always logged to BSE log file when evolving MS merger products.

@ilyamandel I thought I had some questions about where we need to check for evolving MS merger products when there has been a stellar merger, but upon checking I think we're covered in all places, so the fix here should be all that's needed.

I am probably responsible for this problem manifesting - I have a vague recollection of adding a check for stellar merger towards the end of BaseBinaryStar::EvaluateBinary() thinking, at the time, "We don't need to do this if we've had a stellar merger". Well, we do if the stellar merger was a MS merger and we're evolving MS merger products... This is the check, now modified, at (now) line 2911 in BaseBinaryStar.cpp.

While investigating this problem I noticed that we don't always log a "TIMESTEP_COMPLETED" record to the BSE_Detailed_Output file for the very last timestep. We do log a "FINAL_STATE" record which has all the info required, but users (and particularly I think the detailed plotting code) look for the "TIMESTEP_COMPLETED" record type - so I changed the code to ensure a "TIMESTEP_COMPLETED" record is always logged, even if has the same information as the "FINAL_STATE" record.

### DIFF
--- a/src/BaseBinaryStar.cpp
+++ b/src/BaseBinaryStar.cpp
@@ -2908,8 +2908,8 @@ void BaseBinaryStar::EvaluateBinary(const double p_Dt) {
         }
     }
 
-    if (!StellarMerger()) {                                                                                             // stellar merger?
-                                                                                                                        // no - continue evolution
+    if (!StellarMerger() || (HasOneOf({ STELLAR_TYPE::MASSLESS_REMNANT }) && OPTIONS->EvolveMainSequenceMergers())) {   // check stellar merger or evolving MS mergers
+                                                                                                                        // continue evolution
         if ((m_Star1->IsSNevent() || m_Star2->IsSNevent())) {
             EvaluateSupernovae();                                                                                       // evaluate supernovae (both stars) if mass changes are responsible for a supernova
             (void)PrintDetailedOutput(m_Id, BSE_DETAILED_RECORD_TYPE::POST_SN);                                         // print (log) detailed output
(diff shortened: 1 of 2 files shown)


jeffriley:
Thanks @SimonStevenson - target branch is now correct :-)

SimonStevenson (approved):
Thanks for the fix Jeff! I have run the code and my example system is now correctly recorded in BSE_Supernovae. All good to go from me.

Issue #1324, opened 2025-01-20
Pulsar formed from stellar merger product not recorded in BSE_Pulsar_Evolution


SimonStevenson:
**Describe the bug**
Pulsar formed from stellar merger product not recorded in BSE_Pulsar_Evolution

This is similar to issue #1317 I posted recently.

I think I have tracked this down and have a solution. Will post a PR shortly.

The EVOLUTION_STATUS in `BaseBinaryStar::Evolve` was recorded as MASSLESS_REMNANT due to the `IsMRandRemant()` check. I have added an additional check to see if the user is evolving pulsars, and if so if the stellar remnant is a neutron star, then keep evolving.

Once I had done this, I found that I got hundreds of thousands of 1 Myr time steps. I tracked this down to `MR.h` which has `CalculateTimestep` rather than `ChooseTimestep`. Fixing this resolved this issue.

I have also added a check to see if the binary is bound to `BaseBinaryStar::ChooseTimestep` so that the GW and tides stuff is not evaluated for merged/unbound binaries. 

**Label the issue**
Please label the 'severity' and 'urgency' of this issue. You can choose:

`urgency_moderate` - This is a moderately urgent issue
`severity_moderate` - This is a moderately severe bug

**To Reproduce**
This is the system I have been using to test this:

```
./COMPAS -n 1 --mode BSE --initial-mass-1 14.0 --initial-mass-2 1.0 --semi-major-axis 0.1 --evolve-main-sequence-merger TRUE --evolve-pulsars TRUE --pulsar-birth-magnetic-field-distribution LOGNORMAL --pulsar-birth-spin-period-distribution UNIFORM --radial-change-fraction 0.01 --remnant-mass-prescription FRYER2012
```

**Expected behavior**
Pulsar should be recorded in BSE_Pulsar_Evolution.

**Versioning (please complete the following information):**
 - OS: MacOS v 12.4
 - COMPAS v03.12.00

ilyamandel:
Thank you for fixing this with #1325 , @SimonStevenson !

PR #1325, opened 2025-01-20
Allow for evolution of pulsar formed from main sequence stellar merger

SimonStevenson:
Fixes issue #1324


ilyamandel (approved):
Thank you for the fixes, @SimonStevenson -- all look great!

PR #1326, opened 2025-01-28
Fixes initialisation in MS_gt_07::Initialise() for CORE_MASS_PRESCRIPTION::SHIKAUCHI

jeffriley:
MS_gt_07::Initialise() now allows for CH stars that spin down when initialisation done for CORE_MASS_PRESCRIPTION::SHIKAUCHI.

@brcekadam I haven't tested this other than for compilation - can you check that the CORE_MASS_PRESCRIPTION::SHIKAUCHI code functions as you expect.  If you can find a CH star that spins down, you could check for that too (but don't worry too much if you can't find one easily).

### DIFF
--- a/src/BaseBinaryStar.cpp
+++ b/src/BaseBinaryStar.cpp
@@ -2850,19 +2850,16 @@ void BaseBinaryStar::EmitGravitationalWave(const double p_Dt) {
  */
 double BaseBinaryStar::ChooseTimestep(const double p_Multiplier) {
 
-    // Timesteps required by individual stars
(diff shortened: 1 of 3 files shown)


ilyamandel (approved):
Looks good to me! 
@jeffriley , I leave it to you and @brcekadam to decide whether to merge now or wait for Adam's review. :)

brcekadam (approved):
Thanks @jeffriley! I tested it and everything looks good. I wasn't able to find a CH star that spins down, but there shouldn't be an issue anymore.

PR #1328, opened 2025-02-03
Minor enhancement to MS core mass calculations

brcekadam:
I have made a few changes to the main sequence core mass prescription:

- As suggested by @ilyamandel, I renamed `--main-sequence-core-mass-prescription SHIKAUCHI` option to `BRCEK`. I will be writing a paper describing the new implementation, and since we’ve added some additional functionality on top of the Shikauchi et al. (2024) core mass calculations, it will probably make more sense to rename the prescription so that users can find relevant details more easily.

- I've made some changes to the core mass calculations themselves. The use of ODE solver was actually not necessary after all, @ryosuke-hirai found an easier way to do it, which should slightly speed up the calculations. We are assuming that the time steps are small enough that the change in core mass is small, which should be always true. In the case of thermal timescale mass transfer, the core does not have the time to react to the change, and we can assume that relevant quantities stay constant. So we don’t have to use ODE solver either.

- I have lowered the threshold for core mass calculations and rejuvenation from 15 Msun down to 1.5 Msun (roughly the mass at which stars start to have convective cores). The main reason why we could not do it before was the new luminosity prescription, so I ended up using the default luminosity value (from Hurley) in the core mass calculations for stars with `m_MZAMS` lower than 15 Msun. This actually seems to work quite well and allows all calculations even for lower mass stars, while keeping the HR diagram intact.

- I have made the change that `m_Mass0` is always updated in the beginning of the HG if `BRCEK` core mass prescription is used and `m_MZAMS` is above the threshold. I noticed that in the case of MS accretors, the core mass was not being updated at all. So the full core mass evolution would be calculated, but since this value was lower than the default value, Mass0 was never updated, and we would end up with the same core mass as if the `ZERO` prescription was used.

HRdiagram
This is how the HR diagram looks for solar metallicity tracks. The dashed lines show the default Hurley stellar tracks and solid lines show the new implementation. New luminosity prescription kicks in at 15 Msun, but core mass calculations are now allowed down to 1.5 Msun.


jeffriley (changes requested):
Thanks @brcekadam.  Just a quick look - I'll take a more detailed look later, but for now:

In `What's New`: I get the changes you made, but given the date on the `What's New` entry and the gap between then and now, there is a possibility for confusion. Maybe it's worth a new entry just to point out we changed the option.  (I don't think we need to bother deprecating SHIKAUCHI` given the short timespan).

Also, for the default yaml file, just changing an entry value doesn't change the header that indicates what version of COMPAS created the yaml file. I know it works, but it could cause confusion.  Maybe just use the new version of COMPAS to create a new yaml file and then the headers will be correct.

In `HG::CalculateLuminosityOnPhase()`:

    if (... (utils::Compare(m_MZAMS, 15.0) <= 0))

What does the `15.0` refer to?  Instead of having a magic number in the code, should it be in `constants.h`? I'm not suggesting it must be, just asking you to consider whether it should be. If it doesn't go into `constants.h` I'd like to see a comment in the code describing why `15.0` is used.

Also in `HG::CalculateLuminosityOnPhase()`, we may use `m_Mass` to calculate `LTMS` instead of `p_Mass`, but we will always us `p_Mass` to calculate `LEHG` - not suggesting it shouldn't be that way, just asking for a sanity check for consistency.

ilyamandel (commented):
BaseBinaryStar::ResolveMainSequenceMerger()
- why do we only use the updated core mass when both stars exceed the mass threshold?  should we do it if just one does so (for that star)?

HG::CalculateLuminosityOnPhase()
- This seems to make some very specific assumptions about what values are being used for p_Mass vs m_Mass when calling this; at least state these assumptions in the comments, but are they warranted?
- ditto for HG::CalculateRadiusOnPhase()

brcekadam:
Hi @jeffriley and @ilyamandel, I'll make the requested changes! To answer your questions:

The number `15.0` is the ZAMS mass from which we can use the Shikauchi et al. (2024) luminosity prescription. I've used it a few times throughout and left comments, but forgot in this case... I'll make sure to leave a comment in each case. It might not be necessary to define it in `constants.h`, this number should never be changed by the user.

In `BaseBinaryStar::ResolveMainSequenceMerger()`, we have slightly different ways of determining the final hydrogen mass when the new core mass prescription is used and when it's not. I did not really want to mix the two approaches if one star has an accurate core mass value and the other doesn't... but maybe that wouldn't cause any issues, I just haven't looked at mergers much.

I also looked at the changes after PR #1331, and unfortunately it does cause unphysical stellar tracks when during the MS hook the temperature rises, but the luminosity decreases (see the plot below). 

So I was thinking, what if instead of modifying `HG:UpdateInitialMass()`, we do `m_Mass0 = std::max(Mass0ToMatchDesiredCoreMass(this, desiredCoreMass), m_Mass);` in `HG:Initialise()`, so we prevent `m_Mass0` from ever dropping below `m_Mass`? Now this would mean that in some cases, we do the core mass calculations during the MS, but just end up using the slightly higher core mass from Hurley in the end. If the star goes through significant mass loss, the calculations would still be important. This would keep the stellar tracks physical and prevent any jumps in luminosity, and I don't need to make any changes to `HG::CalculateLuminosityOnPhase()`. I would still need to keep the `p_Mass` vs `m_Mass` thing in `HG::CalculateRadiusOnPhase()`, since `m_Mass` is used for radius calculations on the MS, but `m_Mass0` is used in the HG, so this prevents any jumps in radius.

HRdiagram_WithWinds

brcekadam:
> So I was thinking, what if instead of modifying HG:UpdateInitialMass(), we do m_Mass0 = std::max(Mass0ToMatchDesiredCoreMass(this, desiredCoreMass), m_Mass); in HG:Initialise(), so we prevent m_Mass0 from ever dropping below m_Mass?

Hmm actually, that's not a good solution either. In MS accretors, we do want to allow `m_Mass0` to be less than `m_Mass`, otherwise the core can end up being overestimated by quite a lot if we use the Hurley core mass at the end of MS. So we should keep the change in `HG:UpdateInitialMass()`, but then I'm not sure if there is a way to fix the stellar tracks for these lower mass stars.

ilyamandel:
@brcekadam :

> So I was thinking, what if instead of modifying HG:UpdateInitialMass(), we do m_Mass0 = std::max(Mass0ToMatchDesiredCoreMass(this, desiredCoreMass), m_Mass); in HG:Initialise(), so we prevent m_Mass0 from ever dropping below m_Mass?

I worry the issue with this solution goes beyond accretors -- it's a problem for any star whose core mass at the end of the MS is lower than you would expect from the Hurley formulae given its total mass.  The reason we want Mass0 is that it's a proxy for the core mass on the HG, i.e., it determines the luminosity...

ilyamandel:
Hi @brcekadam ,
It looks like we have a solution in the Slack channel -- do you want to polish off this PR?  [Note: we will probably end up using different delta's for accreting stars than for mass losing stars anyway, given how that study is going...]

brcekadam:
Hi @ilyamandel, 
Yes, I have the changes mostly ready now! I was just waiting for PR #1335 to be merged first :-)

ilyamandel:
OK.  Simon found a possible issue with #1335 that I'd like to investigate further. Will aim to finish in the next day or two.

brcekadam:
Hi @jeffriley, 
I just pushed the last few changes! Let me know if you'd like me to make some further changes.

ilyamandel (approved):
Thank you, @brcekadam !

jeffriley (approved):
Thanks @brcekadam - all good.

jeffriley:
@avivajpeyi we seem to be failing here installing texlive... Thoughts?

avivajpeyi:
See https://github.com/TeamCOMPAS/COMPAS/pull/1339 :)

jeffriley:
> See 1339 :)

Oh - I thought the fix was already in :-)

jeffriley:
@avivajpeyi 1339 approved and merged - same problem.  Did I miss something?

avivajpeyi:
Huh strange, it worked on that branch, and looks like the ci is still running on dev...


https://github.com/TeamCOMPAS/COMPAS/actions/runs/13407641721/job/37450422362


It shouldn't be an issue anymore, but I can take a look at it once I'm back at my computer tomorrow

brcekadam:
I just synced this branch with dev. Now it should be working :-)

jeffriley:
@avivajpeyi stand down - I think @brcekadam fixed it :-)

PR #1330, opened 2025-02-04
Gwlandscape python plot backend

TomReichardt:
Adding functionality to the python plotting scripts to enable the output plots to be displayed in the GWLandscape service. Writing tests for all of the added functions.


ilyamandel (approved):
Thank you for the updates and the tests, @TomReichardt !

PR #1331, opened 2025-02-07
Fixes to Mass0 treatment on HG; fixes to van den Heuvel diagram plotting; update to name of "COMPAS" PPISN prescription

ilyamandel:


jeffriley (approved):
@ilyamandel looks good - just a couple of things that you can choose to ignore:

(a) should you announce the deprecation of "COMPAS" in favour of "WOOSLEY" in `What's New`?

(b) a technical programming point (in a Python program!):

In `plot_detailed_evolution.py`, I am not convinced the bitwise and operator `&` is the best choice in (e.g.) lines 584-591 (there may be other examples - these lines are just the ones I saw).

The bitwise and operator `&` ANDs two binary numbers together and sets each bit of the result to `1` only if the corresponding bits of both operands are `1`.  While that may produce the desired result here (because of how TRUE and FALSE are represented), I think the boolean and operator `AND` should be used instead.  Not only is it (I think) the correct operator to use, the boolean operators support short-circuiting, whereas the bitwise operators don't.

Short-circuiting is not just about performance - it's a useful tool.  For example, this conditional statement:

    if x > 0.0 and (y / x) > 3.0: do something...

will evaluate correctly when x = 0.0, whereas this conditional statement:

    if x > 0.0 & (y / x) > 3.0: do something...

will cause a divide-by-zero error when x = 0.0 and terminate the program at that point.

I suspect you've just followed what was already in the code, but what's in the code is really not right (even though it works).

ilyamandel:
Thank you very much for the review, @jeffriley !  And you were right, I just copied & in python without thinking (I very rarely write python code, so I am not sure I even knew what the logical AND operator was).  Now fixed.

Issue #1333, opened 2025-02-11
No pulsar output generated for single star


SimonStevenson:
**Describe the bug**
When running a single massive star (that becomes a neutron star/pulsar), COMPAS does not produce a SSE_Pulsar_Evolution file, despite `--evolve-pulsars TRUE`

Context: I was planning on using this set up to test changes to the pulsar evolution routines in COMPAS. 

It seems that there is no such thing as an SSE_Pulsar_Evolution output in the code. I think this should be added (I will do this and submit a PR to fix this issue).

Note to self: we should also check the stopping condition for single stars to make sure that they continue evolving if `--evolve-pulsars TRUE` (similar to issue #1324 )

**Label the issue**
`urgency_low`      - This issue is not urgent
`severity_minor`    - This is a minor bug with minimal impact

**To Reproduce**
Run the following command:
```
./COMPAS --mode SSE -n 1 --initial-mass 10.0 --pulsar-magnetic-field-decay-timescale 1.0 --evolve-pulsars TRUE --random-seed 0
```
The resultant COMPAS_Output.h5 file does not contain an SSE_Pulsar_Evolution

**Expected behavior**
COMPAS_Output.h5 should contain SSE_Pulsar_Evolution for single stars that produce pulsars when `--evolve-pulsars TRUE`

**Versioning (please complete the following information):**
 - OS: MacOS v12.4
 - COMPAS - recent dev [e.g., v03.12.04]

SimonStevenson:
I've made a start on this on this branch:
https://github.com/TeamCOMPAS/COMPAS/tree/single_psr_output

I have added a SSE_Pulsar_Evolution output (by analogy with the existing outputs) and edited the stopping condition for single stars to allow neutron stars/pulsars to continue evolving if EvolvePulsars is True. 

Current issues:
1) Despite naming it SSE_Pulsar_Evolution, the pulsar output ends up being called BSE_Pulsar_Evolution. Not sure why.
2) The pulsar parameters are initialised, but never updated for single stars, because NS::UpdateMagneticFieldAndSpin() is only called in BaseBinaryStar::EvaluateBinary() (and relies on binary quantities). I could try writing an equivalent function to go in Star::Evolve. Would that work?

Updated command line used for testing:
```
[...]

PR #1334, opened 2025-02-12
Add single psr output

SimonStevenson:
Add SSE PSR Output to resolve issue https://github.com/TeamCOMPAS/COMPAS/issues/1333 . See that issue for further context and details. 

Current outstanding issue to be resolved before merging: Output appears with name BSE_Pulsar_Evolution rather than SSE_Pulsar_Evolution. I called it SSE_Pulsar_Evolution in `LogTypedefs.h`. I assume it is picking up the name from the `--logfile-pulsar-evolution` option (which defaults to BSE_Pulsar_Evolution), but the same is true for SSE_Supernovae (default value for `--logfile-supernovae` is BSE_Supernovae) and that correctly shows up as SSE_Supernovae. I assume I just missed something somewhere. [EDIT: now solved. Indeed, I missed something in `Options.h`]

Example test line:
```
./COMPAS --mode SSE -n 1 --initial-mass 10.0 --pulsar-magnetic-field-decay-timescale 1000.0 --evolve-pulsars TRUE --random-seed 0 --maximum-evolution-time 1000.0 --pulsar-birth-magnetic-field-distribution LOGNORMAL --pulsar-birth-spin-period-distribution NORMAL
```

Some example test plots:
single_psr_test_B_v_t
single_psr_test_f_v_t


SimonStevenson:
Also realised I need to add something to the What's New page, will do that now

SimonStevenson:
Figured out the name issue

ilyamandel (approved):
Hi @SimonStevenson ,
Looks good, happy to approve.  Two things for your consideration:

1. Star::EvolveOneTimeStep

I found the following a bit hard to parse.  

else if (m_Star->IsOneOf(COMPACT_OBJECTS) && !OPTIONS->EvolvePulsars()){                            // If star is a compact object and we are not evolving pulsars
                evolutionStatus = EVOLUTION_STATUS::DONE;                                                       // yes - we're done	                evolutionStatus = EVOLUTION_STATUS::DONE;                                                       // we are done
            }
            else if (m_Star->IsOneOf(WHITE_DWARFS) || m_Star->StellarType() == STELLAR_TYPE::BLACK_HOLE){       // If compact object is a WD or BH, we're done even if we are evolving pulsars
                evolutionStatus = EVOLUTION_STATUS::DONE;
            }	           
            
            
What if you instead just do this?

 else if (m_Star->IsOneOf(WHITE_DWARFS) || m_Star->StellarType() == STELLAR_TYPE::BLACK_HOLE ||  (m_Star->StellarType() == STELLAR_TYPE::NEUTRON_STAR && !OPTIONS->EvolvePulsars() ) ){       // If a WD or BH, or an NS but not evolving pulsars, we're done 
                evolutionStatus = EVOLUTION_STATUS::DONE;
            }	            
            
            
2. LogTypedefs.h

The list seems to be in lexicographic order, except for this new addition -- so maybe keep it ordered?

    SSE_DETAILED_OUTPUT,
    SSE_SUPERNOVAE,
    SSE_SWITCH_LOG,	    
    SSE_SYSTEM_PARAMETERS,
    SSE_PULSAR_EVOLUTION

SimonStevenson:
Thanks for the review Ilya!

1. Thanks, I agree that's a bit simpler than what I had.
2. Fixed!  

I've implemented and tested these changes. Will now merge.

ilyamandel (approved):
Thanks, @SimonStevenson !

PR #1335, opened 2025-02-12
Nuclear Timescale MT

ilyamandel:


jeffriley (approved):
@ilyamandel Looks good - only concern is use of -inf. It looks like the return of -inf is only used to force the value < 0 so conditionals will go a particular way? That's fine as long as we don't ever start using the -inf in calculations and propagate the value through to other calculations etc....

ilyamandel:
Hi @jeffriley , 
-inf is used to make sure that it's smaller than any finite number that comes from CalculateZetaRocheLobe().

jeffriley:
Yeah, got it. Just as long as future programmers don't propagate the -inf.
Maybe a note in the function documentation?

PR #1340, opened 2025-02-20
Fix of convective envelope mass calculation; fix of nuclear timescale mass transfer for giants

ilyamandel:
- Fixed typo in implementation of Tonset for convective envelope mass calculation (cf. Mandel, Hirai, Picker, 2024)
- Corrected radial estimates for mass losing Giant Branch stars
- Added new functionality, CalculateRadiusOnMassChange(), to streamline code
- Fixed typo in calculation of binding energy of secondary's envelope in the 2-stage CE prescription

### DIFF
--- a/src/BH.h
+++ b/src/BH.h
@@ -59,6 +59,8 @@ protected:
     double  CalculateMassLossRate()                                                 { return 0.0; }                                                     // Ensure that BHs don't lose mass in winds
     double  CalculateMomentOfInertia() const                                        { return (2.0 / 5.0) * m_Mass * m_Radius * m_Radius; }
     double  CalculateRadiusOnPhase() const                                          { return CalculateRadiusOnPhase_Static(m_Mass); }                   // Use class member variables - returns radius in Rsol
+    double  CalculateRadiusOnPhase(double p_Mass, double p_Luminosity) const        { return CalculateRadiusOnPhase(); }        // not a meaningful calculation for BH, ignore arguments
+    
 };
 
 #endif // __BH_h__
--- a/src/BaseBinaryStar.cpp
+++ b/src/BaseBinaryStar.cpp
@@ -1591,7 +1591,7 @@ void BaseBinaryStar::ResolveCommonEnvelopeEvent() {
             
             // stage 1: convective envelope removal on a dynamical timescale; assumes lambda = lambda_He (this still uses the Picker convective envelope mass fit to estimate lambda)
             double lambda1    = m_Star1->CalculateConvectiveEnvelopeLambdaPicker(convectiveEnvelopeMass1, maxConvectiveEnvelopeMass1);
-            double lambda2    = m_Star1->CalculateConvectiveEnvelopeLambdaPicker(convectiveEnvelopeMass2, maxConvectiveEnvelopeMass2);
+            double lambda2    = m_Star2->CalculateConvectiveEnvelopeLambdaPicker(convectiveEnvelopeMass2, maxConvectiveEnvelopeMass2);
             
             double k1         = m_Star1->IsOneOf(COMPACT_OBJECTS) ? 0.0 : (2.0 / (lambda1 * alphaCE)) * m_Star1->Mass() * (mass1 - endOfFirstStageMass1) / m_Star1->Radius();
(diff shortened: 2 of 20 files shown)


jeffriley (changes requested):
@ilyamandel Compile warnings?  Also, why are the pytests failing?

ilyamandel:
Hi @jeffriley ,
I don't see any compile warnings when I compile on MacOS.  I use the standard Makefile and the following makeCOMPAS:

`make -j4 GSLINCDIR=/opt/homebrew/Cellar/gsl/2.7.1/include GSLLIBDIR=/opt/homebrew/Cellar/gsl/2.7.1/lib BOOSTINCDIR=/opt/homebrew/Cellar/boost/1.85.0/include BOOSTLIBDIR=/opt/homebrew/Cellar/boost/1.85.0/lib HDF5INCDIR=/opt/homebrew/Cellar/hdf5/1.14.3_1/include HDF5LIBDIR=/opt/homebrew/Cellar/hdf5/1.14.3_1/lib`

ilyamandel:
On the other hand, it's not just pytests failing, it's a segmentation fault when running on our default binary.  Yikes.   Definitely something to investigate!

jeffriley:
> I don't see any compile warnings when I compile on MacOS.

Hmmm.  The build on github shows compile warnings - it uses Linux and g++

jeffriley:
@ilyamandel Maybe what got pushed isn't what you thought got pushed...  Do you get the segmentation fault locally?

ilyamandel:
Yes, the github branch is the same as mine.
No, I can't reproduce the github compiler warnings locally, but I have a guess at what might fix them, so let me try that.  A bit annoying though to debug in the dark -- I wish I could get the same compiler warnings here.
Yes, I can reproduce a seg fault, which hadn't come up in my checks (different binary parameters).  Working on it...

ilyamandel:
Compilation warnings fixed, now working on seg fault

ilyamandel:
OK, seg fault diagnosed.  We are in an infinite loop when we call
CHeB::CalculateRadiusOnPhase(mass, luminosity).

This in turn calls CHeB::CalculateRadiusOnPhase(mass,luminosity,tau).
Which calls CHeB::CalculateRadiusAtBluePhaseStart(mass).
Which calls CalculateRadiusAtHeIgnition()
Which sits in GiantBranch.cpp, but calls CalculateRadiusOnPhase(mass, LHeI).
And the latter calls CHeB::CalculateRadiusOnPhase(mass, luminosity).

Now, this wasn't an issue before because the last step didn't happen until we made CalculateRadiusOnPhase(mass, luminosity) virtual -- in other words, when it was called from GiantBranch::CalculateRadiusAtHeIgnition(), that presumably resolved to the GiantBranch version of CalculateRadiusOnPhase(mass,luminosity), not the CHeB version, and all was well.  But now, with the virtual keyword added, we have a loop.

Now, it's easy to fix this: If I change the following line in GiantBranch::CalculateRadiusAtHeIgnition()
double RGB_LHeI  = CalculateRadiusOnPhase(p_Mass, LHeI);
to instead read
double RGB_LHeI  = GiantBranch::CalculateRadiusOnPhase(p_Mass, LHeI);
the infinite loop is gone and there's no seg fault.  

I just need to understand whether that's the desired radius calculation... ;-)

jeffriley:
> ... that presumably resolved to the GiantBranch version of CalculateRadiusOnPhase(mass,luminosity), not the CHeB version

Yes, it would have.  Since `GiantBranch::CalculateRadiusOnPhase(mass,luminosity)` was *not* `virtual` prior to this PR, any call to `CalculateRadiusOnPhase(mass,luminosity)` would resolve to the version in the source file from which it was called - in this case `GiantBranch`.  So changing the following line in `GiantBranch::CalculateRadiusAtHeIgnition()`:

    double RGB_LHeI = CalculateRadiusOnPhase(p_Mass, LHeI);

to instead read

    double RGB_LHeI = GiantBranch::CalculateRadiusOnPhase(p_Mass, LHeI);

will restore the previous behaviour.

jeffriley:
> I just need to understand whether that's the desired radius calculation... ;-)

Oh, well that's a different question from restoring the previous behaviour :-)

ilyamandel:
Indeed. :-)
I'll try to check a few more things before resubmitting this, just in case
-- may not happen until later.

jeffriley (approved):
All good - thanks @ilyamandel

Issue #1327, opened 2025-01-30
Mass transfer treatment of low mass donor star


yuzhesong:
There still seems to be issue in mass loss and mass transfer in some neutron star binaries. 

The attached system forms a neutron star + He white dwarf binary. There are a few issues that concerns the result: 

1. The companion star loses ~0.9 Msun in the time step when it forms the HeWD (seen in the first plot) and almost 0.2 Msun is accreted onto the neutron star (seen in the second plot). It seems the mass accretion onto the NS is too extreme in this case. 
2. The orbital separation of the binary increases significantly for this case when the companion star loses most of its mass to form the HeWD.

In this particular case we would expect the mass accretion onto the neutron star to be milder, with the neutron star less massive than that produced in the model. We'd also expect the final HeWD to be heavier than 0.1 Msun with less material ejected, so that the final binary separation to be smaller than it is in the current version. 

Looking at the donor star in single star evolution (third plot), the core mass initially jumps to some value around 0.2 Msun, then grows over a short time to ~0.6 Msun. When such a star is in a binary, we think that the treatment of mass transfer in COMPAS changes as soon as the star has that initial core mass though, such that the (envelope) mass transferred is too high, and the final HeWD mass ends up being too low (0.1 Msun v 0.6 Msun).

In trying to solve #1076 @ilyamandel has asked this: 

> Now, at the moment, we only have nuclear timescale MT enabled for MS donors -- we assume that evolved donors always lose their entire envelope on MT.
> Is that a reasonable assumption? 

We think that according to this result, the answer is no, this does not seem like a reasonable assumption, and it leads to unphysical behaviour. @SimonStevenson thinks that the issue is that the mass transfer proceeds on a similar timescale as the core mass grows. So we can't assume one is faster than the other.

`urgency_moderate` - This is a moderately urgent issue

`severity_moderate` - This is a moderately severe bug

Using COMPAS v3.10.05: 
The compasConfig_bin.txt can be used to produce the plot for the behaviour of the star in binary system; compasConfig_sin.txt for the behaviour of the star in single star. 

![Image](https://github.com/user-attachments/assets/1cb0fafd-970c-495c-9c9a-29b5e46e116c)

![Image](https://github.com/user-attachments/assets/f5aee05e-8aa4-486f-ab7c-1ec1edf67302)

![Image](https://github.com/user-attachments/assets/7369789c-aa59-435e-953d-e8a191e28dcb)

[...]

PR #1342, opened 2025-02-28
Fix to issue #1327: partial envelope removal on nuclear timescale MT from giants now enabled

ilyamandel:
See illustration plots in #1327


jeffriley (approved):
Looks good to me

SimonStevenson:
Awesome, thanks Ilya. This looks like it does exactly what I think it should. Happy to approve this PR. I will do some more testing of populations and try to reproduce the Porb-MWD relation from Tauris & Savonije.

Issue #24, opened 2019-11-29
Fixed time steps


ilyamandel:
Fixed-width time steps yield stripes on the HR diagram (exacerbated by dividing the fixed-width time steps by 10, as demonstrated by Alejandro below )

<img width="401" alt="image" src="https://user-images.githubusercontent.com/512213/69864748-b9b19f80-12f3-11ea-86f5-05497d7f4a9d.png">

The adaptive step sizing may fix this; to be tested.

ilyamandel:
Since new COMPAS does not use fixed timesteps, there is no reason to expect this to be an issue.  Closing in the absence of evidence that there is a problem here.

SimonStevenson:
I'm not sure I understand what you mean by "new COMPAS doesn't use fixed timesteps". The timestepping is fundamentally the same as in Legacy COMPAS (and in BSE), with suggested timesteps based on the evolutionary phase of the star(s), constrained by limits on changes to the properties of the star(s).  In practice, Jeff did fix a few bugs, but I'm not sure we know for certain whether this issue is solved.

jeffriley:
I'm not entirely sure what the issue is :-)  

It is true that I didn't change the mechanism for calculating the timestep other than fixing a couple of defects, but it also true that COMPAS doesn't use fixed timesteps - so I don't really know what @ilya/@alejandro mean when they refer to 'fixed-width timesteps'.

I'm also not sure I understand the plots, or in fact how they were generated - what does "dividing the fixed-width time steps by 10" actually mean?

ilyamandel:
OK, I was under the wrong impression that new COMPAS used adaptive time steps which we discussed at some point (decreasing the time step size if the change in one step is too large, and occasionally proposing time step size increases).  Sorry, not sure why I was so confused.  Anyway, re-opening the issue.

jeffriley:
Wait... 

COMPAS does vary the size of the timestep - it is calculated for binaries in BaseBinaryStar::CalculateDt() and for single stars in BaseStar::CalculateTimestep() (and updated in places for mass loss I think).

So I still don't understand what 'fixed-width timesteps' means.  Or how @avigna divided the timestep by 10 (was the problem plot produced by changing the code or by printing every 10th value or something - either of which might be problematic)?

avigna:
[...]

PR #1344, opened 2025-02-28
Updates to improve convergence without sacrificing computational speed

ilyamandel:
The computational cost has been reduced by about 25% even though the default version now aims for a fractional mass change below 0.001 per time step, a stricter threshold by a factor of 10 than previously.

Note: I'd like to do a few more tests, but am going away for the weekend, so please review without merging this in for now.

### DIFF
--- a/online-docs/pages/User guide/COMPAS output/standard-logfiles-record-specification-binary.rst
+++ b/online-docs/pages/User guide/COMPAS output/standard-logfiles-record-specification-binary.rst
@@ -1320,7 +1320,7 @@ but not both. If both are printed then the file will contain two columns with th
    * - Data type:
      - DOUBLE
    * - COMPAS variable:
-     - Calculated using BinaryConstituentStar::m_StarToRocheLobeRadiusRatio
+     - Calculated using BinaryConstituentStar::StarToRocheLobeRadiusRatio()
    * - Description:
      - Ratio of the primary star’s stellar radius to Roche radius (R/RL), evaluated at periapsis.
    * - Header String:
@@ -1336,7 +1336,7 @@ but not both. If both are printed then the file will contain two columns with th
    * - Data type:
      - DOUBLE
    * - COMPAS variable:
-     - Calculated using BinaryConstituentStar::m_StarToRocheLobeRadiusRatio
+     - Calculated using BinaryConstituentStar::StarToRocheLobeRadiusRatio()
    * - Description:
      - Ratio of the secondary star’s stellar radius to Roche radius (R/RL), evaluated at periapsis.
    * - Header String:
--- a/online-docs/pages/whats-new.rst
+++ b/online-docs/pages/whats-new.rst
@@ -3,6 +3,12 @@ What's new
 
 Following is a brief list of important updates to the COMPAS code.  A complete record of changes can be found in the file ``changelog.h``.
 
+**03.14.00 Mar 3, 2025**
+
+* Updates to improve convergence without sacrificing computational speed
+* Capped total wind mass loss rate at MAXIMUM_WIND_MASS_LOSS_RATE (set to 0.1 Msol/yr) for all prescriptions
+* Changed order of calls in SSE evolution to better match BSE evolution
+
 **03.13.02 Feb 19, 2025**
 
 * Replaced name of ``SHIKAUCHI`` main sequence core mass prescription with ``BRCEK``.
--- a/online-docs/pages/whats-new.rst
+++ b/online-docs/pages/whats-new.rst
@@ -5,7 +5,8 @@ Following is a brief list of important updates to the COMPAS code.  A complete r
 
 **03.14.00 Mar 3, 2025**
 
-* Updates to improve convergence without sacrificing computational speed
+* Updates to improve convergence without sacrificing computational speed, including updates to default mass and radial change fractions 
+per time step and their usage
 * Capped total wind mass loss rate at MAXIMUM_WIND_MASS_LOSS_RATE (set to 0.1 Msol/yr) for all prescriptions
 * Changed order of calls in SSE evolution to better match BSE evolution
 
--- a/src/BaseBinaryStar.cpp
+++ b/src/BaseBinaryStar.cpp
@@ -1929,9 +1929,10 @@ double BaseBinaryStar::CalculateZetaRocheLobe(const double p_jLoss, const double
  *
  * JR: todo: flesh-out this documentation
  *
+ * 
+ * void CalculateWindsMassLoss(double p_Dt)
  *
  * @param   [IN]    p_Dt                      Time step
- * void CalculateWindsMassLoss(double p_Dt)
  */
 void BaseBinaryStar::CalculateWindsMassLoss(double p_Dt) {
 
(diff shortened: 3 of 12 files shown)


SimonStevenson (changes requested):
This isn't looking so good at the moment. 

remnant_mass_relation

Here is the command I'm using to test
```
./COMPAS -n 1 --mode SSE --initial-mass r[1.0,298,0.5] --random-seed 0 --remnant-mass-prescription FRYER2012
```
[Using FRYER2012 just to avoid randomness in the initial-final mass relation]

ilyamandel:
Thanks, Simon -- will work on this further on Monday, away for the rest of
the weekend.

ilyamandel:
OK, the immediate problem is that when LBV winds turn on, since we are using time step guesses based on the mass and radius change from the previous timestep (without LBV winds, when things were changing slowly and large time steps were allowed), we immediately, in one overly large time step with extreme mass loss, deplete most or all of the star. 

Now that the problem is clear, time to think how to address it...  Thinking...

ilyamandel:
OK, almost there.  There were a couple of issues.  

@jeffriley : One was that the order of operations in SSE and BSE was different.  BSE computes wind mass loss first, then evolves the stars (changing their radii and luminosities).  This makes it possible to estimate beforehand how much mass the stars will lose in winds and choose the time step appropriately, to ensure that not too high a fraction of mass is lost.  SSE actually did it in the opposite order, first evolving the stars (changing radii and luminosities), then computing mass loss rates.  That makes it impossible to account in advance for whether stars will experience LBV winds and pick the timestep appropriately (if LBV winds suddenly turn on because the star crosses the Humphries-Davidson limit, we need to take much shorter time steps to adequately resolve the mass loss rate).  It also means that stars in wide binaries may evolve differently from SSE stars.  I changed SSE to follow BSE order, which means I can precompute the time step required to reasonably calculate winds.

@SimonStevenson, @ryosuke-hirai  : The second issue is that LBV winds are *really* strong.  They can reach a couple of solar masses per year in the default prescription.  This seems unphysical.  It also makes it impossible to resolve such winds since our default minimal timestep is 1 year: we always lose more than the desired fraction of mass per time step.  I've capped LBV winds to 0.1 solar masses per year in all LBV prescriptions.

ilyamandel:
- Capped total wind mass loss rate at MAXIMUM_WIND_MASS_LOSS_RATE (set to 0.1 Msol/yr) for all prescriptions
- Changed order of calls to stellar evolution and wind mass loss in SSE to match BSE
- Updated documentation

More testing to come tomorrow (so feel free to review, but don't merge yet)

ilyamandel:
Not there yet...

ilyamandel:
Ready for review, but still doing some further testing

ilyamandel:
SSE seems to be converged.   I am comparing pre-SN total mass as a function of ZAMS mass for two runs on which one has time steps decreased by an initial factor of 10.  Other than one time point on the very edge of an extreme winds change (to be expected), the behaviour is within acceptable margins.

The runs being compared are
```
./COMPAS --mode SSE --initial-mass r[1.0,298,0.5] --random-seed 0 
./COMPAS --mode SSE --initial-mass r[1.0,298,0.5] --random-seed 0 --timestep-multiplier 0.1
```

image

ilyamandel:
I checked with BSE, intentionally taking a very wide binary to avoid mass transfer, so results should be essentially identical to those above:

```
./COMPAS --initial-mass-1 r[1.0,298,0.5] --initial-mass-2 1 -a 100 --random-seed 0
./COMPAS --initial-mass-1 r[1.0,298,0.5] --initial-mass-2 1 -a 100 --random-seed 0 --timestep-multiplier 0.1
```
 
 They are:
image

SimonStevenson (approved):
Thanks for the changes Ilya. I've made a remnant mass relation and plotted a few HR diagrams and everything looked sensible. Perhaps as expected, the results are considerably different to those obtained using the previous time stepping.

jeffriley (changes requested):
Placeholder - changes coming...

jmerritt1:
I'm currently testing with settings relevant to compare with Figure 7, will review soon. Thanks again for the prompt fix.

jmerritt1:
image

image

image

In addition to significantly changing remnant masses, and in some cases remnant types, this (Times branch) has also removed the unexplained "chatter" present in the last plot (v3.07 with change-fractions 0.005)

jeffriley:
@ilyamandel 

Looks good (pending testing).  I just pushed some minor formatting changes.

Some questions in the code (please remove once resolved):

BaseStar::ResolveMassLoss() - see question at BaseStar.cpp line 2808
BaseStar::CalculateMassAccretedForCO() - see question in function description (BaseStar.cpp line 2936)
BaseStar::CalculateTimestep() - see question at BaseStar.cpp line 4355

And a request:

changelog and whatsnew - you also changed the mass and radial change fractions - I think you should note that.

EDIT: Actually, ignore the second question for `BaseStar::CalculateMassAccretedForCO()` - I don't think we can/should move it.  And maybe the name isn't as bad as I thought when I first wrote the question (a while ago) - we could probably leave it.

jmerritt1 (approved):
Changes look good. Replicated the initial vs. final mass plots (Figure 7). 

Edit: I also generated HR Diagrams, where despite a 10x more lenient maximum radial-change-fraction, points on HG, and as LBV, are now less sparse. Rightward incursions into the LBV regime are also significantly limited, which is a positive side effect? Besides final masses, there are no other major changes that I'm noticing here: 
image

and v3.07 for comparison:
image

ilyamandel:
Thanks for the test, @jmerritt1 !

Is it expected that PESSIMISTIC would yield larger remnant masses than MERRITT2024 for high-mass progenitors at some metallicities?  Is this due to VMS winds (Sabhahit vs Bestenlehner)?

ilyamandel:
@jeffriley ,
Thank you very much for your review!
A lot of the cosmetic changes you made are to code I hadn't touched in this PR, but they look OK.
Regarding your questions:
BaseStar::ResolveMassLoss() - see question at BaseStar.cpp line 2808
- I believe the order is correct in the one place where it matters, for HG stars.
BaseStar::CalculateMassAccretedForCO() - see question in function description (BaseStar.cpp line 2936)
- I think you are correct that it could be moved, but it's not really connected to this PR, so I'd prefer not to have to repeat the various tests.  I'll leave your comment to be considered later (along with some of the other *Ilya* remarks in the code that I haven't yet addressed).
BaseStar::CalculateTimestep() - see question at BaseStar.cpp line 4355
- If a timescale is negative or zero, that means something went wrong in that calculation, and we should not be using it to set the step size.  So that choice was intentional, and no changes are needed.

I'll update the changelog and whatsnew shortly.

jeffriley (approved):
All good - thanks @ilyamandel

ilyamandel:
BSE convergence tests show small differences (typically at few percent level in terms of various outcomes, such as number and types of SNe, DCOs formed, etc.).  This means there is still work to be done on convergence test, so I will not close issue #24 yet.  However, even though convergence is not yet perfect, @SimonStevenson , @jeffriley and I agreed that things are now looking sufficiently good to merge this in.

jmerritt1:
> Is it expected that PESSIMISTIC would yield larger remnant masses than MERRITT2024 for high-mass progenitors at some metallicities? Is this due to VMS winds (Sabhahit vs Bestenlehner)?

Yes, despite being called PESSIMISTIC, Bestenlehner gives less mass loss at high Z, and this must be responsible for the jump above 100 Msol. While we overhaul the results and discussion in the paper, we can also decide if this is, in fact the best set of prescriptions to choose. 

You mentioned adding PESSIMISTIC to COMPAS as a `mass-loss-prescription` option, we can certainly do this, if users would find it useful to override the four on-phase options.

Issue #1345, opened 2025-03-08
Divide-by-zero in BaseStar::CalculateTimestep()


jeffriley:
**Describe the bug**
Divide-by-zero in `BaseStar::CalculateTimestep()`.

In `BaseStar::CalculateTimestep()` the following code sometimes results in a divide-by-zero:

    // there is a chance that mass loss from winds is much faster than previously estimated if, say, LBV winds have turned on
    // we therefore precompute the mass loss rate to avoid taking an overly long timestep, despite the extra computational costs
    double massChangeWinds = m_Mass - CalculateMassLossValues(dt, false);
    dt = min(dt, OPTIONS->MassChangeFraction() * (dt * m_Mass / massChangeWinds));    

The problem occurs in the instance I first noticed because one of the stars is a BH and `BaseStar::CalculateMassLossValues()` returns an unchanged mass because `CalculateMassLossRate()` returns `0.0` for BHs, but note that the problem will also occur if `OPTIONS->UseMassLoss()` is FALSE in `BaseStar::CalculateMassLossValues()`.

Maybe a sanity check here for 0.0 mass loss?

Run COMPAS v03.14.00 with:

    ./compas --random-seed 1741412874 -n 1

to see the problem when a BH is involved.

 - OS: Ubuntu 22.04.2 LTS
 - COMPAS v03.14.00

@ilyamandel I think this is a side-effect of your recent changes, so assigning to you.

ilyamandel:
Thanks for the catch, @jeffriley !
Resolved in #1346 .

PR #1346, opened 2025-03-08
Added a check to prevent a divide-by-zero error from the previous PR

ilyamandel:
Resolves issue #1345


jeffriley (approved):
Thanks @ilyamandel

Issue #1348, opened 2025-03-13
Divide-by-zero in EAGB::CalculateTimescales()


jeffriley:
**Describe the bug**
Running COMPAS v03.15.00 with:

    ./compas -n 1 --random-seed 0

results in a divide-by-zero error in `EAGB::CalculateTimescales()`, with the following stack trace:
 
    EAGB::CalculateTimescales(double, std::vector<double, std::allocator<double> >&)
    TPAGB::CalculateTimescales(double, std::vector<double, std::allocator<double> >&)
    Remnants::CalculateTimescales(double, std::vector<double, std::allocator<double> >&)
    Remnants::CalculateTimescales()
    BaseStar::UpdateAttributesAndAgeOneTimestepPreamble(double, double, double)
    BaseStar::UpdateAttributesAndAgeOneTimestep(double, double, double, bool, bool)
    Star::UpdateAttributesAndAgeOneTimestep(double, double, double, bool, bool)
    Star::AgeOneTimestep(double, bool)
    BaseBinaryStar::EvolveOneTimestep(double)
    BaseBinaryStar::Evolve()
    BinaryStar::Evolve()
    EvolveBinaryStars()
    main

The line of code in `EAGB::CalculateTimescales()` at which the divide-by-zero occurs is:

    timescales(tinf1_FAGB) = tBAGB + ((1.0 / (p1 * gbParams(AHe) * gbParams(D))) * PPOW((gbParams(D) / LBAGB), p1_p));

The problem occurs because the variable `LBAGB` is 0.0, having been calculated at this line in `EAGB::CalculateTimescales()`: 

    double LBAGB = CalculateLuminosityAtBAGB(p_Mass);

because `p_Mass` is passed to `EAGB::CalculateTimescales()` as 0.0.

The short explanation is that at some stage during the evolution of the binary, star 1 becomes a neutron star, and in the course of evolution we call `BaseStar::UpdateAttributesAndAgeOneTimestep()` which calls `CalculateTimescales()` which for the NS calls `Remnants::CalculateTimescales(`), which in turn calls `TPAGB:CalculateTimescales()` with `m_Mass0` as the mass parameter, which calls `EAGB::CalculateTimescales()` with `p_mass` = 0.0, leading to the divide-by-zero error.

[...]

PR #1349, opened 2025-03-14
Fix to issue #1348

ilyamandel:


ilyamandel:
It's meaningless to compute GB timescales for NSs and BHs -- so I turned off that computation. :)

jeffriley (approved):
Works for me - thanks @ilyamandel

PR #1352, opened 2025-03-15
added orbital AM and system velocity vectors to SN output

### LINKED ISSUE #1289: Output current orbital angular momentum direction in original x-y-z coordinates
The solution to #1243 provides the velocity in the original pre-first SN orbital plane coordinates -- but we also need the orbit in the same coordinate system in order to compare the alignment between the two.

reinhold-willcox:


ilyamandel (changes requested):
Hi @reinhold-willcox ,
The code changes look good, but could you please update the documentation?  I believe you need to remove the orbital inclination from the list of outputs and add the orbital AM vector -- see https://compas.readthedocs.io/en/latest/pages/User%20guide/COMPAS%20output/standard-logfiles-record-specification-binary.html .  You may also want to add a mention to whatsnew.

reinhold-willcox:
Done

ilyamandel (approved):
Hi @reinhold-willcox ,

In comments like these, does "birth reference frame" refer to the original orientation of the binary at birth, or the reference frame immediately before the SN?  This is relevant for the case of a second SN in the binary.  Could you please clarify?

> x-component of the normalized orbital AM unit vector, calculated after a SN in the birth reference frame (with z parallel to preSN orbital AM, and x parallel to the LRL vector)

PR #1353, opened 2025-03-15
Tides Updates

veome22:
This PR has one major improvement to the behavior of `KAPIL2024` tides, which should avoid making drastic changes to the binary due to a misguided time step. The rest are minor changes. Here is a summary:

1. If too large of a time step is taken for any reason [e.g. if the stellar type changed after ChooseTimestep() but before ProcessTides()], force `KAPIL2024` tides to only take the largest possible time step such that the spins, semi-major axis, and eccentricity change by at most `TIDES_MAXIMUM_ORBITAL_CHANGE_FRAC` of their current values.
2. When computing the Dynamical component of `KAPIL2024` tides from a radiative core + convective envelope boundary, use only the `CalculateConvectiveCoreRadius()` value to infer the absence of a convective core. This change is in response to a test I did where the convective core for a ~ $1 M_\odot$ star where the convective core radius was zero as expected, but the mass was non-zero.
3. Added two new log outputs: `Radius_Core@CO` and `Radius_Total@CO`. These are meant to help with estimates of spins, angular momenta, etc. at compact object formation.
4. Minor typo fix to `TIDES_MINIMUM_FRACTIONAL_NUCLEAR_TIME`.
5. An update to the online docs, now that `KAPIL2024` tides are compatible with CHE.


ilyamandel (changes requested):
Thank you, @veome22 !

A few comments:

1.  You are not really limiting the timestep, because other things are still proceeding on the initially selected timestep.  You are actually limiting the maximal changes to the system properties due to tides.  I think it's useful to describe things in those terms to avoid confusion.  Personally, I'd create a variable called something like 
double fraction_tidal_change = 1.0;
and reduce it if need be (more on that below), then apply all tidal changes as
m_Star2->SetOmega(m_Star2->Omega() + fraction_tidal_change*(DOmega2Dt * p_Dt * MYR_TO_YEAR));
m_Eccentricity = m_Eccentricity + fraction_tidal_change*((DEccentricity1Dt + DEccentricity2Dt) * p_Dt * MYR_TO_YEAR);
etc.

2. I don't understand the comparisons in 
utils::Compare(DOmega1Dt * tides_Dt, TIDES_MAXIMUM_ORBITAL_CHANGE_FRAC * OrbitalAngularVelocity())
Why is the rate of tidal spin up of a star being compared to the fractional change in the orbital velocity?  Maybe you are trying to reuse a constant, but I think it would be cleaner to define a new constant even if it has the same value -- say TIDES_MAXIMUM_ROTATION_CHANGE_FRAC.

3. Then you might write something like the following:
 fraction_tidal_change = std::min(fraction_tidal_change, (TIDES_MAXIMUM_ROTATION_CHANGE_FRAC*m_Star1->Omega() ) / (DOmega1Dt*p_Dt * MYR_TO_YEAR) );
  fraction_tidal_change = std::min(fraction_tidal_change, (TIDES_MAXIMUM_ROTATION_CHANGE_FRAC*m_Star2->Omega() ) / (DOmega2Dt*p_Dt * MYR_TO_YEAR) );
   fraction_tidal_change = std::min(fraction_tidal_change, (TIDES_MAXIMUM_ORBITAL_CHANGE_FRAC*m_SemiMajorAxis ) / ((DSemiMajorAxis1Dt + DSemiMajorAxis2Dt) *p_Dt * MYR_TO_YEAR) );
etc.
[You can, of course, combine the mins into a single line]
   
3. Please document the new outputs in the online documentation pages.

veome22:
Hi @ilyamandel , thanks for going through the PR!

1. I think your suggestion makes sense. I will introduce a modifier for the tidal strength to avoid confusion.
2. The comparison to `OrbitalAngularVelocity()` instead of stellar spins is actually intentional. If the stellar spins are very small compared to the orbital frequency (say, at the beginning of evolution), we want to be able to take larger time steps until we approach synchronization. It doesn't matter if the star spins up quickly until it has the possibility of over-spinning and exceeding the synchronization rate (or shooting too far below synchronization when spinning down). If you are happy with this logic, I won't define any other constants for the rotational limit.

ilyamandel:
> The comparison to OrbitalAngularVelocity() instead of stellar spins is actually intentional. If the stellar spins are very small compared to the orbital frequency (say, at the beginning of evolution), we want to be able to take larger time steps until we approach synchronization.  

Conversely, you'll end up making only very small fractional changes if the stars are rotating significantly super-synchronously -- but if that's intentional, I am OK with that. :)

ilyamandel (approved):
Changes look good -- thanks, @veome22 !

PR #1355, opened 2025-03-16
Reinstate proper inheritance for CalculateTimescales()

jeffriley:
Defect repair
- Added prototype for CalculateTimescales(const double p_Mass, DBL_VECTOR &p_Timescales) to NS.h to reinstate proper inheritance (compiler warning -Woverloaded-virtual; introduced in v03.15.01)
- Removed call to CalculateTimescales() from both NS::Initialise() and BH::Initialise() (superfluous since v03.15.01)

### DIFF
--- a/src/BH.h
+++ b/src/BH.h
@@ -46,7 +46,7 @@ public:
 protected:
     
     void Initialise() {
-        CalculateTimescales();                                                                                                                          // Initialise timescales
+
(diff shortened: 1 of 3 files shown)


ilyamandel (approved):
Thanks, @jeffriley , and sorry for not catching the compiler warning (did not come up on MacOS).

PR #1356, opened 2025-03-17
Added sensitivity as an option to detection_matrix.py

SimonStevenson:
This PR should resolve an issue that @jmerritt1 has been having using detection_matrix.py to compute detection rates. The issue was that he was inadvertently using 'O1' sensitivity instead of the desired sensitivity ('O3'). This PR add the sensitivity as an option. 

@jmerritt1 can you try this PR and let me know if it works for you. You should be able to modify `sens` in the line below and it should work.

```
detection_matrix = DetectionMatrix.from_compas_output(
        compas_fname,  save_plots=False,
        chirp_mass_bins=50, redshift_bins=100,
        cosmological_parameters=dict(aSF=0.01, dSF=4.70, mu_z=-.23, sigma_z=0), sens='O1' # change me to 'O3'
    )
```


jmerritt1:
Thanks a ton, this is now actually using O3 sensitivity, with rates going up by a factor of ~5. 
image

jmerritt1 (approved):
Adds sensitivity option rather than having to change the script, functions as intended. Thanks for the catch.

PR #1358, opened 2025-03-19
Removed deprecated mass loss prescriptions

SimonStevenson:
This should resolve issue #1357 . I have removed the deprecated mass loss options (and added a new ZERO mass loss option for WR stars for consistency with other mass loss types).

PR #1359, opened 2025-03-20
Update docs - remove NONE mass loss options

SimonStevenson:
Forgot to update the docs in #1358 . This PR should update the docs to match the code (remove the NONE mass loss prescription options, and add ZERO to the list of WR-mass-loss-prescription options).


ilyamandel (approved):
Thank you, @SimonStevenson !

Issue #1354, opened 2025-03-16
BaseStar::CalculateLogBindingEnergyLoveridge() returns log(BindingEnergy) ~+/-1E10


jeffriley:
`BaseStar::CalculateLogBindingEnergyLoveridge()` sometimes returns log(BindingEnergy) ~+/-1E10.  This causes overflows when raised to power 10 (in `BaseStar::CalculateLambdaLoveridgeEnergyFormalism()`).  Seems to only happen when radius is ~1E-3 or lower (seems a little low...).

**To Reproduce**

Instrument `BaseStar,cpp` to print `m_Mass`, `m_Radius`, and `logBindingEnergy ` just before the return statement in `BaseStar::CalculateLogBindingEnergyLoveridge()`, and run:

    ./compas -n 1 --random-seed 14
    
    COMPAS v03.15.00
    Compact Object Mergers: Population Astrophysics and Statistics
    by Team COMPAS (http://compas.science/index.html)
    A binary star simulator
    
    Go to https://compas.readthedocs.io/en/latest/index.html for the online documentation
    Check https://compas.readthedocs.io/en/latest/pages/whats-new.html to see what's new in the latest release
    
    Start generating binaries at Sun Mar 16 14:46:41 2025
    
    m_Mass = 8.713363708228811e-01, m_Radius = 9.499999408351632e-03, logBindingEnergy = 4.817894008594936e+10
    m_Mass = 8.713363708228811e-01, m_Radius = 9.499999408351632e-03, logBindingEnergy = 5.700361114132716e+10
    m_Mass = 3.566165023247284e+00, m_Radius = 1.806501687910545e+02, logBindingEnergy = 4.731090851497682e+01
    m_Mass = 3.566165023247284e+00, m_Radius = 1.806501687910545e+02, logBindingEnergy = 4.737776185956845e+01
    
    0: Stars merged: (Main_Sequence_>_0.7 -> Carbon-Oxygen_White_Dwarf) + (Main_Sequence_>_0.7 -> Naked_Helium_Star_Hertzsprung_Gap)
    
    Generated 1 of 1 binaries requested
    
    Simulation completed
    
    End generating binaries at Sun Mar 16 14:46:41 2025
    
    Clock time = 0.048837 CPU seconds
    Wall time  = 0000:00:00 (hhhh:mm:ss)

and

    ./compas -n 1 --random-seed 5
    
    COMPAS v03.15.00
[...]

Issue #1311, opened 2024-12-11
What to do on super-critical accretion-induced spin up


ilyamandel:
The current treatment of accretion induced spin-up can cause a star to rotate at a super-critical velocity.  That is unphysical.  I would propose to cap the maximum rotation to Keplerian, with two options that both explicitly conserve angular momentum:
(a) the excess angular momentum is immediately deposited into the orbit (assuming an efficient feedback mechanism);
or
(b) accretion becomes non-conservative once the accretor reaches critical rotation.  

Thoughts welcome; tagging folks for discussion rather than implementation.

themikelau:
I think we should have both, with (b) being the default as it matches what is commonly assumed in detailed binary calculations. (a) could be justified by the Paczynski (1991) disk model, where angular momentum is transported from the star to the outer edge of an accretion disk, and returned to the orbit via tides.

lee0517-hub:
Regarding (b), two possible implementation approaches might be considered:

The first approach is similar to how the BPASS code (Eldridge et al. 2017) handles accretion-induced spin-up. In this case, the accretor reaches critical rotation and stops accreting mass once the accreted mass exceeds 5% of its initial mass (Eldridge et al. 2011). Recently, the SEVN code has used this condition to determine whether an accretor can CHE([Dall’Amico et al. 2025](https://ui.adsabs.harvard.edu/abs/2025arXiv250104778D/abstract)).

The second approach involves applying a reduction factor to the accretion rate for rotating stars, defined as 1 − Ω/Ωcrit.  This ensures that stars rotating at Ωcrit will no longer accrete mass (e.g., [Shao & Li 2014](https://iopscience.iop.org/article/10.1088/0004-637X/796/1/37)). Additionally, it can be assumed that the remaining mass is ejected from the binary system in the form of an isotropic wind, carrying away the corresponding orbital angular momentum (this should help mitigate the orbital widening caused by mass transfer).

On another note, I have a potentially related question. In a very simple model of accretion-induced CHE that I recently worked on (https://github.com/TeamCOMPAS/COMPAS/compare/dev...lee0517-hub:acc-CHE?expand=1), it was found that COMPAS's mass transfer scheme seems to easily cause the accretor to exceed critical rotation because the secondary star receives too much mass at once. Perhaps limiting the mass transfer process by the critical rotation of the accretor could resolve this.

Finally, regarding the second question @ilyamandel mentioned in email:

[...]

Issue #1308, opened 2024-12-10
Incorrect response to mass transfer when stellar rotation is taken into account


ilyamandel:
The recent implementation of accretor tidal spin up and donor tidal spindown violates the conservation of angular momentum, because I just added extra angular momentum to the accretor and removed angular momentum from the donor without accounting for a corresponding change in the angular momentum of the binary.

Basically, we are implementing the point-particle equation for binary separation evolution on mass transfer -- but, of course, if we are adding rotational angular momentum to the accretor and removing rotational angular momentum from the donor, we need to also account for those.

ilyamandel:
Fixed in #1360, so closing, but two things to note:
1. On stable mass transfer, I correct the orbital separation for the angular momentum lost/gained by stars *after* finding a root in which the donor just fits into its Roche lobe (too messy otherwise); that means it may not fit the Roche lobe exact once this correction is applied
2. I do not apply this to mass loss in common envelopes.  The alpha-lambda prescription is not intended to conserve angular momentum, anyway (we just assume that the right amount of angular momentum is carried away by the lost mass), and I wasn’t sure whether we should consider the rotational angular momentum lost by the donor in the second phase of the 2-stage prescription (if we do, we would need a fit for the moment of inertia of the radiative intershell, not just its mass and radius).  @ryosuke-hirai , what do you think?

Issue #1253, opened 2024-10-24
Add new convective envelope criterion


ilyamandel:
Use Picker+ convective CE mass prescription and 10% of envelope being convective threshold from Romagnolo+.

ilyamandel:
@SimonStevenson -- leaving this for you unless you tell me otherwise ;-)

ilyamandel:
Added ENVELOPE_STATE_PRESCRIPTION::CONVECTIVE_MASS_FRACTION (default threshold of convective envelope by mass to label envelope convective is 0.1, can be set with --convective-envelope-mass-threshold) in #1360 , closing.

PR #1360, opened 2025-03-22
Accretion, etc.

ilyamandel:
- Changed CalculateLambdaLoveridge() to return 1.0 for non-giant-branch stars to avoid meaningless results where the formalism is ill defined (resolves issue #1354)
- All binding energies and lambdas are now computed on request, not stored  in memory
- Added ENVELOPE_STATE_PRESCRIPTION::CONVECTIVE_MASS_FRACTION (default threshold of convective envelope by mass to label envelope convective is 0.1, can be set with --convective-envelope-mass-threshold), resolves issue #1253
- Stable mass transfer now conserves angular momentum after accounting for the rotational angular momentum lost or gained by the stars (resolves issue #1308)
- Imposed Keplerian rotation limit on mass-gaining stars (resolves issue #1311):
- Response depends on new --response-to-spin-up option; default (KEPLERIAN_LIMIT) forces mass transfer to become non-conservative once star (approximately) reaches super-critical rotation
- alternatively, with TRANSFER_TO_ORBIT variation, the star continues to accrete, but excess angular momentum is deposited in the orbit


ilyamandel:
Marked as draft until it's been tested -- but feel free to comment in the meantime :)

SimonStevenson (changes requested):
Starting review. More to come.

Running `./COMPAS --help' gives:
```
--response-to-spin-up
  Response to spin-up prescription (Options: [], default = 'KEPLERIAN_LIMIT')
```
Should print allowed options.

SimonStevenson:
Do either `--response-to-spin-up KEPLERIAN_LIMIT ` or `TRANSFER_TO_ORBIT` recover the previous behaviour, or is this a non-backwards compatible change?

SimonStevenson:
Plots comparing the two models for a BBH progenitor. Commands used:
```
./COMPAS -n 1 --response-to-spin-up KEPLERIAN_LIMIT --initial-mass-1 35.0 --initial-mass-2 30.0 --orbital-period 500.0 --detailed-output TRUE

python /Users/spstevenson/Github/COMPAS_review_accretion/COMPAS/compas_python_utils/detailed_evolution_plotter/plot_detailed_evolution.py ./COMPAS_Output_1/Detailed_Output/BSE_Detailed_Output_0.h5 --dont-show --outdir ./COMPAS_Output_1
```
detailedEvolutionPlotKEPLERIAN

and
```
./COMPAS -n 1 --response-to-spin-up TRANSFER_TO_ORBIT --initial-mass-1 35.0 --initial-mass-2 30.0 --orbital-period 500.0 --detailed-output TRUE

python /Users/spstevenson/Github/COMPAS_review_accretion/COMPAS/compas_python_utils/detailed_evolution_plotter/plot_detailed_evolution.py ./COMPAS_Output_2/Detailed_Output/BSE_Detailed_Output_0.h5 --dont-show --outdir ./COMPAS_Output_2
```
detailedEvolutionPlotTRANSFERTOORBIT

SimonStevenson:
For the convective envelope prescription, I got an error:
```
./COMPAS -n 1 --envelope-state-prescription CONVECTIVE_MASS_FRACTION --initial-mass-1 35.0 --initial-mass-2 30.0 --orbital-period 500.0 --detailed-output TRUE --convective-envelope-mass-threshold 0.1 --random-seed 0
```
yields
```
ERROR:  in Hertzsprung_Gap::virtual ENVELOPE HG::DetermineEnvelopeType() const: Unknown envelope state prescription, (ObjectId = 6)
0: Evolution stopped because an error occurred: (Main_Sequence_>_0.7 -> Hertzsprung_Gap) + (Main_Sequence_>_0.7 -> Main_Sequence_>_0.7)
```

SimonStevenson:
I ran a small population test using the KEPLERIAN_LIMIT response (command below) and the resulting BBH distribution is a bit funky

COMPAS_Output h5Rate_Infomu00 035_muz-0 23_alpha0 0_sigma00 39_sigmaz0 0

```
./COMPAS -n 100000 --response-to-spin-up KEPLERIAN_LIMIT --metallicity-distribution LOGUNIFORM --chemically-homogeneous-evolution-mode NONE --pulsational-pair-instability-prescription HENDRIKS
```

ilyamandel:
Hi @SimonStevenson ,
Thank you for the tests!  Definitely more things for me to do.  One quick question in the meantime:

> Do either --response-to-spin-up KEPLERIAN_LIMIT  or TRANSFER_TO_ORBIT recover the previous behaviour, or is this a non-backwards compatible change?

I agree that there should be a backwards compatible option; happy to add one, NO_LIMIT, which will simply allow the star to reach arbitrary super-Keplerian spins.  I will also make TRANSFER_TO_ORBIT the default, because it will allow for continued accretion (like NO_LIMIT, so similar to previous behaviour) but without the unphysical behaviour of arbitrarily high spins.  Would you be OK with that?

SimonStevenson:
> I agree that there should be a backwards compatible option; happy to add one, NO_LIMIT, which will simply allow the star to reach arbitrary super-Keplerian spins. I will also make TRANSFER_TO_ORBIT the default, because it will allow for continued accretion (like NO_LIMIT, so similar to previous behaviour) but without the unphysical behaviour of arbitrarily high spins. Would you be OK with that?

This sounds good to me! It's useful to have a backwards compatible option (for comparison, testing etc), even if we discourage using it for science.

veome22 (commented):
Hi @ilyamandel,

On the subject of making `TRANSFER_TO_ORBIT` the default option, would this be philosophically inconsistent with the default tidal prescription of `NONE`? I am nervous about the physical interpretation of transferring angular momentum into the orbit from a super-critically rotating star without a tidal mechanism to do so. Is there some other reason we would expect this orbital change to happen (apart from tidal interactions)?

ilyamandel:
OK, there were two bugs.  One was an incomplete fix to Loveridge lambda calculations (now, hopefully, properly fixed); the other was that, in some cases, the stellar masses were not updated correctly on mass loss (introduced inadvertently when limiting angular momentum accretion in this PR).  Both of these are now fixed.  Now looking into the points raised above by @SimonStevenson .

ilyamandel:
OK, I think this solves all of the issues noted above by @SimonStevenson as well as the ones I was aware of, so it's ready for review.

@jeffriley : there's one line in the auto-created yaml file which I can't figure out how to fix, maybe you can spot where I made a mistake:
`#    --envelope-state-prescription: 'LEGACY'                               # Default: ''                            # Options: ['CONVECTIVE...
`
Why is the default showing up as empty here, even though it's correctly listed as LEGACY just to the left?  I must be missing something somewhere, but can't see what it is...

@veome22 : the idea behind TRANSFER_TO_ORBIT is that mass accretion can continue even once the accretor's rotation reaches the critical value, but there's no further spin-up of the accretor and total angular momentum is correctly conserved.  Popham & Narayan (1991, https://ui.adsabs.harvard.edu/abs/1991ApJ...370..604P) suggest this could be due to efficient outward angular momentum transport in the accretion disk.  While the physics are indeed related to the physics behind tides, this doesn't directly concern stellar tides, so I think it's OK to combine this option with the NONE option for tides if the user wishes to do so.

Marking this ready for review.

SimonStevenson (approved):
Thanks for the changes Ilya. I have rerun the tests I performed previously and it seems to have addressed the issues I found, so approving.

veome22:
@ilyamandel I'm still running a few small tests, but I wanted to quickly flag that last I checked, the `--response-to-spin-up KEPLERIAN_LIMIT` option does not conserve angular momentum during stable mass transfer. I understand that this option is meant to not modify the orbit directly, but just since angular momentum is lost differently than accreted, the secondary always gains more angular momentum than the primary loses during a primary->secondary SMT episode.

On a practical note, the angular momentum of a non-rotating primary would be trivially zero, which is true for the default configuration of COMPAS (without tidal rotation). In this case, I am not sure how to even define the angular momentum conservation constraint.

Do you think the current recipe for angular momentum change in donor vs accretor stars is okay? 

The binary I tested was:
```
./COMPAS/src/COMPAS --random-seed 0 -n 1 --initial-mass-1 35.0 --initial-mass-2 25.0 --orbital-period 17.0 --chemically-homogeneous-evolution-mode PESSIMISTIC --tides-prescription NONE --maximum-number-timestep-iterations 1999999 --mass-change-fraction 0.001 --radial-change-fraction 0.001 --detailed-output TRUE --response-to-spin-up KEPLERIAN_LIMIT
```

And here was the result of the first SMT episode:

```
Mass change in primary:  -21.71244689891135 solMass 
Mass change in secondary:  21.71244689891135 solMass
Semi-Major Axis before vs after MT: 115.62435750672832 solRad  vs  199.86841715590887 solRad 

  Angular Momentum change in primary star:  0.0 AU2 solMass / yr 
Angular Momentum change in secondary star:  17.610106984473493 AU2 solMass / yr 
Angular Momentum change in orbit:  -201.47197960933272 AU2 solMass / yr  

Total Angular Momentum change due to MT:  -183.86573710771762 AU2 solMass / yr

```

veome22 (changes requested):
Also, I'm still seeing an error with the `--envelope-state-prescription CONVECTIVE_MASS_FRACTION` option.

Here's the binary I'm running:
```
./COMPAS/src/COMPAS --random-seed 0 -n 1 --initial-mass-1 26.854 --initial-mass-2 26.42 --semi-major-axis 0.6032 --metallicity 0.0001 --chemically-homogeneous-evolution-mode PESSIMISTIC --tides-prescription NONE --maximum-number-timestep-iterations 1999999 --mass-change-fraction 0.001 --radial-change-fraction 0.001 --detailed-output TRUE --logfile-type 'CSV' --envelope-state-prescription CONVECTIVE_MASS_FRACTION
```

```

COMPAS v03.17.00
A binary star simulator

Start generating binaries at Sat Mar 29 20:36:14 2025

ERROR:  in Core_Helium_Burning::virtual ENVELOPE CHeB::DetermineEnvelopeType() const: Unknown envelope state prescription, (ObjectId = 10)

0: Evolution stopped because an error occurred: (Main_Sequence_>_0.7 -> Naked_Helium_Star_MS) + (Main_Sequence_>_0.7 -> Core_Helium_Burning)

Generated 1 of 1 binaries requested

Simulation completed

End generating binaries at Sat Mar 29 20:36:22 2025

Clock time = 7.98579 CPU seconds
Wall time  = 0000:00:08 (hhhh:mm:ss)
```

jeffriley:
ERROR:  in Core_Helium_Burning::virtual ENVELOPE CHeB::DetermineEnvelopeType() const: Unknown envelope state prescription, (ObjectId = 10)

Is because this case:

    case ENVELOPE_STATE_PRESCRIPTION::CONVECTIVE_MASS_FRACTION:

is missing the `break` statement - so we're falling through to the `default`.  I can push a fix for that later.

ilyamandel:
Sorry about the missing break statements -- @SimonStevenson pointed me to the issue in HG.cpp, where I fixed it, but I neglected to check other classes.  Apologies, now fixed -- and thank you, @veome22 !

ilyamandel:
> I wanted to quickly flag that last I checked, the --response-to-spin-up KEPLERIAN_LIMIT option does not conserve angular momentum during stable mass transfer. I understand that this option is meant to not modify the orbit directly

This option *is* meant to update the orbital in order to conserve angular momentum, so if it's not conserved, I've messed something up. :(  Thank you for the test parameters, I'll work on this.  Marking as draft again until I've had a chance to test this.

jeffriley:
> @jeffriley : there's one line in the auto-created yaml file which I can't figure out how to fix

Not your fault @ilyamandel.  The immediate cause of the problem was that you inserted "--envelope-state-prescription" into a couple of option descriptions.  The ultimate cause of the problem is that the yaml file code parses the Boost help output to extract default values etc., and one of your descriptions just happened to put the "--envelope-state-prescription" string at the start of a new line (what are the odds...?), and that tricked my parser into thinking it was the help record for the "--envelope-state-prescription" option.  I changed the parse to take the help formatting into account.

veome22:
Hi @ilyamandel, I have another question about angular momentum conservation during stable mass transfer. When the mass transfer is non-conservative (e.g. from HG onto a BH), how is the angular momentum loss from the orbit calculated? In my tests, I am seeing very significant losses of angular momenta, but I don't think it would be physical if the only angular momentum being lost is from the donor star.

The reason I ask is that when I run some binaries with and without tides, I see very different amounts of angular momenta being lost to non-conservative stable mass transfer, even though both sets of systems may have the same total initial angular momentum before SMT. I can't see why this should be the case if angular momentum loss depends on the specific angular momentum at the surface of the donor, but this assumption may be misguided.

See SMT2 below:

```

./COMPAS --random-seed 0 -n 1 --initial-mass-1 35.0 --initial-mass-2 25.0 --orbital-period 17.0 --chemically-homogeneous-evolution-mode PESSIMISTIC --tides-prescription NONE --maximum-number-timestep-iterations 1999999 --mass-change-fraction 0.001 --radial-change-fraction 0.001 --detailed-output TRUE --logfile-type 'CSV' --response-to-spin-up TRANSFER_TO_ORBIT
```

```
 Time (Myr), Event,                            M1 (M_o), type1, M2 (M_o), type2, a (R_o),   e
   0.000053   Zero-age main-sequence, metallicity Z=0.0142   35.000     1     25.000     1    108.925  0.000
   5.496349   Star 1: MS->HG                    32.202     2     24.325     1    115.616  0.000
   5.497630   Stable mass transfer from 1 to 2   10.486     7     46.037     1    246.922  0.000
   5.497630   Star 1: HG->HeMS                  10.486     7     46.037     1    246.922  0.000 
  6.136872   Star 1: HeMS->HeHG                 9.570     8     45.075     1    255.451  0.000 
  6.166243   Star 1 undergoes supernova and forms a BH    9.433    14     45.027     1    256.324  0.000 
  8.257062   Star 2: MS->HG                     9.433    14     41.069     2    276.415  0.000 
  8.259185   Stable mass transfer from 2 to 1    9.433    14     14.623     7     16.995  0.000 
  8.259185   Star 2: HG->HeMS                   9.433    14     14.623     7     16.995  0.000 
  8.831839   Star 2: HeMS->HeHG                 9.433    14     10.285     8     20.735  0.000 
  8.859061   Star 2 undergoes supernova and forms a BH    9.433    14     10.111    14     20.921  0.000 
15444.951451   Double compact object (BH+BH) merging in 1.54e+04 Myr    9.433    14     10.111    14     20.921  0.000
 

SMT1:

Mass change in primary:  -21.71244689891135 solMass 
Mass change in secondary:  21.71244689891135 solMass 
 
Angular Momentum change in primary star:  0.0 AU2 solMass / yr 
Angular Momentum change in secondary star:  47.64647968845568 AU2 solMass / yr 
Angular Momentum change in orbit:  -47.64557032529442 AU2 solMass / yr 
 
Semi-Major Axis before vs after MT: 115.62435750672832 solRad  vs  246.92212914179268 solRad 
Total Angular Momentum change due to MT:  -4.538067855719419e-06 AU2 solMass / yr

 

 

SMT2:

Mass change in primary:  0.000102489763847 solMass 
Mass change in secondary:  -26.376190125143857 solMass 
 
Angular Momentum change in primary star:  8.52960533e-07 AU2 solMass / yr 
Angular Momentum change in secondary star:  0.0 AU2 solMass / yr 
Angular Momentum change in orbit:  -338.5245061632909 AU2 solMass / yr 
 
Semi-Major Axis before vs after MT: 276.797066480755 solRad  vs  16.995204423418496 solRad 
Total Angular Momentum change due to MT:  -338.5309986309088 AU2 solMass / yr

```

ilyamandel:
Hi @veome22,

I ran the exact binary you recommended:

./COMPAS --random-seed 0 -n 1 --initial-mass-1 35.0 --initial-mass-2 25.0 --orbital-period 17.0 --chemically-homogeneous-evolution-mode PESSIMISTIC --tides-prescription NONE --maximum-number-timestep-iterations 1999999 --mass-change-fraction 0.001 --radial-change-fraction 0.001 --detailed-output TRUE --response-to-spin-up KEPLERIAN_LIMIT

Mass lost from primary star:  -23.269908912325626 solMass
 Mass gained in secondary star:  23.269908912325626 solMass
  Angular Momentum change in primary star:  0.0 AU2 solMass / yr
 Angular Momentum change in secondary star:  18.369262316719688 AU2 solMass / yr
 Angular Momentum change in orbit:  -231.77763074105025 AU2 solMass / yr
 Total Angular Momentum change due to MT:  -213.41281420722623 AU2 solMass / yr

I cannot reproduce the results you quoted.  The first mass transfer in this case is a very non-conservative MT from a HG primary to a MS secondary, made non-conservative precisely because the secondary is not allowed to accrete once it reaches close to break-up spin.  The primary loses 32.2024-10.4862=21.7162 solar masses, but the secondary only gains 26.7572-24.325=2.4322.  Because the mass transfer is so non-conservative, a very significant amount of angular momentum is lost from the system.  This means the system is not a good test of angular momentum conservation, since the loss of angular momentum from the system must be accounted for.  

[In asnwer to your second question, the bulk of the angular momentum lost here is taken from the orbit, rather than the donor star.  Even if both stars had zero rotation, non-conservative mass transfer will carry away angular momentum, changing the orbital properties, and similarly for winds.  In fact, the only difference between winds and non-conservative mass transfer is whether the material lost from the system carries away the specific angular momentum of the donor or the accretor.]

I instead ran

./COMPAS --random-seed 0 -n 1 --initial-mass-1 35.0 --initial-mass-2 25.0 --orbital-period 17.0 --chemically-homogeneous-evolution-mode PESSIMISTIC --tides-prescription NONE --maximum-number-timestep-iterations 1999999 --mass-change-fraction 0.001 --radial-change-fraction 0.001 --detailed-output TRUE --response-to-spin-up TRANSFER_TO_ORBIT

With the TRANSFER_TO_ORBIT variation, the secondary is allowed to continue accreting even once it reaches break-up frequency, but angular momentum is forced to be conserved.  Indeed, the same episode of mass transfer now becomes conservative (to a part in 10,000).  And the total angular momentum is conserved after this mass transfer episode.  This is also what you found in the first mass transfer in your later example.  

However, I did find a typo (now fixed) while performing these tests, so I am very grateful to your insistence on these checks.

To the best of my knowledge, this PR is now ready to be pulled, if you agree and are comfortable with approving.

jeffriley (changes requested):
I'll review this today

jeffriley (approved):
All good, thanks @ilyamandel (I pushed a commit with a little code tidy-up)

veome22:
Hi @ilyamandel, thank you for bearing with all my tests 😃 

I think the current model for angular momentum loss during non-conservative mass transfer sounds sensible. Also whatever you changed most recently has fixed the issue of inconsistent amounts of angular momentum loss between simulations with different tidal prescriptions. I'm good with this PR now!

veome22 (approved):

ilyamandel:
Hi @jeffriley ,

It looks like you added a print statement in MainSequence.h.  I suspect this came up during testing.  It shouldn't ever be called, but if it is, the print is probably not in the standard format -- this should either throw a warning / error if you want to alert the user, or not print at all...  I'd suggest removing this print statement in the next PR (not urgent, since I don't expect it to ever be called).

double CalculateRadiusOnPhase(const double p_Mass, const double p_Luminosity) const { std::cout << "MS::CalculateRadiusOnPhase(LUM) called\n"; return Radius(); } // Not a meaningful function for MS stars

jeffriley:
Ah yes, missed that - thanks. I'll remove it.

PR #1361, opened 2025-04-01
Fixes and enhancements to BRCEK core mass prescription

brcekadam:
A few fixes and additions to the MS core mass prescription:

- I combined MainSequence::CalculateRadiusOnPhase() and MainSequence::CalculateRadiusOnPhaseTau() into one function CalculateRadiusOnPhase(p_Mass, p_Tau). As far as I could tell, this should have no impact on the functionality, but it makes the code cleaner and made my implementation much easier :-)

- There was a bug in my initial implementation when the radius was always slightly higher than the RL radius during nuclear timescale mass transfer due to effectively aging the star too soon. That should now be fixed.

- We are now able to track the surface helium/hydrogen abundance on the MS when the star gets so stripped that the processed material ends up on the surface (when total mass reaches the initial convective core mass). We then scale the radius to slow down any subsequent mass transfer and prevent the total mass from ever reaching the current core mass.


jeffriley (changes requested):
Thanks @brcekadam.  A few comments/questions:

(a) combining `MainSequence::CalculateRadiusOnPhase()` and `MainSequence::CalculateRadiusOnPhaseTau()` into `CalculateRadiusOnPhase(p_Mass, p_Tau)` makes me a little nervous...

(i) in `MainSequence.h` we had

    double CalculateRadiusOnPhase() const { return CalculateRadiusOnPhase(m_Mass, m_Age, m_RZAMS0); }
	
but now have

    double CalculateRadiusOnPhase() const { return CalculateRadiusOnPhase(m_Mass, m_Tau); }
	
with `p_RZAMS` from the original `MainSequence::CalculateRadiusOnPhase()` removed, with that value being calculated as:

    double RZAMS = CalculateRadiusAtZAMS(p_Mass); 

from the original `MainSequence::CalculateRadiusOnPhaseTau()`.

Is that correct?  I don't think `CalculateRadiusAtZAMS(p_Mass)` is guaranteed to be the same as `m_RZAMS0` - is that a deliberate change?

(ii) we now have two functions with the same signature that should be passed different parameters:

    double CalculateRadiusOnPhase(const double p_Mass, const double p_Tau)

    double CalculateRadiusOnPhase(const double p_Mass, const double p_Luminosity)
	
They look the same to the compiler and linker (double CalculateRadiusOnPhase(double, double)) - I haven't done a detailed check of the code, but it makes me a little nervous wondering how an ambiguous call might be resolved...

Can you check the code, and/or do some regression testing to make sure that change doesn't have any insidious side-effects?


(b) in `BaseBinaryStar::ChooseTimestep()` you calculate the radial expansion timescale for each star:

    double radialExpansionTimescale1 = m_Star1->Mass() / m_Star1->CalculateNuclearMassLossRate();
    double radialExpansionTimescale2 = m_Star2->Mass() / m_Star2->CalculateNuclearMassLossRate();

`BaseStar::CalculateNuclearMassLossRate()` calculates the radial expansion timescale and returns:

    return m_Mass / radialExpansionTimescale;

We have a function in the `BaseStar` class that calculates the radial expansion timescale (`CalculateRadialExpansionTimescale()`) - but we want to use a different calculation here (and in `BaseStar::CalculateNuclearMassLossRate()`).
Rather than have a function to do something, but then do it differently in the code occasionally, I think we should define a new function to calculate the radial expansion timescale in a mass transfer regime, and use that in both `BaseStar::CalculateNuclearMassLossRate()` and `BaseBinaryStar::ChooseTimestep()`.  Not only does that (I think) make it more clear (and callable from elsewhere in the code if/when necessary), it means `BaseBinaryStar::ChooseTimestep()` can call the new function directly to calculate the radial expansion timescale, rather than calling `BaseStar::CalculateNuclearMassLossRate()` and dividing by the mass to get back to the timescale (seems a little unnecessarily convoluted - and uses compute cycles we don't need to use...).


(c) in `BaseBinaryStar::ChooseTimestep()` we calculate `dt` based on some criteria (some of this was there prior to this PR), and we have comments such as:

`// limit time step for stars losing mass on nuclear timescale`, and
`// yes - reduce timestep if necessary to ensure that the orbital separation does not change by more than ~1% per timestep due to GW emission`

then just before we return from `BaseBinaryStar::ChooseTimestep()` we do this:

    dt *= p_Multiplier;	

I know it's kind-of caveat utilitor, but the user doesn't necessarily know in what order we do things in `BaseBinaryStar::ChooseTimestep()`.  Are we really "limiting" or "ensuring" (see comments above) if we later increase `dt` by multiplying by the timestep multiplier?  Should "limiting" or "ensuring" come *after* multiplying y the timestep multiplier?  I don't know the answer btw, just posing it :-)

brcekadam:
Hi @jeffriley, thank you for your review!

(a) 
(i) On the main sequence, `m_Mass0` is equal to `m_Mass` (looking at `MainSequence::UpdateInitialMass()`, which gets called in `BaseStar::ResolveMassLoss()`), and `m_RZAMS0` is calculated from `m_Mass0` at every time step in `MainSequence::EvolveOneTimestepPreamble()`. So `m_RZAMS0` should always be the same as `CalculateRadiusAtZAMS(m_Mass)`.

(ii) That’s a good point. I tried testing it on current `dev` branch, and `MainSequence::CalculateRadiusOnPhase(double, double)` is never called. Looking at the code, `CalculateRadiusOnPhase(const double p_Mass, const double p_Luminosity)` seems to be contained in the classes that it is relevant for. But I understand how it's not ideal to have two functions with the same signature that should be passed different parameters, so I'm happy to make some adjustments if we want to completely avoid this situation. 

(b) Yes, that’s a good idea! I can define the new function.

(c) Hmm, I guess we are not exactly “limiting” or “ensuring”. Maybe we can update the description of `--timestep-multiplier` to specify that it applies to the final, optimally chosen time step. If the multiplication happens before “limiting” or “ensuring", I feel like it could lead to confusion about why the time steps don’t always change (e.g. if a user picks a multiplier that’s too high). Additionally, applying the multiplier at the end is probably more useful for testing, as it avoids setting a hard upper limit on the time step and needing to change the code to experiment with larger steps :-)

jeffriley:
@brcekadam 

> (i) On the main sequence, m_Mass0 is equal to m_Mass (looking at MainSequence::UpdateInitialMass(), which gets called in 
> BaseStar::ResolveMassLoss()), and m_RZAMS0 is calculated from m_Mass0 at every time step in 
> MainSequence::EvolveOneTimestepPreamble(). So m_RZAMS0 should always be the same as CalculateRadiusAtZAMS(m_Mass).

Ok, but you are calculating `RZAMS` as:

    double RZAMS= CalculateRadiusAtZAMS(p_Mass);

where `p_Mass` is a parameter passed to `MainSequence::CalculateRadiusOnPhase()`.

> m_RZAMS0 is calculated from m_Mass0 at every time step in MainSequence::EvolveOneTimestepPreamble()

Let me nuance that a little. `m_RZAMS0` is calculated from `m_Mass0` at *the start of* every time step in `MainSequence::EvolveOneTimestepPreamble(`).

Prior to this change, the mass used to calculate `m_RZAMS0` in `MainSequence::EvolveOneTimestepPreamble()` was `m_Mass0` *at the start of the timestep*. Are you certain, or perhaps certain that it doesn't matter, that the mass used to calculate `RZAMS` in `MainSequence::CalculateRadiusOnPhase()` (`p_Mass`) is the same as `m_Mass0` at the start of the timestep?  If it isn't then `RZAMS` as calculated in `MainSequence::CalculateRadiusOnPhase()` will  not be the same as `m_RZAMS0` as calculated in `MainSequence::EvolveOneTimestepPreamble()`, will it?  That may not matter (hence my earlier statement (`or perhaps certain that it doesn't matter`)) - I don't know.

> (ii) That’s a good point. I tried testing it on current dev branch, and MainSequence::CalculateRadiusOnPhase(double, double)
> is never called. Looking at the code, CalculateRadiusOnPhase(const double p_Mass, const double p_Luminosity) seems to be
> contained in the classes that it is relevant for. But I understand how it's not ideal to have two functions with the same 
> signature that should be passed different parameters, so I'm happy to make some adjustments if we want to completely
> avoid this situation.

Yes, I think we should do something - change the name or the signature.  It could turn into a subtle, insidious, problem if people make changes in the future that unwittingly cause the wrong function to be called.  It's probably ok as it is (maybe...), but could bite us later - I'd rather just head that possibility off.  Having multiple functions with the same signature is not really good practice - I think we'd just be asking for trouble.

> (c) Hmm, I guess we are not exactly “limiting” or “ensuring”. Maybe we can update the description of --timestep-multiplier to
> specify that it applies to the final, optimally chosen time step. If the multiplication happens before “limiting” or “ensuring", I
> feel like it could lead to confusion about why the time steps don’t always change (e.g. if a user picks a multiplier that’s too
> high). Additionally, applying the multiplier at the end is probably more useful for testing, as it avoids setting a hard upper
> limit on the time step and needing to change the code to experiment with larger steps :-)

I've never liked the timestep multiplier - it's a blunt instrument that can do as much harm as good :-)  I think it's a fine idea in the hands of someone who knows what's happening.  There are places in the code where we clamp the timestep to a lower bound (the absolute minimum timestep) - I'm pretty sure we do that after applying the timestep multiplier, so that the timestep never goes below the absolute minimum (I haven't checked, but I think that's what we do).  But we don't clamp it on the upper end - I guess we don't have an upper bound that we can easily define at run time, but it bothers me that we say we're limiting the timestep to some value, perhaps to ensure that we limit other things (like mass loss/transfer, radial change, etc.), then (possibly) go and increase the timestep without checking that we're not going to undo what we just said we did.  I think it's fine to have it there, but I think we should probably make sure that users know (i.e. change the documentation for the `--timestep-multiplier` option) that it's really just a test/debug thing and should be used with discretion, care, and at their own peril...  Used injudiciously it could cause unexpected results (and confusion...).  You don't have to change the documentation in this PR - one of us can do that later.

ilyamandel:
Just to jump in on 

> Should "limiting" or "ensuring" come after multiplying y the timestep multiplier? I don't know the answer btw, just posing it :-)

I added --timestep-multiplier a while ago, and I do find that it's useful as a debugging tool, so I'd like to keep it -- and keep it precisely the format it now has (multiply the timesteps after they are chosen by other tools).  In other words, I am happy to change the documentation, but not the code behaviour.

In fact, just to make sure this one is addressed, I've changed the online docs in #1360 from

**--timestep-multiplier** |br|
Multiplicative factor for timestep duration. |br|
Default = 1.0

to

**--timestep-multiplier** |br|
Multiplicative factor for timestep duration.  This multiplier is applied after the timesteps are chosen using other program options such as ``--radial-change-fraction`` and ``--mass-change-fraction``, and will therefore override expected behaviour.  This option is primarily intended for debugging/testing of convergence issues rather than for production runs. |br|
Default = 1.0

I also changed the help string for timestep multiplier from
Timestep multiplier for SSE and BSE
to 
Timestep multiplier for SSE and BSE on top of other choices, for use in debugging.

ilyamandel (commented):
@brcekadam : the reason for the existence of CalculateRadiusOnPhaseTau() is precisely because CalculateRadiusOnPhase(const double p_Mass, const double p_Luminosity) already existed, so I wanted something with a different name to avoid ambiguity and possibly incorrect behaviour, hence the addition of this new function.  Sorry, I probably should have documented that better.  But I agree with @jeffriley's concern.

ilyamandel (changes requested):
@brcekadam : the reason for the existence of CalculateRadiusOnPhaseTau() is precisely because CalculateRadiusOnPhase(const double p_Mass, const double p_Luminosity) already existed, so I wanted something with a different name to avoid ambiguity and possibly incorrect behaviour, hence the addition of this new function.  Sorry, I probably should have documented that better.  But I agree with @jeffriley's concern.

brcekadam:
I see what you mean @jeffriley, now I understand. I took a closer look, and the mass used to calculate `RZAMS` in `MainSequence::CalculateRadiusOnPhase() (p_Mass)` is **not** the same as `m_Mass0` at the start of the time step every time the function is called. The difference seems to be very small, but I also can't guarantee that it has no impact. 

So how about this: I put the third parameter back in, so we would have `MainSequence::CalculateRadiusOnPhase(p_Mass, p_Tau, p_RZAMS)` (this would also resolve the issue of two functions having the same name and signature). Then in `MainSequence.h`, we would have

`double CalculateRadiusOnPhase() const { return CalculateRadiusOnPhase(m_Mass, m_Tau, m_RZAMS0); }`
and
`double CalculateRadiusOnMassChange(double p_dM) { return CalculateRadiusOnPhase(m_Mass + p_dM, m_Tau, CalculateRadiusAtZAMS(m_Mass + p_dM)); }`

brcekadam:
@jeffriley, @ilyamandel 

I have now implemented the changes! Let me know if you find any other issues :-)

jeffriley (changes requested):
Started review - will complete in next day or two.

One request at the moment:

Can we document, (say) in the description of `MainSequence::CalculateMainSequenceCoreMassBrcek()`, why the mixing core mass cannot be greater than 90% of total mass (comment line `777` in `MainSequence.cpp`)?  I saw the brief discussion in the slack channel, but I think we should put the reason in the code so future readers of the code know why.  And should that fraction (`0.9`) be put in `constants.h` and given a symbolic name? Might it be used elsewhere in the code, or is it only ever likely to be here?. If we do put it in `constants.h` then perhaps put the explanation of why it is `0.9` there.

jeffriley:
@brcekadam I have completed my review.  Just the note above, and the following three items:

(1) In BaseBinaryStar::ChooseTimestep() you have:

        // limit time step for stars losing mass on nuclear timescale
        double radialExpansionTimescale1 = m_Star1->CalculateRadialExpansionTimescaleDuringMassTransfer();
        double radialExpansionTimescale2 = m_Star2->CalculateRadialExpansionTimescaleDuringMassTransfer();
        if (utils::Compare(radiusToRL1 * (1.0 + 0.5 * OPTIONS->RadialChangeFraction()), 1.0) > 0)
            dt = std::min(dt, 0.5 * OPTIONS->RadialChangeFraction() * radialExpansionTimescale1);
        if (utils::Compare(radiusToRL2 * (1.0 + 0.5 * OPTIONS->RadialChangeFraction()), 1.0) > 0)
            dt = std::min(dt, 0.5 * OPTIONS->RadialChangeFraction() * radialExpansionTimescale2);

There is no need to calculate `radialExpansionTimescale1` and `radialExpansionTimescale2` if they are not going to be used.  Either move the calculation for each inside the appropriate conditional, or (imo better) change the code to do this:

        // limit time step for stars losing mass on nuclear timescale
        if (utils::Compare(radiusToRL1 * (1.0 + 0.5 * OPTIONS->RadialChangeFraction()), 1.0) > 0)
            dt = std::min(dt, 0.5 * OPTIONS->RadialChangeFraction() * m_Star1->CalculateRadialExpansionTimescaleDuringMassTransfer());
        if (utils::Compare(radiusToRL2 * (1.0 + 0.5 * OPTIONS->RadialChangeFraction()), 1.0) > 0)
            dt = std::min(dt, 0.5 * OPTIONS->RadialChangeFraction() * m_Star2->CalculateRadialExpansionTimescaleDuringMassTransfer());

That way we're not creating local variables that get used once in the very next statement and never again - reduces overhead (i.e. improves performance (slightly, depending on compiler optimisation - but slightly for perhaps thousands of timesteps for millions of binaries...)), code clutter, increases clarity.


(2) In MainSequence::CalculateRadiusTransitionToHG(), change the constant `1` to `1.0` in this statement:
   
    return (radiusAtHookStart * (1 - p_Tau) + radiusTAMS * (p_Tau - 0.99)) / 0.01;


(3) In `MainSequence::UpdateMainSequenceCoreMass()` you have:

    double tMS = m_Timescales[static_cast<int>(TIMESCALE::tMS)];

Have a look at (e.g.) `MainSequence::CalculateTauOnPhase()` (the very next function in `MainSequence.cpp`):

    double MainSequence::CalculateTauOnPhase() const {
    #define timescales(x) m_Timescales[static_cast<int>(TIMESCALE::x)]  // for convenience and readability - undefined at end of function
        return std::max(0.0, std::min(1.0, m_Age / timescales(tMS)));
    #undef timescales
    }

and the use of the #define and #undef.  That construct lets us use (e.g.) `timescales(tMS)` as shorthand for `m_Timescales[static_cast<int>(TIMESCALE::tMS)]`, without incurring the overhead of creating a local variable and moving the value from the m_Timescales` array into it - the substitution is done at compile time. I think you should use it in `MainSequence::UpdateMainSequenceCoreMass()`.

ilyamandel:
Thanks for the changes, @brcekadam !
I'll wait for you to address @jeffriley's requests before approving.

jeffriley (approved):
Thanks @brcekadam !

ilyamandel (approved):
Hi @brcekadam ,

I'm happy to approve.  Just one question: given the changes in the radius functions, have have you checked the radial evolution of stars that follow the old (non-BRCEK) prescription?  Ideally, given that our goal is to be able to reproduce old results (except when we are fixing bugs ;-) ) with old options, we should be able to get ~identical results for SSE, and perhaps even BSE in the case of case A mass transfer, when old options are used -- or at least to understand and accept the differences.  If you could make such a check before pulling this in (or if you've made it already), that would be awesome.

Minor suggestion:
In changelog.h:
"total mass now never reaches the core mass" -- would it be technically more correct to say "core mass now never reaches the total mass"?  or perhaps "total mass is now never reduced to the convective core mass"?

brcekadam:
Hi @ilyamandel, 

I checked the detailed outputs for a few single stars and for a system with case A mass transfer, and the results (using the old prescription) are the same between the two version. And you are right about the phrasing in `changelog.h`, it should be the other way around. I fixed it, so now it should be ready! Thank you both for your comments :-)

Issue #1366, opened 2025-04-05
Inconsistent comments on NS spin-up during CE


ilyamandel:
BaseBinaryStar::CalculateMassTransfer() has the following lines:

// Check for recycled pulsars. Not considering CEE as a way of recycling NSs.
	if (!m_CEDetails.CEEnow && m_Accretor->IsOneOf({ STELLAR_TYPE::NEUTRON_STAR })) {                      // accretor is a neutron star
    m_Donor->SetRLOFOntoNS();                                                // donor donated mass to a neutron star
    m_Accretor->SetRecycledNS();                                              // accretor is (was) a recycled NS
	}

However, the new code NS::UpdateMagneticFieldAndSpin() does adjust NS spins during a CE.  The code and comment above contradict each other.  

According to @yuzhesong, if
--neutron-star-accretion-in-ce
is not set to ZERO then the neutron star is getting recycled, so that block of code should be updated.

ilyamandel:
Fixed in #1367 , closing.

PR #1367, opened 2025-04-06
Fix to Issue #1366 

yuzhesong:
This PR should provide fixes to issue #1366 by:
1. labelling the neutron star as recycled if it accretes mass from the donor star through common envelope, when --neutron-star-accretion-in-ce is not set to ZERO. 
2. adding a new output option CE_ONTO_NS to label the donor star to have transferred mass onto the NS through common envelope.


ilyamandel (changes requested):
BaseBinaryStar.cpp:

else if (m_CEDetails.CEEnow && OPTIONS->NeutronStarAccretionInCE() != NS_ACCRETION_IN_CE::ZERO) { // accretor is a neutron star, system is in CE

Why is this true only if the accretor is a NS?  This would evaluate to true in CE and if the appropriate option is set, whatever the accretor type.  I don't think that's the intended behavior.

ilyamandel (changes requested):
**CE_ONTO_NS**

Why do we need this as a separate output?  This is readily available by parsing the CE output info.  I think we should remove unnecessary outputs that can already be trivially obtained from other outputs, not add new ones.  (Yes, I'd prefer to remove RLOF_ONTO_NS , too...)

jeffriley (approved):
Made a couple of cosmetic changes.  I'll approve, but @ilyamandel's comments need to be addressed.

yuzhesong:
@ilyamandel 

> BaseBinaryStar.cpp:
> 
> else if (m_CEDetails.CEEnow && OPTIONS->NeutronStarAccretionInCE() != NS_ACCRETION_IN_CE::ZERO) { // accretor is a neutron star, system is in CE
> 
> Why is this true only if the accretor is a NS? This would evaluate to true in CE and if the appropriate option is set, whatever the accretor type. I don't think that's the intended behavior.

Ah you are right, I need to add the m_Accretor->IsOneOf({ STELLAR_TYPE::NEUTRON_STAR }) criteria in the else if statement. 


> **CE_ONTO_NS**
> 
> Why do we need this as a separate output? This is readily available by parsing the CE output info. I think we should remove unnecessary outputs that can already be trivially obtained from other outputs, not add new ones. (Yes, I'd prefer to remove RLOF_ONTO_NS , too...)

I have no objections removing both CE_ONTO_NS and RLOF_ONTO_NS. Will do so in the next commit.

ilyamandel (approved):
Looks good, @yuzhesong  -- thank you!

PR #1369, opened 2025-04-07
Fix for #1365

### LINKED ISSUE #1365: Rotational frequency input: Hz or rad/s ?
The rotational frequency initial values (--rotational-frequency, --rotational-frequency-1, --rotational-frequency-2) are specified in the help as having units of Hz, which makes sense for frequency.  However, in fact, these values are used to set the angular velocity omega, which has units of radians per second.  There is thus a factor of 2*pi mismatch between the behaviour described in the help and the actual behaviour.  

We should either change the help and documentation to correctly reflect that the user is inputting omega in rad/s, or include a factor of 2*pi in code such as the following line from BaseStar.cpp:

    m_OmegaZAMS                                = p_RotationalFrequency >= 0.0                           // valid rotational frequency passed in?
                                                    ? p_RotationalFrequency                             // yes - use it
                                                    : CalculateZAMSAngularFrequency(m_MZAMS, m_RZAMS);  // no - calculate it

veome22:
1. Added a factor of $2 \pi$ to convert `--rotational-frequency, --rotational-frequency-1, --rotational-frequency-2` inputs to angular velocity in BaseStar.cpp.
2. Corrected a typo (I think) in the documentation for `BaseStar::CalculateOmegaCHE()` in BaseStar.cpp, where the output was labeled as rad/s instead of rad/yr.

Note: when the `BaseStar()` constructor is called, the rotational frequency is already converted from Hz to 1/yr, so only an extra factor ot $2 \pi$ was needed to convert to rad/yr. Omega is defined in units of rad/yr everywhere in COMPAS.

### DIFF
--- a/src/BaseStar.cpp
+++ b/src/BaseStar.cpp
@@ -140,7 +140,7 @@ BaseStar::BaseStar(const unsigned long int p_RandomSeed,
     
     m_OmegaCHE                                 = CalculateOmegaCHE(m_MZAMS, m_Metallicity);
     m_OmegaZAMS                                = p_RotationalFrequency >= 0.0                           // valid rotational frequency passed in?
-                                                    ? p_RotationalFrequency                             // yes - use it
+                                                    ? _2_PI * p_RotationalFrequency                     // yes - convert to rad/s and use it
                                                     : CalculateZAMSAngularFrequency(m_MZAMS, m_RZAMS);  // no - calculate it
     m_AngularMomentum                          = CalculateMomentOfInertiaAU() * m_OmegaZAMS;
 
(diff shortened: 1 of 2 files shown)


jeffriley (approved):
Thanks @veome22 !

ilyamandel:
@veome22 :
Thank you for this!
However, there's probably one more copy-paste typo in the comments that should be fixed:

/*
 * Calculate the minimum rotational frequency (in yr^-1) at which CHE will occur
 * for a star with ZAMS mass MZAMS
 ...
 * double CalculateOmegaCHE(const double p_MZAMS, const double p_Metallicity)
...
 * @return                                      Initial angular frequency in rad*yr^-1
 
 That's the minimum angular frequency for CHE being returned, not the *initial* angular frequency...

PR #1372, opened 2025-04-13
Add option "--timestep-multipliers" to enable more granular, phase-dependent, timestep multipliers

jeffriley:
Enhancement:
  - Add option "--timestep-multipliers" to enable more granular, phase-dependent, timestep multipliers
  - Added maximum allowed value for options `--timestep-multiplier` and `--timestep-multipliers`
 
@ilyamandel Rather than complaining that `--timestep-multiplier` is a blunt instrument, I decided to sharpen it a little.  I know it's not the highest priority change, but I had an itch that needed scratching ;-)  I hope it's useful - if not, don't merge it :-)

I have added a new option  `--timestep-multipliers` that allows the user to specify phase-dependent timestep multipliers (one for each stellar type).  `--timestep-multiplier` still exists, and can be used in conjunction with `--timestep-multipliers`: `--timestep-multiplier` acts as a "global" multiplier, while `--timestep-multipliers` has a value for each stellar type.  The timestep is now multiplied by the values specified by `--timestep-multiplier`, and the phase-appropriate value specified by `--timestep-multipliers`.  The default for `--timestep-multiplier`, and each phase-dependent value for `--timestep-multipliers`, is `1.0`. Doing it this way, I think, affords more flexibility.

The `--timestep-multipliers` option is a vector option, and shorthand notation is allowed.  See https://compas.readthedocs.io/en/latest/pages/User%20guide/Program%20options/program-options-vector-options.html for documentation on vector options (page updated with this PR, but the existing page gives a reasonable explanation).

I have also set a maximum value (in `constants.h` for timestep multipliers.  I have set the maximum value to `1000.0` - feel free to change that to whatever you think it should be.  (We could also change the option name if you think another is better - it is close to the existing `--timestep-multiplier`).

Example:

    ./compas --timestep-multipliers [0.1,100.0,,2.75,3.1,0.01,,40.23,,,0.001,,2,1000.0,,2,] --timestep-multiplier 1.25

The example above specifies `1.25` for the "global" multiplier, and values for some of the stellar types - some are defaulted and will be `1.0`.  In this example, the timestep value for stellar type `MS_LTE_07` will be multiplied by `1.25` (the global multiplier), then by `0.1` (the phase-dependent multiplier).  Similarly, the timestep value for stellar type `CORE_HELIUM_BURNING` will be multiplied by `1.25`, then by `3.1`.  And the timestep value for stellar type `THERMALLY_PULSING_ASYMPTOTIC_GIANT_BRANCH` will be multiplied by `1.25`, then by `1.0` (the phase-dependent default value).

As I said, I know it's not a high priority, and very few people will use it, but I hope it will be useful for convergence testing...  And now my itch is scratched ;-)


ilyamandel (approved):
Hi @jeffriley ,

Looks great (modulo one possible logical mistake, see below)!  Glad you got to scratch your itch. ;-)

Approving now, but please make the following minor changes before pulling in:

- Merge in dev (I had just approved and pulled another PR which was submitted before yours);

- In program-option-list-defaults.rst, "multicative" -> "multiplicative"

- You added a new option, but I didn't see an update to yaml.h or a new default yaml file created

- In BaseBinaryStar.cpp:
`dt *= OPTIONS->TimestepMultiplier() * (dt1 < dt2 ? OPTIONS->TimestepMultipliers(static_cast<int>(m_Star1->StellarType())) : OPTIONS->TimestepMultipliers(static_cast<int>(m_Star2->StellarType()))) * p_Factor;`

I don't think this is what you want.  Imagine that dt1<dt2, but the timestep multiplier for dt2 (because of the phase that it's in) is much, much smaller than for dt1; then the user would expect you to take a step based on dt2 times its multiplier, but you'll base it on dt1 with the current logic. 

- In Options.h, remove extra "the" in "the the user"

jeffriley:
Thanks @ilyamandel 

> You added a new option, but I didn't see an update to yaml.h or a new default yaml file created

Oops - completely forgot.  I'll fix that (and the typos mentioned).

> In BaseBinaryStar.cpp:

    dt *= OPTIONS->TimestepMultiplier() * (dt1 < dt2 ? OPTIONS->TimestepMultipliers(static_cast<int>(m_Star1->StellarType())) : OPTIONS->TimestepMultipliers(static_cast<int>(m_Star2->StellarType()))) * p_Factor;

> I don't think this is what you want. Imagine that dt1<dt2, but the timestep multiplier for dt2 (because of the phase that it's in) is > much, much smaller than for dt1; then the user would expect you to take a step based on dt2 times its multiplier, but you'll base > it on dt1 with the current logic.

Well... here's my rationale for doing that.  We're evolving a binary (evolving single stars is a bit more straightforward - there's only one dt and one multiplier...), so the timestep we choose is the smaller of the timestep for star1 (dt1) and star2 (dt2), and given that, I choose the multiplier relevant to the stellar type of the star whose timestep we are using to evolve the binary.  If we are using the timestep for star1 to evolve the binary, I'm not sure dt2 and the multiplier relevant to star2 is the intuitive choice.

EDIT: ... because we say we apply the multiplier after we do everything else.  But since you are most likely the primary user of this functionality I am completely happy to be swayed by your judgement :-)

ilyamandel (commented):
I think the logical option would be to replace 

double dt1 = m_Star1->CalculateTimestep();

with 

double dt1 = m_Star1->CalculateTimestep() *  OPTIONS->TimestepMultipliers(static_cast<int>(m_Star1->StellarType())) ;

and similarly for dt2.

Then, on line 3013, just have 

dt *= OPTIONS->TimestepMultiplier() * p_Factor;

ilyamandel:
Yes, I recognise that this means that if, say, the timestep was set by tides instead of stellar evolution, --timestep-multipliers may have limited effect, unlike --timestep-multiplier which would still multiply the timestep.  But, in my mind, this is the more self-consistent option.

jeffriley:
@ilyamandel Just noticed there is still an "multicative" in the options list - there must have been two and I only fixed one (I'm sure I fixed it...).  Anyway, I'll fix the remaining one on the next PR.

PR #1376, opened 2025-04-28
Updated coefficients from Shikauchi et al. (2024), fixed surface and core abundances for HeMS and HeHG

brcekadam:
Updated coefficients from Shikauchi et al. (2024), required for MS core mass calculations. Only the preprint is available online, so see our Slack channel for reference. 

This PR also fixes the behavior of surface and core helium abundance for HeMS and HeHG stars. HeMS stars now have the surface helium abundance set to 1-Z, and HeHG stars have the core helium abundance set to zero.


ilyamandel (approved):
Looks good -- thank you, @brcekadam !

Issue #1380, opened 2025-05-02
Non-positive Nanjing lambda


ilyamandel:
Nanjing lambda can be non-positive in version 03.18.02, which should not happen

To reproduce, run 

> ./COMPAS -n 1 --metallicity-distribution LOGUNIFORM --random-seed 46

The output is 

ERROR:  in Core_Helium_Burning::double BaseStar::CalculateBindingEnergy(double, double, double, double) const: Lambda <= 0.0; Binding energy = 0.0, (ObjectId = 6)
0: Evolution stopped because an error occurred: (Main_Sequence_>_0.7 -> Core_Helium_Burning) + (Main_Sequence_<=_0.7 -> Main_Sequence_<=_0.7)

ilyamandel:
The issue was with the Loveridge lambda, not the Nanjing lambda.  Fixed in #1381, closing.

PR #1381, opened 2025-05-02
Fix for issue #1380

ilyamandel:
Fix for issue #1380: Loveridge lambdas could occasionally be zero because of an overflow in the binding energy calculation.


jeffriley (approved):
All good

PR #1387, opened 2025-05-08
Fix for issue #1378

### LINKED ISSUE #1378: Non-random supernova kicks when specifying seeds using grid
**Describe the bug**
When sending a run using the command line + a grid file containing nothing but a list of --random-seeds, all resultant supernova kicks have the same SN_Kick_Magnitude_Random_Number(1) and SN_Kick_Magnitude_Random_Number(2). Other random quantities (masses, semimajor axis) work properly. When sending a run with one specific seed, using just the command line works properly also for the SN kick.

**To Reproduce**
1. Correct behavior:
./COMPAS --number-of-systems 1 --mode BSE --initial-mass-function KROUPA --initial-mass-min 10 --initial-mass-max 100 --mass-ratio-distribution FLAT --mass-ratio-min 0.1 --mass-ratio-max 1 --semi-major-axis-distribution FLATINLOG --semi-major-axis-min 0.1 --semi-major-axis-max 1000 --metallicity 0.0142 --logfile-type CSV --fryer-supernova-engine RAPID --maximum-neutron-star-mass 2.0  --kick-magnitude-sigma-CCSN-NS 265 --kick-magnitude-sigma-ECSN 30 --kick-magnitude-sigma-USSN 30 --common-envelope-mass-accretion-prescription CONSTANT --common-envelope-mass-accretion-constant 0 --kick-magnitude-distribution MAXWELLIAN --random-seed 1745928160

2. Incorrect behavior:
./COMPAS --mode BSE --initial-mass-function KROUPA --initial-mass-min 10 --initial-mass-max 100 --mass-ratio-distribution FLAT --mass-ratio-min 0.1 --mass-ratio-max 1 --semi-major-axis-distribution FLATINLOG --semi-major-axis-min 0.1 --semi-major-axis-max 1000 --metallicity 0.0142 --logfile-type CSV --fryer-supernova-engine RAPID --maximum-neutron-star-mass 2.0  --kick-magnitude-sigma-CCSN-NS 265 --kick-magnitude-sigma-ECSN 30 --kick-magnitude-sigma-USSN 30 --common-envelope-mass-accretion-prescription CONSTANT --common-envelope-mass-accretion-constant 0 --kick-magnitude-distribution MAXWELLIAN --grid grid.txt

Where grid.txt contains just
--random-seed 1745928160

For this particular seed, the former results in a DCO while the latter becomes unbound.

**Versioning:**
 - OS: Ubuntu 22.04 (on WSL2)
 - COMPAS v03.13.04 and COMPAS v03.14.01

**Additional context**
Ran into this while trying out --tides-prescription KAPIL2024, will post about it separately.

jeffriley:
Fix for issue #1378: reinstate 'false' fallback option for 'kick-magnitude-random*' options (mistakenly changed to 'true' in v03.00.00)

### DIFF
--- a/src/Options.h
+++ b/src/Options.h
@@ -1458,9 +1458,9 @@ public:
     double                                      KickMagnitude1() const                                                  { return OPT_VALUE("kick-magnitude-1", m_KickMagnitude1, true); }
     double                                      KickMagnitude2() const                                                  { return OPT_VALUE("kick-magnitude-2", m_KickMagnitude2, true); }
 
-    double                                      KickMagnitudeRandom() const                                             { return OPT_VALUE("kick-magnitude-random", m_KickMagnitudeRandom, true); }
-    double                                      KickMagnitudeRandom1() const                                            { return OPT_VALUE("kick-magnitude-random-1", m_KickMagnitudeRandom1, true); }
-    double                                      KickMagnitudeRandom2() const                                            { return OPT_VALUE("kick-magnitude-random-2", m_KickMagnitudeRandom2, true); }
+    double                                      KickMagnitudeRandom() const                                             { return OPT_VALUE("kick-magnitude-random", m_KickMagnitudeRandom, false); }
+    double                                      KickMagnitudeRandom1() const                                            { return OPT_VALUE("kick-magnitude-random-1", m_KickMagnitudeRandom1, false); }
+    double                                      KickMagnitudeRandom2() const                                            { return OPT_VALUE("kick-magnitude-random-2", m_KickMagnitudeRandom2, false); }
 
     STR_VECTOR                                  LogClasses() const                                                      { return m_CmdLine.optionValues.m_LogClasses; }
     std::string                                 LogfileCommonEnvelopes() const                                          { return m_CmdLine.optionValues.m_LogfileCommonEnvelopes; }
--- a/src/Options.h
+++ b/src/Options.h
@@ -70,12 +70,18 @@ const std::string NOT_PROVIDED_STR(1, static_cast<char>(NOT_PROVIDED_CHAR));
 //       grid line
 //
 //    2. if 'fallback' is 'true':
-//           the value specified on the commandline IFF the user did not specify the 
-//           option on the grid line but did specify the option on the commandline
+//           the value specified on the commandline if the user did not specify the 
+//           option on the grid line (regardless of whether they specified the option
+//           on the commandline).  In this case, if the user did not speify a value on
+//           the commandline, the commandline value is set according to the default
+//           behaviour for the option, and the grid line value is set from that.  Note
+//           that for options whose default behavious is to draw a random number, this
+//           will only be done once for the commandline value, and each grid line that
+//           falls back to the commandline will take the same value as the commandline.
+//           Consider using 'fallback' = 'false' for those cases.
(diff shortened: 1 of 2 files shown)


ilyamandel:
Hi @jeffriley ,

I am a bit confused.

Options.h describes OPT_VALUE as follows:

// To reiterate: by using the OPT_VALUE macro, the value of the option returned will
// be (in order of priority):
//
//    1. the value specified on the grid line IFF the user specified the option on the 
//       grid line
//
//    2. if 'fallback' is 'true':
//           the value specified on the commandline IFF the user did not specify the 
//           option on the grid line but did specify the option on the commandline
//       else if 'fallback' is 'false':
//           the default value for the option
//
//    3. the default value for the option

Specifying false would make COMPAS ignore the command-line value if the user does not specify the option in the grid file but does specify it on the command line.  Why is that the right approach in this case?  And how would it solve Aldana's issue (she wasn't specifying it on the command line)?

jeffriley:
@ilyamandel I understand your confusion, and it's precisely why it took me a little while to push this fix!  At first it seems a little counter-intuitive, but it's actually logical.  Here's what happens:

(a) In the case that fallback is TRUE:

If the option is not specified on the gridline, we fallback to the commandline.  If the option is not specified on the commandline, we fallback to the default value for the option.

(b) In the case that fallback is FALSE:

If the option is not specified on the gridline, we fallback directly to the the default value for the option.

For most options, (a) is the behaviour we want.  But not here - which is one of the reasons I implemented the fallback parameter in the first place (and why I should have known not to change the value to TRUE for these options).

For this option (any of the 'kick-magnitude-random*' variants), what behaviour is most logical (actually, what behaviour would the user most likely want)?  If the user specifies a value for the option on a grid line we'll use that, but if the users doesn't specify a value on the grid line, then what behaviour do we want (or do we think the user wants)?

If the user specified a value for the option on the command line, and we fallback to that, then every grid line that falls back gets the same value, and that's probably what the user wants if they specified (what will be) a default value on the commandline.  But if the use didn't specify a value for the option on the commandline, what happens if we fallback to the commandline?  The default behaviour is that we draw a random number - but we only do that once for the commandline, so then every grid line that falls back to the commandline gets the same value - the random number we drew for the commandline.  That's the behaviour Aldana was seeing.

The correct thing to do, in my opinion - and the thing I think users expect - is to not fallback to the default commandline value, but fallback directly to the default behaviour for the option, which is to draw a random value.  That way, each grid line that does not specify a value gets a different random number.  So fallback = FALSE is what we want here.

Seeing the documentation for the OPT_VALUE macro here reminded me that I should have changed it.  If you look at the functionality, this comment is wrong:

// 2. if 'fallback' is 'true':
// the value specified on the commandline IFF the user did not specify the
// option on the grid line but did specify the option on the commandline

We actually fallback to the commandline value if the user did not specify the option on the grid line, regardless of whether they specified a value on the commandline.  It's subtle, but it's the same thing in all cases expect this type of option (where the default behaviour draws a random number). I'll push a change t hat corrects that description.

ilyamandel (approved):
Thank you for the explanation, Jeff.  It was indeed the comment about falling back to the command line on false only if the command line contains that option which confused me.  I am approving, but please clarify that comment before pulling.

PR #1388, opened 2025-05-09
Added more outputs

ilyamandel:
- Added several outputs to provide more information and aid code testing: SEMI_MAJOR_AXIS_POST_STAGE_1_CE (for the 2-stage CE); BINDING_ENERGY_CONVECTIVE_ENVELOPE, CONVECTIVE_ENV_MASS, LAMBDA_CONVECTIVE_ENVELOPE (convective envelope binding energy,  mass, and lambda, respectively, Picker prescription)
- Ensure that the MANDEL main sequence core mass prescription never provides a core mass exceeding the total mass
- Corrected several erroneous comments, cleaned up a few function definitions
- Updated and corrected documentation


jeffriley (approved):
Thanks @ilyamandel  - looks good.  Just one suggested change:

On line 293 of `BaseStar.cpp`, use `std::ignore` (https://en.cppreference.com/w/cpp/utility/tuple/ignore)

I'll approve now so you can merge after making the change.

EDIT: also, looks like a typo in Options.h (behavious -> behaviour(s))

EDIT: wait - you didn't change Options.h - why are we picking up a spelling mistake there now?

ilyamandel:
Thank you for the std::ignore suggestion; implemented.  I think the behavious typo was actually from the change you (@jeffriley ) just made, and which I rolled in via "git merge dev" after I already started on my PR, so it shows up in the changes relative to when I originally branched off dev; anyway, fixed that too, so will merge now.

PR #1389, opened 2025-05-11
Corrected and simplified GiantBranch::CalculateRadialExtentConvectiveEnvelope()

ilyamandel:


jeffriley (approved):
Why bother with the local variable `convectiveCoreRadius`?

ilyamandel:
Thanks, @jeffriley .  I agree, that was a leftover, cleaned.

PR #1390, opened 2025-05-13
Tides updates

veome22:
There are a few major changes to the `KAPIL2024` tidal prescription here.
1. I realized I had been using the indices $l,m,n$ inconsistently in the expansions for equilibrium and dynamical tides. I have now renamed all the tidal Love number variables to have the indices $(n,m)$, and I have gone back and used $l=2$ everywhere. I also simplified the equations following the paper draft, which makes the computation ~30% faster.
2. Renamed the prescription to `KAPIL2025` to be consistent with the paper submission. This may break some other folks' code, but I hope not many people have been using this option before the companion paper comes out anyway.


ilyamandel (changes requested):
Hi @veome22 :

A suggestion: deprecate KAPIL2024 rather than deleting it completely; this way, users who pick it won't immediately get non-working code, not until we delete it.  @jeffriley already added functionality for this, you should be able to find an example to copy.

// (l=2, n=1, m=0), Gravity Wave dissipation from core boundary
I am a bit confused: can n be larger than l?

// (l=2, n=1, m=0), Gravity Wave dissipation from envelope boundary is always 0.0 since m=0.0
Does m in m=0.0 refers to something else than the azimuthal m=0?  Clarify

Looks great otherwise!

ilyamandel (commented):
Oh, one more thing: since the option name changed, can you re-create the .yaml file?

ilyamandel:
@veome22 : Do we have *any* outputs that could be put into, say, DetailedOutput that would describe the current strength of tidal coupling?  I think it would be quite useful to be able to print out the coupling strengths / rates of rotational period evolution...

veome22:
@ilyamandel,
>A suggestion: deprecate KAPIL2024 rather than deleting it completely; this way, users who pick it won't immediately get non-working code, not until we delete it. @jeffriley already added functionality for this, you should be able to find an example to copy.

Sure, deprecating the KAPIL2024 option sounds like a good idea. I will figure out how to do it, but it looks like I need to add an entry in `std::vector<std::tuple<std::string, std::string, std::string, bool>> deprecatedOptionValues` in `Options.h`.



> I am a bit confused: can n be larger than l?

Yes. $l$ and $m$ are the usual degree and order of the spherical harmonic potential of the tidally deformed body, and these indices follow the relation $|m| < l$. However, $n$ is the order of the orbital motion, and there is essentially no restriction on $n$ for eccentric orbits. According to [Ogilvie 2014](https://www.annualreviews.org/content/journals/10.1146/annurev-astro-081913-035941), if we truncate terms smaller than $e^2$, then we need to consider $n \leq l+2 = 4$. In practice we keep terms only up to $n=3$.



> Does m in m=0.0 refers to something else than the azimuthal m=0? Clarify

$m$ is the azimuthal wave number. I will clarify in the comments.



>Oh, one more thing: since the option name changed, can you re-create the .yaml file?

I'd be happy to, but I have no idea how to do this... Do I just run COMPAS with the `--create-YAML-file` argument?



> Do we have any outputs that could be put into, say, DetailedOutput that would describe the current strength of tidal coupling? I think it would be quite useful to be able to print out the coupling strengths / rates of rotational period evolution...

Yes, we actually have a lot of candidates. I have been storing the imaginary tidal Love numbers (`m_Star1->CalculateImKnmTidal`) from equilibrium (`CalculateImKnmEquilibrium`) and dynamical (`CalculateImKnmDynamical`) tides for my own testing. I could make any number of these outputs available for the DetailedOutput. 
`BINARY_PROPERTY::SYNCHRONIZATION_TIMESCALE` and `BINARY_PROPERTY::CIRCULARIZATION_TIMESCALE` already exist, but I believe they are calculated using Hurley+ 2002, and then never used. I could update those to be consistent with the KAPIL2025 model of tides, although they would be meaningless for NONE and PERFECT tides. The same is true for the Love numbers. Let me know if I should go ahead and add any of these.

ilyamandel:
Hi @veome22 ,

> m is the azimuthal wave number. I will clarify in the comments.

Great.  Maybe don't write it as 0.0 then -- the wave number is an integer, so this notation confused me into thinking m was something else.

On outputs: yes, I think it would be awesome to add a few additional outputs.  It's fine if some of them are 0 when your tidal model is not used, as long as that's well documented.  I'd add at least some of them to detailed output but not elsewhere -- people can decide themselves whether to add it elsewhere (can be done via log file specification at runtime).  We should definitely use your tides for BINARY_PROPERTY::SYNCHRONIZATION_TIMESCALE and BINARY_PROPERTY::CIRCULARIZATION_TIMESCALE rather than the Hurley ones that are never used -- that's just going to confuse people.  And whatever you print out already is probably the right thing to have available to other users, and would also make life easier for you in the long term (no need to modify the code by hand for your runs, which is good practice, anyway) -- so please do add them.  Thank you!

veome22:
Hi @ilyamandel and @jeffriley! I have made all the suggested changes, and also added the synchronization and circularization timescales + tidal quantities to the list of outputs. 

To store the tidal Love numbers as output variables, I have had to define different outputs for each star x each tidal harmonic, for a total of 8 variables per tidal dissipation mechanism. This means that there are probably a lot of superfluous lines of code, and the documentation might also get ugly. Before I finalize the tidal ImKnm components and add them to the online docs, I wanted to see if there's a better way to do what I've done.

`BaseStar::CalculateImKnmDynamical`, `BaseStar::CalculateImKnmEquilibrium`, etc. belong to the `BaseStar` class, which suggests that I should store them under stellar properties rather than binary properties. This would cut the number of variables by half. However, the only way to actually evaluate them (I think) is through `BaseBinaryStar`. This is because we need to pass the semi-major axis, orbital velocity, and companion mass to these functions. Since I don't know how to access binary properties or companion properties from BaseStar.cpp, I can't see a way to calculate and store the tidal Love numbers as `ANY_STAR_PROPERTY` instead of `BINARY_PROPERTY` variables. Is there an elegant way to evaluate the strength of tides in each star within `BaseStar.cpp`? Or should I just stick with my current version?

jeffriley (changes requested):
Thanks @veome22.  

I haven't completed my review yet, but here a a few things to think about while I think about how best to handle the the tidal Love numbers as output variables, and also complete the review of the rest of the code.

I suggest you change this type of construct:

    (w10 < 0.0 ? -std::abs(s10_8_3) : s10_8_3)

to:

    std::copysign(s10_8_3, w10)

throughout.  Note the `std::copysign()` function is for floating point values only, not ints.  We could trivially implement a template function to do the same for ints, but I don't think we need to just yet (at least I didn't see any).

So, no offense intended, but I do not like the variable names used in the tides functions :-).  I get why you've used the ones you have, and I have no suggestion for better names - I just don't like them :-), but could we change this type of thing:

    double one_minus_alpha_4 = oneMinusAlpha_2 * oneMinusAlpha_2;

to perhaps

    double oneMinusAlpha_4 = oneMinusAlpha_2 * oneMinusAlpha_2;

Our naming convention for variable names is to use camel case where possible, not snake case (I get the _# suffixes).

Where you have this sort of thing:

    double dynPrefactor     = 3.207452512782476;                                                        // 3^(11/3) * Gamma(1/3)^2 / 40 PI
    double m_l_factor_22    = 2.0 / (2.0 * (2.0 + 1.0)) / std::cbrt(2.0 * (2.0 + 1.0));                 // m * (l(l+1))^{-4/3}

it would be worth considering using `constexpr` for both.  The compiler will handle constants in an efficient way (or should), but something like `double x = 2.0;` is not guaranteed to be evaluated at compile time - the compiler might leave it to be evaluated at run time, which means it could be evaluated many times if is in a function that is called one or more times per timestep. `constexpr` is guaranteed to be evaluated at compile time, so changing the two statements shown above to use `constexpr` guarantees the evaluation and substitution is done once only - at compile time.

**Note to self: there are places in the code (probably throughout the code) where we do things like:

    double BaseStar::CalculateOStarRotationalVelocityAnalyticCDF_Static(const double p_Ve) {
    
        double alpha  = 4.82;
        double beta   = 1.0 / 25.0;
        double mu     = 205.0;
        double sigma  = 190.0;
        double iGamma = 0.43;
    
        boost::math::inverse_gamma_distribution<> gammaComponent(alpha, beta); // (shape, scale) = (alpha, beta)
        boost::math::normal_distribution<> normalComponent(mu, sigma);
    
        return (iGamma * boost::math::cdf(gammaComponent, p_Ve)) + ((1.0 - iGamma) * boost::math::cdf(normalComponent, p_Ve));
    }

We really should change those to `constexpr` too - so the evaluation and substitution is done once only (at compile time).  Not pointing any fingers - I am almost certainly responsible for a lot (if not most) of them because I think it actually makes the code clearer - I just didn't think about `constexpr` at the time.

More later.

ilyamandel (changes requested):
Program list options documentation currently says

>Options: { NONE, PERFECT, KAPIL2024 } |br|
>...
>``KAPIL2025`` uses the prescription from Kapil+ (2025). |br|
Should be KAPIL2025 under options.


When you have things like 
> case TIDES_PRESCRIPTION::KAPIL2025:
you want to have 
> case TIDES_PRESCRIPTION::KAPIL2024:
on the previous line so that (for now, while the option is deprecated but not removed) the same thing happens.

Happy to leave it to you and @jeffriley to decide on the new output names -- but whatever they are, they should be added to the documentation and explained there (I realise you were intentionally delaying this).   Also, the description of the synchronization timescale argument should be updated, since it's now correctly split into two.  

Oh, and I agree that it makes sense to treat all tides as binary properties.  There's no meaning to tides for single stellar evolution.

jeffriley:
@veome22 , @ilyamandel 

Just to clarify:

> When you have things like
>
> case TIDES_PRESCRIPTION::KAPIL2025:
> you want to have
> case TIDES_PRESCRIPTION::KAPIL2024:
> on the previous line so that (for now, while the option is deprecated but not removed) the same thing happens.

This is not true.

I have code in the Options class that parses the commandline (or grid file line) before we hand the commandline (or grid file line)  over to Boost - that parsing code is where I make substitutions for ranges and sets etc. (Boost knows nothing about ranges and sets for program options, so I needed to write code to expand the ranges and sets and pass each iteration of the expansions to Boost).  Same for shorthand options - that's something the Options code does before handing over to Boost (Boost knows nothing about shorthand).  That code also handles deprecated option names - I wrote code to look for deprecated option names, and/or option values, and replace the deprecated strings with the new values (given in the deprecation tables).

Bottom line is, if `KAPIL2024` is listed in the `deprecatedOptionStrings` table in `Options.h`, with `KAPIL2025` as the replacement, you only need to list `KAPIL2025` in the code - if the user enters `KAPIL2024` my code with change it to `KAPIL2025` and (for the first time it is changed) issue a warning to the user that they used a deprecated option (`KAPIL2024` never makes it past the Options code).

EDIT: Oh wait - `KAPIL2024` and `KAPIL2025` are option values for the `tides-prescription` option, so the entry should be in the `deprecatedOptionValues` table in `Options.h`  (my mistake). But the same applies - you don't need to account for it in the code (other than putting it in the deprecation table(s) in `Options.h`)

ilyamandel:
Thank you for the clarification, @jeffriley , and apologies for my ignorance!  
@veome22 , then my request boils down to a documentation update to explain the new outputs once you and @jeffriley agree on the names.

ilyamandel:
@jeffriley , @veome22 -- have you agreed on names for the new variables?  (Sorry, I don't mean to rush you, and I know Jeff is travelling... ;-) )

jeffriley:
@ilyamandel I'm still in Oz for a couple of days.  I don't think there's much we can do right now about the Love number constants - I think just leave them as @veome22 has them.  I'll just make sure I don't want any other changes and submit my review later tonight or tomorrow.

veome22:
Thanks @ilyamandel and @jeffriley ! I will also be busy starting tomorrow, so I will take some time in the next few hours to roll in all the suggested changes from before and the tidal outputs. I'll let you know when the PR is ready for final approval!

jeffriley (approved):
Ok, I think we're good.  Can we just make sure that each of the functions that has magic numbers/constants in it is documented showing the paper(s) that the constants come from?  You may have already done that - I haven't checked them all.  I'll approve pending fixing the things already pointed out.

EDIT: and just in case, you'll need to merge dev, and then a new yaml file, fix `changelog.h` and `what's new`.

veome22:
All done! Also, I like v3.20.00 as the version to base the new COMPAS methods paper on 😃.
Thanks for the help @jeffriley

ilyamandel (approved):
Looks great, @veome22 !

Though the version for the paper will be a slightly later one, with the WD PR rolled in. :)

PR #1394, opened 2025-05-21
Allow binned-cosmic-integrator to also include BNS, BHNS

avivajpeyi:
Main changes were to the 'BBHPopulation' class. This is now a `binary_population`. 

User can select which set of DCOs from a list of valid DCOs they'd like to be included in the binned cosmic integrator (different from the FastCosmicIntegrator, which does this in a vectorized format, but not memory efficient).
```
DCO_GROUPS = dict(
    BBH=[BH, BH],
    BNS=[NS, NS],
    BWD=[WD, WD],
    NSBH=[NS, BH],
    WDNS=[WD, NS],
    WDBH=[WD, BH],
)
```


avivajpeyi:
Hmmm seems like the `fake_compas_output` needs to include 'Mass@ZAMS(1)' 


```
def test_analytical_vs_numerical_star_forming_mass_per_binary(fake_compas_output, tmpdir, test_archive_dir):
        np.random.seed(42)
        m1_max = M1_MAX
        m1_min = M1_MIN
        m2_min = M2_MIN
        fbin = 1
    
>       numerical = star_forming_mass_per_binary(fake_compas_output, m1_min, m1_max, m2_min, fbin)

py_tests/test_total_mass_evolved_per_z.py:70: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
compas_python_utils/cosmic_integration/totalMassEvolvedPerZ.py:1[50](https://github.com/TeamCOMPAS/COMPAS/actions/runs/15174806606/job/42672872620#step:8:51): in star_forming_mass_per_binary
    multiplicationFactor, _ = totalMassEvolvedPerZ(**locals())
compas_python_utils/cosmic_integration/totalMassEvolvedPerZ.py:136: in totalMassEvolvedPerZ
    MassEvolvedPerZ = retrieveMassEvolvedPerZ(path)
compas_python_utils/cosmic_integration/totalMassEvolvedPerZ.py:113: in retrieveMassEvolvedPerZ
    m1s = (allSystems['Mass@ZAMS(1)'])[()]
...

>   ???
E   KeyError: "Unable to synchronously open object (object 'Mass@ZAMS(1)' doesn't exist)"

```

SimonStevenson (changes requested):
Thanks for the changes Avi! I've checked that CosmicIntegration.py runs ok and produces the demo outputs. It doesn't really test any of the new functionality though (i.e., for BNS or WD binaries; by default it focuses on BBHs as before). I wonder if the example population there could be made to demonstrate some more of this functionality.

What was the main motivation for these changes? To keep up with #1368 ? Or is this independent of that?

It might be useful to explicitly set the dcos_included=['BBH'] option for BinaryPopulation.from_compas_h5 in the CosmicIntegration.py demo so that it is obvious that is where the selection happens. I didn't spot that immediately. 

Given the additions of WD binaries in BinaryPopulation, should there be frac WDWD, frac_WDNS and frac_WDBH in generate_mock_population similar to frac_BNS etc? I guess those can be 0 by default. 

If so, we will then also need to check that Pdet = 0 for WD containing binaries, since Pdet is explicitly for LIGO-type detectors that will not detect WD binaries. We should still be able to calculate cosmological formation rates for these though. I've raised a similar concern in #1368 

That's all I spotted, once you've answered those I'm happy to approve

avivajpeyi:
Hey Simon! Thanks for the quick review :)

> What was the main motivation for these changes?

There were two reasons:

1. In Jeff and Ilya’s 2023 paper on the NN surrogate for the cosmic integrator, Jeff used BBH, NSBH, and BNS populations. Until now, my followup work with Jeff (using GPs) has only involved BBHs, so I need to extend the code to support BNS and NSBH.

2. I added WDs to keep the binned-cosmic-integrator aligned with Melanie’s work :) 

> It might be useful to explicitly set the dcos_included=['BBH'] option...

Good call! I’ll make that change in the CosmicIntegration.py demo so it’s clear where the selection happens.


**Re: WD binaries** 
Would it be easier if i disable the WDs for now / raise an error if they are used? 


Thanks again! I’ll push these updates shortly.

avivajpeyi:
BTW @SimonStevenson  -- the 'CosmicIntegration.py' is run by the CI as well  -- in case you don't want to manually rerun it 
https://github.com/TeamCOMPAS/COMPAS/blob/dev/py_tests/test_example_notebooks.py

SimonStevenson (approved):
Thanks Avi!

SimonStevenson:
>  the 'CosmicIntegration.py' is run by the CI as well -- in case you don't want to manually rerun it
https://github.com/TeamCOMPAS/COMPAS/blob/dev/py_tests/test_example_notebooks.py

Good to know!

PR #1395, opened 2025-05-22
Added functionality to create new System Snapshot logfile

jeffriley:
Added functionality to create new System Detailed Output logfile
Writing to the System Detailed Output logfile is triggered by system age and/or simulation time passing thresholds set by new program options (see below).  New program options added:

`--logfile-system-detailed-output`: specifies the name of the System Detailed Output logfile (default is "BSE/SSE]_System_Detailed_Output")
`--logfile-system-detailed-output-record-types`: specifies the enabled record types for System Detailed Output logfile (default is all types)
`--system-detailed-output-age-thresholds`: specifies the age thresholds for System Detailed Output logfile
`--system-detailed-output-time-thresholds`: specifies the time thresholds for System Detailed Output logfile

Removed stashing to SSE SNe file - printing is now done directly (stashing no longer required since we now don'r save/revert).

@ilyamandel : age thresholds are compared to `m_Age`. Time threshold are compared to `m_Time`. We can add an option for `m_Tau` - or swap with `m_time` if you want.

The age and  time  threshold options can't use ranges/sets - each iteration of ranges and sets generates a new binary/star, so that wouldn't work.  I've made them vector options (see https://compas.readthedocs.io/en/latest/pages/User%20guide/Program%20options/program-options-vector-options.html).

If you just want to specify one threshold value, you can use (e.g.) `--system-detailed-output-age-thresholds 5.2`, but if you want to specify several age thresholds you would use (in shorthand notation) e.g. `--system-detailed-output-age-thresholds [5.2,12.7]`

See documentation of new options for how they work - also documentation in the code.

As I'm typing this I remembered I wanted to sanity check the the threshold values the user input - I check for >= 0.0, but I wanted to put the upper bound as `OPTIONS->MaxEvolutionTime()`, but forgot.  I can do that while you're reviewing.

I think I have everything, but could easily have forgotten something.  

EDIT: btw,  if this is the functionality you want (or if we make it the functionality you want), we can add new options for thresholds on any attribute of a binary or star - so we can trigger writing of a record based on any attribute.

EDIT: new filename and option names:

`--logfile-system-snapshot-log`: specifies the name of the System Snapshot logfile (default is `[BSE/SSE]_System_Snapshot_Log`)
`--logfile-system-snapshot-log-record-types`: specifies the enabled record types for System Snapshot logfile (default is all types)
`--system-snapshot-age-thresholds`: specifies the age thresholds for System Snapshot logfile
`--system-snapshot-time-thresholds`: specifies the time thresholds for System Snapshot logfile


ilyamandel (commented):
I wonder if "system detailed output" is too similar to "detailed output" and may cause confusion.  What if we instead call it something like "snapshot" (since we are doing snapshots at particular times)?

I hadn't even thought of using age rather than time as a check for printing; I can't think of a use case for that now, but let's keep it -- maybe one will present itself :-)

Why does it look like you deleted switchlog from standard-logfiles-record-types.rst?

changelog.h: "since we now don'r save/revert": don'r -> don't

ilyamandel:
For the future, the option of thresholding on any feature sounds cool.  For example, BH HMXBs are really thresholds on a MS star that's 80% Roche Lobe filling (well, and the companion being a BH, so maybe not so simple...).  But let's leave this until we have a clear use case. :-)

jeffriley:
Thanks @ilyamandel . I am very open to a new name - I didn't like system-detailed-output.  I meant to comment on that, but I was in a hurry.  Happy to take suggestions - "system-snapshot-log" works for me - ok?

I was't sure whether you wanted age as well as or instead of time - it wasn't hard to put it in and it serves as a template for other stuff maybe.  Happy to add more options for other attributes as we need them - thr structure is in place now.

I may have accidentally deleted the switchlog file standard-logfiles-record-types.rst when I was making my changes - I'll take a look. 

I'll try to get this done tonight.

ilyamandel:
I like system-snapshot-log! :) 
Ping me once you've made the changes and I should be able to approve promptly, since I've now reviewed everything else.
[Disclaimer: to be honest, I have not reviewed the technical changes -- e.g., to Log.cpp and Log.h -- carefully; I presume the proof is in the pudding here.  I did review the logic for deciding on when to log, and it looks fine.]

jeffriley:
@ilyamandel Changes per your review pushed.  I think I got everything, but it would be prudent to run some tests :-)

> Why does it look like you deleted switchlog from standard-logfiles-record-types.rst?

We don't have record types for the switch-log files - that's why it is missing (it was never there).

ilyamandel (approved):
Thank you, @jeffriley !

PR #1397, opened 2025-05-30
Included unit conversion in WhiteDwarfs::CalculateEtaPTY()

ilyamandel:
- Included unit conversion in WhiteDwarfs::CalculateEtaPTY() [Piersanti+ 2014 expects the input in units of 10^-8 Msun/yr, the code was using the value in units of Msun/Myr, despite the incorrect comment that it was in Msun/yr]
- All critical mass ratios now return the HURLEY_HJELLMING_WEBBINK_QCRIT_WD for white dwarfs and 0 for other remnant donors (only stable mass transfer -- though we should never have mass transfer from neutron stars or black holes :) ) as fix for issue #1385


brcekadam:
Thanks for the fix, @ilyamandel! 

Just one thing, there is a forgotten `double` in `Remnants.h` -- that's probably why the compile test fails.

ilyamandel:
Oops, that's what I get for trying to do two things at once.  Fixed, thank you, @brcekadam !

nrsegovia:
> * Included unit conversion in WhiteDwarfs::CalculateEtaPTY() [Piersanti+ 2014 expects the input in units of 10^-8 Msun/yr, the code was using the value in units of Msun/Myr, despite the incorrect comment that it was in Msun/yr]

Oops, my bad.

>     * All critical mass ratios now return 0 for remnant donors (only stable mass transfer) [resolves issue [Ge stability criteria accidentally applied to WDs #1385](https://github.com/TeamCOMPAS/COMPAS/issues/1385) ]

Wouldn't it be better to default to the `HURLEY_HJELLMING_WEBBINK` value in the case of WDs? That is closer to the "common" value (e.g., Marsh+ 2004).

ilyamandel:
Great suggestion, @nrsegovia !  Done.

ilyamandel:
@nrsegovia -- if you are OK with this change, please feel free to approve. :)

nrsegovia (approved):
@ilyamandel looks good to me :)

PR #1398, opened 2025-05-30
Fix for online docs compile failure

jeffriley:
The onlie docs are currently not compiling.  The problem is an error in the formatting of file

../online-docs/pages/User guide/COMPAS output/standard-logfiles-record-specification-binary.rst

introduced in https://github.com/TeamCOMPAS/COMPAS/pull/1390 (spaces at start of the "..flat-table::" declaration for SYNCHRONIZATION_TIMESCALE_2)

This change should fix the problem (the docs compile ok locally).

Just a reminder to check compilation of the docs locally after making changes - instructions are pinned to the `devel_compas_documentation` slack channel.


veome22 (approved):
Oops, thank you for fixing this!

Issue #1305, opened 2024-12-05
Incorrect orbital evolution on mass loss


ilyamandel:
When a binary (no tides, no GWs, no supernovae, no mass transfer) loses mass in winds or pulsations on timescales much longer than the orbital timescale, M_total * a should be conserved.  I've tested this before, but now it's not conserved, again. :(

The following example was created with `./COMPAS -n 1 --detailed-output --random-seed 1733362410 `

Part of the issue is that the binary evolution on wind mass loss is unnecessarily approximate, yielding wiggles (not visible on the plot without zooming in).  I fixed that in #1304  (draft for now).  But there's clearly a bigger issue when the star loses a lot of mass in pulsations on the TPAGB, leading to full envelope loss.  To be fixed [I'll work on this, just listing @SimonStevenson , @reinhold-willcox , @jeffriley FYI .]

![image](https://github.com/user-attachments/assets/cba95044-401b-484c-8929-698965a07d5d)

ilyamandel:
Fix completed, maximum fractional change in a*M is now 0.0few percent (see plot below).

The main issue was that we would sometimes create a WD as an outcome of a supernova: if the Hurley criteria called for an ECSN, but we didn't meet the mass transfer history threshold for an ECSN proposed by @reinhold-willcox , ResolveElectronCaptureSupernova() would yield an ONeWD.  That is inconsistent with the Hurley evolution in the HR diagram, and is also inconsistent with the idea that a WD is formed by gradual mass loss, not a rapid, explosive SN.  For now, the fix is that if the ECSN criteria are not met, we go through a "standard" SN.

ilyamandel:
![image](https://github.com/user-attachments/assets/8bb68ee4-2ccb-4bd3-bb02-50b4f6becf37)

PR #1400, opened 2025-06-18
Resolves an issue that appeared in #1305 with some TPAGB stars in the ECSN range (but not satisfying ECSN conditions) exploding as CCSNe

ilyamandel:
In #1305, I tried to address a situation where TPAGB stars entered the ResolveSupernova() function but were not, in fact, supposed to explode in a supernova because they did not satisfy the ECSN criteria (most likely because --allow-non-stripped-ECSN was false).  Prior to that fix, we allowed these to become white dwarfs, but that created problems such as the impact of a sudden mass change associated with instantaneous envelope removal from the binary (e.g., a Blaauw kick, a non-conservation of a*M, etc.).  In that fix, I opted to treat all systems that did not qualify for ECSN as regular core-collapse SNe.  However, as @pauldisberg pointed out, this created a different set of problems, such as an abundance of low-kick NSs forming from stars that really should not have experienced CCSNe at all.  In this update, I am fixing the treatment properly, by modifying the calculation of whether a star should go supernova to return false in cases when neither the ECSN nor CCSN criteria are satisfied.  To achieve this, I added a new parameter (threshold mass, generally expected to be MCH or MECS) to CalculateCoreMassAtSupernova_Static().  I also removed McSN from GBParams, instead computing it on the fly when needed.

Finally, I cleaned up documentation in a couple of places, such as adding output column descriptions directly to Quick Links for easier navigation.

I am also starting to upload some more Matlab post-processing scripts; these do not require review (and are still works in progress).

### DIFF
--- a/compas_matlab_utils/BSEDetailedOutput.m
+++ b/compas_matlab_utils/BSEDetailedOutput.m
@@ -1,0 +1,72 @@
+file='~/Work/COMPAS/src/COMPAS_Output/Detailed_Output/BSE_Detailed_Output_0.h5';
+
+tic
+time=h5read(file,'/Time');
+MThistory=h5read(file,'/MT_History');
+Z=h5read(file,'/Metallicity@ZAMS(1)');
+mass1=h5read(file,'/Mass(1)');
+massCO1=h5read(file,'/Mass_CO_Core(1)');
+mass2=h5read(file,'/Mass(2)');
+massCO2=h5read(file,'/Mass_CO_Core(2)');
+massHe1=h5read(file,'/Mass_He_Core(1)');
+massHe2=h5read(file,'/Mass_He_Core(2)');
+luminosity1=h5read(file,'/Luminosity(1)');
+luminosity2=h5read(file,'/Luminosity(2)');
+type1=h5read(file,'/Stellar_Type(1)');
+type2=h5read(file,'/Stellar_Type(2)');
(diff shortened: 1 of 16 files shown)


jeffriley (approved):
Looks good - approved

ilyamandel:
Just to prove that all is still well with conservation of a*Mtot for binaries making WDs without tides, GWs, or mass transfer (plot made with latest code version):

image

pauldisberg (approved):
Approved: NS kick distributions look good, the peak caused by stars that would have gone ECSN if they were stripped but now undergo CCSNe is reduced.

ilyamandel:
Thanks, @pauldisberg !  If you happen to have a plot showing the impact of this change (say, the single NS kick distributions before and after the change with default assumptions, i.e., without allowing for ECSNe from non-stripped donors), could you post it here -- useful to have such information handy when we try to remember a year later why we made this change. :-)

pauldisberg:
v_kick_comparison
As requested: the kick distributions for individual NSs (through ./COMPAS -n 100000 --mode SSE --random-seed 1714388113). The figure shows interpolated histograms with bins of 10 km/s, for varying v_NS and sigma_kick (Mandel & Müller 2020), for the COMPAS version before Ilya's fix (v03.20.02) and the version after Ilya's fix (cf. Fig. 2 of Kapil et al. 2023).

PR #1401, opened 2025-06-23
Fixes to CHE and MS mergers with BRCEK core mass prescription

brcekadam:
CHE and MS mergers were not behaving properly when BRCEK core mass prescription is used, so I made a few changes:

- Core mass calculations are now disabled for CH stars, and Hurley's luminosity prescription is always used (unless corrected via --enhance-CHE-lifetimes-luminosities)
- Total hydrogen mass after MS merger was not calculated correctly
- We now have a way of determining the initial convective core mass after full mixing, which happens after MS merger or in a CH star that spun down


ilyamandel (changes requested):
BaseBinaryStar::ResolveMainSequenceMerger()::
 
        // final hydrogen mass in the merger product, assuming that the lost mass comes from the envelope
        finalHydrogenMass      = hydrogenMass1 + hydrogenMass2 - (m_Star1->Mass() + m_Star2->Mass() - finalMass) * initialHydrogenFraction;
        
Would it be safer to user HydrogenAbundanceSurface(), in case that's different from the initial hydrogen fraction because of previous MT history, or because stars used to be CHE before merging?  Though I guess in that case it's not clear which star's envelope  mass is removed from -- maybe the one with the higher hydrogen surface abundance under the assumption that's the one with a fluffier envelope?

ilyamandel (commented):
MainSequence::CalculateInitialMainSequenceCoreMass()
The two prescriptions are potentially discontinuous for stars at ZAMS and just after ZAMS; should we just use @ryosuke-hirai 's prescription for all stars?

brcekadam:
@ilyamandel 

That's a good point about the hydrogen mass calculations! Now that I think about it, there are a few things that need adjusting in the hydrogen mass calculation for the individual stars. It currently doesn't take into account that the total mass can drop below the initial core mass. It should probably look something like this

`hydrogenMass = HydrogenAbundanceCore() * MainSequenceCoreMass() + (HydrogenAbundanceCore() + HydrogenAbundanceSurface()) * (std::min(InitialMainSequenceCoreMass(), Mass()) - MainSequenceCoreMass()) / 2.0 + initialHydrogenFraction * std::max(Mass() - InitialMainSequenceCoreMass(), 0.0);`

Then for the final hydrogen mass we could use

`finalHydrogenMass = hydrogenMass1 + hydrogenMass2 - (m_Star1->Mass() + m_Star2->Mass() - finalMass) * std::max(m_Star1->HydrogenAbundanceSurface(), m_Star2->HydrogenAbundanceSurface());`

Concerning your second comment, I kept it as two separate prescriptions to keep the full Shikauchi+ framework consistent. @ryosuke-hirai's prescription doesn't have a metallicity dependance and the MESA models were done for a slightly different overshooting parameter. There could potentially be a discontinuity for core masses of MS mergers at ZAMS and just after ZAMS, but this still might be the safer option than changing the initial core mass prescription for all stars.

ilyamandel:
Sounds good, @brcekadam -- ping me once you've made/tested that modification, and I'll promptly approve.

brcekadam:
Thanks @ilyamandel, I just pushed the change!

ryosuke-hirai:
I agree with @brcekadam that we can keep the Shikauchi formula for now, but we could update everything to the f_mix(Mmix,Yc) form down the line.

ilyamandel (approved):
Hi @brcekadam ,

Looks good, approved.

I haven't pulled because I am not certain whether

double hydrogenMass1 = m_Star1->HydrogenAbundanceCore() * m_Star1->MainSequenceCoreMass() + (m_Star1->HydrogenAbundanceCore() + m_Star1->HydrogenAbundanceSurface()) * (std::min(m_Star1->InitialMainSequenceCoreMass(), m_Star1->Mass()) - m_Star1->MainSequenceCoreMass()) / 2.0 + initialHydrogenFraction * std::max(m_Star1->Mass() - m_Star1->InitialMainSequenceCoreMass(), 0.0);

the last mention of initialHydrogenFraction should be m_Star1->HydrogenAbundanceSurface() , and similarly for star 2.  Feel free to adjust if you think it's warranted and then merge, or just merge directly.

brcekadam:
You're absolutely right, @ilyamandel! I'll fix that.

PR #1404, opened 2025-06-25
Update to Maxwellian kicks; new log normal (Disberg&Mandel 2025) kicks

ilyamandel:
- The MAXWELLIAN NS CCSN kick changed from the Hobbs value of 265 km/s to 217 km/s based on 48 younger than 10 Myr pulsars with proper motions from Disberg & Mandel (2025) sample; corrects Hobbs+ 2005 missing Jacobian
- Implemented a LOGNORMAL NS CCSN kick magnitude distribution based on Disberg & Mandel, 2025


jeffriley (approved):
All good from me - approved

SimonStevenson:
I know this has been merged already, but here is my test plot. Made using a population of single stars.
compare_LOGNORMAL_MAXWELLIAN_kick_dist_cdf

```
./COMPAS --mode SSE -n 50000 --kick-magnitude-distribution MAXWELLIAN --initial-mass-min 8.0 --initial-mass-max 30.0
```
and
```
./COMPAS --mode SSE -n 50000 --kick-magnitude-distribution LOGNORMAL --initial-mass-min 8.0 --initial-mass-max 30.0
```

ilyamandel:
Thank you, @SimonStevenson !

Your plots are consistent with the curves used when making the fits (attached).

image

One caveat is that Disberg & Mandel (2025) recommend cutting off the distribution at 1000 km/s.  We don't currently do this in the COMPAS code.  @pauldisberg -- should we add that cut?

pauldisberg:
@ilyamandel Yes, I think adding the cut would be a good idea, because (1) kicks above 1000 km/s are not really expected based on SN simulations, (2) we don't find any pulsar with a maximum likelihood velocity above 1000 km/s, and (3) the log-normal part above 1000 km/s is relatively small anyway.

ilyamandel:
@pauldisberg -- OK, I'll add that cut in a future update.

ilyamandel:
Actually, just posted the fix now, before I forget -- see #1405 .

PR #1406, opened 2025-06-25
Fix for issue #400

### LINKED ISSUE #400: Clarify reference metallicity for SSE models and winds
ZSOL in constants.h refers to the Hurley models, and should really be called ZREFERENCEHURLEY or some such, not ZSOL, to avoid confusion.  For winds, we should go back to the individual papers — e.g., the WR scaling according to Vink & De Kotter (2005) is a relative scaling with metallicity, so the exact value of ZSOL doesn’t matter.  Of course, it does matter for the prefactor — which, in Vink+ 2001, is taken as Zsol=0.019.  So, if we want to fairly use these wind models, we should scale them to whatever Zsol was assumed there.  [For WR winds, we already have a free multiplicative factor, so I am not too worried about those.]. Anyway, sounds like we should add an issue to go through our wind models and check that we are using the appropriate reference metallicity for each one — and do away with the ambiguous ZSOL.

brcekadam:
Fix for issue #400. I removed the ambiguous ZSOL value, and now we will have `ZSOL_HURLEY = 0.02`, `ZSOL_ANDERS = 0.019`, `ZSOL_ASPLUND = 0.0142`. I checked the prescriptions and made sure that correct Zsol value is used (hopefully I didn't miss any).


jeffriley (dismissed):
Thanks @brcekadam.  Just one request: 

Can you provide functions for `ZSOL_ANDERS` and `ZSOL_ASPLUND` that mirror `LogMetallicityXi()` for `ZSOL_HURLEY`, and use those? Otherwise we might be taking the log of the same number many thousands of times over the course of the evolution of each system - when we really only need to do it once (log is an expensive function).

You can change the name of the function if you want - it's `LogMetallicityXi()` because Hurley used 'xi' in his paper (from memory).  Or maybe just use `LogMetallicityXi_HURLEY()` etc. Whatever works.

brcekadam:
That's a good point, @jeffriley! I've just pushed the changes, and it should be fixed :-)

ilyamandel (approved):
Looks great, @brcekadam !

Apologies for merging in another PR just before yours (you'll have to merge in dev and increment the version number in order to avoid a changelog conflict).

brcekadam:
No worries, @ilyamandel! :-) It should be ready now.

PR #1410, opened 2025-07-17
Enhancements- deprecated MERRITT2024 in favour of , plus others

jeffriley:
Enhancement:
- Changed mass loss prescription MERRITT2024 to MERRIT2025; deprecated MERRIT2024
- Added datestrings to vectors supporting deprecation in Options.h to allow timely removal
- Added version strings for gsl, boost, and HDF5 to COMPAS splashscreen (for now, gsl & hdf5 are installed versions, boost is version compiled with COMPAS)


ilyamandel (approved):

PR #1411, opened 2025-07-21
Fixed random kick draw for MULLERMANDEL prescription

ilyamandel:


pauldisberg (approved):
image
Kicks generated with `./COMPAS -n 100000 --mode SSE --main-sequence-core-mass-prescription BRCEK --muller-mandel-kick-multiplier-NS 630 --muller-mandel-sigma-kick-NS 0.45 --random-seed 1714388113` (black histogram) match the manually calculated kicks using the Mandel & Müller prescription. Approved

PR #1419, opened 2025-08-08
Changes for isuues 1413 and 1414

### LINKED ISSUE #1413: Minimum secondary mass
We don't want to stop on a failed check against minimum secondary mass (which should just be used for a sampling range) when the user provides binary parameters explicitly; we should only check against MINIMUM_INITIAL_MASS .

Meanwhile, the MINIMUM_MASS_SECONDARY constants is not used and should be removed altogether.

### LINKED ISSUE #1414: Change to default record types in output files
We define record types for the various output files.  Currently only the detailed output and pulsar evolution files have multiple record types defined, so this issue pertains to those files, but the solution implemented should allow for record types to be added to the other files.

Record types are described in the online documentation at:

https://compas.readthedocs.io/en/latest/pages/User%20guide/COMPAS%20output/standard-logfiles-record-types.html

Currently, the default is that all record types are enabled for all files.  This can lead to a substantial increase in size, particularly for the detailed output files.  Not all users always want all record types.  The initial impetus for including intra-timestep record types in the detailed output files was two-fold:

(a) some users expressed a desire to see the values of some stellar/binary attributes at various times *during* a timestep, and
(b) they can be a good debugging resource

Users can turn record types that they don't want off - see the explanation at the bottom of the page linked above.  For example, for the detailed output files, if a user just wants record type 4 (`TIMESTEP_COMPLETED`), they would use `--logfile-detailed-output-record-types 8` (2^(4-1)).  The other record types they might want are `INITIAL_STATE` (type 1) - this is the state before the first timestep (which won't be included when just type 4 is specified), and `FINAL_STATE` (type 5).  `FINAL_STATE` will be the same as the last `TIMESTEP_COMPLETED` record, but it's there for completeness (and in case someone turns type 4 off.  No good reason to turn 4 off, but it can be...).  To include `INITIAL_STATE`, `TIMESTEP_COMPLETED`, and `FINAL_STATE` the user would use `--logfile-detailed-output-record-types 25`  (2^(1-1) + 2^(4-1) + 2^(5-1)).

I think it would be better to set the default record types to the ones we think will commonly be used (e.g., for detailed output files, types 1, 4, and 5, as above), and let users add record types (via the mechanism described above and on the page linked above) whever they need/want them.

My proposal would be to change the default record types for:

(a) the detailed output files to types 1, 4, and 5
(b) the pulsar evolution files to type 3 (and maybe make the record type names consistent between SSE and BSE)

Thoughts?

jeffriley:
Changes for issue 1413:
- Deprecated options
      . "--minimum-mass-secondary" in favour of "--minimum-sampled-secondary-mass"
      . "--initial-mass-min"       in favour of "--initial-mass-function-min"
      . "--initial-mass-max"       in favour of "--initial-mass-function-max"
      . "--initial-mass-power"     in favour of "--initial-mass-function-power"
 - Default value for "--initial-mass-function-min" remains at 5.0 Msol
- Default value for "--initial-mass-function-max" remains at 150.0 Msol
- Default value for "--initial-mass-function-power" remains at 0.0
- Default value for "--minimum-sampled-secondary-mass" remains at 0.1 Msol
- User supplied value for "--minimum-sampled-secondary-mass" checked against values of constants MINIMUM_INITIAL_MASS and MINIMUM_INITIAL_MASS
- Secondary mass, whether input by user, sampled, or calculated from M1 & q, checked against constant MINIMUM-INITIAL-MASS
 - Online docs "program-options-list-defaults" and "standard-logfiles-record-specification-optios" pages updated
- Whatsnew page updated

                                          
Changes for issue 1414:
- Changed [SSE/BSE]_PULSAR_RECORDTYPE::DEFAULT to [SSE/BSE]_PULSAR_RECORDTYPE::PRE_SN
- don't want DEFAULT anymore, but wanted to preserve numbering.  Neither PRE_SN nor POST_SN are currently used (POST_SN was), so we could rename them one day.
- Removed default parameter from BaseStar::PrintPulsarEvolutionParameters() and BaseBinaryStar::PrintPulsarEvolutionParameters()
- Changed default record types for:
      - SSE and BSE detailed output files to include only record types 1, 4, & 5 (INITIAL_STATE, TIMESTEP_COMPLETED, and FINAL_STATE)
      - SSE and BSE pulsar evolution files to include only record type 3 ((pulsar) TIMESTEP_COMPLETED)

@ilyamandel I think I captured what you wanted for issue 1413 (I was interrupted and came back to the code changes a few times, but hopefully I remembered where I was when I was interrupted each time...).  Let me know if I've forgotten anything or missed the mark.


jeffriley:
I'm suspicious of what checks were actually run...

ilyamandel:
Thanks, @jeffriley, will check tomorrow.

ilyamandel (approved):
Looks great, @jeffriley !  Just two minor things prior to merging this in:

The program option documentation for --initial-mass-function mentioned --initial-mass-function-min (and -max) only for UNIFORM; they are also used for SALPETER, KROUPA, and POWERLAW -- so would be more consistent to mention them for all or for none

At the bottom of the same file (online-docs/pages/User guide/Program options/program-options-list-defaults.rst), where we list all options, initial-mass-power should now be initial-mass-function-power

jeffriley:
Thanks @ilyamandel - I knew I'd miss something in the documentation :-)

I know my changes compile without errors (or warnings) here, but we don't seem to be doing the build check in github.  I'm happy to merge (when I've fixed the documentation) - are you?

ilyamandel:
I am happy to merge too; in fact, merging now. :-)

PR #1420, opened 2025-08-09
Minor fixes

### LINKED ISSUE #1211: Do we need --mass-loss-prescription NONE (now ZERO)?

jeffriley:
(i) Deprecated option ``-use-mass-loss`` in favour of ``--mass-loss-prescription``:

Instead of using ``--use-mass-loss`` or ``--use-mass-loss true`` to enable mass loss, then specifying the mass loss prescription to be used with ``--mass-loss-prescription``, mass loss can be enabled using ``--mass-loss-prescription`` with any valid prescription (that is not ``zero``), and disabled with ``--mass-loss-prescription zero`` instead of ``use-mass-loss false``.

Note: users might see a few odd error message if they use ``--use-mass-loss`` and ``--mass-loss-prescription`` in the same command.  Errors they may see are:

"Commandline Options error: option '--mass-loss-prescription' cannot be specified more than once"
"Commandline Options error: the required argument for option '--mass-loss-prescription' is missing"

Both are because the commandline parsing code in COMPAS replaces deprecated options before handing the commandline over to boost to parse for options, so we may have duplicated ``--mass-loss-prescription``, and when we do that we may leave it without a parameter (because ``-use-mass-loss`` doesn't need "true" (though it can have it).

I could spend some time and energy (not insignificant) preventing those errors, but I don't think they'll cause too much confusion, and should be rarely seen.
     
(ii) Added compiler flag "-Wno-vla-cxx-extension" to "CXXFLAGS" in Makefile to suppress compiler extension warning

(iii) Fixed online docs for omissions in v03.22.02:
         - fixed description for "--initial-mass-function" in "program-options-list-defaults.rst", and
         - changed "--initial-mass-power" to "--initial-mass-function-power" in "program-options-list-defaults.rst"

### DIFF
--- a/compas_python_utils/preprocessing/compasConfigDefault.yaml
+++ b/compas_python_utils/preprocessing/compasConfigDefault.yaml
@@ -1,5 +1,5 @@
 ##~!!~## COMPAS option values
-##~!!~## File Created Fri Aug  8 15:57:19 2025 by COMPAS v03.22.02
+##~!!~## File Created Sun Aug 10 13:16:13 2025 by COMPAS v03.23.00
 ##~!!~##
 ##~!!~## The default COMPAS YAML file (``compasConfigDefault.yaml``), as distributed, has
 ##~!!~## all COMPAS option entries commented so that the COMPAS default value for the
@@ -25,7 +25,6 @@ booleanChoices:
 
     ### STELLAR PROPERTIES
 #    --check-photon-tiring-limit: False                                    # Default: False
-#    --use-mass-loss: True                                                 # Default: True
 #    --enable-rotationally-enhanced-mass-loss: False                       # Default: False
 #    --enhance-CHE-lifetimes-luminosities: True                            # Default: True
 #    --expel-convective-envelope-above-luminosity-threshold: False         # Default: False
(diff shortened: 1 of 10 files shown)


ilyamandel (approved):
Looks great, thanks, @jeffriley !  

I removed a repeated line from an earlier PR from changelog.h .

One optional suggestion: I think removing the use-mass-loss option should lead to an updated yaml.h and default yaml file.  Yes, I realise the option is only deprecated, so nothing would go wrong with the old yaml file -- but I think of the yaml file as an "advertisement" of the available options, and if we don't want users to use deprecated options any more, we shouldn't advertise them.

jeffriley:
I agree re yaml file @ilyamandel - completely slipped my mind.  I'll push a new yaml file a bit later.

ilyamandel:
P.S. One more thing: please remove --use-mass-loss from the list at the bottom of program-options-list-defaults.rst

PR #1422, opened 2025-08-18
MALTSEV and Piersanti fixes

ilyamandel:
- In the MALTSEV SN prescription, treat wind-stripped stars as if they experienced case B mass transfer
- Limit the output of CalculateEtaPTY() [Helium accretion efficiency onto WDs from Piersanti+ 2014, A3] to be in [0,1]


jeffriley (approved):
All good.

PR #1430, opened 2025-08-20
fix buildpath (CI fix)

avivajpeyi:
The CI is currently failing on dev due to the build object being placed in:

src/bin/COMPAS

This PR ensures that src/COMPAS is still availible.


jeffriley (approved):
Thanks @avivajpeyi

PR #1433, opened 2025-09-02
First implementation of HAMSTARS mass transfer efficiency

ilyamandel:


themikelau (approved):
The Hamstars stuff looks good to me. I understand this is a simplified implementation. Do we want to call this a different name in anticipation for a future, full implementation?

ilyamandel:
Thanks, @themikelau ! 
I suggest keeping the simple name for now and adjust later if/when we implement something else?
BTW, @SimonStevenson , @jmerritt1 -- note this now also contains a fix to OB/WR mass loss for stars which combine the two with weights (credit to @jeffriley for spotting this issue)

PR #1435, opened 2025-09-25
Update references, include COMPAS methods paper 2

ilyamandel:


ilyamandel:
@SimonStevenson -- ping on the review, since I think this was a change you suggested. :)

jeffriley (approved):
This looks good to me - thanks @ilyamandel

PR #1436, opened 2025-10-23
Renamed MS core prescription option ZERO to HURLEY

brcekadam:
Just a change to the option `--main-sequence-core-mass-prescription ZERO`, which will now be called `HURLEY`.


ilyamandel (approved):
Looks great -- thanks, @brcekadam !  
@jeffriley -- pinging you just to let you know this will be a change to roll in.

PR #1437, opened 2025-10-26
USSN treatment in M&M; extend ScaleMassLossWithSurfaceHeliumAbundance beyond CHE

ilyamandel:
- Added option --USSN-kicks-override-mandel-muller ; if set to true, use user-defined USSN kicks (as a fixed value) in lieu of the Mandel & Muller kick prescription for USSNe
- Replaced --scale-CHE-mass-loss-with-surface-helium-abundance with the more general --scale-mass-loss-with-surface-helium-abundance (applies to all MS stars, not just CHE stars)
- Updated rotational velocity solver to use boost root finder


jeffriley (approved):
All good @ilyamandel .  I made two small fixes - I'll leave it to you to merge.

When I started reviewing my first suggestion was going to be changing "overwrite" to "override", but as if by magic, that was done :-)

PR #1439, opened 2025-10-28
fix: add -Wno-vla-extension (note, -Wno-vla-cxx-extension might be needed in the future)

avivajpeyi:
@SimonStevenson is getting some new warnings when building

> I am getting a new compiler warning with latest dev
> warning: unknown warning option '-Wno-vla-cxx-extension'; did you mean '-Wno-vla-extension'? [-Wunknown-warning-option]
> 

> I see several new warnings when compiling COMPAS with latest dev:
Log.cpp:1323:22: warning: variable length arrays in C++ are a Clang extension [-Wvla-cxx-extension]
 1323 |             bool buf[bufSize];
      |                      ^~~~~~~
Log.cpp:1323:22: note: read of non-const variable 'bufSize' is not allowed in a constant expression
Log.cpp:1272:13: note: declared here
 1272 |     size_t  bufSize         = p_H5file.dataSets[p_DataSetIdx].buf.size();                                                   // size of write buffer
      |             ^
Log.cpp:1329:27: warning: variable length arrays in C++ are a Clang extension [-Wvla-cxx-extension]
 1329 |             short int buf[bufSize];
      |                           ^~~~~~~
Log.cpp:1329:27: note: read of non-const variable 'bufSize' is not allowed in a constant expression
Log.cpp:1272:13: note: declared here
 1272 |     size_t  bufSize         = p_H5file.dataSets[p_DataSetIdx].buf.size();                                                   // size of write buffer
      |             ^
Log.cpp:1336:21: warning: variable length arrays in C++ are a Clang extension [-Wvla-cxx-extension]
 1336 |             int buf[bufSize];
      |                     ^~~~~~~
Log.cpp:1336:21: note: read of non-const variable 'bufSize' is not allowed in a constant expression
Log.cpp:1272:13: note: declared here
 1272 |     size_t  bufSize         = p_H5file.dataSets[p_DataSetIdx].buf.size();                                                   // size of write buffer

I think we need different flags based on clang versions.... 
Clang < 17 → -Wno-vla-extension
Clang ≥ 17 → -Wno-vla-cxx-extension

(for context -- my Jetbrains IDE is what suggested me to use `Wno-vla-cxx-extension` in place of `Wno-vla-extension`

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -90,15 +90,8 @@ ifneq ($(filter fast staticfast,$(MAKECMDGOALS)),)
     OPTFLAGS += $(OPTFLAGS_SPECIFIC)
 endif
 
-
-EXTRA_WARN_SUPPRESS := \
-	-Wno-vla-extension \ # might need -Wno-vla-cxx-extension for clang>18?
-    -Wno-unused-parameter \
-    -Wno-unused-variable \
-    -Wno-sign-compare \
-    -Wno-reorder-ctor \
-    -Wno-unused-function
-
+# might need -Wno-vla-cxx-extension for clang>18?
+EXTRA_WARN_SUPPRESS := -Wno-vla-extension #-Wno-unused-parameter -Wno-unused-variable -Wno-sign-compare -Wno-reorder-ctor -Wno-unused-function
 
 CXXFLAGS := -std=c++17 -Wall $(OPTFLAGS) $(CXXFLAGS_SPECIFIC) $(EXTRA_WARN_SUPPRESS)
 ICFLAGS := -I$(GSLINCDIR) -I$(BOOSTINCDIR) -I$(HDF5INCDIR) -I.
--- a/src/Makefile
+++ b/src/Makefile
@@ -91,7 +91,7 @@ ifneq ($(filter fast staticfast,$(MAKECMDGOALS)),)
 endif
 
 # might need -Wno-vla-cxx-extension for clang>18?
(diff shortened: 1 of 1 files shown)


avivajpeyi:
Hey @jeffriley  -- lets just foget about the changing flag for clang>18 for now? Ive left a comment in case this becomes an issue in the future.

jeffriley (commented):
Thanks @avivajpeyi !

A few questions:

Why have you added `-Wno-sign-compare`?  I think if we are generating these warnings we should fix the code - I know sometimes it's easier just to ignore the warning (or turn it off), but comparing a signed int to an unsigned (even `size_t`) is a bit of a code smell imo.

You've added `-Wno-unused-variable` - we should remove any unused variables, not just ignore their presence.

You've added `-Wno-unused-parameter`.  Are you seeing any of these warnings (I'm not with gcc on linux)?  For C++17 and later we can add the ` [[maybe_unused]]` attribute.  Not sure if that's available in clang (but I suspect it would be).

Ditto for `-Wno-unused-function` - although tbh we should remove any unused functions - if we find we need them later we can reinstate them from github.

You've added `-Wno-reorder-ctor`.  Are you seeing any of these warnings (I'm not with gcc on linux)?  If you are, we should fix the code - that's a bit of a code smell imo, and could cause problems in the future if there's a dependency on member initialisation.

avivajpeyi:
OK -- lets just remove the extra warning-surpressment

jeffriley (approved):
Great - that works :-)  Thanks!

Issue #1446, opened 2025-12-15
Theta and phi variables are flipped when assigning isotropic kicks


ilyamandel:
Theta and phi variables are flipped when assigning isotropic kicks, potentially giving us unintended, non-isotropic kick distributions. 

Discovered by @pauldisberg , easy fix but urgent.

Kicks are returned as (theta, phi), but incorrectly parsed as (phi, theta):

utils.cpp, DrawKickDirection():
return std::make_tuple(theta, phi);

Options.cpp:
std::tie(phi1, theta1) = utils::DrawKickDirection(m_KickDirectionDistribution.type, m_KickDirectionPower);
std::tie(phi2, theta2) = utils::DrawKickDirection(m_KickDirectionDistribution.type, m_KickDirectionPower);

ilyamandel:
Fixed in #1447 , closing

PR #1447, opened 2025-12-15
Fix to kick direction bug

ilyamandel:
Fix for issue #1446: Theta and phi variables are flipped when assigning kicks, potentially giving unintended kick distributions


pauldisberg (approved):
image

Did a small run with "./COMPAS -n 1000 --mode BSE --main-sequence-core-mass-prescription BRCEK --muller-mandel-kick-multiplier-NS 630 --muller-mandel-sigma-kick-NS 0.45 --kick-direction-distribution ISOTROPIC --random-seed 1714388113". Distributions of theta and phi look as expected for isotropic kicks.

jeffriley (approved):

PR #1450, opened 2026-01-28
Updated default mandelmuller kickparams

pauldisberg:
Updated the default Muller Mandel kick parameters values to v_NS = 630 km/s and sigma_kick = 0.45, from Disberg+2026 (previously from Kapil+2023). The sigma_kick value has been changed for both NSs and BHs.


ilyamandel (changes requested):
Hi @pauldisberg ,
This looks good, but as you rightly said, you also need to update the documentation.  The documentation is in the same github repo; see ../online-docs/pages/... In this case, I'd also update whats-new.rst to let users know about this change.  Since you spotted a mistake in the Hobbs parameter description in the documentation, could you please that, too?  Finally, since the default settings have changed, please update the yaml file -- you'll want to run something like
./COMPAS --create-YAML-file ../compas_python_utils/preprocessing/compasConfigDefault.yaml

ilyamandel (approved):
@pauldisberg : looks good.  I tweaked whatsnew to remove the mention of the Hobbs update in 03.28.00 -- that was actually done in 03.20.06 (even though the documentation was not everywhere correctly updated, the code was), and there's already a record of that in whatsnew.

PR #1451, opened 2026-01-30
Major changes to tidal prescriptions

veome22:
Major Updates:
1. Introduced a new tidal prescription called `ZAHN1977`, based mostly on [Zahn (1977)](https://ui.adsabs.harvard.edu/abs/1977A%26A....57..383Z/abstract), although this should be mostly the same as [Hurley+ (2002)](https://ui.adsabs.harvard.edu/abs/2002MNRAS.329..897H/abstract). I wrote this option for my own internal testing, so please feel free to point out any code inefficiencies/ bad documentation.
2. Updated the `KAPIL2025` prescription to allow for pseudo-synchronization based on Eq. (45) of [Hut (1981)](https://ui.adsabs.harvard.edu/abs/1981A%26A....99..126H/abstract), which should also help with [this](https://github.com/TeamCOMPAS/COMPAS/discussions/1382). 
3. Renamed `KAPIL2025` to `KAPIL2026`. 

Minor updates:
Changes to the dynamical tides equations in `KAPIL2026`.


ilyamandel (commented):
Thanks, @veome22 !

This will be a fairly quick skim, because I am about to go offline for 10 days.  I trust that @jeffriley will review once he is back from his work trip to Italy, so I am marking this as comments, not review -- once @jeffriley approves, I am happy. ;-)  The non-cosmetic ones are starred.

* changelog.h -- you need to actually change the version number on the last line :-), and then re-create the YAML file with that version (currently, it claims it was created with 3.28.00).

KAPIL2025->KAPIL2026 -- I guess our methods will be a bit wrong, but we can live with that ;-)

* I am quite worried about updating previous values in BaseBinaryStar.cpp, lines 2800-2804.  Normally, these previous values are meant to hold the values from the previous time step.  Now, we may be updating these multiple times per time step, which could create problems elsewhere.  Jeff will be fixing this in his "state vector" PR, but can we avoid putting this in here for now, because it might cause unanticipated side effects?  If you need to store pre-tide values to be used just a few lines later in the same function (lines 2811-2815), save them as local variables.

*Speaking of lines 2811-2815 in BaseBinaryStar.cpp, I don't quite understand what you are doing here and why.  Aren't the equations above intended to conserve angular momentum?  If they don't, shouldn't you be computing the angular momentum of the orbit using conservation laws, rather than just capping it (what if the eccentricity is too small to conserve total angular momentum rather than too large)?  Also, I can't easily see  what lines 2811-2815 are doing; can you write a mathematical equation in the comments above, or refer to one from the literature?


There seem to be multiple commented out equations in the Zahn tidal prescription (BaseBinaryStar.cpp, lines 2876 onwards).  Some appear to be pseudo-code and are presumably there as explanations, but others just seem like code that was commented out (say, lines 2918-2920).  Please check and remove any equations that were just temporary code.

veome22:
Thanks for your comments @ilyamandel!

> changelog.h -- you need to actually change the version number on the last line :-), and then re-create the YAML file with that version (currently, it claims it was created with 3.28.00).

Indeed, should be fixed now!

> I am quite worried about updating previous values in BaseBinaryStar.cpp, lines 2800-2804. [...] If you need to store pre-tide values to be used just a few lines later in the same function (lines 2811-2815), save them as local variables.

I only need them locally, so I have converted them to local variables. 

> There seem to be multiple commented out equations in the Zahn tidal prescription (BaseBinaryStar.cpp, lines 2876 onwards). 

I forgot to remove these comments earlier, but they're gone now. Thanks for catching this!

> Speaking of lines 2811-2815 in BaseBinaryStar.cpp, I don't quite understand what you are doing here and why. Aren't the equations above intended to conserve angular momentum? 

This is a more complicated issue to address. In principle the secular equations were obtained by angular momentum conservation, as stated in Hut (1981). However, I found that angular momentum was not being conserved in practice, and this only happened for $e>0$. The ecc limit is a stop-gap solution, but it works completely. 
As far as I understand, a physically motivated solution to this problem should really limit spins rather than eccentricity, since I think the da/dt and de/dt equations are constructed perturbatively, and then $d\Omega$/dt is computed by angular momentum conservation. I am happy to implement this and push an update after some testing.

veome22:
Hi @jeffriley! I'd love your advice, leaving a couple questions here for when you are back:

1. The latest COMPAS build checks fail when running pytests with the following error: 
`ModuleNotFoundError: No module named 'py_tests'`. Is there something I messed up, or is this a git / COMPAS issue?

2. The more I look at my `ZAHN1977` code the more I dislike it. My main concern is that I do not follow the same infrastructure we developed for the other tidal model, where each star (e.g. `BaseStar.cpp` or `Remnants.cpp`) gets its own `ImKnmDynamical` and `ImKnmEquilibrium`. I could do something like this for the `ZAHN1977` model too, and it might also save some compute time / memory for when the tides should not be applied. Is this worth doing now, or does this seem like overkill?

jeffriley:
@avivajpeyi or @reinhold-willcox could one of you look at why the py_tests are failing here (module not found?)?

jeffriley:
@veome22 

> The latest COMPAS build checks fail when running pytests with the following error ...

Hopefully @avivajpeyi or @reinhold-willcox can fix that - or at least tell us what the problem is.

> The more I look at my ZAHN1977 code the more I dislike it ...

I'll take a look.

reinhold-willcox:
@jeffriley I can take a look later this week, but the py_tests are out of my expertise so I can't promise I can figure out what's causing error.

jeffriley:
> I can take a look later this week, but the py_tests are out of my expertise so I can't promise I can figure out what's causing error.

Thanks @reinhold-willcox - I think @avivajpeyi implemented the py-tests, but wasn't sure if you were across them - I am certainly not...

avivajpeyi:
Strange, I can try to make some time tomorrow to look at this! Give. That none of the python code has changed, I don't see why this is happening...

Also strange because we can see that the postprocessing to create the detailed evolution plot worked as shown above.

veome22:
Hi @avivajpeyi! Were you able to look into the pytest issue at all? 
Please let me know if I can help somehow.

ilyamandel:
@veome22 -- short of changing the prescription to fix A.M. conservation, can you write out (either here or in a comment in the code) what lines 2811-2815 in the version of BaseBinaryStar.cpp that I reviewed were meant to do?  I might have been in a hurry, but I didn't see how this limit would ensure the conservation of total angular momentum -- can you let me know exactly what equation your code is implementing?  

Otherwise, happy to approve.

veome22:
> @veome22 -- short of changing the prescription to fix A.M. conservation, can you write out (either here or in a comment in the code) what lines 2811-2815 in the version of BaseBinaryStar.cpp that I reviewed were meant to do? I might have been in a hurry, but I didn't see how this limit would ensure the conservation of total angular momentum -- can you let me know exactly what equation your code is implementing?


Hi @ilyamandel , I have set up a short Overleaf doc outlining the equations of angular momentum conservation, both for the previous (eccentricity-based) and the updated (spin-based) formalisms. Please find it [here](https://www.overleaf.com/read/zpfddbmbdcsx#839428), and let me know if I can clarify anything further.

ilyamandel (approved):
Thank you, @veome22 !  I am happy with this explanation (which presumably will find its way into the paper, at least in some shortened form) and the revised code.  
I don't recall any more when we actually update the moment of inertia of the star -- in principle, it does change from one time step to another, so if we are updating everything at once, as @jeffriley will do, we will need to check this maths carefully -- but happy to approve this version.  Thanks again!

ilyamandel (commented):
The new value of lConv is a bit inconsistent with how we are estimating tConv, but OK. ;-)

jeffriley:
@veome22 I have reviewed your changes and I approve, subject to @ilyamandel's comments.

> The more I look at my ZAHN1977 code the more I dislike it. My main concern is that I do not follow the same infrastructure 
> we developed for the other tidal model, where each star (e.g. BaseStar.cpp or Remnants.cpp) gets its own
> ImKnmDynamical and ImKnmEquilibrium. I could do something like this for the ZAHN1977 model too, and it might also
> save some compute time / memory for when the tides should not be applied. Is this worth doing now, or does this seem
> like overkill?

If you have time (and the inclination), I think it would be good to make both treatments behave similarly.  I am in the middle (actually closer to the start...) of a rewrite, so this is going to be rewritten to some extent anyway, but starting that rewrite from a base where like things follow the same pattern would be a good thing.

Now, we just have to get the py_tests working. @avivajpeyi, any ideas?  Although we could just merge anyway...

ilyamandel:
@veome22 , are you OK with merging this in, so the code matches the arXiv paper?

veome22:
Thanks @jeffriley and @ilyamandel!  I would like the PR to be merged ASAP to match the paper version, and then I can re-factor/update the code as necessary later on. At that point, I can also update the arxiv (and published) version of the paper to match any future updates.

Please feel free to go ahead and merge if that sounds alright.

veome22:
> The new value of lConv is a bit inconsistent with how we are estimating tConv, but OK. ;-)

@ilyamandel, yes, but I think the old l_conv ~ R_env was also [inconsistent with t_conv](https://r-izzard.surrey.ac.uk/doc/lectures/binary_stars/3/convective_turnover.pdf). If I understand correctly, the new l_conv might even be closer to the R_env/3 assumption used to obtain t_conv in Rasio (1996) and Hurley (2002). In the future, it is probably worth updating t_conv and l_conv to be calculated self-consistently in COMPAS and remove the variable definition inside the tides code.

PR #1452, opened 2026-02-09
Modified the Max Moe sampler (sampleMoeDiStefano.py) and tested that it is correct

BalexBalex:
Tests done: 1) The masses of singles and binary components follow Kroupa for a 1e6MSun sample, 2) The result is identical to the analogous COSMIC code implementation, 3) The other quantities (M1, q, Porb, ecc) and pairs of such quantities are distributed indistinguishably between the COMPAS and COSMIC versions, 4) Changing MMin in COMPAS correctly truncates the primary mass to be above MMin, without affecting other correlations.
MassSetComparison_COSMIC_COMPAS_Init
MassSetComparison

### DIFF
--- a/compas_python_utils/preprocessing/sampleMoeDiStefano.py
+++ b/compas_python_utils/preprocessing/sampleMoeDiStefano.py
@@ -393,15 +393,13 @@ def createParameterDistributionsAndSampler(testing=False):
     # Slope = -2.3 for M1 > 1 Msun
     fM1 = np.power(M1, -2.3)
 
-    # Slope = -1.6 for M1 = 0.5 - 1.0 Msun -> Slope = -2.3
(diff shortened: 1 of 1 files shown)


ilyamandel (changes requested):
Thank you, @BalexBalex !  
The changes and result look great, but I think the current comments make things more confusing (e.g., # Slope = -1.6 for M1 = 0.5 - 1.0 Msun, when it's really -2.3 in the code and -2.3 is the intended Kroupa value).  Can we update the comments and remove the commented out code?

BalexBalex:
Updated it, just left a short note to show where the code was modified!

ilyamandel (approved):
Thank you, @BalexBalex !  Looks great!

PR #1455, opened 2026-02-18
Fix invalid python imports in pytests

### LINKED ISSUE #1454: Python import path / working-directory issue in conftest.py in py_test ?-> maybe GitHub check do not complete due to this
**Describe the bug**
pr-1368 and a few earlier PRs around the Cosmic Integration seems to have changed the `py_test/conftest.py` file in some way that it now seems to have an issue with the Python import path / working-directory issue in conftest.py? 

There's also a double / copy of the same line: 
`# Testvalues used in test_total_mass_evolved_per_z defined in py_tests/test_values.py
from py_tests.test_values import MAKE_PLOTS, M1_MIN, M1_MAX, M2_MIN, F_BIN

# Testvalues used in test_total_mass_evolved_per_z defined in py_tests/test_values.py
from py_tests.test_values import MAKE_PLOTS, M1_MIN, M1_MAX, M2_MIN, F_BIN`

**To Reproduce**
1. Go to https://github.com/TeamCOMPAS/COMPAS/blob/dev/py_tests/conftest.py 
2. see changes since https://github.com/TeamCOMPAS/COMPAS/commit/4ea8b9e124984d748ba79c9ef519dbc383d1ccca 
3. see error in https://github.com/TeamCOMPAS/COMPAS/actions/runs/22083882124/job/63814365416?pr=1448 

Specifically, this and earlier changes does from `py_tests.test_values import ...`
But I think that in the GitHub Actions job, Python’s sys.path doesn’t include the directory that contains py_tests/, so import py_tests fails.

@jeffriley or others might know best how to fix this neatly (and whether it needs to be fixed?) 
maybe something like adding 
```
ROOT = Path(__file__).resolve().parents[1]  # directory that contains py_tests/
sys.path.insert(0, str(ROOT))

from py_tests.test_values import MAKE_PLOTS, M1_MIN, M1_MAX, M2_MIN, F_BIN
```

but not sure, as the pytest suite seemed to be working before, so it might be something different.

avivajpeyi:
Close #1454 

1. Fixed pytest import errors by removing from py_tests... style imports.
2. Replaced silent cosmic-integration import fallbacks with a clear setup error message (install with -e . / -e '.[dev]').
3. Made mass-range handling work for tiny test datasets and aligned DCO naming (BHBH).

### DIFF
--- a/compas_python_utils/cosmic_integration/ClassCOMPAS.py
+++ b/compas_python_utils/cosmic_integration/ClassCOMPAS.py
@@ -3,10 +3,16 @@ import numpy as np
 import h5py as h5
 import os
 import sys
-# Get the COMPAS_ROOT_DIR var, and add the cosmic_integration directory to the path
-compas_root_dir = os.getenv('COMPAS_ROOT_DIR')
-sys.path.append(os.path.join(compas_root_dir, 'compas_python_utils/cosmic_integration'))
-import totalMassEvolvedPerZ as MPZ
+from pathlib import Path
+
+try:
+    from . import totalMassEvolvedPerZ as MPZ
+except ImportError:
+    # Fallback for direct script execution: add this module's directory to sys.path.
+    ci_dir = Path(__file__).resolve().parent
+    if str(ci_dir) not in sys.path:
+        sys.path.append(str(ci_dir))
+    import totalMassEvolvedPerZ as MPZ
 
 
 class COMPASData(object):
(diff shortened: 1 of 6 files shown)


LiekeVanSon (approved):
Looks good to me! Note sure, but I might have been to blame introducing this in the latest update to the CosmicIntegration scripts, if so, sorry, but thanks for cleaning this up!

Related: do we maybe want a changelog for the python scripts as well? (this might help prevent such incidents in the future)

ilyamandel (approved):
Sorry for the delayed check -- looks great, @avivajpeyi !  
(@jeffriley explained that some of these are reverting previous incorrect changes, where your comments on a PR didn't come through somehow... :( )

PR #1456, opened 2026-02-18
Update makefile for auto-boost detection for macOS (Linux/windows left alone)

avivajpeyi:
This pr tries to improve the makefile to
1. Automatically detect Darwin (MacOS) and configures include/library paths for GSL, Boost, and HDF5 using Homebrew defaults.
2. Switches GSL, Boost, and HDF5 directory assignments to ?=, allowing users to override paths via environment variables.
3. On macOS, -lboost_system is now conditionally omitted if the library file is missing, preventing link-time errors. (for new macs)

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -1,4 +1,6 @@
 CPP := g++
+UNAME_S := $(shell uname -s)
+HOMEBREW_PREFIX := $(shell brew --prefix 2>/dev/null)
 
 # Build configuration: DOCKER_BUILD=1 for Docker, otherwise local
 ifeq ($(DOCKER_BUILD),1)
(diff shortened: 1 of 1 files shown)


jeffriley (changes requested):
Thanks @avivajpeyi 

You changed ":=" to "?=" in a few lines, e.g.

GSLINCDIR ?= /include
GSLLIBDIR ?= /lib

I think that's fine (the difference afaik is that ":=" sets the variable value immediately, whereas "?=" only sets the variable value if it doesn't already contain a value.

Where you check for "Darwin" and set the values of:

GSLINCDIR
GSLLIBDIR
BOOSTINCDIR
BOOSTLIBDIR
HDF5INCDIR
HDF5LIBDIR

you use ":=" - isn't that going to override anything the use specified on the commandline?  I think if the user went to the trouble of specifying directories on the commandline we should honour that - given that they specify them, I think we should presume they know where their libraries and include files are.  I think we should only try to automatically detect where they are if the user didn't specify them on the commandline.  It may be that you can find where the standard libraries and include files are, but maybe the user wants to use specific libraries and include files that may not be where the standard files are installed.  Am I understanding your changes correctly?

ilyamandel:
@avivajpeyi , just a ping in case you hadn't seen @jeffriley 's review

avivajpeyi:
Oops, I missed this earlier! thanks Jeff, these are good points.

Yes, your description of `:=` vs `?=` matches my understanding as well: `:=` assigns the value immediately, while `?=` only assigns it if the variable hasn't already been set.

Also I may be misunderstanding, but I thought command-line variable assignments in GNU make already take precedence over ordinary assignments in the makefile unless `override` is used, so I don’t think the `:=` vs `?=` in the Darwin block would affect something like:

>    make GSLINCDIR=/include/balh

(because the command-line value should still win.)
https://www.gnu.org/software/autoconf/manual/autoconf-2.64/html_node/Macros-and-Submakes.html


That was also why I didn't set the Darwin block to `?=` initially. I assumed command-line values would still take precedence, so using `:=` there would not override something the user explicitly set.

My use of `?=` earlier was mainly to provide sensible defaults while allowing users to override them if they have custom install locations.

**Would you like me to change the Darwin-block assignments to `?=` as well,**?
or can you be more explicit about what you would like me to change?

jeffriley:
> Also I may be misunderstanding, but I thought command-line variable assignments in GNU make already take precedence
> over ordinary assignments in the makefile unless override is used

@avivajpeyi you're absolutely right - I missed that :-)

jeffriley (commented):
All good - approved

jeffriley (approved):

ilyamandel:
We are still seeing some CI failures...

PR #1458, opened 2026-03-06
Clarify that T_eff,crit differs from T_* in Sander et al. 2023 Eq. 1 (defined at Rosseland optical depth τ=20)

julia-haynes:
This updates the comment to clarify that the effective temperature used in the correction corresponds to the critical/sonic-point temperature (T_eff,crit), which differs from T_* as defined in Sander et al. 2023 Eq. 1 at a Rosseland continuum optical depth of τ = 20.

### DIFF
--- a/src/BaseStar.cpp
+++ b/src/BaseStar.cpp
@@ -2328,8 +2328,8 @@ double BaseStar::CalculateMassLossRateWolfRayetSanderVink2020(const double p_Mu)
  * as a function of effective temperature, according to the
  * prescription of Sander et al. 2023 (https://arxiv.org/abs/2301.01785)
  * 
- * Use the correction given in Eq. 18, with the effective temperature
- * (what they refer to as T_\star in Eq. 1) as T_eff,crit
+ * Use the correction given in Eq. 18, with the stellar effective temperature used * as a proxy for T_eff,crit (the effective temperature at  
+ * temperature at the critical/sonic point). Using T_eff as a proxy for T_eff,crit introduces a small approximation.
  * 
  * 
  * double CalculateMassLossRateWolfRayetTemperatureCorrectionSander2023(const double p_Mdot)


ilyamandel:
Thank you for that.  There are some typos in the revised text (e.g., * in the middle of a line, "temperature" repeated one time too many).  But my real question is about the last comment:

"Using T_eff as a proxy for T_eff,crit introduces a small approximation." 

Do you have any information supporting this?  It would be great to know how much of a difference it actually makes, i.e., how different the two temperatures are.

ilyamandel:
@julia-haynes -- gentle ping on my question above.

ilyamandel:
We'll aim to add some comments on this into the revised paper version; not sure if the code is a right place for this discussion.  If there's no more feedback from @julia-haynes , I may close this PR.

ilyamandel:
Comments to be added to the updated paper version, but the code doesn't seem to be the right place for them.  Closing this issue.

ilyamandel:
P.S. For completeness, here's the new text in the winds paper:

> Inevitably, we must make a number of simplifications to detailed models in order to adapt them to the limited information available in the evolutionary tracks used for rapid population synthesis.  For example, we take $T_\mathrm{eff}$ to be the effective temperature predicted by the rapid stellar evolution models \citep[][]{Hurley:2000MNRASSSE,COMPASTeam:2021tbl}, while \citet{Sander:2020MNRAS} use the temperature at the Rosseland continuum optical depth of 20 as a proxy for the temperature at the critical point.  Comparing our simplified prescriptions to the detailed models of \citet{Sander:2020MNRAS,Sander:2023aap} suggests that we match the detailed mass loss rate predictions for WR stars with a mass of 30 M$_\odot$, underpredict the mass loss rate for more massive stars (by $\sim 0.2$ dex at 40 M$_\odot$) and overpredict the mass loss rate for less massive stars (by $\sim 0.4$ dex at 20 M$_\odot$). Of course, even detailed models are incomplete.  For example, \citet{Sander:2020MNRAS} do not account for further radius dependencies \citep[e.g.,][]{Sander:2023aap} and dynamic inflation effects observed in multi-dimensional atmosphere models \citep[e.g.,][]{Moens:2022,Moens:2025}. Currently, there are no mass loss formulae consistently describing the mass loss of the full variety of observed WR and He stars, and we are thus limited to the available descriptions.

Issue #1424, opened 2025-08-18
Strange indenting in documentation


reinhold-willcox:
**Describe the bug**

Extremely minor, should be lowest priority. Some of the binary properties at the following page are indented, others are not. It seems unintentional, so I would suggest removing indents. 

https://compas.readthedocs.io/en/latest/pages/User%20guide/COMPAS%20output/standard-logfiles-record-specification-binary.html

`urgency_low`      - This issue is not urgent
`severity_minor`    - This is a minor bug with minimal impact

**Versioning (please complete the following information):**
 - COMPAS v03.23.01

ilyamandel:
I looked but can't figure out why this is happening... @jeffriley ?

jeffriley:
I'll take a look

ilyamandel:
Thanks, @jeffriley !

jeffriley:
I believe the odd formatting is because the tables (each specification) are being centred by Sphynx (I think) when the html files are built.  I will change the COMPAS css file to left-align tables (I hope that doesn't adversely affect other pages, but if it does we can be more specific in the markdown files and css file).

Attempted fix in PR [#1460](https://github.com/TeamCOMPAS/COMPAS/pull/1460)

jeffriley:
I believe this is now fixed.

jeffriley:
Ah... except that the "Back to top" markers are problematic now.  I'll take a look at those tomorrow :-)

ilyamandel:
@jeffriley , did you ever figure this one out?  (Not urgent, just curious.)

jeffriley:
Sorry @ilyamandel, I thought I had replied to this (I meant to :-)

Short answer: no.

Longer answer: the changes I made to fix the problem worked on my desktop system, but failed when deployed to readthedocs. That would seem to indicate that the readthedocs build environment on my system is different from the deployed build system. What I plan to do - soon I hope - is figure out what the differences are, and update the deployed system to match mu system.

PR #1460, opened 2026-03-09
Modified docs css file to left-align tables

jeffriley:
Attempted fix for issue #1424.  I don't see the issue when I use my local doc files (I see it it in  the online documentation, but it is not apparent if I compile and view the docs locally...), so I can only hope this fixes the issue.  If not, I'll try something else.


ilyamandel (approved):
Thank you!

PR #1461, opened 2026-03-09
Documentation fix for left alignment of tables

jeffriley:
Re issue https://github.com/TeamCOMPAS/COMPAS/issues/1424.  The previous fix (left aligning tables) lines the items up correctly, but it broke some other formatting.  This is an attempt to fix that ricochet problem - just for the list of binary properties in this PR, and if it works I'll apply it to other files as necessary in a new PR.

EDIT: I can't see the problem in my local build, so need to test it with the online docs


ilyamandel (approved):
Thanks, @jeffriley !

PR #1466, opened 2026-03-15
Fix for issue 1463

### LINKED ISSUE #1463: LAMBDA_DEWI (Claeys+2014) produce negative common envelope λ in CHeB and EAGB stars
Hi,

There is a sign error in Appendix A of Claeys et al. 2014 in the prescription for the common-envelope parameter λ for CHeB and EAGB stars. It is the prescription used when the command-line option --common-envelope-lambda-prescription LAMBDA_DEWI is selected.

For these stars with an envelope mass greater than 1 solar mass, the formula of the paper is written as
λ = 2 × min(0.8, 1.25 − 0.15 log₁₀ L, λ₃),
with
λ₃ = min(−0.9, 0.58 + 0.75 log₁₀ M) − 0.08 log₁₀ L.

Because λ₃ includes min(−0.9, …), it becomes negative, and λ is therefore also negative. The implementation in COMPAS follows the paper, which leads to a negative λ error and stops the simulation when this case occurs.

In the latest version of J. Hurley's BSE implementing this prescription, there is no minus sign in front of the 0.9 in the min function, which I believe to be the correct implementation.

`urgency_moderate`

`severity_moderate`

See this screenshot for the version, an example of initial parameters leading to the negative lambda error, and the full command line:
Image

brcekadam:
Fixes a typo from Claeys+2014 common-envelope lambda prescription. This sign error is mentioned in e.g. Appendix E.1 from [Marchant+2021](https://ui.adsabs.harvard.edu/abs/2021A%26A...650A.107M/abstract).

Closes #1463

### DIFF
--- a/src/CHeB.cpp
+++ b/src/CHeB.cpp
@@ -108,7 +108,7 @@ void CHeB::CalculateTimescales(const double p_Mass, DBL_VECTOR &p_Timescales) {
  */
 double CHeB::CalculateLambdaDewi() const {
 
-    double lambda3 = std::min(-0.9, 0.58 + (0.75 * log10(m_Mass))) - (0.08 * log10(m_Luminosity));                          // (A.4) Claeys+2014
+    double lambda3 = std::min(0.9, 0.58 + (0.75 * log10(m_Mass))) - (0.08 * log10(m_Luminosity));                           // (A.4) Claeys+2014 with corrected typo (see e.g. Appendix E.1 in Marchant+2021)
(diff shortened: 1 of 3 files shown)


ilyamandel (approved):
Looks good -- thank you, @brcekadam !

jeffriley (approved):
Looks good - thanks @brcekadam 

Oh - @ilyamandel beat me to it :-)

ilyamandel:
@avivajpeyi -- note that the CI seems to be failing again?

PR #1467, opened 2026-03-18
New compas debugging_utils file

reinhold-willcox:
Added my own compasUtils file, which I use for lots of post-processing and testing. It was previously in `misc/unsupported_utils`, so I am hoping to get it elevated to the level of the other utils. 

This contains two major functions that help a lot with inspecting COMPAS data. 
- printCompasDetails: takes the hdf5 output (optionally with specific seeds or masks) and prints a pandas df of the whole output (including units). Makes it visually much easier to analyze the output from a jupyter notebook. 
- getEventStrings: takes the hdf5 output and generates a string per seed that summarizes the events in the binary. Events include SMT, CEE, SNe, and mergers, and these are listed chronologically with a funny but simple syntax (described in the function). Combined with np.unique(), this is a great way to quickly see which binary channels dominate and which are rare. 

I expect that there will be changes requested, so it is probably not yet ready to be merged in, but happy to hear feedback on how to improve / align these scripts with the other python utilities.


ilyamandel (commented):
I leave the review to folks who use python (@SimonStevenson ? @avigna ? ).

Consider including supported python tools in the documentation in https://compas.readthedocs.io/en/latest/pages/User%20guide/Post-processing/post-processing.html

avivajpeyi:
Looks cool!  I would suggest adding some more doctrings at the top, adding some pytests to ensure that this is maintained/doesn't break in the future. 

Also, how is this different from  [h5view](https://github.com/TeamCOMPAS/COMPAS/blob/dev/compas_python_utils/h5view.py)?

Can they be merged together?


Finally, if you can add it to the setup.py entry points, its quite easy to get the auto-documentation tools to build docs like so

https://compas.readthedocs.io/en/latest/pages/User%20guide/Post-processing/hdf5/post-processing-h5view.html


I have some time, im happy to help with this if youd like :)

nrsegovia:
As I suggested in @ilyamandel's post on Slack, I think it might be useful to add type hinting and docstring styling (similar to numpy) to bring the python code style closer to what we have for C++ files. 

Perhaps we can start here by just adding type hints?

compas_python_utils/compasUtils.py
```diff
@@ -0,0 +1,355 @@
+import h5py as h5
+import numpy as np
+import pandas as pd
+from numpy.dtypes import StringDType
+
+
+########################################################################
+# ## Function to print the data from a given COMPAS HDF5 group in a readable pandas template
+
+def printCompasDetails(data, *seeds, mask=()):
```
nrsegovia, line 10: Do I understand correctly that this function name might be a bit misleading? Outside the jupyter environment it would just return the data frame, not print it. 

I also wonder if `*seeds` could be replaced for an input list, it might be more readable for large sets of seeds.

compas_python_utils/compasUtils.py
```diff
@@ -0,0 +1,355 @@
...
+    
+        allSeeds = data[seedVariableName][()]
+        seedsMask = np.isin(allSeeds, seeds)
+        if len(seeds) == 0: # if any seed is included, do not reset the mask
+            seedsMask = np.ones_like(allSeeds).astype(bool)
+        if mask == ():
+            mask = np.ones_like(allSeeds).astype(bool)
+        mask &= seedsMask
+
+        df = pd.DataFrame.from_dict({param: data[param][()][mask] for param in list(data.keys())}).set_index(seedVariableName).T
```
nrsegovia, line 32: There are a couple of values being repeated, e.g., `('SEED' in list_of_keys)` and `list(data.keys())`

compas_python_utils/compasUtils.py
```diff
@@ -0,0 +1,355 @@
...
+
+    else: # No seed parameter, so do custom print for Run Details
+
+        # Get just the keys without the -Derivation suffix - those will be a second column
+        keys_not_derivations = []
```
nrsegovia, line 37: Can be swapped for a one line definition without affecting readability, e.g., `keys_not_derivations = [x for x in key_list if "-Derivation" not in x]` or similar

ilyamandel:
@reinhold-willcox -- please see the comments above.

reinhold-willcox:
Ya, I'll take care of it (Nicolas has also written to me directly with some pointers). 

This is a fairly low priority for me, so I'll try to address this after the semester ends.

ilyamandel:
@reinhold-willcox , any progress on this PR?

reinhold-willcox:
@ilyamandel, sorry, haven't had a chance to dive into this yet, but I should have a window in the next few weeks.

compas_python_utils/compasUtils.py
reinhold-willcox: Hey, finally getting around to this. The *seeds input is particularly handy because you can supply a single seed, multiple comma-separated seeds, or a list / array of seeds, so it's very flexible in this format. But I'm happy to make that clearer in the docstrings.

compas_python_utils/compasUtils.py
reinhold-willcox: agreed

compas_python_utils/debugging_utils.py
```diff
@@ -0,0 +1,440 @@
...
+        The values of each parameter in the H5Group, excluding those removed by `seeds` or `mask`.
+
+    Notes
+    -----
+    Designed for easier visualization in a jupyter notebook, for inspection / debugging purposes. 
+    If both `seeds` and `mask` are supplied, resulting systems must pass both filters.
+
+    Examples
+    --------
+    >>> mt_data = h5.File('COMPAS_Output.h5')['BSE_RLOF'])
```
nrsegovia: I think that `)` at the end should be removed

compas_python_utils/debugging_utils.py
```diff
@@ -0,0 +1,440 @@
...
+    
+    >>> mt_seeds = mt_data['SEED'][()]
+    >>> print_compas_details_dataframe(mt_data, mt_seeds[:50])
+    [output of all BSE_RLOF events occuring in the first 50 seeds]
+    
+    >>> cee_events = mt_data['CEE>MT'][()] == 1 # needed to convert to boolean mask
+    >>> print_compas_details_dataframe(mt_data, mt_seeds[:50], mask=cee_events)
+    [output of all Common Envelope events occuring in the first 50 seeds]
+    """
+    list_of_keys = list(data.keys())
```
nrsegovia: This might be unnecessary. It seems that most of the time you do `x in list_of_keys`, but `x in data` would produced the same behavior if I remember correctly

compas_python_utils/debugging_utils.py
```diff
@@ -0,0 +1,440 @@
...
+
+    # Check if SEED parameter exists in data
+    if ('SEED' in list_of_keys) | ('SEED>MT' in list_of_keys): 
+        seed_variable_name='SEED' if ('SEED' in list_of_keys) else 'SEED>MT' #SEED>MT is a relic from older versions, but we leave this in for backwards compatibility
+
+        # If `seeds` or `mask` arguments supplied, create the relevant mask
+        all_seeds = data[seed_variable_name][()]
+        seeds_mask = np.isin(all_seeds, seeds)
+        if len(seeds) == 0: # If `seeds` argument is not supplied, set the default mask
```
nrsegovia: `seeds` does not have a default value, so shouldn't it always be supplied? Otherwise you would get an error

compas_python_utils/debugging_utils.py
```diff
@@ -0,0 +1,440 @@
...
+        if mask == ():
+            mask = np.ones_like(all_seeds).astype(bool)
+        mask &= seeds_mask
+        df = pd.DataFrame.from_dict({param: data[param][()][mask] for param in list_of_keys}).set_index(seed_variable_name).T
+
+    else: # No seed parameter, currently only applies to the RunDetails H5Group 
+        keys_not_derivations = [key for key in list_of_keys  if '-Derivation' not in key] # Get just the keys without the -Derivation suffix
+        
+        # Some parameter values are string types, formatted as np.bytes_, need to convert back
+        def convert_strings(param_array):
```
nrsegovia: I wonder if this *must* be defined within the function. I think that at some point you (or the user) could need this utility, so exposing it would be better. Also, I'm being picky but perhaps you could change the check from `isinstance` to something like `np.issubdtype(param_array.dtype, np.bytes_)`, though this assumes that `param_array` is a numpy array (I'm not entirely sure it is the case). This way you wouldn't check the first element, but the array itself.

compas_python_utils/debugging_utils.py
```diff
@@ -0,0 +1,440 @@
...
+
+        # If `seeds` or `mask` arguments supplied, create the relevant mask
+        all_seeds = data[seed_variable_name][()]
+        seeds_mask = np.isin(all_seeds, seeds)
+        if len(seeds) == 0: # If `seeds` argument is not supplied, set the default mask
+            seeds_mask = np.ones_like(all_seeds).astype(bool)
+        if mask == ():
+            mask = np.ones_like(all_seeds).astype(bool)
+        mask &= seeds_mask
+        df = pd.DataFrame.from_dict({param: data[param][()][mask] for param in list_of_keys}).set_index(seed_variable_name).T
```
nrsegovia: If the default is to include all seeds, wouldn't this result in a "wide" table? Which also impacts pandas performance. I guess this is required to add units as a string column, but it might be worth to consider returning a separate dictionary with units in case the user wants to check them.

compas_python_utils/debugging_utils.py
```diff
@@ -0,0 +1,440 @@
...
+        seed_events.sort(key=lambda ev:(ev[1], event_ordering.index(ev[0])))  # sort the events by time and event type (mt_data before sn_data if at the same time)
+        returned_seeds[idx] = seed                                            # record the seed in the seeds array being returned
+        returned_events[idx] = seed_events                                    # record the events for this seed in the events array being returned
+    return returned_seeds, returned_events                                    
+
+
+###########################################
+# ## Produce strings of the event histories
+
+stellar_type_dict = {
```
nrsegovia: I see that this is a dictionary for simplified stellar type names, so perhaps you want to modify the name for something obvious like `simplify_types_dict`. Not critical, though.

compas_python_utils/debugging_utils.py
```diff
@@ -0,0 +1,440 @@
...
+    use_int_stypes : bool, optional
+
+    Returns
+    -------
+    event_strings : array of EventHistoryString's
+        strings of the event history of each system
+
+    Notes
+    -----
+    Exactly one of data or all_events must be included. If neither, nothign is returned.
```
nrsegovia: typo in `nothing`

nrsegovia (commented):
@reinhold-willcox it all looks great, I just left a couple of comments/suggestions. I'm still not entirely convinced by `*seeds` but it seems to work just fine, so I won't dwell on that. I do wonder why some sections of the code seem to prefer lists over numpy arrays, though. And as a final style-related comment, the PEP8 has some recommendations regarding line length:

> Limit all lines to a maximum of 79 characters.
> For flowing long blocks of text with fewer structural restrictions (docstrings or comments), the line length should be limited to 72 characters.

(I have never cared much about this until I recently started adding a length limit to my own code, and it looks much tidier)

compas_python_utils/debugging_utils.py
reinhold-willcox: *seeds works the same as *args, so they are all optional. Not sure what the best way to indicate this is, but it is meant to run fine if it is not supplied.

compas_python_utils/debugging_utils.py
reinhold-willcox: In my experience, it's extremely valuable to have the units right there in the same dataframe. This is meant to be a debugging tool, so having that information right in sight as you look at the other columns has been really revealing. It does return a very wide table if you include all the seeds from some large run, but then the pandas (or possibly jupyter) width limit in practice replaces many of these with an ellipsis.

reinhold-willcox:
> @reinhold-willcox it all looks great, I just left a couple of comments/suggestions. I'm still not entirely convinced by `*seeds` but it seems to work just fine, so I won't dwell on that. I do wonder why some sections of the code seem to prefer lists over numpy arrays, though. And as a final style-related comment, the PEP8 has some recommendations regarding line length:
> 
> > Limit all lines to a maximum of 79 characters.
> > For flowing long blocks of text with fewer structural restrictions (docstrings or comments), the line length should be limited to 72 characters.
> 
> (I have never cared much about this until I recently started adding a length limit to my own code, and it looks much tidier)

Thanks @nrsegovia . I've gone through and implemented your changes, except where I've raised discussions above. As for lists vs arrays, this often depends on the use case. It's easier to append to lists when you don't know how long they will be at the outset, while the arrays are nicer for speed. If there's a specific place you think I should swap one for the other, I'm happy to take a look.

reinhold-willcox:
> Looks cool! I would suggest adding some more doctrings at the top, adding some pytests to ensure that this is maintained/doesn't break in the future.
> 
> Also, how is this different from [h5view](https://github.com/TeamCOMPAS/COMPAS/blob/dev/compas_python_utils/h5view.py)?
> 
> Can they be merged together?
> 
> Finally, if you can add it to the setup.py entry points, its quite easy to get the auto-documentation tools to build docs like so
> 
> https://compas.readthedocs.io/en/latest/pages/User%20guide/Post-processing/hdf5/post-processing-h5view.html
> 
> I have some time, im happy to help with this if youd like :)

h5view is a command line tool, this one is for use in jupyter notebooks. They are similar conceptually in that they both provide a view into the data, but they are very different under the hood so I'm not sure it makes sense to combine them. 

I'm also not sure what to do for the setup.py functionality. I added the file to the setup.py entry points, but all the other files seem to allow for a command line argument parser, which is not relevant for this file. Can you advise on what I should be putting in main()?

ilyamandel:
@reinhold-willcox -- I leave the review of the python code to @nrsegovia , who is clearly far more knowledgable than I am.  Just want to flag that if this is becoming a supported python utility, it should be added to the documentation at the same level as other supported python utils -- see https://compas.readthedocs.io/en/latest/pages/User%20guide/Post-processing/post-processing.html .

compas_python_utils/debugging_utils.py
nrsegovia: I see, I was not aware of its behavior. It seems using a single * groups elements in a tuple, which explains `len(seeds)==0`. I'm not sure about the potential effects that having a list or array inside a tuple would have, but at least `np.isin` seems to behave correctly. I also think that moving mask before *seeds might be a good idea, up to you.

ilyamandel (changes requested):
One more comment.  This seems to have somehow picked up changes to other files (changelog.h, COWD.cpp, my Matlab code, etc.).  I suspect that's just because other things changed in the meantime, but I am always worried that we will miss something that really did change or revert an unrelated file.  Is there a way to make this look clean, so only changes actually made as part of this PR show up?

nrsegovia:
> Thanks @nrsegovia . I've gone through and implemented your changes, except where I've raised discussions above. As for lists vs arrays, this often depends on the use case. It's easier to append to lists when you don't know how long they will be at the outset, while the arrays are nicer for speed. If there's a specific place you think I should swap one for the other, I'm happy to take a look.

Looks good to me. Sorry for taking so long to check! I have the same concerns as @ilyamandel regarding other files showing up as modified in this PR

reinhold-willcox:
@ilyamandel @nrsegovia No idea why the other files were showing up here, the identical local branches for me were showing the correct file diffs. In the end, changing the base branch to something else then changing it back to dev seems to have solved it, so this looks like a weird github quirk/bug. 

@ilyamandel I'm happy to add some documentation. Also, I'm still iterating with @avivajpeyi on the best way to implement unit tests for this section of the code.

avivajpeyi:
Hey @reinhold-willcox how do these look for unit tests?

avivajpeyi (approved):
Looks good! You could add some automated markdown docs in the future if you'd like

ilyamandel (commented):
Thanks, @reinhold-willcox !

The only thing preventing me from changing "Changes requested" to "Approved" is the documentation. ;-)

reinhold-willcox:
@ilyamandel I've added in the documentation, by cleaning up the notebook for post-processing demos. I actually wrote this notebook ages ago on the same functions, but now it's cleaner and works with the updated syntax. 

@avivajpeyi I'm happy with your unit tests. If we decide we need more, we can create some. 

Note that there are now a large number of commits. This is because I accidentally added a file that was too large for github, and had to do some fancy git command to remove it. This shouldn't have an impact on the repo though, since you can see that only 6 files are changed.

ilyamandel (approved):
I was wondering if you wanted to add a pointer directly to https://compas.readthedocs.io/en/latest/pages/User%20guide/Post-processing/post-processing.html -- but happy to approve.

reinhold-willcox:
@ilyamandel I don't follow. That pages already points to the Basic COMPAS Data Analysis notebook. Unless I'm misunderstanding you?

ilyamandel:
@reinhold-willcox : all good, I hadn't appreciated that DataAnalysis.py would take the place of [DataAnalysis.html.  ](https://compas.readthedocs.io/en/latest/pages/User%20guide/Post-processing/notebooks/DataAnalysis.html)

PR #1475, opened 2026-04-16
docs: add notes on GH images

### LINKED ISSUE #1434: Add details of github-image in docs
The github image of compas is now usable:
https://github.com/TeamCOMPAS/COMPAS/pkgs/container/compas

One can use it in the same way as a docker image (onld difference is that github hosts the image instead of docker).

We should probably add some documentation about this, maybe here
https://compas.readthedocs.io/en/latest/pages/Developer%20guide/Developer%20build/docker-developer.html

avivajpeyi:
Close #1434


ilyamandel (approved):
This look awesome -- thank you, @avivajpeyi !

Issue #1468, opened 2026-03-27
NS mass exceeding CO core mass


pauldisberg:
For a run of COMPAS v03.29.02, initiated with `./COMPAS -n 1000000 --random-seed 1714388113 --mode BSE --muller-mandel-kick-multiplier-NS 5 --muller-mandel-sigma-kick-NS 0.45 --kick-direction-distribution ISOTROPIC`, I obtain one system where the NS mass exceeds the pre-SN CO core mass (which I believe is not as intended).

The SEED of this system is 1714665937, and the NS mass exceeds the CO core mass by 0.03740726 Msun (see attached code below).

<img width="1262" height="473" alt="Image" src="https://github.com/user-attachments/assets/ba5dd801-4ee2-4be6-b914-d0e3c54d8c8c" />

ilyamandel:
Hi @pauldisberg , 

I can't reproduce the issue using the example you provided.  I ran the following with COMPAS v03.29.02.  Note that there's no need to run 1M binaries as an example; the following should be sufficient:
`./COMPAS -n 1 --random-seed 1714665937 --logfile-type CSV  --mode BSE --muller-mandel-kick-multiplier-NS 5 --muller-mandel-sigma-kick-NS 0.45 --kick-direction-distribution ISOTROPIC`
I get an ECSN with a 1.26 Msun remnant mass and a ~1.38 Msun progenitor CO core.

That said, the behaviour is plausible, since the code only insists on the remnant mass not being greater than the He core mass, rather than the CO core mass, in GiantBranch::CalculateRemnantNSMassMullerMandel().  So it's plausible to get a NS remnant that's more massive than the CO core mass.  

The question, then, is whether to require that NSs have a remnant mass less than the CO core mass of the progenitor, or to simply cut off the kick to be zero when they don't (as we do for BHs already).  The former is probably the more sensible choice, especially since, as I just checked, none of the NSs had remnant masses less than the CO core mass in Bernhard's models that inspired our prescription.  What do you think?

Aside: should we also take this opportunity to update the M&M NS kicks in COMPAS to your latest re-calibration?

ilyamandel:
@pauldisberg , please see my questions above.

pauldisberg:
Apologies I forgot to submit my answer @ilyamandel. 

[...]

PR #1477, opened 2026-04-19
M&M 2020 implementation improvement

ilyamandel:
Corrected the M&M NS remnant mass prescription to never return a remnant mass larger than the CO core mass (see issue #1468)


pauldisberg (approved):

ilyamandel:
@jeffriley -- just flagging this for you so it's reflected on your branch.

jeffriley:
@ilyamandel Thanks for the heads up - a couple of things:

(a) since this was an enhancement (was it? The changelog entry says it was - could probably go either way?), the version should have gone to v03.30.00 rather than v03.29.04)

(b) your changes mean that the p_HeCoreMass parameter to GiantBranch::CalculateRemnantNSMassMullerMandel() is superfluous and should be removed.

PR #1481, opened 2026-05-25
Fix to non-random supernova kick angles

AldanaGrichener:
Ran into another bug similar to issue #1378 (while working on the CBD PR, still forthcoming 🙂), applied the analogous fix to PR #1387.

(Missed it back then because I bypassed #1378 by using a bash script which sent each seed individually)

### DIFF
--- a/src/Options.h
+++ b/src/Options.h
@@ -72,7 +72,7 @@ const std::string NOT_PROVIDED_STR(1, static_cast<char>(NOT_PROVIDED_CHAR));
 //    2. if 'fallback' is 'true':
 //           the value specified on the commandline if the user did not specify the 
 //           option on the grid line (regardless of whether they specified the option
-//           on the commandline).  In this case, if the user did not speify a value on
+//           on the commandline).  In this case, if the user did not specify a value on
 //           the commandline, the commandline value is set according to the default
 //           behaviour for the option, and the grid line value is set from that.  Note
 //           that for options whose default behaviours is to draw a random number, this
--- a/src/Options.h
+++ b/src/Options.h
@@ -1748,12 +1748,12 @@ public:
 
     void                                        ShowHelp()                                                              { PrintOptionHelp(!m_CmdLine.optionValues.m_ShortHelp); }
 
-    double                                      SN_MeanAnomaly1() const                                                 { return OPT_VALUE("kick-mean-anomaly-1", m_KickMeanAnomaly1, true); }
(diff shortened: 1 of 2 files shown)


ilyamandel (commented):
@jeffriley , I am a bit puzzled by the description of OPT_VALUE.  It says:

//.          if the user did not speify a value on
//           the commandline, the commandline value is set according to the default
//           behaviour for the option, and the grid line value is set from that.  Note
//           that for options whose default behaviours is to draw a random number, this
//           will only be done once for the commandline value, and each grid line that
//           falls back to the commandline will take the same value as the commandline.
//           Consider using 'fallback' = 'false' for those cases.

[aside: note typo in speify, should be specify :) ]

My question is, how is the programmer to know whether this is the first time that OPT_VALUE is called for this option, and so whether a random-number should be regenerated or not?

jeffriley:
> My question is, how is the programmer to know whether this is the first time that OPT_VALUE is called for this option, and
> so whether a random-number should be regenerated or not?

If I understand the question correctly @ilyamandel, the value for the option is set (once) at parse time in Options.cpp (and that's where we call RAND->Random()), OPT_VALUE just retrieves the value when called.

(EDIT: So the programmer doesn't need to know how many times the getter (which uses OPT_VALUE) is called - it will always return the same value.)

jeffriley (approved):
Looks good.

I fixed the typo :-)

jeffriley (changes requested):
I changed my mind.  changelog needs to be updated - record the change and change the patch/fix number.

ilyamandel:
@jeffriley : The comment says:
// Note
// that for options whose default behaviours is to draw a random number, this
// will only be done once for the commandline value, and each grid line that
// falls back to the commandline will take the same value as the commandline.
// Consider using 'fallback' = 'false' for those cases.
This code said exactly that -- false -- before Aldana's fix.  Why was this the wrong code, given the code comment?

jeffriley:
@ilyamandel 

> This code said exactly that -- false -- before Aldana's fix. Why was this the wrong code, given the code comment?

Is one of us looking at the wrong changes?  Didn't Aldana change the fallback values from "true" to "false"?

Or am I completely misunderstanding your question?

ilyamandel:
You are right, @jeffriley . :)

jeffriley (approved):
Thanks @AldanaGrichener !

AldanaGrichener:
You are welcome :)

PR #1483, opened 2026-08-10
2-detector selection effects in Matlab post-processing

ilyamandel:


ilyamandel:
@pauldisberg -- can you approve this so we don't have too many things hanging around?  Or @avigna, since you like Matlab? :) [We are taking a lax attitude to the Matlab code for now, since it's not "official" / documented.]

pauldisberg (approved):
Approved

Only question I have is concerning the comment about the SNR threshold: "should really be 10 for network, but 8 matches Reed's calculations more closely according to Michael Fulgoni": is this because Reed set this value at 8 and if so why? (but happy to discuss in-person)

ilyamandel:
@pauldisberg : My understanding is that Michael set the threshold to 10.0 as suggested by Reed, but it seems that I am able to best reproduce their results with a threshold of 8.0.  Not sure whether it's because of a different waveform model, approximate detector placement, or for some other reason...

Issue #1484, opened 2026-08-21
Grid interpolation for Ge's mass transfer stability model


Dylan-Hebrail:
The interpolation of the q_crit grids for mass transfer stability with Ge's model is giving weird results (--critical-mass-ratio-prescription GE or GE_IC). 

`severity_moderate` 

`urgency_moderate` 

The figure below shows post first stable mass transfer binaries with a mass ratio q~1, thermal timescale in red, nuclear timescale in green. The donors are HG or MS stars. The vertical lines show the donor masses with q_crit values in the grid and match a saw tooth-like upper limit of the red points tracing where mass transfer gets unstable.

<img width="617" height="448" alt="Image" src="https://github.com/user-attachments/assets/6c47bbe0-66ee-4430-965d-457a2285060b" />

Lines 1320-26 in MainSequence.cpp, in `MainSequence::InterpolateGeEtAlQCrit`:

`double interpolatedQCritUpperEff = (logMass < logLowerMass) ? qCritUpperEffLowerMass
                                         : (logMass > logUpperMass) ? qCritUpperEffUpperMass
                                         : qCritUpperEffLowerMass + (logUpperMass - logMass) / (logUpperMass - logLowerMass) * (qCritUpperEffUpperMass - qCritUpperEffLowerMass);
        double interpolatedQCritLowerEff = (logMass < logLowerMass) ? qCritLowerEffLowerMass
                                         : (logMass > logUpperMass) ? qCritLowerEffUpperMass
                                         : qCritLowerEffLowerMass + (logUpperMass - logMass) / (logUpperMass - logLowerMass) * (qCritLowerEffUpperMass - qCritLowerEffLowerMass);`

The two numerators `(logUpperMass - logMass)` should be `(logMass - logLowerMass)`. As they currently are, `logMass = logUpperMass` would give `qCritLowerMass`, and `logMass = logLowerMass` would give `qCritUpperMass`, with an inverted scaling in `logMass` explaining the saw tooth aspect of the figure.

If this comment about the mass is correct, there are similar issues regarding the interpolation of the radius and beta above and below the code interpolating the mass in the same function.

**Versioning**
COMPAS v03.29.05 (gsl v2.8, boost v1.92.0, HDF5 v2.2.0)

ilyamandel:
Hi @Dylan-Hebrail ,
Thank you for finding this issue and carefully investigating it!
@reinhold-willcox , I believe this is your code -- would you be willing to take a look at it?

[...]

PR #1486, opened 2026-09-09
Fix Ge qCrit interpolation

Dylan-Hebrail:
This fixes the interpolation in MainSequence::InterpolateGeEtAlQCrit(), addressing both problems presented in #1484.

The radius and mass interpolation factors used (upper - x) / (upper - lower), which reversed the interpolation. These are changed to (x - lower) / (upper - lower).

The beta interpolation selected the beta interval [0, 0.5] or [0.5, 1], but used the global beta value as the interpolation factor. This change rescales beta locally within the selected interval before interpolating.

The modified code builds successfully using Clang on macOS. I tested it by producing the same plot as shown in Issue #1484, it now looks like this:
image

The change removed the sawtooth aspect matching the grid points.

Addresses #1484.

### DIFF
--- a/src/MainSequence.cpp
+++ b/src/MainSequence.cpp
@@ -1305,26 +1305,28 @@ double MainSequence::InterpolateGeEtAlQCrit(const QCRIT_PRESCRIPTION p_qCritPres
         double logRadius = log10(m_Radius);
         double qCritUpperEffLowerMass = (logRadius < lowerLogRadiusLowerMass) ? qUppLowLow
                                       : (logRadius > upperLogRadiusLowerMass) ? qUppLowUpp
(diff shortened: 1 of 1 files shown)


ilyamandel (changes requested):
Hi @Dylan-Hebrail ,
This looks great, thank you so much for doing this (and sorry for the very late response -- took me a long time to catch up after travelling).  
Just one request: could you please also update changelog.h to assign this a new version number and a brief comment?

reinhold-willcox:
Looks good! Could you also make the identical fixes to the relevant parts of the HeMS file?

reinhold-willcox (approved):

Dylan-Hebrail:
Hi @ilyamandel, @reinhold-willcox,

I updated the PR with your requests! In HeMS, the relevant function only interpolates for R and M so I changed only these two interpolation factors.

jeffriley (approved):
All good from me.  Thanks Dylan!

ilyamandel (approved):
Thank you, @Dylan-Hebrail !

PR #1487, opened 2026-09-15
Updates to debugging_utils

reinhold-willcox:
The get_event_strings function (and the related subfunctions) now allow for an optional seeds argument to be passed. Previously, these functions would only operate on the entire COMPAS dataset, which can become quite slow for large datasets. 

There is also some minor tidying to the module.


ilyamandel (approved):
Thanks, @reinhold-willcox .  I am not a regular python user, but happy to approve since nobody else is jumping in. :)

reinhold-willcox:
Thanks Ilya! There's something going wrong in the pychecks though, I'll fix that first.

PR #1488, opened 2026-09-21
Improve lint pipeline

ulgens:
Several improvements for the lint (previously named `precommit-checks`) pipeline.

* Pins and updates GHA dependencies, resolve outdated Node.js warnings.
* Replaces pre-commit with prek.


ilyamandel:
Thanks, @ulgens !  
@avivajpeyi , could you please review this one?

avivajpeyi:
Cool! Im not sure what the pros/cons of switching from precommit to prek are, but I see that its working fine... shall we merge it in?

Screenshot 2026-09-23 at 5 03 20 pm

avivajpeyi (approved):

ulgens:
@avivajpeyi In most cases, it's a trivial change but I prefer prek thanks to how bad pre-commit's maintenance is. prek runs faster, has better integration with uv, supports parallelization... and apart from a couple of cases - which doesn't affect the repo, developers can use the same config file with both prek and pre-commit. It's pretty much a free performance and experience upgrade.

PR #1489, opened 2026-09-21
Fix bibtex formatting

ulgens:
Before:

Screenshot 2026-09-21 at 17 00 27 UTC@2x

After:

Screenshot 2026-09-21 at 17 00 55 UTC@2x


ilyamandel (approved):
Thanks for the fix!

